PR #183, opened 2014-05-09
Replace xml_stream with exml

fenek:
This is a reincarnation of Rafał's `exml` branch. The original was too difficult for rebase/merge with current master. :( We can compare performance of modern `exml` and MongooseIM with `xml_stream`-based version. Perhaps with this we could avoid virtually duplicating functionality in another C module?


michalwski:
@pzel, any progress?

pzel:
Until (this serious exml bug)[https://github.com/esl/exml/issues/11] is fixed, we should put the project on hold.

michalwski:
Fixes #414

ppikula:
Performance comparison against master branch: https://snapshot.raintank.io/dashboard/snapshot/prLZfftfcq2XhSflK8LcbyiqZZ0tITey

First run - master, 
Second  - this PR

ppikula:
## Related pull requests:
* https://github.com/esl/exml/pull/14 - requires version bump(release)
* https://github.com/esl/escalus/pull/79 - probably the same in case of this lib

## Minor changes:
* refactor `translate.erl` and add unit test tests
* print output of unit tests output during travis build to prevent killing by travis
* convert some strings to binaries(like lang or version attribute in stream start element)

## Big changes:
* remove expat port - it was used for parsing incoming xml
* remove xml.c - NIF which was responsible for encode xmlel to string
* exml escapes all cdatas and attributes, previously it was done "on demand"
* constraint  on stanza size

## Load tests results: 

Interactive summary:
* [master](https://snapshot.raintank.io/dashboard/snapshot/vo1jd6Hhv7tkbWzILfxoJ1Lfkhamskro)
* [exml](https://snapshot.raintank.io/dashboard/snapshot/FGiU8yUA0CYzHYlMov7n2ucWAvQHHnSK) 

Recon statistics: https://gist.github.com/ppikula/b8c1b9fb77644216e1a6

Long story short: It seems that the current way(master branch) is more efficient in terms of CPU&Memory. Moreover, delivery time is also better in the current version. However these difference  are not so big  ~5%.  

In my opinion, it is worth considering for sake of unification, because right now we are using **3 different!** modules/libraries:
* exml for the bosh&websockets traffic
* xml.c NIF for `#xmlel{} -> binary`
* expat port for `binary -> #xmlel`

BTW, p1 also cleaned it up by moving everything to NIFs, so we can first unify and then make it fast:) 

## Things/ideas to do after merge:
* completely remove xml.erl module - move some functions to jlib or create a new module. In case of "duplicated functions"(get_attr_s, get_cdata_s, etc) use only the ones from the exml library.
* profile & improve exml - ex. either move all escaping to NIF or do it in Erlang, bump reductions.  


Thoughts and comments are welcomed!

ppikula:
I rebased the branch. Regarding increased memory usage, we found a cause, which was an error in the exml library. It was creating refcounted binaries even if they were smaller than 64 bytes. I case of JIDs it generate a big overhead. @studzien  Can you attach some graphs, from the latest tests?

studzien:
Sorry for the delay.

without exml:
https://snapshot.raintank.io/dashboard/snapshot/bLFjor2NY4hgdQh2yqEiwCqRgVNF0wS1

with exml (with refc binaries):
https://snapshot.raintank.io/dashboard/snapshot/jJ30kYS9pZHFCHKeIDkHjH72CBO0uAlk

with exml (no refc binaries):
https://snapshot.raintank.io/dashboard/snapshot/s5SIUroNeufvZ4s1fgGVz0yXMXamM8wx

PR #315, opened 2014-11-20
fails CI tests when there are some auto skipped cases

mentels:
prints problematic suite names

DO NOT MERGE


pzel:
This PR is stale -- I marked it as `do-not-merge` and I'll remove it in a couple days unless you want to continue work, @mentels.

michalwski:
That's mine code actually and can be deleted.

PR #348, opened 2015-01-09
Fix bosh response when origin isn't set

rgrinberg:
If origin isn't explicitly it should just be *. The current behavior is
to not give a response at all.

You can test with `curl -X OPTIONS -v 'http://localhost:5280/http-bind'`. Without this commit you will not get a response from the server.

### DIFF
--- a/apps/ejabberd/src/mod_bosh.erl
+++ b/apps/ejabberd/src/mod_bosh.erl
@@ -186,7 +186,7 @@ init(_Transport, Req, _Opts) ->
 
 -spec info(info(), req(), rstate()) -> {'ok',req(),_}.
 info(accept_options, Req, State) ->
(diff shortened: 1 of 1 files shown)


lavrin:
After some consideration and a glance into the relevant spec I think it's better to stay strict. Please refer to 6.2.1 of [Cross-Origin Resource Sharing W3C Recommendation](http://www.w3.org/TR/2014/REC-cors-20140116/#resource-preflight-requests). The relevant excerpt is:

> 1. If the Origin header is not present terminate this set of steps. The request is outside the scope of this specification.

rgrinberg:
Hmm, to me being strict isn't the problem here but rather not returning a response which is the confusing part. I'd expect at least a 400, bad request to be returned.

PR #355, opened 2015-01-18
Fix Detach Command Typo

justanshulsharma:


ppikula:
The description is correct

PR #357, opened 2015-01-19
[DON'T MERGE] Fix `mongooseimctl started`

lavrin:
`reltool.config.script` referenced application `mongoose` instead
of `mongooseim`. Such application doesn't exist, hence wasn't included
in the release.
Since `ejabberd_ctl.erl` used by `mongooseimctl` looks in
`application:loaded_applications/0` for name `mongooseim`,
the subcommand `started` looped until all retries were exhausted and
always returned a failure, irrelevant of the fact that the server was
already running in the background.

### DIFF
--- a/rel/reltool.config.script
+++ b/rel/reltool.config.script
@@ -57,7 +57,7 @@ IncludeApps = lists:map(fun(App) -> {app, App, [{incl_cond, include}]} end, Apps
 [{sys, [
         {lib_dirs, ["../apps", "../deps"]},
         {incl_cond, exclude},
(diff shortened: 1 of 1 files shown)


lavrin:
I'll leave out the [DON'T MERGE] tag once I can prepare a test case.

michaelcentralway:
:+1: just discovered this as well.
Commit which changed this is: https://github.com/esl/MongooseIM/commit/e45d9fde6ac5f38b5098e404b8658b7fe3b1abe5

PR #378, opened 2015-02-17
Riak base

michalwski:
This PR:
- adds simple connectivity layer with Riak
    - allows to specify more than one pool of riak workers
- abstracts some of riak operations in mongoose_riak module 
- adds `ejabberd_auth_riak` module


mongoose-im:
travis is using test branch `riak-base` from https://github.com/esl/ejabberd_tests/tree/riak-base

michalwski:
This PR goes on top of #426 now

apps/ejabberd/src/ejabberd_auth_riak.erl
```diff
@@ -0,0 +1,269 @@
...
+                     binary(),
+                     fun()) -> boolean().
+check_password(LUser, LServer, Password, Digest, DigestGen) ->
+    case do_get_password(LUser, LServer) of
+        false ->
+            false;
+        #scram{} = Scram ->
+            scram:check_digest(Scram, Digest, DigestGen, Password);
+        PassRiak when is_binary(PassRiak) ->
+            ejabberd_auth:check_digest(Digest, DigestGen, Password, PassRiak)
```
ppikula: The check_password/3 function has  `_` clause.  Do we need the same clause here?

apps/ejabberd/src/mongoose_riak_sup.erl
```diff
@@ -0,0 +1,117 @@
...
+    {ok, _} = supervisor:start_child(ejabberd_sup, ChildSpec).
+%%--------------------------------------------------------------------
+%% @doc
+%% Starts the supervisor
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec(start_link(integer(), list()) ->
+    {ok, Pid :: pid()} | ignore | {error, Reason :: term()}).
+start_link(Workers, PoolSpec) ->
```
ppikula: Should be PoolsSpec instead of PoolSpec. By the way I am not a fan of creating separate pools for every riak node.
I think It should be handled by an external load balancer or DNS server.  I think that the mentioned solution simplifies the logic on our side and allows to easily expand  Riak cluster  - one can add a new riak node and then register it in the load balancer. In the current solution we need to restart/reload MiM nodes and I couple of riak nodes go down the number of active connections will shrink (or even take MiM down?) while the load balancer would just open new connections to the remaining nodes.

ppikula:
I don't  see any changes in the `doc` directory ;)

apps/ejabberd/src/ejabberd_auth_riak.erl
michalwski: We don't, it's covered in `do_get_password/2`.

apps/ejabberd/src/mongoose_riak_sup.erl
michalwski: I agree with you :) The implementation was done before we changed our mind into LB approach. Fortunately all you describe is possible with current implementation. It's as simple as setting only one endpoint in the list of riak servers which is the LB for example.
DNS based solution will not be that simple with Erlang. As far as I remember some internal functions sorts the list of IP address for given name so you will always get the same IP unless you choose the IP on your own.

michalwski:
There is no `doc` yet because right now I'm focused on implementation only :)

ppikula:
Any updates regarding docs ? I think this is very important to not make the doc worse than it is now and
adding undocumented stuff is step towards that situation.  I won't merge it without the doc :).

apps/ejabberd/src/mongoose_riak_sup.erl
ppikula: Can we remove this multi-pool logic?  It think It would simplify the logic on our side, especially situations when one of the nodes goes down or extending/shrinking the Riak cluster. 

With one pool we give one simple guarantee  - as long as I can have one pool alive I will run. If something is wrong I will go down. With multi-pool we may start thinking how many nodes down means failure, should Mongoose try to restore the dead pools after certain amount of time or how to add a new one etc.  For example, In case of MAM I doesn't make sense to serve users if we cannot store their messages.

All in all, I'm a fan of one-pool approach here and removing all possibilities to setting more that one pool. Doing that we forcing users to go in the right direction :). If somebody needs to test, he/she can do that with one Riak node or use HAproxy on localhost.  

Regarding DNS it depends on DNS server. I think that ROUTE 53 can return only 1 IP and also is able to do health checks.

I would like to hear what other guys think about that ! If it turns out that I'm alone in this opinion, I will merge it :)

ppikula:
### RFC - auto reconnecting pool?

There is one thing that bothers me (not sure how to solve it),  and I would like to here your opinion @esl/mongooseim-developers, because
it is a base PR for other riak features.

Right now when Riak is down, the pool after couple of supervisor's restarts goes down and
it won't reconnect automatically.

The question is: what should we do?

Of course the answer is: It depends on use case :)

#### Option 1: Use auto reconnect feature from riakc

We are consistent with odbc pool behaviour - the pool is constantly trying to reestablish connections and is even
capable of buffering some requests till connection is reestablished - it can
be set using [client_options](https://github.com/basho/riak-erlang-client/blob/master/include/riakc.hrl#L30).
I tested it, again manually, seems to work. There is one problem, the riak library doesn't
log the fact that connection was lost and reestablished. @michalwski has some workaround to this.
Of course we still have logs from failed writes/reads, but we don't see exact moment when the outage started. Do we really need that ? 

#### Option 2: Take down the server

It depends on the use case, but sometime it doesn't make sense to process messages -
it is better to crash early and give the feedback to client applications. In this
case we are not pretending that we saved messages in the archive and then  oops - they are gone
because our riak cluster was down at that time. On the other hand, in case of authentication
backend it is ok to continue processing, we only won't be able to authenticate new sessions.
Maybe separate pools per each module?

Riak's selling point is its high availability, so if whole Riak cluster is unavailable,
something must be really wrong or it indicates network problems. Maybe it is a
good idea to stop the node?

#### Option 3: Make it configurable? What & How?

We can put strategy in the `riak_server` options (take down or retry). Does it make
sense? Does it introduce confusion?

I would vote for option 1, we keep configuration simple, it is the same as in case of odbc. 

### Summary of my recent changes:
* added entries in the docs about configuring riak connection and the database itself + added description for remaining db backends
* converted the pool of pools to one pool
* made the pool reloadable via our config reload mechanism
* changed Riak's configuration on Travis I added {delete_mode, immediate} option to the ensure that tombstones are deleted immediately not after 3secs, otherwise some of our tests may fail - escalus thinks that it removed users, but in fact it needs to wait 3 secs ...  (more [here](http://basho.com/posts/technical/riaks-config-behaviors-part-3/) ).
 I'm not happy with this workaround, because we are testing not the real world case, I guess. On the other hand waiting 3 secs between every test case is pointless. I am not a Riak expert so, if someone has opinion on that, please share.

michalwski:
I also vote for option 1. We some additional effort we can add logs to riakc process (thanks to `sys:install` function)  informing when connection was closed and when reconnection attempts were made. From my experience it's always good to know when exactly communication with external service was broken and re-established.

pzel:
I also vote for #1.
By the way, do we have a consistent method of raising alarms from
MongooseIM?
Now that most installations in the wild are actually instrumented with
Wombat, (and Wombat can send emails, khm khm), maybe we should start
notifying the alarm registry that something has gone wrong?

-S

On Mon, Aug 17, 2015 at 11:25 AM, Michał Piotrowski <
notifications@github.com> wrote:

> I also vote for option 1. We some additional effort we can add logs to
> riakc process (thanks to sys:install function) informing when connection
> was closed and when reconnection attempts were made. From my experience
> it's always good to know when exactly communication with external service
> was broken and re-established.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/esl/MongooseIM/pull/378#issuecomment-131742022>.
>



-- 
Simon Zelazny
Erlang Solutions, Ltd.

michalwski:
I'm not sure if most installations are instrumented with Wombat. Did you count all the installations done without our knowledge? They are thousands I'd say :D and not necessarily using Wombat (yet :D ).
Anyway, alarm registry sounds good to me but we should not assume we have Wombat installed.

pzel:
Sure -- my point was rather that alarms are an easy and extensible way to notify operators about serious failures, and they cost nothing in terms of performance. Anyway, it's another topic that should be addressed separately, sorry for side-tracking the conversation!

lavrin:
:+1: for option 1 because of consistency with ODBC pooling. Moreover, server shutdown is something that requires manual intervention, which is labour, therefore cost. I'd say that providing a limited service (e.g. no archive) is a much lesser impact/deficiency from the end-user's point of view. IMO option 3 is subject to YAGNI.

Also :+1: for alarms.

michalwski:
:+1: from me for merge

PR #383, opened 2015-02-19
Remove obsolete XEP-0091 timestamp

gmodarelli:
Hi guys, at my company we are developing a mobile chat based on MongooseIM. For our android client we are using the ASmack library and we get an date parsing error when the client receives an offline message.

The offending date is the one passed inside the x tag of the [XEP-0091](http://xmpp.org/extensions/xep-0091.html) extension. Since I noticed it has obsoleted, I thought I could have removed it from MongooseIM.

But I'm a Ruby developer and know just the very basics of Erlang. I've managed to remove the <x> tag from the message stanza, recompiled the repo and the server works. No date parsing error on the client anymore.

But I don't know if I need to add some tests around my changes or if the changes are correct (ie. do I need to remove/add something else?). I'd love to contribute to the project and I'd like to hear your feedback.

Bye

### DIFF
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -1768,9 +1768,9 @@ process_presence_probe(From, To, StateData) ->
                     Packet = xml:append_subtags(
                                StateData#state.pres_last,
                                %% To is the one sending the presence (the target of the probe)
-                               [jlib:timestamp_to_xml(Timestamp, utc, To, <<>>),
+                               [jlib:timestamp_to_xml(Timestamp, utc, To, <<>>)]),
                                 %% TODO: Delete the next line once XEP-0091 is Obsolete
-                                jlib:timestamp_to_xml(Timestamp)]),
+                                %% jlib:timestamp_to_xml(Timestamp)]),
                     case privacy_check_packet(StateData, To, From, Packet, out) of
                         deny ->
                             ok;
@@ -2957,8 +2957,9 @@ add_timestamp({_,_,Micro} = TimeStamp, Server, Packet) ->
         false ->
             %% TODO: Delete the next element once XEP-0091 is Obsolete
             TimeStampLegacyXML = timestamp_legacy_xml(Server, Time),
-            TimeStampXML = jlib:timestamp_to_xml(Time),
-            xml:append_subtags(Packet, [TimeStampLegacyXML, TimeStampXML]);
+            %% TimeStampXML = jlib:timestamp_to_xml(Time),
+            %% xml:append_subtags(Packet, [TimeStampLegacyXML, TimeStampXML]);
+            xml:append_subtags(Packet, [TimeStampLegacyXML]);
         _ ->
             Packet
     end.
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -2953,13 +2953,13 @@ add_timestamp({_,_,Micro} = TimeStamp, Server, Packet) ->
     Time = {D,{H,M,S, Micro}},
     case xml:get_subtag(Packet, <<"delay">>) of
         false ->
-            TimeStampLegacyXML = timestamp_legacy_xml(Server, Time),
-            xml:append_subtags(Packet, [TimeStampLegacyXML]);
+            TimeStampXML = timestamp_xml(Server, Time),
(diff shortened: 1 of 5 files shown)


michalwski:
I suggest removing the function (`jlib:timestamp_to_xml/1`) generating obsolete timestamp also.

gmodarelli:
Thanks @michalwski

apps/ejabberd/src/jlib.erl
```diff
@@ -54,7 +54,6 @@
          parse_xdata_submit/1,
          timestamp_to_iso/1, % TODO: Remove once XEP-0091 is Obsolete
```
michalwski: Just noticed that there is another obsolete function. I'd remove it also and see if our tests run by travis passes.

michalwski:
Regarding tests there is [offline_SUITE](https://github.com/esl/ejabberd_tests/blob/master/tests/offline_SUITE.erl#L57) so you can modify the only test case there to ensure that the legacy timestamp is not present.
If you fork ejabberd_tests and create branch of the same name as the branch from which this PR was sent then travis-ci will automatically use your branch for tests.

michalwski:
And what's most important thanks for your contribution @gmodarelli :+1:

gmodarelli:
Ok @michalwski I'll try to write the tests and I'll remove the other obsolete method. I'm glad to contribute :+1:

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -2955,10 +2953,8 @@ add_timestamp({_,_,Micro} = TimeStamp, Server, Packet) ->
     Time = {D,{H,M,S, Micro}},
     case xml:get_subtag(Packet, <<"delay">>) of
         false ->
-            %% TODO: Delete the next element once XEP-0091 is Obsolete
             TimeStampLegacyXML = timestamp_legacy_xml(Server, Time),
```
studzien: I think that somebody confused the binding/function name and named the modern format as legacy. It's a good opportunity to fix it :)

gmodarelli: @studzien I hope the new name is ok. Thanks for the feedback

apps/ejabberd/src/jlib.erl
gmodarelli: @michalwski I removed this one too.

gmodarelli:
@michalwski I'm still not able to write tests for the project, sorry. I need to study Erlang more :) But I will!

michalwski:
Looks like some of the obsolete functions where used in other place(s).

Take a look at tests output:
https://travis-ci.org/esl/MongooseIM/jobs/52133477#L1543
and here's more details about the error:
https://travis-ci.org/esl/MongooseIM/jobs/52133477#L1750

This happened for all jobs except one which doesn't run this test.

michalwski:
I didn't meant to restore the obsolete function :) The best approach would be to fix the place where the function is used.

gmodarelli:
@michalwski ok :) I'll try to replace it.

gmodarelli:
@michalwski sorry if this is taking me so long. I had doubts on how to proceed. The obsolete function (timestamp_to_iso/1) is only used by the build_random_password/1 function in the mod_admin_extra_accounts module.

So I replaced the call to the function and generated the date in the build_random_password itself.

Please let me know if this is the right way to proceed. Thank you!

michalwski:
I think you can put there date in other format not necessary the obsolete one. This function is generating random password for banned user so in fact we can put any password there, even without date.

gmodarelli:
Ok cool, so we could remove the date. I'm just wondering if we have restriction on the format of the random password. Can we just remove the date part or we need to replace it with something else?

Thank you.

michalwski:
I don't know about any restriction. This format was inherited from ejabberd. The only reason to setup such password is to prevent banned user from authentication. The format just brings the server administrator info that user with such password was banned, when and from what reason.

gmodarelli:
Ok, I'll add the date back (with this format 1997-07-16T19:20:30.45+01:00) so we won't break code depending on it, even if it is just for logging or debugging purpose.

gmodarelli:
I was not able to add the timezone offset, so I opted for the universal time.

michalwski:
Ok, UTC is enough for that.

michalwski:
Thanks for the fix!

gmodarelli:
Thank you for helping me with it!

PR #390, opened 2015-03-01
Create ./etc if doesn't exist for quickrun

rgrinberg:
Doing `$ make quickrun` doesn't work on a clean build because ./etc
doesn't exist


michalwski:
Thanks @rgrinberg !

PR #409, opened 2015-03-20
add gcm dependency, ability to push message to hardcoded registration id... [DON'T MERGE]

vooolll:
... on offline message, gcm configurable via ejabberd cfg


vooolll:
Right now there is no possibility to get token automaticly from user, I suppose , I need to extend some modules for that, for instance if odbc authentication is used , I need to store registration_id/device_token in user table as not required column, in case the http authentication used one more request should be send for getting device_token and so on with other authentication backends. Am I right?

michalwski:
Something is wrong with all jobs on travis. Looks like the release was not generated: https://travis-ci.org/esl/MongooseIM/jobs/55129930#L875. You need to add gcm to deps in rebar.config.

vooolll:
@michalslaski I fixed this, waiting for build.

michalwski:
ok

BTW I'm @michalwski and not @michalslaski but you were close :D

vooolll:
@michalwski  I am sorry :D michalwski

vooolll:
May I add custom parameter to authentication xml or it will brake something? I need token to make a push.

michalwski:
Adding custom parameter to auth xml is not good idea. How about you store the token in private storage (mod_private) and read it (or even cache somehow in the c2s process) when needed?

vooolll:
@michalwski I try to save it with mod_private, but it seems like I do something weird 

`-define(BACKEND, (mod_private_backend:backend())).`

I define the BACKEND macro, 

then I try to do:
`Token = ?BACKEND:get_data(To#jid.luser, To#jid.lserver, <<"user:token">>, []),`

but It causes error:

2015-03-26 08:47:34.556 [error] <0.618.0>@ejabberd_hooks:run1:240 {undef,[{mod_private_mnesia,get_data,[{jid,<<"bob">>,<<"localhost">>,<<>>,<<"bob">>,<<"localhost">>,<<>>},<<"localhost">>,"user:token",[]],[]},{mod_offline_push,send_notice,3,[{file,"src/mod_offline_push.erl"},{line,40}]},{safely,apply,3,[{file,"src/safely.erl"},{line,19}]},{ejabberd_hooks,run1,3,[{file,"src/ejabberd_hooks.erl"},{line,236}]},{ejabberd_sm,route,3,[{file,"src/ejabberd_sm.erl"},{line,108}]},{ejabberd_local,route,3,[{file,"src/ejabberd_local.erl"},{line,139}]},{ejabberd_router,route,3,[{file,"src/ejabberd_router.erl"},{line,78}]},{ejabberd_c2s,session_established2,2,[{file,"src/ejabberd_c2s.erl"},{line,1098}]}]}

As I see in mod_private there is ?BACKEND:init(Host,Opts) but I actually don't want to initialize mod_private backend twice, I actually want to use the one, which already initialized. Is there any other backend independent way to extract private xml?

michalwski:
The function `?BACKEND:get_data/4` is not exported. Please use functions `?BACKEND:multi_set_data/3` and `?BACKEND:multi_get_data/3`.

vooolll:
Currently I have the following format for private xml,

    `<notification xmlns="notification:prefs">
      <token>Some token here</token>
    </notification>`

Last version mod_offline_push module can get this xml of sender and push the notification to the specified token. When I finish the IOS client preparation I will add the `<type>` tag (for mobile platform)

vooolll:
@michalwski Could you help me a bit, I can't catch why my build is not passing?

michalwski:
I've restarted the failed job. That's probably because of some time-sensitive tests. Sometimes they fail when travis is too loaded.

vooolll:
@michalslaski is there way to identify it, myself without asking you, it is a bit annoying I think.

michalwski:
You can always set up travis for your fork so then you'll be able to restart jobs when needed.

vooolll:
@michalwski where is a good place user should store his apple push notification certificates?

PR #421, opened 2015-04-09
Add list of supported XEPs to README; update documentation links

pzel:
😸


fenek:
I think there should be a full name of each XEP included, as I think it will be more friendly to many people looking for specific features. :) It could be easy to remember "Message Archive Management" for someone but "XEP-0313" could be impossible. :)

ppikula:
Can somebody rerun the tests(manually verify each XEP ... ) and merge it if they pass?

PR #426, opened 2015-04-13
Auth refactoring

michalwski:
Refactoring of auth-related modules.

One of the most significant changes is that all auth backend modules expect `ejabberd:luser()` and `ejabberd:lserver()` now in callbacks defined by `ejabberd_gen_auth` behaviour. In other words they don't do `jlib:nodeprep(User)` or `jlib:nameprep(Server)`

All changes taken from PR #378


mongoose-im:
travis is using test branch `auth-refactoring` from https://github.com/esl/ejabberd_tests/tree/auth-refactoring

PR #429, opened 2015-04-15
Add mod_offline_stub to prevent <service-unavailable/>

lavrin:
...when MAM is used, but mod_offline is not.


lavrin:
No tests are available for this functionality.

PR #436, opened 2015-05-05
add config option to enable external auth behavior for http auth

Stelminator:


Stelminator:
We have a use case for HTTP Auth, which lets us deliver auth as a service, but it doesn't make sense to implement `get_password/2`.  This results in the server advertising DIGEST-MD5, which most clients will prefer, but it cannot work.  I'd prefer to be able to just select SASL mechanisms per host, but just being able to flip the `store_type/1` to `external` here gets me the behavior I need: only PLAIN auth being advertised.

I'm not sure what to do about `plain_password_required/0`.

PR #441, opened 2015-05-13
mod_mam cassandra: create worker pool per node

### LINKED ISSUE #440: mod_mam_con_ca_arch not playing nice in cluster environment
When shutting down a node, the whole worker pool gets deleted, which is shared between nodes:

https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mod_mam_con_ca_arch.erl#L133

An easy fix would be to have one group per node:
https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mod_mam_con_ca_arch.erl#L236   should rather be -> {mam_ca, node(), Host}.

(same holds for the mod_mam_muc_ca_arch module)

sstrigler:
Fixes #440

### DIFF
--- a/apps/ejabberd/src/mod_mam_con_ca_arch.erl
+++ b/apps/ejabberd/src/mod_mam_con_ca_arch.erl
@@ -233,7 +233,7 @@ select_worker(Host, UserJID) ->
     end.
 
 group_name(Host) ->
-    {mam_ca, Host}.
+    {mam_ca, node(), Host}.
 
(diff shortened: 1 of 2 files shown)


michalwski:
@sstrigler do you think you can integrate cassandra tests on travis? @arcusfelis started the work some time ago but didn't have chance to finish it.

sstrigler:
@michalwski haha, no, probably not, sorry :dash:

michalwski:
:disappointed:

Issue #435, opened 2015-05-05
mod_offline race condition


sstrigler:
If a user A sends a message to user B exactly that moment while this user is logging in it may happen that user B's mod_offline spool returns no message but also user B would not receive user A's message and it would get queued at the mod_offline spool. Next time user B logs in they would receive the message.

sstrigler:
lol https://support.process-one.net/browse/EJAB-1268

PR #447, opened 2015-05-27
[#435] fix race condition in mod_offline

sstrigler:
See https://github.com/esl/MongooseIM/issues/435 for details.


michalwski:
This is superseded by #1213 and associated race condition fixes.

Issue #432, opened 2015-04-24
Ubuntu + Debian: please use /etc and /var/log


COLABORATI:
Wow, that was weird, I was looking for the logfiles and I found that the ubuntu packages of mongooseim has everything below /usr/lib/mongooseim. Well, I could change the init script, but what happens on next update? It would  be great if you configured the debian + ubuntu packages to use the usual places, e.g. /etc/mongooseim and /var/log - THANKS!

kjellski:
I'd just like to say that would be really great! :+1:

COLABORATI:
Hi, I would like to ask if you just want to keep it like that or if you will change this in the next release, because I am writing some automation and provisioning stuff around this and would like to save myself a round of testing and fiddling around. 

This is why it would be great if the deb packages would use the expected paths ASAP - but if you just want to use the /usr/lib/mongooseim for some reason I do not understand, then it would be great to read some info from you about this, so I can better plan for this. 

Yes, I know, I am annoying you, sorry. You know, I do not want to package this myself, this is why I am asking you to change the paths, and I know this will take about 15 minutes for you, but will save me (and billions of other admins!) hours or possibly days of pain. Pleeease. Thank you very much!

lavrin:
We are aware of this being a pain in the ass, believe me. This is the place where you're wrong, though:

> I know this will take about 15 minutes for you.

The rabbit hole is deeper than you can see from above the edge. I'll try solving this in the next week.

kjellski:
@lavrin thanks for your efforts, I know it's hard to do these kind of maintenance things! :+1: for every effort!

lavrin:
To give you some context, #352 is the reason why it's not a 15 minute fix.

arcusfelis:
Probably we cause create symlinks "etc/ejabberd.cfg => /etc/ejabberd.cfg" and "log/ => /var/log/ejabberd" inside post-scripts of the packages.

kzemek:
AFAIK this is fixed for some time now @fenek , pending confirmation

fenek:
It's fixed, at least as of 3.3.0

Issue #352, opened 2015-01-13
Setting `lager_file_backend` path to non-default breaks `ejabberd_loglevel:set/1`


lavrin:
There are two problems with changing a `lager_file_backend` path in `app.config`:

- `ejabberd_loglevel:set/1` breaks,
- there's a discrepancy between `ejabberd_app:get_log_path/0` (same as `ejabberd_loglevel:log_path/0` seen in the logs below) and the path set in `app.config` for `lager_file_backend`.

In other words, there's no consistent (or any properly functioning?) interface for operations teams to set the logging path for MongooseIM to anything other than the default.

Please compare the two console logs. The first one uses the default path `log/ejabberd.log` and doesn't cause issues:

```
$ cat etc/app.config
[
 {lager, [
          {handlers, [
                      {lager_console_backend, [info, {lager_default_formatter,[{eol, "\r\n"}]}]},
                      %% file is shadowed by ejabberd.cfg
                      {lager_file_backend, [{file, "log/ejabberd.log"}, {level, info}, {size, 2097152}, {date, "$D0"}, {count, 5}]}
                     ]}
         ]}
].

%% vim: filetype=erlang
17:56:03 erszcz @ x4 : ~/work/esl/mongooseim/rel/mongooseim (master)
$ bin/mongooseimctl live
Exec: /Users/erszcz/work/esl/mongooseim/rel/mongooseim/erts-6.1/bin/erlexec -boot /Users/erszcz/work/esl/mongooseim/rel/mongooseim/releases//mongooseim -embedded -config /Users/erszcz/work/esl/mongooseim/rel/mongooseim/etc/app.config -args_file /Users/erszcz/work/esl/mongooseim/rel/mongooseim/etc/vm.args -- console
Root: /Users/erszcz/work/esl/mongooseim/rel/mongooseim
Erlang/OTP 17 [erts-6.1] [source] [64-bit] [smp:4:4] [async-threads:5] [kernel-poll:true]

2015-01-13 17:56:23.889 [notice] <0.183.0>@lager_file_backend:128 Changed loglevel of log/ejabberd.log to info
2015-01-13 17:56:23.889 [info] <0.7.0> Application lager started on node mongooseim@localhost
2015-01-13 17:56:23.910 [info] <0.7.0> Application mnesia exited with reason: stopped
2015-01-13 17:56:23.930 [info] <0.7.0> Application mnesia started on node mongooseim@localhost
[...]

PR #448, opened 2015-05-27
Codify ejabberd_loglevel behaviour in tests and allow for changing logging path

lavrin:
Changes:
- added `ejabberd_loglevel_SUITE` which tests `ejabberd_loglevel` external API,
- removed `ejabberd_app:get_log_path` and `ejabberd_loglevel:log_path`,
- made Lager config in `app.config` the definitive place for configuring MongooseIM logging - you can use any backends available and use multiple backends of the same type - the hardcoded assumptions about backend types are lifted,
- made `etc/app.config` and `bin/mongooseim` templates so that the log path can be set uniformly for logs from within the server as well as logs created by the runner script (i.e. `bin/mongooseim`).

Solves #352 and partially addresses #432.


mongoose-im:
travis is using test branch `log-path-fix` from https://github.com/esl/ejabberd_tests/tree/log-path-fix

lavrin:
Ready to merge!!111

ppikula:
Great code, man!

PR #460, opened 2015-06-29
Riak vcard

michalwski:
Riak backend for mod_vcard.

This is done on top of #378


mongoose-im:
travis is using test branch `riak-vcard` from https://github.com/esl/ejabberd_tests/tree/riak-vcard

apps/ejabberd/src/ejabberd_binary.erl
```diff
@@ -28,3 +29,16 @@ string_to_binary(S) when is_list(S) ->
     end;
 string_to_binary(B) when is_binary(B) ->
     B.
+
+
+join([], _) ->
```
fenek: Maybe this could be more optimal:
```erlang
join(Parts, Sep) when is_integer(Sep) -> join(Parts, <<Sep>>);
join(Parts, Sep) -> join(Parts, Sep, <<>>).

join([], _Sep, Acc) -> Acc;
join([Part], _Sep, Acc) -> <<Acc/binary, Part/binary>>;
join([Part | RParts], Sep, Acc) -> join2(RParts, Sep, <<Acc/binary, Part/binary, Sep/binary>>).
```

michalwski: Actually some time ago I saw some micro-benchmarking done by @arcusfelis showing that `io_list_to_binary` is faster than incrementally built binary.

fenek: Ah, true, I think I saw it too. I wonder if it still holds with current OTP versions (AFAIR those were done with R16).

PR #463, opened 2015-07-07
Refactor ejabberd_c2s:handle_info/3

lavrin:
Split `ejabberd_c2s:handle_info/3` into multiple functions, remove redundant code, extract named predicates, reformat.

### DIFF
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -1205,11 +1205,9 @@ handle_routed(_, _From, _To, Packet, StateData) ->
 handle_routed_iq(From, To, Packet = #xmlel{attrs = Attrs}, StateData) ->
     case jlib:iq_query_info(Packet) of
 	#iq{xmlns = ?NS_LAST} ->
-	    LFrom = jlib:jid_tolower(From),
-	    LBFrom = jlib:jid_remove_resource(LFrom),
-	    HasFromSub = ( ( ?SETS:is_element(LFrom, StateData#state.pres_f)
-			     orelse ?SETS:is_element(LBFrom, StateData#state.pres_f) )
-			   andalso is_privacy_allow(StateData, To, From, #xmlel{name = <<"presence">>}, out) ),
+	    HasFromSub = ( is_subscribed_to_my_presence(From, StateData)
+			   andalso is_privacy_allow(StateData, To, From,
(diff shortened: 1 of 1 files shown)


apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -1369,6 +1186,157 @@ handle_info(Info, StateName, StateData) ->
...
+	    {true, Packet#xmlel.attrs, StateData};
+	deny ->
+	    {false, Packet#xmlel.attrs, StateData}
+    end;
+handle_routed(_, _From, _To, Packet, StateData) ->
+    {true, Packet#xmlel.attrs, StateData}.
+
+handle_routed_iq(From, To, Packet = #xmlel{attrs = Attrs}, StateData) ->
+    case jlib:iq_query_info(Packet) of
+	%% TODO: Support for mod_last / XEP-0012. Can we move it to the respective module?
```
michalwski: I don't thinks. The hook in its current version doesn't contain state date which is used by the mod_last functionality (as far as I remember).
This could be possible via hook only if we add new type of hook extracting the c2s state if needed and passing it as an argument to the handler.

lavrin:
Looking at [coverage stats](https://coveralls.io/builds/2993815/source?filename=apps%2Fejabberd%2Fsrc%2Fejabberd_c2s.erl) it seems a relatively high fraction of the refactored code is covered and **nothing broke!** I propose merging this as is and pave the way for future `ejabberd_c2s` cleanups.

apps/ejabberd/src/ejabberd_c2s.erl
lavrin: I'm not sure yet, but after a quick glance it seems `StateData` here is only used for checking whether `From` is subscribed to *our* presence (which can be told outside of C2S by looking at the roster) and privacy list check. This would be more costly when done from `mod_last` (extra roundtrips to the DB / roster table), but would slim c2s down. If necessary a new hook is OK in my opinion and doesn't introduce the extra overhead.

pzel:
I concur -- but if you want to really set a precendent and pave the way, let's squeeze those tests in there as well ;)

apps/ejabberd/src/ejabberd_c2s.erl
michalwski: You are right, but I'm more concern about the performance here that's why I think new hook would be better for such cases.

PR #464, opened 2015-07-13
Issue:

aberke:
When sessions not cleanly closed, SM_BACKEND sometimes keeps old sessions for which SID process has died.  In this case, instead of handling packet as a packet sent to offline user, session manager attempts to send message to dead PID and message goes nowhere.
Fix:
When fetching user PIDs, filter out dead PIDs and delete sessions referring to dead PIDs.

### DIFF
--- a/apps/ejabberd/src/ejabberd_sm.erl
+++ b/apps/ejabberd/src/ejabberd_sm.erl
@@ -664,29 +664,55 @@ route_message(From, To, Packet) ->
 
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-
 -spec clean_session_list([sid()]) -> [sid()].
 clean_session_list(Ss) ->
-    clean_session_list(lists:keysort(#session.usr, Ss), []).
+    BadSessions = find_bad_sessions(Ss),
+    lists:foreach(
+        fun(#session{sid=Sid, usr={User, Server, Resource}=Usr}) ->
+            ?DEBUG("closing dead session with SID: ~p, User: ~p", [Sid, Usr]),
+            spawn(?MODULE, close_session, [Sid, User, Server, Resource])
+        end,
+        BadSessions),
+    lists:filter(fun(S) -> not lists:member(S, BadSessions) end, Ss).
 
+-spec find_bad_sessions([sid()]) -> [sid()].
+find_bad_sessions(Ss) ->
+    lists:merge(find_dead_sessions(Ss), find_old_sessions(Ss)).
 
--spec clean_session_list([sid()],[sid()]) -> [sid()].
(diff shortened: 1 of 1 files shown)


pzel:
Hey @aberke, first of all: thanks for taking the time to spot the bug and fix the code.
Your pull request looks cool, but I was wondering whether we could make the whole mechanism cleaner. There's already `clean_session_list/1`, which ignores SIDs that are be outdated.
Now, we're adding another function that does something similar, but cleans the data store for real.

For the record, I like your appoach much better :)

I have a couple of open-ended questions:
   - Which code path leads to sessions not being closed properly? Could we plug the hole there?
   - Can we perhaps get rid of the hacky `clean_session_list/1` and guarantee that the session manager always returns live `pid()`s?
   - `process_info/1` seems heavy, and the session manager is a known performance hotspot for MongooseIM. Maybe a non-crashing `safe_is_process_alive/1` would be better here?
```
safe_is_process_alive(P) when is_pid(P) -> erlang:is_process_alive(P);
safe_is_process_alive(_) -> false.
```
   - Is it scalable at all to check this every time we get a session list? Remeber, we could be verifying `pid()`s that are living on another node in the cluster, and this means addtitional network round-trips.

aberke:
Thanks for your feedback and consideration.

Based on your feedback, I am going to amend this PR (let me know if you instead prefer follow-up commits).

- My concern with `is_process_alive/1` is that it will return a false negative (`false`) if the process is alive on another node.  Do you see a hole in my concern?

- Re "Which code path leads to sessions not being closed properly? Could we plug the hole there?"  I'm not sure.  I would assume the issue is around MongooseIM processes not shutting down cleanly (and then not calling `ejabberd_sm:close_session`) while the `?SM_BACKEND` continues to run and hold sessions.  I happen to be using `ejabberd_sm_redis` as the backend.  I don't see an easy way to fix this if it is indeed the problem...

- I think closing the sessions that are deemed old in `clean_session_list/1` could be useful to keep this issue from reoccurring, but closing the session requires the Server and Resource.

As mentioned, I'll amend this PR to change `clean_session_list/1`  to use `clean_out_dead_sessions/3` so that we can follow-up from there.
What do you think?

pzel:
Hey, it looks like my half-baked theories were off the mark. Doing some local testing with two nodes, I learned that neither `process_info/1` nor `is_process_alive/1` work across node boundaries. (`sys:get_status/1` does, as it uses regular message passing).
I'll look into ejabberd_c2s's terminate function tomorrow and see if I can find the leak. It would be awesome if you could reproduce your bug, so we won't end up hiding breakage, but actually fixing it.

aberke:
Hey @pzel please take a look at the amended changes.  I'll trust you as the expert on whether to use `process_info` vs `is_process_alive` vs _sys:get_status`.

aberke:
Will reopen a better PR

aberke:
Please see new PR: https://github.com/esl/MongooseIM/pull/478

PR #465, opened 2015-07-14
make c2s a better place

ppikula:


michalwski:
I suggest going one step further and refactor the cases in cases in cases ...

pzel:
Once we have proper indentation, we can refactor. And also, more than one person will be ablo refactor concurrently. Let's merge this first, without additions.

PR #470, opened 2015-08-01
Use supervisor2 as a dependency

saa:


pzel:
It seems that https://github.com/odo/supervisor2 is the main repo for supervisor2. Is there a specific reason why you're including your private fork?

saa:
1) Last commit work only on erlang 18
2) And last update without last commit for support erlang 18 is Jan 10, 2014

pzel:
I see. As much as I'd like to merge your work, I think the best option is to first ensure that odo/supervisor2 works on all the Erlang versions we care about, then point MongooseIM to depend on odo/supervisor2. 

We've had issues with private repos being deleted in the past. It's a lot of trouble for non-Erlangers who have MongooseIM installations in the wild, when they suddenly find they can't build a release for some reason.

If you could make a pull request to odo, I will support it and help you get it through.

If you need inspiration on how to support different functions (`erlang:now/0` vs `erlang:timestamp/0`) on different Erlang/OTP versions, take a look at https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/ejabberd_binary.erl and the relevant option set in https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/rebar.config#L6

saa:
You right. Ok, I'll make pull request. In MongooseIM need create new PR or just update?

pzel:
Just update, I'll be watching ;)

michalwski:
Any updated on this?

saa:
@michalwski @pzel sorry, was problems with free time. I will soon push fixes.

michalwski:
Please re-open when you update according to the comments.

PR #471, opened 2015-08-03
Added download section

Nyco:
Added a small download section for those who don't know how to handle source code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,10 @@ For a quick start just
 [download the pre-built package](https://www.erlang-solutions.com/downloads/download-mongooseim)
 that suits your platform.
 
+Download
+--------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,6 +15,10 @@ For a quick start just
 [download the pre-built package](https://www.erlang-solutions.com/downloads/download-mongooseim)
 that suits your platform.
 
+Download
+--------
+Feel free to download MongooseIM at: https://www.erlang-solutions.com/downloads/download-mongooseim
```
lavrin: The same link is 5 lines above ;) Though renaming the section to __Download__ might be a good idea.

Nyco: Ah, misleading then... a "Quickstart guide" should be a guide, right? ;-) So, as a user, I would expect a doc...

ppikula:
merged in #472

PR #472, opened 2015-08-03
Renamed quickstart section, added platorms

Nyco:
A better PR?
The section title was misleading. And the packages to expect are named.


lavrin:
Thanks!

PR #473, opened 2015-08-03
Pointing to experimental Docker image

Nyco:
It would be nice to give some visibility to the Docker image, even though it is still not official, and still experimental. Also a call for contribution is good.

What do you think?


lavrin:
Always good to let the world know :)

PR #474, opened 2015-08-03
try to use new travis infrastructure

ppikula:


ppikula:
I tried to move mim tests to container based infrastructure,but I encountered problem with LDAP. The LDAP setup scripts require sudo to override randomly generated LDAP password. Right now, LDAP is not available as a service and Travis guys said, they are not planning to do this in the near future.
They also gave me a hint how to use their newest infrastructure (see a74599f commit), which is in alpha state and can't even run Erlang ;].  They will announce it on their blog, when it is ready, then we can rerun these tests.

PR #478, opened 2015-08-04
When sessions not cleanly closed, SM_BACKEND sometimes keeps old sess…

aberke:
…ions for which SID process has died. In this case, instead of handling packet as a packet sent to offline user, session manager attempts to send message to dead PID and message goes nowhere.

Fix:
When fetching user PIDs, filter out dead PIDs and delete sessions referring to dead PIDs.
Builds off clean_session_list by not only finding recent user sessions, but then filtering those for only live sessions
Closes all sessions with reason 'replaced' that are not in the list of clean sessions
Uses erlang:process_info/2 to check if session died.  If process from other node then that other node will be responsible for closing that session.

### DIFF
--- a/apps/ejabberd/src/ejabberd_sm.erl
+++ b/apps/ejabberd/src/ejabberd_sm.erl
@@ -671,32 +671,67 @@ route_message(From, To, Packet) ->
 
 -spec clean_session_list([sid()]) -> [sid()].
 clean_session_list(Ss) ->
-    clean_session_list(lists:keysort(#session.usr, Ss), []).
+    ?DEBUG("clean_session_list Ss: ~p",[Ss]),
+    CleanSessions = alive_session_list(recent_session_list(Ss)),
+    BadSessions = lists:filter(fun(S) -> not lists:member(S, CleanSessions) end, Ss),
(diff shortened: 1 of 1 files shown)


aberke:
@pzel Can you please take a look at this?  Thanks!

apps/ejabberd/src/ejabberd_sm.erl
```diff
@@ -671,32 +671,67 @@ route_message(From, To, Packet) ->
...
     end.
 
+alive_session_list(Ss) ->
+    Self = self(),
+    ?DEBUG("~p find alive sessions in session list: ~p with nodes ~p", [Self, Ss, nodes()]),
+    lists:filter(fun(#session{sid={_, Pid}}) when (Pid =:= Self) ->
+        true;
+        (#session{sid={_, Pid}=SID}) ->
+            ?DEBUG("getting process_info for pid ~p", [Pid]),
+            case catch erlang:process_info(Pid, status) of
```
arcusfelis: You can execute process_info for local nodes only with this kind of check `node(Pid) =:= node()`.

PR #490, opened 2015-08-18
Improved cleanup after dead node

michalwski:
This supersedes #52.

Many thanks to @lavrin for some ideas.


apps/ejabberd/test/mongooseim_cleaner_SUITE.erl
```diff
@@ -0,0 +1,46 @@
...
+
+all() ->
+    [sanity_check,
+     cleaner_runs_hook_on_nodedown].
+
+%%
+%% Tests
+%%
+
+sanity_check(_) -> ok.
```
ppikula: ?

lavrin:
Looks good to me. I still don't get why `global:trans` sometimes works with `Retries = 0` and sometimes doesn't... but let's stay at the variant which is known to work.

PR #495, opened 2015-08-27
Refactor ejabberd_c2s:wait_for_stream/2

lavrin:
Local tests and coverage reports show almost complete coverage of the refactored FSM state. Testing against the version just before refactoring shows similar coverage and the same tests pass without modification.

I couldn't come up with a test case which would fall into `wait_for_stream({xmlstreamelement, _}, ...)` clause - if the first element sent on a newly opened socket is not `<stream:stream>` it still gets passed to the function wrapped as `{xmlstreamstart, _, _}`, not `{xmlstreamelement, _}`. See https://github.com/esl/ejabberd_tests/tree/element-before-stream-start for a tried but finally dropped approach. It's probably something lower (even lower than `ejabberd_receiver`, probably XML parser in `xml`).

Ready to merge once Travis passes.

### DIFF
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -240,19 +240,20 @@ wait_for_stream({xmlstreamerror, _}, StateData) ->
 wait_for_stream(closed, StateData) ->
     {stop, normal, StateData}.
 
-handle_stream_start({xmlstreamstart, _Name, Attrs}, StateData) ->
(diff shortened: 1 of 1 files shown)


mongoose-im:
travis is using test branch `refactor-c2s-wait-for-stream` from https://github.com/esl/ejabberd_tests/tree/refactor-c2s-wait-for-stream

michalwski:
Please create a corresponding PR for ejabberd_tests.

PR #497, opened 2015-09-01
Compatibility with OTP 18.0

fenek:
This PR makes MongooseIM compatible with OTP 18.0.

### DIFF
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -1316,7 +1316,7 @@ handle_routed_presence(From, To, Packet = #xmlel{attrs = Attrs}, StateData) ->
             process_presence_probe(From, To, NewState),
             {false, Attrs, NewState};
         <<"error">> ->
-            NewA = remove_element(jlib:jid_tolower(From), State#state.pres_a),
+            NewA = ?SETS:del_element(jlib:jid_tolower(From), State#state.pres_a),
             {true, Attrs, State#state{pres_a = NewA}};
         <<"invisible">> ->
             Attrs1 = lists:keydelete(<<"type">>, 1, Attrs),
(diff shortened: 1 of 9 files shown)


michalwski:
Please add Erlang/OTP 18 to travis jobs, if available already.

mongoose-im:
travis is using test branch `OTP18` from https://github.com/esl/ejabberd_tests/tree/OTP18

michalwski:
@fenek, could you please create a corresponding PR in ejabberd_tests?

fenek:
Rebased with master. Travis, be a good boy!

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -2002,18 +2002,6 @@ presence_broadcast_first(From, StateData, Packet) ->
             StateData#state{pres_a = As}
     end.
 
-
--spec remove_element(E :: 'error' | tuple(),
-                     Set :: gb_set()) -> gb_set().
-remove_element(E, Set) ->
-    case ?SETS:is_element(E, Set) of
```
michalwski: Why do we need to check if the element exists? Currently we use `gb_sets` here and according to doc - if the element doesn't exist, unchanged set is returned.

fenek: We don't, that's why I removed this lines. :)

michalwski: But what's the point? The `gb_sets:del_element/2` will not crash if the removing element is not in the set. I don't think this check is necessary here.
Also `sets` module will cope with this. And if we are to use other set implementation this will be followed by tests.
BTW do we have an explicit test where we remove element from a set, but the element doesn't exist there?

michalwski: Ups, sorry, somehow this was not marked as removed line, or it's just me.... Any way, no more comments on this :)

Issue #494, opened 2015-08-26
Allow mongooseimctl command to support parameters with white spaces


teseo:
I don't know if this is possible at all:

```
add_rosteritem localuser localserver user server 'Name LastName' 'Group Name' B
```

Many thanks!

michalwski:
Closed by #500

PR #500, opened 2015-09-08
commandline arguments with whitespace

Joe-ESL:
When invoking `mongooseimctl` with arguments containing spaces, instead of needing two levels of nested quoting, just use one. See the issue for an example: #494. Personally I think we could read command-line arguments more elegantly, with something like:

```sh
ARGS=""
for ARG in "$@"
do
    if [ "$ARG" = '--' ]
    then
    	break
    else
    	ARGS+='"'$ARG'" '
    fi
done
```

Or using a `case`.

### DIFF
--- a/apps/ejabberd/src/ejabberd_ctl.erl
+++ b/apps/ejabberd/src/ejabberd_ctl.erl
@@ -62,6 +62,8 @@
 -type format_type() :: binary() | string() | char().
 -type cmd() :: {CallString :: string(), Args :: [string()], Desc :: string()}.
 
+-define(ASCII_SPACE_CHARACTER, $\s).
+
 %%-----------------------------
 %% Module
 %%-----------------------------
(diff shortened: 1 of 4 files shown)


mongoose-im:
travis is using test branch `iss494` from https://github.com/esl/ejabberd_tests/tree/iss494

apps/ejabberd/src/ejabberd_ctl.erl
```diff
@@ -62,6 +62,8 @@
 -type format_type() :: binary() | string() | char().
 -type cmd() :: {CallString :: string(), Args :: [string()], Desc :: string()}.
 
+-define(ASCII_SPACE_CHARACTER, $\s).
```
fenek: It's worth using here too: https://github.com/esl/MongooseIM/pull/500/files#diff-5378040ac449fa544bb91213c8dd41e5R394

Joe-ESL: Addressed in 6868962, but made the change for all SPACE literals, just for consistency.

PR #501, opened 2015-09-08
[skip ci] Fix some spec errors

lavrin:
No functional changes, hence no Travis.

### DIFF
--- a/apps/ejabberd/src/gen_mod.erl
+++ b/apps/ejabberd/src/gen_mod.erl
@@ -50,9 +50,9 @@
 
 -include("ejabberd.hrl").
 
--record(ejabberd_module, {module_host :: ejabberd:server(),
-                          opts :: list()
-                         }).
+-record(ejabberd_module, {module_host, opts}).
+-type ejabberd_module() :: #ejabberd_module{module_host :: {module(), ejabberd:server()},
+                                            opts :: list()}.
 
 %% -export([behaviour_info/1]).
 %% behaviour_info(callbacks) ->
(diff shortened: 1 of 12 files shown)


apps/ejabberd/src/gen_mod.erl
```diff
@@ -50,9 +50,9 @@
 
 -include("ejabberd.hrl").
 
--record(ejabberd_module, {module_host :: ejabberd:server(),
-                          opts :: list()
-                         }).
+-record(ejabberd_module, {module_host, opts}).
+-type ejabberd_module() :: #ejabberd_module{module_host :: {module(), ejabberd:server()},
```
fenek: Why have you moved field type specs to new type definition from record definition?

lavrin: Because of match spec construction a dozen of lines down, i.e. this:

```erlang
gen_mod.erl:325: Record construction #ejabberd_module{module_host::{'$1',_},opts::'_'} violates the declared type of field opts::'undefined' | [any()]
```

I admit it's not ideal, because now one might actually construct an invalid record and Dialyzer won't complain. On the other hand, if an input parameter towards a function is specced properly (that is with `ejabberd_module()`, not `#ejabberd_module{}`) then Dialyzer will complain about passing around such invalid records.

fenek: Fair enough. Although I used to add `| atom()` to such record definitions with comment "it's for match spec".

fenek:
:+1:

michalwski:
@fenek how this PR connects to what you've done in #508. Is it safe to merge this PR before yours?

fenek:
At first glance some changes are exactly the same and some are not. I will compare them on Monday (I really need tiny break from Dialyzer :)).

fenek:
Merged into #508

Issue #504, opened 2015-09-10
mochijson2: Missing xmerl otp lib


jonathanve:
Hello, Guys

When installing mongooseim, I get this from the erl shell:

$  mongooseimctl debug
$ (mongooseim@localhost)1> mochijson2:decode_test().
** exception error: undefined function xmerl_ucs:from_utf16be/1
     in function  mochijson2:tokenize_string/3 (src/mochijson2.erl, line 400)
     in call from mochijson2:decode1/2 (src/mochijson2.erl, line 277)
     in call from mochijson2:json_decode/2 (src/mochijson2.erl, line 272)
     in call from mochijson2:decode_test/0 (src/mochijson2.erl, line 566)

At /usr/lib/mongooseim/lib/ there is no xmerl installed.

jonathanve:
I solved it by including the official otp module and added to the ebin/ dir where mongooseim reads it
https://raw.githubusercontent.com/erlang/otp/maint/lib/xmerl/src/xmerl_ucs.erl

michalwski:
What Erlang version do you use?
Why exactly did you try `mochijson2:decode_test()` function? Does decoding doesn't work in general?

jonathanve:
Hello. I am using Erlang 17.5. I am writing a custom module hook and while using mochijson2:encode, I got the error. Missing xmerl_ucs. Decoding and encoding json parameters that include utf8 characters is not working. It seems mongooseim custom erlang does not have xmerl included.

ppikula:
Hey, instead of manual copying you can  add `xmerl` here: https://github.com/esl/MongooseIM/blob/master/rel/reltool.config.script#L8 and then rerun `make rel`, this should include whole xmerl application in the release.

jonathanve:
Thanks, that's really helpful @ppikula . I used https://github.com/ppikula/mongooseim-docker and I created a custom dockerfile using it. I will have to create a custom mongooseim docker container with the change, which it's fine.

jonathanve:
@ppikula should xmerl be included in mongooseim stable branch?

ppikula:
Yeah, I think so. As far as I can see, the xmerl is used in regular functions like `json_encode_string` not only in the mochijson2' tests.  

Do you want to make a pull request?

jonathanve:
PR created:
https://github.com/esl/MongooseIM/pull/506

michalwski:
Fixed by #506

PR #506, opened 2015-09-11
add xmerl otp lib for json encode/decode functions

jonathanve:
https://github.com/esl/MongooseIM/issues/504


michalwski:
`mochijson2` is not the only problem caused by missing `xmerl`. There are also some places in the code where `xmerl_regexp:sh_to_awk` is called.

Thanks!

PR #508, opened 2015-09-14
Fixes based on Dialyzer reports

fenek:
Goes on top of #497


fenek:
PR is very large but each commit is pretty atomic so the review can be performed on per-commit basis.

apps/ejabberd/include/ejabberd.hrl
```diff
@@ -60,11 +55,11 @@
 -define(CRITICAL_MSG(Format, Args),
     lager:critical(Format, Args)).
 
--record(session, {sid      :: ejabberd_sm:sid(),
-                  usr      :: ejabberd:simple_jid(),
-                  us       :: ejabberd:simple_bare_jid(),
-                  priority :: integer(),
-                  info     :: list()
+-record(session, {sid,
```
michalwski: Why did you remove record field types?

apps/ejabberd/include/ejabberd.hrl
```diff
@@ -18,8 +18,6 @@
 %%% 02111-1307 USA
 %%%
 %%%----------------------------------------------------------------------
--ifndef(MONGOOSEIM_EJABBERD_HRL).
```
michalwski: Wasn't such a line useful? We include ejaberd.hrl from almost every file. Maybe I don't understand the prev mechanism correctly, but to mean it meant - do not define the macros if they are already defined.

apps/ejabberd/include/mod_muc_room.hrl
```diff
@@ -88,7 +88,15 @@
...
-                          }).
+-record(muc_online_users, {
+          us,
+          room,
+          host
+         }).
+
+-type muc_online_users() :: #muc_online_users{
+        us :: ejabberd:simple_bare_jid(),
+        room :: mod_muc:room(),
```
michalwski: Why is it better to define field type here and not where the record is defined?

apps/ejabberd/src/cyrsasl.erl
```diff
@@ -36,15 +36,12 @@
 
 -include("ejabberd.hrl").
 
--record(sasl_mechanism, {mechanism :: mechanism(),
+-record(sasl_mechanism, {mechanism :: mechanism() | atom(), % atom() used in ets:select
```
michalwski: Can any atom be used in the ets:select or only some well defined like '_' f.e. ?

apps/ejabberd/src/cyrsasl.erl
```diff
@@ -36,15 +36,12 @@
...
                          module :: sasl_module(),
-                         password_type :: plain | digest | scram
+                         password_type :: plain | digest | scram | atom() % atom() used in ets:select
                         }).
--type sasl_module() :: cyrsasl_anonymous
-                     | cyrsasl_digest
-                     | cyrsasl_plain.
+-type sasl_module() :: atom().
```
michalwski: I think possible sasl modules are actually well defined. What's the reason behind using type `atom` here?

apps/ejabberd/src/ejabberd.app.src
```diff
@@ -2,7 +2,7 @@
 
 {application, ejabberd,
  [{description, "ejabberd"},
-  {vsn, {cmd, "echo 2.1.8+mim-`git describe --always --tags`"}},
+  {vsn, {cmd, "echo 2.1.8-mim-`git describe --always --tags`"}},
```
michalwski: Why `-` is better than `+`. As far as I remember we introduced `+` to separate ejabberd base version from mongooseim version. I don't mind `-` but just wanted to know why we'd like to introduce this change.

apps/ejabberd/src/ejabberd_auth.erl
```diff
@@ -62,11 +62,7 @@
 
 -export_type([authmodule/0]).
 
--type authmodule() :: ejabberd_auth_anonymous
-                    | ejabberd_auth_external
-                    | ejabberd_auth_internal
-                    | ejabberd_auth_ldap
-                    | ejabberd_auth_odbc.
+-type authmodule() :: atom().
```
michalwski: Similar question to the one regarding sasl modules.

apps/ejabberd/src/ejabberd_auth_internal.erl
```diff
@@ -56,11 +56,11 @@
 
 -include("ejabberd.hrl").
 
--record(passwd, {us :: ejabberd:simple_bare_jid(),
-                 password :: binary()
+-record(passwd, {us :: ejabberd:simple_bare_jid() | atom(), % atom() for mnesia:dirty_select
```
michalwski: Is it possible to specify the atom(s) used in `mnesia:drity_select` ?

lavrin:
The trick for defining record types which can also be used as match specs (for `ets:select`) is this:

```erlang
-record(r, {field_a,
            field_b}).
-type r() :: #record{field_a :: atom(),
                     field_b :: binary()}.
```

Then:

1. We don't have to do cumbersome `field_x :: actual_type() | '_'` to properly handle building match specs.
2. Dialyzer doesn't complain about `#r{field_a = '_'}`, so we lose some type safety, but don't get false positives when building match specs.
3. Dialyzer still complains when passing `r() / #r{}` instances to properly specced functions (i.e. functions have to be specced with `-spec f( r() ) -> r().` not the record type `#r{}`). In other words, we can build an invalid record instance, but we'll get warnings as soon as we start passing it around.

apps/ejabberd/include/ejabberd.hrl
fenek: This record is used for selecting from Mnesia, so some fields can be set to `'_'` or `'$X'` then. Precise specs are moved here:
https://github.com/esl/MongooseIM/pull/508/files#diff-67c8b548efe99cc4b398cdbd4e7734e0R74

apps/ejabberd/include/ejabberd.hrl
fenek: We don't have any header files including `ejabberd.hrl`, so when it's used, then it's included by `.erl` file. And including it twice would indicade a mistake or sloppy coding. :)
I think the original reason behind this were habits from C world.

apps/ejabberd/include/mod_muc_room.hrl
fenek: Like in case of session record - precise specs directly in the record make Dialyzer unhappy about e.g. `dirty_select`

apps/ejabberd/src/cyrsasl.erl
fenek: I guess just for this module it can be specified as `mechanism() | '$1'`.

apps/ejabberd/src/cyrsasl.erl
fenek: Because the ideal situation would be to have this module completely agnostic of available SASL methods and using `atom()` (OK, maybe `module()` would be a better description) is a first step towards this direction. Take a look - this list lacked `cyrsasl_scram` and some new mech can appear sooner or later.

apps/ejabberd/src/cyrsasl.erl
michalwski: Can we use the same trick you did for session?

apps/ejabberd/src/cyrsasl.erl
michalwski: I know the `cyrsasl_scram` was missing but only because we didn't care about specs before. In my  opinion it would be better to have all the possible cyrsasl modules defined in the type. But maybe some else has other opinion on this. Anyone? @lavrin? @pzel?

apps/ejabberd/include/mod_muc_room.hrl
michalwski: Understood :)

apps/ejabberd/include/ejabberd.hrl
michalwski: Got it :)

apps/ejabberd/src/ejabberd.app.src
fenek: I'm absolutely sure it caused some strange error but I can't reproduce it. Maybe it got fixed by accident by one of the subsequent commits. :) I'll revert the change.

apps/ejabberd/src/ejabberd_auth.erl
fenek: Similar answer - I believe that this module should not care about the list of available backends. It's generic API for auth. `ejabberd_auth_http` was missing, Riak one was missing too. More may come in the future and the truth is - Dialyzer can't really verify it anyway since it comes from the config.

fenek:
@lavrin 
The truth is, often the record is passed to Mnesia instantly after being built. Some refactoring would be required to wrap Mnesia calls and spec the wrapping function. But well, this PR above else focused on getting rid of Dialyzer errors.

apps/ejabberd/src/ejabberd_auth_internal.erl
fenek: I will move all the specs to type definitions, wherever record is used for select.

michalwski: Great! A short comment in the code explaining why we do this will be helpful in future :)

apps/ejabberd/src/cyrsasl.erl
fenek: Yup, will do the same thing for every record.

apps/ejabberd/src/ejabberd_auth.erl
michalwski: Ok, I see. Anyway, `module()` type would be better here, probably :)

apps/ejabberd/src/ejabberd_sm.erl
```diff
@@ -202,7 +215,8 @@ get_user_resources(User, Server) ->
 
 -spec get_session_ip(User :: ejabberd:user(),
```
lavrin: I would opt for the following syntax in situations where the spec/return type is so long that it has to be wrapped on arrow:

```erlang
-spec get_session_ip(User, Server, Resource) -> Result when
      User :: ejabberd:user(),
      Server :: ejabberd:server(),
      Resource :: ejabberd:resource()
      Result :: undefined | {inet:ip_address(), integer()}.
```

This way we have plenty of space for each parameter type and readability of the `-> some_return_type()` is not degraded.
For multiple alternatives which don't make it into a single line we can then break on `|` like this:

```erlang
      Result :: undefined
              | {inet:ip_address(), integer()}.
```

michalwski: +1

apps/ejabberd/src/cyrsasl.erl
lavrin: I agree with @fenek. This is basically an Erlang equivalent of OO-style subtyping, i.e. in such a case we're striving for possibly infinite extensibility (with regard to number of implementations - the interface of each implementation is frozen). We must have a corresponding behaviour describing the set of callbacks expected from a `sasl_module()`, though. It should be `sasl_module() :: module()`.

lavrin:
@fenek
> The truth is, often the record is passed to Mnesia instantly after being built.
With the backend infrastructure we have in place this should seldom (never?) be the case.
For the (I hope) few cases where we still use Mnesia directly, not through a DB-agnostic backend interface, I think we can live with it for some time, but ideally such cases would be eventually refactored.

apps/ejabberd/src/ejabberd_sm.erl
fenek: For me the syntax with `when` is less readable but it's just personal preference. If nobody agrees with me, then I'll change it wherever I spot it during last fixes.

apps/ejabberd/include/adhoc.hrl
```diff
@@ -33,10 +33,10 @@
                          node :: ejabberd:server(),
                          sessionid :: binary(),
                          status :: atom(),
-                         defaultaction = "" :: string(),
+                         defaultaction = <<>> :: binary(),
```
ppikula: By the way, can you rename it to `dafault_action` ? Probably, the same story couple lines above with `sessionid`

fenek: Sure, didn't think about it.

PR #513, opened 2015-09-17
Removed obsolete namespace from mediated invitation stanza

### LINKED ISSUE #223: Obsolete mediated invitation type?
I'm not sure but it seems that `<x xmlns="jabber:x:conference"/>` (see [mod_muc_room.erl](https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mod_muc_room.erl#L3955-L3958)) is obsolete, probably a leftover from pre-XMPP Groupchat 1.0 specification.
XEP-0045 defines mediated invitation as containing `<x xmlns="http://jabber.org/protocol/muc#user"/>` which is also included in the message sent by MongooseIM.
Maybe we should remove the obsolete element? Unless some client apps rely on it...

ludwikbukowski:
Fixes #223

### DIFF
--- a/apps/ejabberd/src/mod_muc_room.erl
+++ b/apps/ejabberd/src/mod_muc_room.erl
@@ -3990,7 +3990,7 @@ unsave_check_invitation(FromJID, Els, Lang,
                   PasswdEl = create_password_elem(StateData),
                   BodyEl = invite_body_elem(FromJID, Reason, Lang, StateData),
                   Msg = create_invite_message_elem(
-                          OutInviteEl, BodyEl, PasswdEl, RoomJID, Reason),
+                          OutInviteEl, BodyEl, PasswdEl, Reason),
                   ejabberd_hooks:run(invitation_sent, Host,
                                      [Host, ServerHost, RoomJID, FromJID, JID, Reason]),
                   ejabberd_router:route(StateData#state.jid, JID, Msg)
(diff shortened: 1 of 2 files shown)


apps/ejabberd/src/mod_muc_room.erl
```diff
@@ -4079,23 +4079,18 @@ invite_body_text(FromJID, Reason, Lang,
...
+create_invite_message_elem(InviteEl, BodyEl, PasswdEl, Reason)
     when is_list(PasswdEl), is_binary(Reason) ->
-    BRoomJID = jlib:jid_to_binary(RoomJID),
     UserXEl = #xmlel{
         name = <<"x">>,
         attrs = [{<<"xmlns">>, ?NS_MUC_USER}],
         children = [InviteEl|PasswdEl]},
-    ConfXEl = #xmlel{
-        name = <<"x">>,
-        attrs = [{<<"xmlns">>, ?NS_XCONFERENCE}, {<<"jid">>, BRoomJID}],
```
michalwski: Please remove also the macro `?NS_XCONFERENCE`. It's defined in jlib.hrl.

ludwikbukowski: I was thinking about it but I found out that this namespace is used in XEP-0249: 'Direct MUC Invitations' which could be implemented in future. I can remove it anyway

PR #517, opened 2015-09-21
Database docs fix

ludwikbukowski:
Description of database configuration in docs was in two seperated locations, so I made them reffering on each other.

### DIFF
--- a/doc/advanced-configuration/database-backends-configuration.md
+++ b/doc/advanced-configuration/database-backends-configuration.md
@@ -28,7 +28,7 @@ mysql -h localhost -u user -p -e 'create database mongooseim'
 mysql -h localhost -u user -p mongooseim < mysql.sql
 ```
 
-You should also configure MySQL database in `[MongooseIM root]/rel/files/ejabberd.cfg` file. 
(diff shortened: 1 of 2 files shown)


michalwski:
Hint for future. Please use `[skip ci]` in commit message while you update doc only. There is no need to trigger travis in this case.

michalwski:
I canceled the travis build as it's not required for this PR.

doc/advanced-configuration/database-backends-configuration.md
```diff
@@ -28,6 +28,9 @@ mysql -h localhost -u user -p -e 'create database mongooseim'
 mysql -h localhost -u user -p mongooseim < mysql.sql
 ```
 
+You should also configure MySQL database in `[MongooseIM root]/rel/files/ejabberd.cfg` file.
```
michalwski: I think it's enough to say "`ejabberd.cfg` file" instead of the path. The thing is that the ejabberd.cfg file described in `Advanced-configuration.md` is a template for the final config file. This, I think, should be stated in the `Advanced configuration`.

Same applies to next 2 changes.

PR #527, opened 2015-09-28
Add xmerl otp lib for json encode/decode functions

michalwski:
This is #506 but to master branch. As discovered today xmerl (and friends) is used in some other places, not only mochijson2.


michalwski:
It may look like it's my own PR but it's not :) all the job was done by @jonathanve. I only applied it on top of master branch.

jonathanve:
good to know is now fixed @michalwski

PR #530, opened 2015-09-29
Change http api cfg

michalwski:
This PR changes the default HTTP API endpoint so there is no risk that someone accidentally exposes the API to everyone.

### DIFF
--- a/rel/files/ejabberd.cfg
+++ b/rel/files/ejabberd.cfg
@@ -137,14 +137,11 @@
 %%
 {listen,
  [
-  %% HTTP endpoints
+  %% BOSH and WS endpoints over HTTP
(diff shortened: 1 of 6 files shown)


rel/files/ejabberd.cfg
```diff
@@ -173,6 +166,19 @@
         ]}
     ]},
 
+  %% MongooseIM HTTP API it's important to start it on localhost
+  %% or some private interface only (not accessible from the outside)
+  %% At least start it on different port which will be hidden behind firewall
+
+  { {{metrics_http_endpoint}} , ejabberd_cowboy, [
```
ppikula: maybe just `http_endpoint`/`rest_endpoint` ? It is not only about metrics endpoint.

PR #531, opened 2015-09-30
mod_ping tests

michalwski:
This PR:
* adds test for mod_ping
* makes iq reply timeout configurable in mod_ping
* waits for the ping process to stop before removing it from supervisor


michalwski:
> can you use commit hash here?
I'll do that as soon as the corresponding escalus PR is merged.

PR #532, opened 2015-09-30
Replacing the Mochijson2 usage with JSX

wende:
All REST API mochijson2 calls replaced with JSX, which embraces usage of utf8 and utf16 special characters.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -19,6 +19,7 @@
   {cowboy, ".*", {git, "git://github.com/ninenines/cowboy.git", "1.0.1"}},
   {folsom, ".*", {git, "git://github.com/boundary/folsom.git", "0.8.2"}},
   {exometer, ".*", {git, "git://github.com/Feuerlabs/exometer.git", "1.1"}},
+  {jsx, ".*", {git, "git@github.com:talentdeficit/jsx.git", "2.7.2"}},
   {mochijson2, ".*", {git, "git://github.com/bjnortier/mochijson2.git", {branch, "master"}}},
   {alarms, ".*", {git, "git://github.com/chrzaszcz/alarms.git", {branch, "master"}}},
   {fusco, ".*", {git, "git://github.com/esl/fusco.git", {branch, "master"}}},
--- a/rebar.config
+++ b/rebar.config
@@ -19,7 +19,7 @@
   {cowboy, ".*", {git, "git://github.com/ninenines/cowboy.git", "1.0.1"}},
   {folsom, ".*", {git, "git://github.com/boundary/folsom.git", "0.8.2"}},
   {exometer, ".*", {git, "git://github.com/Feuerlabs/exometer.git", "1.1"}},
-  {jsx, ".*", {git, "git@github.com:talentdeficit/jsx.git", "2.7.2"}},
+  {jsx, ".*", {git, "https://github.com/talentdeficit/jsx", "2.7.2"}},
   %{mochijson2, ".*", {git, "git://github.com/bjnortier/mochijson2.git", {branch, "master"}}},
   {alarms, ".*", {git, "git://github.com/chrzaszcz/alarms.git", {branch, "master"}}},
   {fusco, ".*", {git, "git://github.com/esl/fusco.git", {branch, "master"}}},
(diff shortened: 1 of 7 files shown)


lavrin:
Why `jsx`? I've very recently compared different JSON libs available - `jiffy` was the winner in my eyes. I'm not saying it is in general, though, I'm just curious.

ppikula:
Why jsx? I was not involved in this issue, but I heard that jiffy library is the fastest. Is there any reasoning behind this choice?

Why do we need the mochi2jsx module? I mean, why there is another layer that only converts from mochi format to the jsx one, especially if we want to get rid off mochi. In this case I would rather fix the `prepare_struct` function and friends to make it work with the jsx format. That is my opinion, maybe others will disagree. 

Another possibility is to introduce "our" JSON representation(wrap the whole struct around record like: #json{struct={...}}, this will help us to distinguish between 2 formats), but I don't think it priority right now and this is not well though :). It would be good to use this approach in the context  of escaping data like:
`odbc_escape(Value) -> #escaped_odbc{value="something"}` then all functions operating on odbc expects this record and deny raw values.

wende:
@ppikula  I thought it's better not to reinvent the whole prepare_struct code just to match the new format of JSX especially that the formats don't differ that much.
Right now it's totally Mochijson2 independent. The only thing that is left is mochijson's structure.
The dependency can be safely removed.

Usage of `mochi2jsx` module is for making the encoding library-agnostic. It's a facade for a case when the JSON lib (JSX) would have to be changed again. 

It's hard for me to tell why JSX because it's not me who made the decision, but as far as I'm concerned JSX is the only library that can handle UTF16

Jiffy:
> A quick note on unicode. Jiffy only understands UTF-8 in binaries. End of story.

Mochi:
> Yet another JSON (RFC 4627) library for Erlang. mochijson2 works
>  with binaries as strings, arrays as lists (without an {array, _})
>  wrapper and it only knows how to decode UTF-8 (and ASCII).

ppikula:
If it can be safely removed, please do. 

Another question: why the mochi format is better that JSX? IMO the second one is more compact. 
 
```
% JSX:      {struct, [ <<"a">> , [[{ <<"b">> ,10 }],2] ] }
% Mochi:   {struct, [{ <<"a">>, [{struct,[{<<"b">>,10}]},2]}]}
```
I know that it requires some work, but we don't have a strict deadline. We can do it properly, and not introduce another layer of code. Moreover, we waste some CPU cycles converting something that could have been in the right format from the beginning.

lavrin:
I'm not aware of any requirement for UTF-16 - our XML parser and all code around use UTF-8. Performance-wise (based on results from https://github.com/si14/erl_json_test.git) `jiffy` and `jsonx` win, though we already pull in `jsx` (which is still better than `mochijson2`) due to some other dependency.
:+1: for @ppikula - a facade / explicit converter to JSON-consumable format is nice. There's no point in converting to Mochijson2 format and only then to the format of the library we're ultimately going to use, though.

rebar.config
```diff
@@ -19,7 +19,8 @@
   {cowboy, ".*", {git, "git://github.com/ninenines/cowboy.git", "1.0.1"}},
   {folsom, ".*", {git, "git://github.com/boundary/folsom.git", "0.8.2"}},
   {exometer, ".*", {git, "git://github.com/Feuerlabs/exometer.git", "1.1"}},
-  {mochijson2, ".*", {git, "git://github.com/bjnortier/mochijson2.git", {branch, "master"}}},
+  {jsx, ".*", {git, "https://github.com/talentdeficit/jsx", "2.7.2"}},
+  %{mochijson2, ".*", {git, "git://github.com/bjnortier/mochijson2.git", {branch, "master"}}},
```
ppikula: Just remove it :)  no need to keep it.

rel/reltool.config.script
```diff
@@ -36,6 +36,7 @@ BaseAppsToInclude = AppsToRun ++
                      ranch,
                      cowlib,
                      bear,
+                     jsx,
                      mochijson2,
```
ppikula: I guess, It has to be removed from here too. Travis build is failing, probably this is the reason.

michalwski:
I'm closing this PR as this is not finished and there is no chance to finish this in near feature.

PR #535, opened 2015-10-02
Bosh GET method

michalwski:
The GET method was implement only for Load Balancer's health checks. For example ELB uses this method.

### DIFF
--- a/apps/ejabberd/src/mod_bosh.erl
+++ b/apps/ejabberd/src/mod_bosh.erl
@@ -169,6 +169,8 @@ init(_Transport, Req, _Opts) ->
             <<"POST">> ->
                 {has_body, true} = {has_body, cowboy_req:has_body(Req2)},
                 {forward_body, Req2};
+            <<"GET">> ->
+                {accept_get, Req2};
             _ ->
                 error({badmatch, {Method, Req2}})
         end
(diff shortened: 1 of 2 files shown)


test/ejabberd_tests/tests/bosh_SUITE.erl
```diff
@@ -59,9 +59,13 @@ suite() ->
     escalus:suite().
 
 essential_test_cases() ->
-    [create_and_terminate_session,
-     accept_higher_hold_value,
-     do_not_accept_0_hold_value].
+    [%create_and_terminate_session,
```
lavrin: This _seems_ to be accidental... Should we actually comment them out?

michalwski: Correct, it's accidental.

apps/ejabberd/src/mod_bosh.erl
```diff
@@ -190,6 +192,14 @@ info(accept_options, Req, State) ->
     Headers = ac_all(Origin),
     ?DEBUG("OPTIONS response: ~p~n", [Headers]),
     {ok, strip_ok(cowboy_req:reply(200, Headers, <<>>, Req2)), State};
+info(accept_get, Req, State) ->
```
fenek: `accept_get` is missing in `info()` type.

michalwski: good catch :)

PR #536, opened 2015-10-02
Keep alive hook

michalwski:
This PR introduces keep alive hook which is run when the client sends XMPP whitespace keep alive packets. This can be useful in mod_ping if we want to reset the timer after such packets.

### DIFF
--- a/apps/ejabberd/src/ejabberd_receiver.erl
+++ b/apps/ejabberd/src/ejabberd_receiver.erl
@@ -344,7 +344,7 @@ process_data(Data,
     State#state{xml_stream_state = XMLStreamState1,
                 shaper_state = NewShaperState}.
 
-maybe_pause(_, #state{c2s_pid = undefinde}) ->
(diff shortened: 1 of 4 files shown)


apps/ejabberd/src/ejabberd_receiver.erl
```diff
@@ -333,24 +333,32 @@ process_data([Element|Els], #state{c2s_pid = C2SPid} = State)
...
-        Pause > 0 ->
-            erlang:start_timer(Pause, self(), activate);
-        true ->
-            activate_socket(State)
-    end,
+    maybe_pause(Pause, State),
     State#state{xml_stream_state = XMLStreamState1,
                 shaper_state = NewShaperState}.
 
+maybe_pause(_, #state{c2s_pid = undefinde}) ->
```
fenek: Typo. Should be `undefined`

michalwski: :+1:

Issue #242, opened 2014-07-10
Negotiation for Stream Compression ignored by server after SASL


climvalencia:
According to [XEP-0170](http://xmpp.org/extensions/xep-0170.html#c2s-compress), Stream Compression may be negotiated after SASL:

1. TLS
2. SASL
3. Stream Compression
4. Resource Binding

(as in Smack XMPP client library)

However ejabberd_c2s only supports:

1. TLS
2. Stream Compression
3. SASL
4. Resource Binding

(as in Psi+ v0.15)

Sending a `<compress/>` packet after SASL results in it being ignored by the server.

Also in [XEP-0138](http://xmpp.org/extensions/xep-0138.html), Stream Compression should be negotiated after SASL.

michalwski:
Implemented in #616 and merged to master branch.

PR #546, opened 2015-10-12
Xep tool

ludwikbukowski:
Tool for fully-automatic production of available XEPs.


apps/ejabberd/src/ejabberd_c2s.hrl
```diff
@@ -1,6 +1,6 @@
 -include_lib("ejabberd/include/mod_privacy.hrl").
 -define(GEN_FSM, p1_fsm_old).
-
+-xep([{xep, 198}, {version, "1.3"}]).
```
studzien: Why in the .hrl file while other c2s XEP attributes are defined in the .erl file?

ludwikbukowski: by mistake... thanks @studzien !

michalwski:
@ludwikbukowski could you please add some doc how to use your tool? How about adding and new target to Makefile that will run your script?

apps/ejabberd/src/mod_amp.erl
```diff
@@ -6,7 +6,7 @@
 %% This work was sponsored by Grindr LLC
 
 -behavior(gen_mod).
-
+-xep([{xep, 79}, {vesion, "1.2"}, {comment, "partial implemented."}]).
```
Dzol: Should be "partially implemented" or "partial implementation".

tools/xep_tool/xep_tool.escript
```diff
@@ -0,0 +1,133 @@
...
+    F = fun(_, {Name, Url}, {Num, BuildingTable}) ->
+        Add = case Num rem 4 of
+                  0 ->
+                      "\n";
+                  _ ->
+                      " "
+              end,
+        {Num + 1, BuildingTable ++ "[" ++ Name ++ "](" ++ Url ++ ") |" ++ Add}
+    end,
+    {_, TableListElement} = maps:fold(F, {1, ""}, Map),
```
studzien: What's the decision behind choosing maps here?

ludwikbukowski: I'm sure I can replace it with `dict`.

studzien:
My two (ok, three) cents:
* I've noticed that the script is making a HTTP request for the same XEP several times if it occurs in more than one module. Maybe it would be a good idea to pick unique values first :)
* If proper permission are set for the ``xep_tool.escript`` it might be executed directly, without calling ``escript``. Personally, I find it more convenient.
* Maybe it would be a good idea to print a standard "usage" line for the script if it's called without any arguments (or with too many), so it's self-explanatory.

tools/xep_tool/xep_tool.escript
studzien: So, what the idea behind ``dict`` then? :)

tools/xep_tool/xep_tool.escript
```diff
@@ -0,0 +1,175 @@
...
+-record(xep, {xep :: xep(), name :: name(), url :: url(), version :: ver()}).
+-record(module_xep, {xeps = [], module :: module()}).
+
+-spec get_xep(module()) -> {module(), list({atom(), any()})}.
+get_xep(Module) ->
+    Info = Module:module_info(attributes),
+    Values = [Value || {xep, Value} <- Info],
+    {Module, Values}.
+
+-spec match_zeros(xep()) -> string().
```
studzien: after changing the returned value to iolist now specs don't match, perhaps also in some other functions

tools/xep_tool/xep_tool.escript
```diff
@@ -0,0 +1,175 @@
...
+    do(?FILE_NAME);
+
+main([InDir, Output]) ->
+    code:set_path(code:get_path() ++ [InDir]),
+    do(Output);
+
+main(_) ->
+    io:format("~n ***  Bad Arguments. "
+              "Try:~nxep_tool.escript <path_to_ebin_files>"
+              "[output_file]~n~nExample usage (from mongooseim repo root):~n"
```
studzien: This prints:
```
 ***  Bad Arguments. Try:
xep_tool.escript <path_to_ebin_files>[output_file]
```
so a space is missing between arguments

apps/ejabberd/src/ejabberd_s2s_in.erl
```diff
@@ -26,7 +26,8 @@
 
 -module(ejabberd_s2s_in).
 -author('alexey@process-one.net').
-
+-xep([{xep, 175}, {version, "1.2"}]).
+-xep([{xep, 198}, {version, "1.3"}]).
```
studzien: I think we don't support the Stream Management here

apps/ejabberd/src/ejabberd_s2s_out.erl
```diff
@@ -26,7 +26,8 @@
 
 -module(ejabberd_s2s_out).
 -author('alexey@process-one.net').
-
+-xep([{xep, 175}, {version, "1.2"}]).
+-xep([{xep, 198}, {version, "1.3"}]).
```
studzien: I think we don't support the Stream Management here

apps/ejabberd/src/ejabberd_service.erl
```diff
@@ -26,7 +26,8 @@
 
 -module(ejabberd_service).
 -author('alexey@process-one.net').
-
+-xep([{xep, 114}, {version, "1.6"}]).
+-xep([{xep, 198}, {version, "1.3"}]).
```
studzien: I think we don't support the Stream Management here

apps/ejabberd/src/mod_amp.erl
```diff
@@ -6,7 +6,7 @@
 %% This work was sponsored by Grindr LLC
 
 -behavior(gen_mod).
-
+-xep([{xep, 79}, {vesion, "1.2"}, {comment, "partially implemented."}]).
```
studzien: a typo in "vesion"

apps/ejabberd/src/mod_offline.erl
```diff
@@ -27,7 +27,11 @@
 
 -module(mod_offline).
 -author('alexey@process-one.net').
-
+-xep([{xep, 160}, {version, "1.0"}]).
+-xep([{xep, 23}, {version, "1.3"}]).
+-xep([{xep, 22}, {version, "1.4"}]).
+-xep([{xep, 85}, {version, "2.1"}]).
+-xep([{xep, 170}, {version, "1.0"}]).
```
studzien: I'm afraid we don't support XEP-0170 at all, see #242

apps/ejabberd/src/mod_vcard.erl
```diff
@@ -33,7 +33,10 @@
 
 -module(mod_vcard).
 -author('alexey@process-one.net').
-
+-xep([{xep, 54}, {version, "1.2"}]).
+-xep([{xep, 55}, {version, "1.3"}]).
+-xep([{xep, 68}, {version, "1.2"}]).
+-xep([{xep, 82}, {version, "1.1"}]).
```
studzien: How does vcard implement/is related to XEP-0082?

studzien:
Generally speaking I'd rather avoid pushing XEP tags wherever we can.
Especially, I think we should include them only in the main feature's module and leave the backends alone.
I believe that in a perfect world we would only use one XEP for a set of modules related to a feature (maybe except informational XEPs).
What do you think?

apps/ejabberd/src/jlib.erl
```diff
@@ -26,7 +26,19 @@
...
 -module(jlib).
 -author('alexey@process-one.net').
-
+-xep([{xep, 59}, {version, "1.0"}]).
+-xep([{xep, 68}, {version, "1.2"}]).
+-xep([{xep, 18}, {version, "0.2"}]).
+-xep([{xep, 175}, {version, "1.2"}]).
+-xep([{xep, 157}, {version, "1.0"}]).
+-xep([{xep, 212}, {version, "1.0"}]).
+-xep([{xep, 114}, {version, "1.6"}]).
```
studzien: I think this should go to `ejabberd_service`

apps/ejabberd/src/jlib.erl
```diff
@@ -26,7 +26,19 @@
 
 -module(jlib).
 -author('alexey@process-one.net').
-
+-xep([{xep, 59}, {version, "1.0"}]).
+-xep([{xep, 68}, {version, "1.2"}]).
+-xep([{xep, 18}, {version, "0.2"}]).
```
studzien: I think this should go to `ejabberd_c2s`

apps/ejabberd/src/jlib.erl
```diff
@@ -26,7 +26,19 @@
...
+-xep([{xep, 175}, {version, "1.2"}]).
```
studzien: I think this should go to `cyrsasl_anonymous`

apps/ejabberd/src/jlib.erl
```diff
@@ -26,7 +26,19 @@
...
+-xep([{xep, 157}, {version, "1.0"}]).
```
studzien: I think this should go to `mod_disco`

apps/ejabberd/src/jlib.erl
```diff
@@ -26,7 +26,19 @@
...
+-xep([{xep, 212}, {version, "1.0"}]).
+-xep([{xep, 114}, {version, "1.6"}]).
+-xep([{xep, 85}, {version, "2.1"}]).
```
studzien: That's a client-side extension, the only server-side module that is aware of it is `mod_offline`, but I'm not sure if it should land there or in a "static" list of supported XEPs.

apps/ejabberd/src/jlib.erl
```diff
@@ -26,7 +26,19 @@
...
+-xep([{xep, 126}, {version, "1.1"}]).
```
studzien: I think this should go to `mod_privacy`

apps/ejabberd/src/jlib.erl
```diff
@@ -26,7 +26,19 @@
...
+-xep([{xep, 170}, {version, "1.0"}]).
```
studzien: I don't think we really support this one atm (see #242)

lavrin:
This PR starts to look like a never-ending story ;) Maybe it's time to merge it if the tool itself is OK and then carry on with annotating the modules with subsequent PRs?

PR #553, opened 2015-10-15
set ssl session_lifetime to 10 minutes

ppikula:
I  found a memory "leak" while load testing, the increased memory usage was caused by TLS processes.
The processes were started by fusco, through which the external API was called. 

I found the solution here: https://github.com/ninenines/gun/issues/68, although I don't know if 10 minutes is a sane value. I will remove the do-no-merge label after testing this value - especially, I need to check if it influence server side(in our case cowboy connections).

More info about the bug: https://support.process-one.net/browse/TSUN-225

### DIFF
--- a/doc/Advanced-configuration.md
+++ b/doc/Advanced-configuration.md
@@ -234,9 +234,19 @@ Section below describes the default options.
 
 # app.config
 
-A file with Erlang application configuration. It can be found in `[MongooseIM root]/rel/files/`. By default only Lager config can be found there. Check [Lager's documentation](https://github.com/basho/lager) for more information.
+A file with Erlang application configuration. It can be found in `[MongooseIM root]/rel/files/`.
+By default only following applications can be found there:
 
-Here you can change logs location and file names (`file`), rotation strategy (`size` and `count`) and date formatting (`date`). Ignore log level parameters - they are overridden with the value in `ejabberd.cfg`.
+* `lager` - check [Lager's documentation](https://github.com/basho/lager) for more information.
+   
+    Here you can change logs location and file names (`file`), rotation strategy (`size` and `count`) 
+   and date formatting (`date`). Ignore log level parameters - they are overridden with the value in `ejabberd.cfg`.
+
+* `ssl` only `session_lifetime` parameter is specified in
+    this file. Its default value is **600s**. This parameter says for how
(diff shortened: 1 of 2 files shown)


ppikula:
Long story short, this option tells how long a ssl session lives in the ssl library cache. The cache TTL is the same for the client side and the server side. As far as I understand the ssl library caches sessions for the tls resumption feature[1][2]. The default value changed 10 times in 17.0, so that might have caused my problems with memory. 

More details about the problem: https://github.com/erlang/otp/pull/867

[1] https://blog.cloudflare.com/tls-session-resumption-full-speed-and-secure/
[2] https://timtaubert.de/blog/2014/11/the-sad-state-of-server-side-tls-session-resumption-implementations/

ppikula:
It seems that they merged my pull request, but  without answering my question :(. Maybe 1h hour is more sane than 10 minutes, I don't really know. If you have resources, it would be good to run a series of tests to pick "the best" value.  The problem occurs when the other party has a short session lifetime, thus I think that an entry in the documentation about it is necessary.

ppikula:
They added upper limit for the session caches in 18.2, consider adding this option to app.config at least it will solve the problem for people who use the newest Erlang.
ref: https://github.com/erlang/otp/commit/f258633921e7dd0bd5760281d2a26a0f937e53c4

michalwski:
Thanks for the update :)

arcusfelis:
Cache is stored in ETS tables `*_ssl_otp_session_cache`. I'm not sure how processes can consume more or less memory depending on cache size.

arcusfelis:
I would suggest to add metrics for cache hit and miss. But I guess it requires changes in the OTP code.

ppikula:
@arcusfelis take look at the  gun's logs in the issue I referenced. As far as I remember, connection processes read from the ets table, so they receive a copy of ets content - it is a full copy when you connect only to one host:port pair(no other destinations in the cache). Note that the cache module is called indirectly mostly by the `CacheCb:*` calls. Correct me if I'm wrong.

All in all you can quite easily reproduce it and see what is exactly going on. For example in nginx you can set ssl_session_timeout to a low value, keeping erlang ssl timeout on 24H level, and keep bombarding with http requests without keepalive.

Yes, metric for that would require changes in OTP or using tracing...

PR #554, opened 2015-10-15
remove folsom, because we don't use it anymore

ppikula:


michalwski:
Did you try exometer reporter after the change? I don't think this will break, but since we don't have tests for that I'd like to be somehow sure reporters still works.

michalwski:
I did it by myself, reporters and subscriptions still work after the change.

PR #558, opened 2015-10-21
Jlib tests and cleaning

michalwski:
This PR creates tests for jlib module and cleans the module code a bit.

### DIFF
--- a/apps/ejabberd/src/jlib.erl
+++ b/apps/ejabberd/src/jlib.erl
@@ -301,7 +301,7 @@ binary_to_jid(J) ->
     binary_to_jid1(J, []).
 
 
--spec binary_to_jid1(binary(), binary()) -> 'error' | ejabberd:jid().
+-spec binary_to_jid1(binary(), [byte()]) -> 'error' | ejabberd:jid().
 binary_to_jid1(<<$@, _J/binary>>, []) ->
     error;
 binary_to_jid1(<<$@, J/binary>>, N) ->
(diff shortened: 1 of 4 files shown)


ppikula:
Great code man!  :+1: 

I see that you started testing the jid functions. What about moving them all to a
separate module - for instance `jid.erl` (vim-wrangler to the rescue ;))?

 By doing that we could be more "ecological"(use less letters, save columns) - `jlib:jid_to_binary`  -> `jid:to_binary`, `jlib:make_jid` -> `jlib:new/make` etc.

michalwski:
Also we can merge functions `jid_tolower/1` and `jid_to_lower/1` as they do the same and both are used from other modules... I think introducing `jid.erl` module should be done in a separate PR.

pzel:
Code reviewed by pair programming.

michalwski:
Thanks @pzel. Please merge this PR if you don't have further comments.

apps/ejabberd/src/jlib.erl
```diff
@@ -368,22 +363,14 @@ jid_to_binary({Node, Server, Resource}) ->
          end,
     S3.
 
--spec is_nodename([] | binary()) -> boolean().
+-spec is_nodename(<<>> | binary()) -> boolean().
 is_nodename([]) ->
```
fenek: This clause should be removed if we are sure this function won't be called for lists.

michalwski: I changed this to empty binary. Empty list was a leftover after switching form lists to binary. This line was not covered by our tests and that's why we didn't catch it before.

PR #566, opened 2015-10-29
Check user allowed hook

michalwski:
The idea behind this PR is to make user blocking/banning easier. Now the first phase is the old acl matching. If this allows a user the the hook `check_user_allowed` is run (as run_fold) with initial value `allow`.

### DIFF
--- a/apps/ejabberd/src/ejabberd_auth.erl
+++ b/apps/ejabberd/src/ejabberd_auth.erl
@@ -220,8 +220,6 @@ do_check_password_loop([AuthModule | AuthModules], Args) ->
 -spec check_digest(binary(), fun(), binary(), binary()) -> boolean().
 check_digest(<<>>, _, <<>>, _) ->
     false; %%empty digest and password
-check_digest(<<>>, _, Password, Passwd) ->
-    Passwd == Password;
 check_digest(Digest, DigestGen, _Password, Passwd) ->
     Digest == DigestGen(Passwd).
 
--- a/apps/ejabberd/src/ejabberd_auth.erl
+++ b/apps/ejabberd/src/ejabberd_auth.erl
@@ -218,18 +218,13 @@ do_check_password_loop([AuthModule | AuthModules], Args) ->
     end.
 
 -spec check_digest(binary(), fun(), binary(), binary()) -> boolean().
-check_digest(Digest, DigestGen, Password, Passwd) ->
-    DigRes = if
-                 Digest /= <<>> ->
-                     Digest == DigestGen(Passwd);
-                 true ->
-                     false
-             end,
-    if DigRes ->
-           true;
-       true ->
-           (Passwd == Password) and (Password /= <<>>)
-    end.
+check_digest(<<>>, _, <<>>, _) ->
+    false; %%empty digest and password
+check_digest(<<>>, _, Password, Passwd) ->
+    Passwd == Password;
+check_digest(Digest, DigestGen, _Password, Passwd) ->
+    ?WARNING_MSG("Computed digest for password ~p is: ~p", [Passwd, DigestGen(Passwd)]),
+    Digest == DigestGen(Passwd).
 
 -spec set_password(User :: ejabberd:user(),
                    Server :: ejabberd:server(),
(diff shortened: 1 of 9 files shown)


apps/ejabberd/src/ejabberd_auth.erl
```diff
@@ -218,18 +218,12 @@ do_check_password_loop([AuthModule | AuthModules], Args) ->
...
-                     false
-             end,
-    if DigRes ->
-           true;
-       true ->
-           (Passwd == Password) and (Password /= <<>>)
-    end.
+check_digest(<<>>, _, <<>>, _) ->
+    false; %%empty digest and password
+check_digest(<<>>, _, Password, Passwd) ->
```
ppikula: what do you think about the following clause:  check_digest(<<>>, _, Password, Password) -> true; ?

Moreover, there is something wrong with this function -  It says check_digest, but it also checks passwords ...  Maybe it is a good moment to clean this mess.

test/ejabberd_tests/tests/legacy_stream_helper.erl
```diff
@@ -0,0 +1,113 @@
...
+    AuthReq = escalus_stanza:iq_set(?NS_AUTH,
+                                    [username(Username), password(Password), res(<<"res">>)]),
+    escalus:send(Conn, AuthReq),
+    Response = escalus:wait_for_stanza(Conn),
+    escalus:assert(is_iq_result, [AuthReq], Response),
+    {Conn, Props, UnusedFeatures}.
+
+res(Res) when is_binary(Res) ->
+    #xmlel{name = <<"resource">>,
+           children = [exml:escape_cdata(Res)]}.
```
ppikula: I think that we no longer need to escape it

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -579,6 +495,66 @@ wait_for_auth({xmlstreamerror, _}, StateData) ->
...
+       "(~w) Forbidden legacy authentication for "
+       "username '~s' with resource '~s'",
+       [StateData#state.socket, U, R]),
+    Err = jlib:make_error_reply(El, ?ERR_JID_MALFORMED),
+    send_element(StateData, Err),
+    fsm_next_state(wait_for_auth, StateData);
+maybe_legacy_auth(JID, El, StateData, U, P, D, R) ->
+    case user_allowed(JID, StateData) of
+        true ->
+            DGen = fun(PW) ->
```
ppikula: I would move the whole(or part) true clause to a separate function

apps/ejabberd/src/ejabberd_auth.erl
michalwski: > what do you think about the following clause: check_digest(<<>>, _, Password, Password) -> true; ?

Yeah, that will be better :)

> Moreover, there is something wrong with this function - It says check_digest, but it also checks passwords ... Maybe it is a good moment to clean this mess.

I know I know... Some day will finally (hopefully) get rid of all the parent's mess. I'll refactor this function in this PR.

test/ejabberd_tests/tests/legacy_stream_helper.erl
michalwski: You might be right :) I'll try.

ppikula:
No changes in the doc directory :). There are sections about hooks, maybe it is worth to describe the new hook.

apps/ejabberd/src/ejabberd_auth.erl
michalwski: I decided to only remove password checking from check_digest function. Other refactoring around this will be done in a separate pull request as they touch many places.

doc/developers-guide/hooks_description.md
```diff
@@ -117,4 +118,19 @@ Number of retries for this transaction is set to 1 which means that in some
...
+                        Server,
+                        allow, [JID]).
+```
+
+This hook is run after authentication when user sends the iq opening a session.
+Handler function are expected to return:
+
+* `allow` if given JID is allowed to open a new sessions
+* `deny` if the JID is not allowed but other handlers should be run
+* `{stop, deny}` if the JID is not allowed but other handlers should be run
```
ppikula: The line is the same as the one above it, is "should not" the correct version?

PR #567, opened 2015-10-29
riak backend for last activity

ppikula:


apps/ejabberd/src/mod_last_riak.erl
```diff
@@ -0,0 +1,61 @@
...
+            {ok, Timestamp, Status};
+        _ ->
+            %??
+            not_found
+    end.
+
+-spec count_active_users(ejabberd:lserver(), non_neg_integer(), '<' | '>') ->
+    non_neg_integer().
+count_active_users(_LServer, _TimeStamp, _Comparator) ->
+    ?ERROR_MSG("cannot count active users in the riak backend", []),
```
fenek: Why? Secondary indexes are natural fit here.

michalwski:
Looks good :) :+1:  The only missing part is the PR desc when you could add some implementation details.

apps/ejabberd/src/mod_last_riak.erl
```diff
@@ -0,0 +1,76 @@
...
+        {error, Reason} ->
+            {error, Reason}
+    end.
+
+-spec count_active_users(ejabberd:lserver(), non_neg_integer()) -> non_neg_integer().
+count_active_users(LServer, TimeStamp) ->
+    Idx = {index, bucket_type(LServer), ?TIMESTAMP_IDX, TimeStamp+1, infinity()},
+    RedMF = {modfun, riak_kv_mapreduce, reduce_count_inputs},
+    Red = [{reduce, RedMF, [{reduce_phase_batch_size, 1000}], true}],
+    {ok, [{0, [Count]}]}  = mongoose_riak:mapred(Idx, Red),
```
fenek: http://docs.basho.com/riak/latest/dev/using/mapreduce/#When-Not-to-Use-MapReduce

michalwski: What do you mean @fenek? Do you suggest that the map reduce here is used in a wrong way? If so how the code should be improved here?

fenek: The link I've pasted states that MapRed should not be used "When you want to query data over an entire bucket." - and comparing each element's timestamp is in fact iterating over entire bucket.
In my previous comment I've recommended using secondary indexes.

michalwski: But to me it looks like we use the secondary index in the map reduce, don't we? Where is the bug in my reasoning? I know that in the worst case when we choose bad timestamp it may go through the entire bucket.

ppikula: Yes, it is given as example [here](http://docs.basho.com/riak/latest/dev/using/2i/#Querying) section  **Count Bucket Objects via $bucket Index**. 

I think the best way to find which way is better is to test it with at least several millions of entries. After discussion with @fenek we suspect, that the only difference will be who will consume more resources, whether will it be MIM  or Riak.

In case of mapred we receive just count - aggregation and actual counting is done on Riak's side. The *get_index_range* function gives us list of keys(which may be very big) and MiM has to calculate the length. It is not that bad, because Riak allows to paginate these results, so we can avoid receiving 10 millions keys at once, by doing many many requests. I guess the mapred approach might be more efficient, but this is my guess only...

PR #568, opened 2015-10-29
Implement jlib:binary_to_jid/1 in NIF

fenek:
Still needs some benchmark to confirm whether there is some real-life improvement, but in microbenchmark the parsing itself is almost 10x faster.

### DIFF
--- a/apps/ejabberd/c_src/jlib.c
+++ b/apps/ejabberd/c_src/jlib.c
@@ -1,0 +1,86 @@
+#include <string.h>
+#include <stdbool.h>
+#include <erl_nif.h>
+#include <ei.h>
+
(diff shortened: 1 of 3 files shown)


michalwski:
Waiting for the benchmark results :) Some time ago when I was improving `binar_to_jid` I used propEr combined with exometer_histograms. Final result of current implementation for 1024 propEr tests was the following:
```erlang
[{n,59},
  {mean,7},
  {min,3},
  {max,48},
  {median,6},
  {50,6},
  {75,7},
  {90,12},
  {95,21},
  {99,23},
  {999,48}]}]
``` 
Values in microseconds. What are your microbenchmark results?

lavrin:
Do we have a profile somewhere showing it actually is a hotspot?

PR #570, opened 2015-10-30
fix crash on bad xml input

### LINKED ISSUE #564: crash on bad XML input
When sending bad data (non XML) on the stream we see a crash now instead of returning a stream error:

```
2015-10-28 15:32:51.771 [error] <0.677.0> gen_server <0.677.0> terminated with reason: no match of right hand value {error,{"syntax error",<<"asdfasdf\r\n">>}} in ejabberd_receiver:process_data/2 line 328
2015-10-28 15:32:51.771 [error] <0.677.0> CRASH REPORT Process <0.677.0> with 0 neighbours exited with reason: no match of right hand value {error,{"syntax error",<<"asdfasdf\r\n">>}} in ejabberd_receiver:process_data/2 line 328 in gen_server:terminate/7 line 804
2015-10-28 15:32:51.771 [error] <0.293.0> Supervisor ejabberd_receiver_sup had child undefined started with {ejabberd_receiver,start_link,undefined} at <0.677.0> exit with reason no match of right hand value {error,{"syntax error",<<"asdfasdf\r\n">>}} in ejabberd_receiver:process_data/2 line 328 in context child_terminated
```

Expected behaviour: Return stream error, then close connection.
```
<?xml version='1.0'?><stream:stream xmlns='jabber:client' xmlns:stream='http://etherx.jabber.org/streams' id='1609254289' from='chat.sqor.com' version='1.0'><stream:error><xml-not-well-formed xmlns='urn:ietf:params:xml:ns:xmpp-streams'/></stream:error></stream:stream>
```

Actual behaviour: Connection is being closed.

Probably bug got introduced when switching to exml.
https://github.com/esl/MongooseIM/blame/master/apps/ejabberd/src/ejabberd_receiver.erl#L328

sstrigler:
fixes #564

### DIFF
--- a/apps/ejabberd/src/ejabberd_receiver.erl
+++ b/apps/ejabberd/src/ejabberd_receiver.erl
@@ -325,13 +325,20 @@ process_data(Data, #state{parser = Parser,
     mongoose_metrics:update([data, xmpp, received, xml_stanza_size], Size),
 
     maybe_run_keep_alive_hook(Size, State),
-    {ok, NewParser, Elems} = exml_stream:parse(Parser, Data),
-    {NewShaperState, Pause} = shaper:update(ShaperState, Size),
-    ValidElems = replace_too_big_elems_with_stream_error(Elems, MaxSize),
-    [gen_fsm:send_event(C2SPid, wrap_if_xmlel(E)) || E <- ValidElems],
-    maybe_pause(Pause, State),
-    State#state{parser = NewParser,
-                shaper_state = NewShaperState}.
+    case exml_stream:parse(Parser, Data) of
+        {ok, NewParser, Elems} ->
+            {NewShaperState, Pause} = shaper:update(ShaperState, Size),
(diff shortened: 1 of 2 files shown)


michalwski:
Thanks for the change. Would perfect if you could write a test case showing there was a problem before and now it's fixed.

sstrigler:
Oh, lol, of course. Not my day. I knew I had forgotten sth. Would have been too easy :p

sstrigler:
And suddenly a rogue testcase appeared.
[ ] continue
[ ] integration

michalwski:
Hmm, still the error case is not covered somehow. Take a look here: https://coveralls.io/builds/3999189/source?filename=apps%2Fejabberd%2Fsrc%2Fejabberd_receiver.erl#L336

sstrigler:
*facepalm*

michalwski:
Thanks @sstrigler :)

PR #573, opened 2015-11-02
Do not create nor update skipped hooks

michalwski:
This PR prevents creating a spiral metric for skipped hook. Such metric is dangerous because it may have lot of (tens of thousands) updates in a minute which can lead to too long message inbox for process holding this metric and can crash `exometer_reporter`. 

Also, this PR adds some tests for metric subscriptions.

### DIFF
--- a/apps/ejabberd/test/mongooseim_metrics_SUITE.erl
+++ b/apps/ejabberd/test/mongooseim_metrics_SUITE.erl
@@ -8,12 +8,12 @@
 
 -compile([export_all]).
 
-all() -> [subscriptions_initialised].
+all() -> [no_skip_metric, subscriptions_initialised].
 
 init_per_suite(C) ->
     application:load(exometer),
     application:set_env(exometer, mongooseim_report_interval, 1000),
-    {Port, Socket} = carbon_server:start(1, 0),
+    {Port, Socket} = carbon_cache_server:start(1, 0),
     Sup = spawn(fun() ->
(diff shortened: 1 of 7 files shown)


apps/ejabberd/test/mongooseim_metrics_SUITE.erl
```diff
@@ -0,0 +1,90 @@
...
+                        supervisor:start_child(ejabberd_sup, Hooks),
+                        supervisor:start_child(ejabberd_sup, C2SSupervisor),
+                        receive
+                            stop ->
+                                ok
+                        end
+                end),
+    Reporters = get_reporters_cfg(Port),
+    application:set_env(exometer, report, Reporters),
+    PortServer = receive
```
ppikula: I think this receive could be encapsulated inside the `carbon_sever` module- ex. wait_wait_for_client or sth similar. Btw we can name it carbon_cache_server. Initially, I thought that it might have something to do with carbon copies.

michalwski: Good idea. I'll do that.

ppikula:
What about testing title's feature - skipped hooks? I see two ways of doing that, either extend carbon_server and make it understand metric reports or use meck to validate if "mongoose_metrics:update/2" has/hasn't been called.

michalwski:
Done. Also I renamed the carbon_server and applied your suggestion.

PR #574, opened 2015-11-05
Remove duplicate code, make abstract xmpp_router

pzel:
I removed 8 instances of identical code from 4 separate modules, and implemented them as a generic `xmpp_router`. The modules now expose their own distinct implementations, under `Module:do_route/3`. These module-specific implemenations were not changed.

Edit: yes, this makes the ?ERROR_MSG in the generic implementation a little less informative, as the trace will always be logged as coming from `xmpp_router`. This is why the implementation module is referenced in the error message itself. I believe the cleanup is worth the tradeoff.

### DIFF
--- a/apps/ejabberd/src/xmpp_router.erl
+++ b/apps/ejabberd/src/xmpp_router.erl
@@ -1,15 +1,21 @@
 -module(xmpp_router).
 -export([route/4]).
--include_lib("ejabberd/include/ejabberd.hrl").
+
+-include("ejabberd.hrl").
 
 
--callback do_route(From :: ejabberd:jid(), To :: ejabberd:jid(), Packet :: jlib:xmlel()) -> ok.
+-callback do_route(From :: ejabberd:jid(), To :: ejabberd:jid(),
+                   Packet :: jlib:xmlel()) -> ok.
+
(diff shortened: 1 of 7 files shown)


michalwski:
That's just awesome men! :+1:

apps/ejabberd/src/xmpp_router.erl
```diff
@@ -0,0 +1,15 @@
...
+-export([route/4]).
+-include_lib("ejabberd/include/ejabberd.hrl").
+
+
+-callback do_route(From :: ejabberd:jid(), To :: ejabberd:jid(), Packet :: jlib:xmlel()) -> ok.
+route(Module,From,To,Packet) ->
+    case (catch Module:do_route(From,To,Packet)) of
+        {'EXIT', Reason} ->
+            ?ERROR_MSG("~p when routing ~ts from ~ts to ~ts in module: ~p",
+                       [Reason, exml:to_binary(Packet),
```
michalwski: I suggest reorganising the log message so that the packet is the last element. It will be more readable. Also I think simple test for the `'EXIT'` case would be good here.

michalwski: And maybe `erlang:get_stacktrace()` can be helpful is such error case.

PR #575, opened 2015-11-09
Admin extra interface

ludwikbukowski:
I tried to make admin interface more hospitable. It should do some checks on input and inform about result of command.
Improved following modules:
- mod_admin_extra_accounts
- mod_admin_extra_last
- mod_admin_extra_private
- mod_admin_extra_roster
- mod_admin_extra_stanza
- mod_admin_extra_stats
- mod_admin_extra_vcard

### DIFF
--- a/apps/ejabberd/src/mod_admin_extra_accounts.erl
+++ b/apps/ejabberd/src/mod_admin_extra_accounts.erl
@@ -36,8 +36,9 @@
     delete_old_users/1,
     delete_old_users_vhost/2,
     ban_account/3,
-    num_active_users/2
-    ]).
+    num_active_users/2,
+    check_account/2,
+    check_password/3]).
 
 -include("ejabberd.hrl").
 -include("ejabberd_commands.hrl").
@@ -53,13 +54,13 @@ commands() ->
                            desc = "Change the password of an account",
                            module = ?MODULE, function = set_password,
                            args = [{user, binary}, {host, binary}, {newpass, binary}],
-                           result = {res, rescode}},
+                           result = {res, restuple}},
         #ejabberd_commands{name = check_password_hash, tags = [accounts],
                            desc = "Check if the password hash is correct",
                            longdesc = "Allowed hash methods: md5, sha.",
                            module = ?MODULE, function = check_password_hash,
                            args = [{user, binary}, {host, binary}, {passwordhash, string}, {hashmethod, string}],
-                           result = {res, rescode}},
+                           result = {res, restuple}},
         #ejabberd_commands{name = delete_old_users, tags = [accounts, purge],
                            desc = "Delete users that didn't log in last days, or that never logged",
                            module = ?MODULE, function = delete_old_users,
@@ -74,7 +75,7 @@ commands() ->
                            desc = "Ban an account: kick sessions and set random password",
                            module = ?MODULE, function = ban_account,
                            args = [{user, binary}, {host, binary}, {reason, binary}],
-                           result = {res, rescode}},
+                           result = {res, restuple}},
         #ejabberd_commands{name = num_active_users, tags = [accounts, stats],
                            desc = "Get number of users active in the last days",
                            module = ?MODULE, function = num_active_users,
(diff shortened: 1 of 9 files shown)


michalwski:
@ludwikbukowski sth's wrong with `ejabberdctl_SUITE:vcard_rw` because all jobs failed except the one using external authentication (where this test case is probably not run).

apps/ejabberd/src/ejabberd_auth.erl
```diff
@@ -246,12 +246,18 @@ set_password(User, Server, Password) ->
 do_set_password(LUser, LServer, _) when LUser =:= error; LServer =:= error ->
     {error, invalid_jid};
 do_set_password(LUser, LServer, Password) ->
-    lists:foldl(
-      fun(M, {error, _}) ->
-              M:set_password(LUser, LServer, Password);
-         (_M, Res) ->
-              Res
-      end, {error, not_allowed}, auth_modules(LServer)).
+    case is_user_exists(LUser, LServer) of
```
michalwski: I think this check should be moved to the main function which is `set_password` unless `do_set_password` is used directly from other places.

apps/ejabberd/src/ejabberd_auth.erl
```diff
@@ -246,12 +246,18 @@ set_password(User, Server, Password) ->
...
+        true ->
+            lists:foldl(
+                fun(M, {error, _}) ->
+                    M:set_password(LUser, LServer, Password);
+                   (_M, Res) ->
+                       Res
+                end, {error, not_allowed}, auth_modules(LServer));
+        false ->
+            {error, user_not_exists}
```
michalwski: I'd change the error reason to `user_does_not_exist`

apps/ejabberd/src/mod_admin_extra_accounts.erl
```diff
@@ -82,30 +83,55 @@ commands() ->
...
+    case ejabberd_auth:is_user_exists(User, Host) of
+        true ->
+            case ejabberd_auth:check_password(User, Host, Password) of
+                true ->
+                    {ok, io_lib:format("Password '~s' for user ~s@~s is correct", [Password, User, Host])};
+                false ->
+                    {incorrect, io_lib:format("Password '~s' for user ~s@~s is incorrect", [Password, User, Host])}
+            end;
+        false ->
+            {not_exists, io_lib:format("Password '~s' for user ~s@~s is incorrect because the account does not exist", [Password, User, Host])}
```
michalwski: `not_exists` -> `does_not_exist` or even `user_does_not_exist`

apps/ejabberd/src/ejabberd_auth.erl
lavrin: Why not just `not_found`?

lavrin: BTW, we have a proper name for `is_user_exists` already, i.e. `does_user_exist`. Is it available in this scope?

michalwski: `does_user_exists` is in backend modules only (at least now). In `ejabberd_auth` we can use `do_does_user_exist` which is local function and it's called form `is_user_exists`

apps/ejabberd/src/mod_admin_extra_private.erl
```diff
@@ -69,18 +69,24 @@ commands() ->
...
+                          attrs = [{<<"xmlns">>,?NS_PRIVATE}],
+                          children = [#xmlel{ name = Element, attrs = [{<<"xmlns">>, Ns}]}] } },
+            ResIq = M:process_sm_iq(From, To, IQ),
+            [#xmlel{ name = <<"query">>,
+                     attrs = [{<<"xmlns">>,<<"jabber:iq:private">>}],
+                     children = [SubEl] }] = ResIq#iq.sub_el,
+            Ret = exml:to_binary(SubEl),
+            {ok, Ret};
+        false ->
+            {user_not_exists, io_lib:format("User ~s@~s not exists", [Username, Host])}
```
michalwski: Same story about the `user_not_exists` atom.

michalwski: Also introducing function `do_private_get` with `ejabberd_auth:is_user_exists` result as first arg may improve code readability here.

apps/ejabberd/src/mod_admin_extra_roster.erl
```diff
@@ -163,12 +163,22 @@ commands() ->
                      Group :: binary() | string(),
                      Subs :: subs()) -> 'error' | 'ok'.
 add_rosteritem(LocalUser, LocalServer, User, Server, Nick, Group, Subs) ->
-    case subscribe(LocalUser, LocalServer, User, Server, Nick, Group, Subs, []) of
-        {atomic, ok} ->
-            push_roster_item(LocalUser, LocalServer, User, Server, {add, Nick, Subs, Group}),
-            ok;
-        _ ->
-            error
+    case ejabberd_auth:is_user_exists(LocalUser, LocalServer) of
```
michalwski: This part definitely requires some refactoring as there is to many nested case clauses.

apps/ejabberd/src/mod_admin_extra_stanza.erl
```diff
@@ -153,6 +160,16 @@ build_packet(message_headline, [Subject, Body]) ->
...
     C2sPid = ejabberd_sm:get_session_pid(Username, Host, Resource),
-    {ok, XmlEl} = exml:parse(Stanza),
-    p1_fsm:send_event(C2sPid, {xmlstreamelement, XmlEl}).
+    case C2sPid of
+        none ->
+            {bad_jid, io_lib:format("User ~s@~s/~s does not exist",[Username, Host, Resource])};
+        _ ->
+            case exml:parse(Stanza) of
+                {ok, XmlEl} ->
+                    p1_fsm:send_event(C2sPid, {xmlstreamelement, XmlEl}),
```
michalwski: Our `ejabberd_c2s` is no longer `p1_fsm` but `p1_fsm_old`. This shouldn't make any difference here but I'd use the correct module to call the process.

apps/ejabberd/src/mod_admin_extra_vcard.erl
```diff
@@ -113,21 +113,44 @@ commands() ->
 %%%
 
 get_vcard(User, Host, Name) ->
```
ludwikbukowski: Remember to add specs for exported functions and also change them when you change function's result value

apps/ejabberd/src/mod_admin_extra_last.erl
```diff
@@ -62,10 +62,25 @@ commands() ->
 
 -spec set_last(ejabberd:user(), ejabberd:server(), _, _) -> 'ok'.
```
ludwikbukowski: remember to modify spec

michalwski:
I see that most (if not all) commands you touched have now result type `{res, restuple}`. This is correct, but not much informative. How about spending some more time and defining more precise result types. A good example maybe command `connected_users_info`. Which you can find in `mod_admin_extra_sessions` module. The result type for this command looks like the following (in shell).

```
mongooseimctl help connected_users_info

  Command Name: connected_users_info

  Arguments:

  Returns: connected_users_info::[ sessions::{ jid::string,
                                               connection::string,
                                               ip::string,
                                               port::integer,
                                               priority::integer,
                                               node::string,
                                               uptime::integer } ]

  Tags: session

  Description: List all established sessions and their information
```

Also I think we can remove the unnecessary empty lines in the help output above.

test/ejabberd_tests/tests/ejabberdctl_SUITE.erl
```diff
@@ -561,6 +737,22 @@ send_message(Config) ->
...
+    escalus:story(Config, [{alice, 1}, {bob, 1}], fun(Alice, Bob) ->
+        {_, Err1} = ejabberdctl("send_message_chat", ["'@@#$%!!.§§£'",
+                                                   escalus_client:full_jid(Bob),
+                                                   "\"Hello bobby!\""], Config),
+        {_, Err2} = ejabberdctl("send_message_headline", ["'%%@&@&@==//\///'",
+                                                       escalus_client:short_jid(Bob),
+                                                       "Subj", "\"Are
+                                                       you there?\""],
+                             Config),
+        Err1 =/=0,
```
erszcz: I noticed this on Travis when testing stuff rebased on top of this PR:

```
ejabberdctl_SUITE.erl:750: Warning: use of operator '=/=' has no effect
ejabberdctl_SUITE.erl:751: Warning: use of operator '=/=' has no effect
```

Which makes sense, since these lines are not assertions! They evaluate to a `boolean()` which is then discarded.

Please rewrite to:

```erlang
true = Err1 =/= 0
```

or 


```erlang
?assertNotEqual(0, Err1)
```

or

```erlang
?assert(Err1 =/= 0)
```

test/ejabberd_tests/tests/ejabberdctl_SUITE.erl
```diff
@@ -577,6 +769,26 @@ send_stanza(Config) ->
...
+        BobJID = <<BobName/binary, $@, Domain/binary, $/, (escalus_client:resource(Bob))/binary>>,
+        Stanza = re:replace(exml:to_binary(escalus_stanza:from(escalus_stanza:chat_to(Alice, "Hi"), BobJID)),
+                            <<?DOUBLE_QUOTE_CHAR>>, <<?SINGLE_QUOTE_CHAR>>, [global, {return, binary}]),
+        StanzaWrong = <<"<iq type='get' id='234234'><xmlns='wrongwrong'>">>,
+        {_, Err} = ejabberdctl("send_stanza_c2s", [WrongBobName, Domain, Resource, <<$\", Stanza/binary, $\">>],
+                               Config),
+        {_, Err2} = ejabberdctl("send_stanza_c2s", [BobName, Domain, Resource, <<$\", StanzaWrong/binary, $\">>],
+                               Config),
+
+        Err =/= 0,
```
erszcz: Similar to the above case.

erszcz:
In general, compiler warnings are there for a reason ;)

test/ejabberd_tests/tests/ejabberdctl_SUITE.erl
ludwikbukowski: Fixed here:
https://github.com/esl/MongooseIM/pull/693

PR #577, opened 2015-11-09
MUC Light

fenek:
Implementation based on https://github.com/xsf/xeps/pull/118

Still lacks docs. :(


lavrin:
Great work!
Alas, not even a single line of high-level developer doc?
I mean distribution model (single point of failure/serialization or completely async routing, if the latter what are message ordering guarantees?), storage choices (persistent vs transient data, what kind of data goes where), high-availability / fault-tolerance / resiliency to node failure (will service continue to operate for the rest of online users when one node in the cluster fails).
Of course we could just read the code... but if you know this stuff, we could also read a description ;)

fenek:
That's why it's still `WIP` and not `review-wanted`. :) This implementation waited so long for "coming out", I didn't want to risk another pause because some other project appears. My current goal is to make it pass current test suite in Travis.

Thanks for the reminder anyway. :)

lavrin:
PDF preview of the XEP proposal is available:
[XEP-xxxx: Multi-User Chat Light.pdf](https://github.com/esl/MongooseIM/files/35614/XEP-xxxx.Multi-User.Chat.Light.pdf)

fenek:
Updated version:
[XEP-xxxx.Multi.User.Chat.Light.pdf](https://github.com/esl/MongooseIM/files/36382/XEP-xxxx.Multi.User.Chat.Light.pdf)

studzien:
Apparently, incoming MUC Light messages are stored in the one-to-one chats archive (since they are routed via ``ejabberd_router:route/3``).
I've created a test case for this here: e5c5f32611f7

The instant solution I can see is to bind to the ``filter_packet`` hook in ``mod_muc_light`` and stop running the hook just before it gets to ``mod_mam``.
Do you think this make any sense?

studzien:
Obviously we want to stop the hook if and only if the message is a MUC Light message :)

michalwski:
Closing and re-opening to trigger travis

PR #579, opened 2015-11-10
Introduce jid module

michalwski:
This PR introduces `jid` module where all jid-specific functions are defined. Previously they were defined in `jlib`, it's still possible to use this module for jid manipulation but this will produce a log message on warning level saying that given function has been deprecated and will be removed.

### DIFF
--- a/apps/ejabberd/src/jid.erl
+++ b/apps/ejabberd/src/jid.erl
@@ -1,0 +1,212 @@
+%%==============================================================================
+%% Copyright 2015 Erlang Solutions Ltd.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%% http://www.apache.org/licenses/LICENSE-2.0
+%%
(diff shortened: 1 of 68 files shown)


michalwski:
Example warning log message generated by the old API is the following:
```
2015-11-12 14:20:32.113 [warning] <0.55.0>@jlib:binary_to_jid:286 This function has been deprecated in MongooseIM 1.6.1 and will be removed in 1.7.0```

File: apps/ejabberd/src/jid.erl [hunk 1]
```diff
@@ -0,0 +1,212 @@
...
+-export([make/1]).
+-export([are_equal/2]).
+-export([from_binary/1]).
+-export([to_binary/1]).
+-export([is_nodename/1]).
+-export([nodeprep/1]).
+-export([nameprep/1]).
+-export([resourceprep/1]).
+-export([to_lower/1]).
+-export([remove_resource/1]).
```
ppikula: maybe `to_bare` instead of `remove_resource`? I think it is more accurate.

ppikula:
:+1:

PR #581, opened 2015-11-12
Better node down cleanup

### LINKED ISSUE #428: Session table cleanup is not sufficient
We have a cleanup procedure for session table when one of the nodes goes down, i.e. `?SM_BACKEND:cleanup(Node)`, but entries in other tables that are also created per session become zombies. The good example is `mod_carboncopy` which relies on `unset_presence_hook` - this hook won't be executed on brutal node shutdown.

This is especially harmful in case of applications that allow multiple resources and the resources are not constant per device.

A possible solution would be to execute new hook in the cleanup function, so modules can perform own cleanup. Existing hooks shouldn't be reused, as `cleanup/1` is executed on every other node in cluster and the module must be aware that the hook can be executed for the same session several times.

fenek:
This PR adds `session_cleanup` hook that is run for every session cleaned up by SM backend. New hook is used by `mod_carboncopy`, `mod_last`, `mod_stream_management` and `ejabberd_auth_anonymous`.

It resolves #428.


michalwski:
How about the TODO in desc? It seems implemented now.

fenek:
Ah, yes, forgot to remove it. Thanks. :)

PR #584, opened 2015-11-12
Add tests and refactor the acl module

### LINKED ISSUE #503: ACLs with a GLOB pattern cause a runtime error
I've used GLOB patterns in my config (not anymore: headcounter/deployment@81bbd3c608f11ef65e6a21ad25c87a6adaf7c4ff) so far, which in my case led to errors when users are trying to change their passwords (because the ACL was checked for `mod_register`).

The issue here is that `acl.erl` [is using `xmerl_regexp:sh_to_awk/1` with a binary instead of a list](https://github.com/esl/MongooseIM/blob/3af6c65111c26167869ae8473dbf3c33ab75ef53/apps/ejabberd/src/acl.erl#L287), which leads to the following error:

```erlang
2015-09-09 18:21:43.699 [error] <0.1099.0>@gen_iq_handler:process_iq:142
  {function_clause,[
    {xmerl_regexp,sh_to_awk_1,[<<"?">>],[{file,"xmerl_regexp.erl"},{line,191}]},
    {xmerl_regexp,sh_to_awk,1,[{file,"xmerl_regexp.erl"},{line,189}]},
    {acl,is_glob_match,2,[{file,"src/acl.erl"},{line,287}]},
    {lists,any,2,[{file,"lists.erl"},{line,1223}]},
    {acl,match_acls,3,[{file,"src/acl.erl"},{line,192}]},
    {mod_register,process_iq,4,[{file,"src/mod_register.erl"},{line,115}]},
    {gen_iq_handler,process_iq,6,[{file,"src/gen_iq_handler.erl"},{line,140}]},
    {gen_iq_handler,handle_info,2,[{file,"src/gen_iq_handler.erl"},{line,196}]}
  ]}
```

ppikula:
closes #503

### DIFF
--- a/apps/ejabberd/test/acl_SUITE.erl
+++ b/apps/ejabberd/test/acl_SUITE.erl
@@ -1,0 +1,229 @@
+-module(acl_SUITE).
+-compile([export_all]).
+
+-include_lib("common_test/include/ct.hrl").
+-include_lib("proper/include/proper.hrl").
+-include_lib("eunit/include/eunit.hrl").
+
+
+all() ->
+    [
+     all_rule_returns_allow,
+     none_rule_returns_deny,
+     basic_access_rules,
+     compound_access_rules,
+     host_sepcific_access_rules,
+     different_specs_matching_the_same_user
+    ].
+
+all_rule_returns_allow(_Config) ->
+    given_acl_started(),
+    given_stringprep_loaded(),
+
+    JID = jlib:make_jid(<<"pawel">>, <<"phost">>, <<"test">>),
+    ?assertEqual(allow, acl:match_rule(global, all, JID)),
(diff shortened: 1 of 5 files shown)


apps/ejabberd/test/acl_SUITE.erl
```diff
@@ -0,0 +1,302 @@
...
+     compound_access_rules,
+     host_sepcific_access_rules,
+     global_host_priority,
+     all_and_none_specs,
+     invalid_spec,
+     different_specs_matching_the_same_user
+    ].
+
+all_rule_returns_allow(_Config) ->
+    given_acl_started(),
```
michalwski: I see this (`acl_started`) is  required by all test cases (as well as stringprep). Would it that bad if we move this to init_per_testcase or even to init_per_suite as I don't see `acl` nor `stringprep` stopping (which I think should happen at the end).

PR #590, opened 2015-11-18
C2s ciphers tests

### LINKED ISSUE #370: c2s - ciphers test
There is an option to specify which ciphers should be used by the tls driver: https://github.com/esl/MongooseIM/blob/master/doc/advanced-configuration/Listener-modules.md#configuration but there is no tests for that:
https://coveralls.io/builds/1891922/source?filename=apps%2Fejabberd%2Fsrc%2Fejabberd_c2s.erl#L182

michalwski:
Fixes #370

These tests are using `openssl s_client` from Erlang VM.

### DIFF
--- a/apps/ejabberd/src/jlib.erl
+++ b/apps/ejabberd/src/jlib.erl
@@ -58,6 +58,7 @@
          timestamp_to_mam_xml/4,
          timestamp_to_iso/2,
          now_to_utc_binary/1,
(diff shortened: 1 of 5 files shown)


pzel:
^ Some cosmetic changes included in my additional commit. If you don't object to the changes and tests pass, I'll merge the branch.

PR #594, opened 2015-11-27
Riak offline

### LINKED ISSUE #586: Riak backend for mod_offline

michalwski:
This PR:
* adds Riak backend for offline storage
* extends offline_SUITE

Resolves #586

### DIFF
--- a/apps/ejabberd/src/mod_offline.erl
+++ b/apps/ejabberd/src/mod_offline.erl
@@ -64,6 +64,15 @@
 -define(MAX_USER_MESSAGES, infinity).
 -define(BACKEND, mod_offline_backend).
 
+-type msg() :: #offline_msg{us :: {ejabberd:luser(), ejabberd:lserver()},
+                          timestamp :: erlang:timestamp(),
+                          expire :: erlang:timestamp() | never,
+                          from :: jid(),
+                          to :: jid(),
+                          packet :: exml:element()}.
+
+-export_type([msg/0]).
+
 -record(state, {host, access_max_user_messages}).
 
 %% ------------------------------------------------------------------
@@ -164,7 +173,7 @@ write_messages(LUser, LServer, Msgs) ->
     end.
 
 -spec is_message_count_threshold_reached(ejabberd:luser(), ejabberd:lserver(),
-                                         integer, integer()) -> boolean().
+                                         integer(), integer()) -> boolean().
 is_message_count_threshold_reached(LUser, LServer, MaxOfflineMsgs, Len) ->
     case MaxOfflineMsgs of
         infinity ->
--- a/apps/ejabberd/src/mod_offline.erl
+++ b/apps/ejabberd/src/mod_offline.erl
@@ -38,7 +38,6 @@
 
 %% Hook handlers
 -export([inspect_packet/3,
-         resend_offline_messages/2,
          pop_offline_messages/3,
          get_sm_features/5,
          remove_expired_messages/1,
@@ -51,6 +50,9 @@
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
+
+%% helpers to be used from backend moudules
+-export([is_expired_message/2]).
 
 -include("ejabberd.hrl").
 -include("jlib.hrl").
@@ -452,7 +454,7 @@ pop_offline_messages(Ls, User, Server) ->
 pop_offline_messages(User, Server) ->
     LUser = jid:nodeprep(User),
     LServer = jid:nameprep(Server),
-    case ?BACKEND:pop_messages(LUser, LServer) of
+    case pop_messages(LUser, LServer) of
         {ok, Rs} ->
             lists:map(fun(R) ->
                 Packet = resend_offline_message_packet(Server, R),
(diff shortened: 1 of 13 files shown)


michalwski:
Firstly this one esl/escalus#86 has to be merged.

test/ejabberd_tests/tests/offline_SUITE.erl
```diff
@@ -18,7 +21,12 @@ all() ->
     [{group, mod_offline_tests}].
 
 all_tests() ->
-    [simple_message].
+    [simple_message,
+     error_message,
+     groupchat_message,
+     headliine_message,
```
fenek: Typo.

test/ejabberd_tests/tests/offline_SUITE.erl
```diff
@@ -86,3 +183,28 @@ login_send_presence(Config, User) ->
...
+    {atomic, ok} = escalus_ejabberd:rpc(ejabberd_config, add_global_option, [OptionName, Settings]),
+    Old.
+
+send_message(Alice, I) ->
+    escalus:send(Alice, escalus_stanza:chat_to(bob, make_chat_text(I))),
+    timer:sleep(100).
+
+
+make_chat_text(I) ->
+    Number = list_to_binary(integer_to_list(I)),
```
fenek: R15 is no longer supported! :D

michalwski: True, the PR was done before ditching the support. I'll change this line now.

apps/ejabberd/src/mod_offline.erl
```diff
@@ -140,33 +153,55 @@ stop(Host) ->
...
             ?ERROR_MSG("~ts@~ts: write_messages failed with ~p.",
                 [LUser, LServer, Reason]),
             discard_warn_sender(Msgs)
     end.
 
-%% Function copied from ejabberd_sm.erl:
+-spec is_message_count_threshold_reached(ejabberd:luser(), ejabberd:lserver(),
+                                         integer(), integer()) -> boolean().
+is_message_count_threshold_reached(LUser, LServer, MaxOfflineMsgs, Len) ->
+    case MaxOfflineMsgs of
```
fenek: I'd say it would be prettier with a match in function head. :)

apps/ejabberd/src/mod_offline_riak.erl
```diff
@@ -0,0 +1,187 @@
...
+
+read_user_idx(LUser, LServer) ->
+    {ok, IdxResult} = mongoose_riak:get_index(bucket_type(LServer), ?USER_IDX,
+                                              LUser, []),
+    IdxResult?INDEX_RESULTS.keys.
+
+write_msg(LUser, LServer, #offline_msg{from = FromJID, packet = Packet,
+                                       timestamp = TimestampIn, expire = Expire}) ->
+    Timestamp = usec:from_now(TimestampIn),
+    Obj = riakc_obj:new(bucket_type(LServer), key(LUser, Timestamp), exml:to_binary(Packet)),
```
fenek: Wouldn't it be more efficient to encode it with `term_to_binary`?

apps/ejabberd/src/mod_offline_riak.erl
```diff
@@ -0,0 +1,187 @@
...
+
+
+-spec bucket_type(ejabberd:lserver()) -> {binary(), ejabberd:lserver()}.
+bucket_type(LServer) ->
+    {<<"offline">>, LServer}.
+
+-spec key(binary(), integer()) -> binary().
+key(LUser, TimestampInt) ->
+    Timestamp = integer_to_binary(TimestampInt),
+    Random = integer_to_binary(random:uniform(1024)),
```
fenek: Not really random without ensuring proper seed first.

apps/ejabberd/src/mod_offline_riak.erl
michalwski: It would, but I wanted to make it easier to read from other backend languages. Some additional backend service f.e.

fenek:
General refactoring could include replacing all `now()` occurences with `os:timestamp()` in old code.

apps/ejabberd/src/mod_offline_riak.erl
fenek: I'm not sure if it's so common for other services to read this data directly. MAM uses `term_to_binary` and I don't think many people complain about this (although maybe I'm wrong).

michalwski: Actually they do. I've seen many questions like  "how to read this from other lang than Erlang"

apps/ejabberd/src/mod_offline_riak.erl
michalwski: True that.

apps/ejabberd/src/mod_offline_riak.erl
```diff
@@ -0,0 +1,187 @@
...
+-include("ejabberd.hrl").
+-include("jlib.hrl").
+-include("mod_offline.hrl").
+-include_lib("riakc/include/riakc.hrl").
+
+-define(TIMESTAMP_IDX, {integer_index, "timestamp"}).
+-define(USER_IDX,      {binary_index, "user"}).
+-define(EXPIRE_IDX,    {integer_index, "expire"}).
+
+-define(INFINITY, 99999999999). %% Wed, 16 Nov 5138 09:46:39 GMT
```
fenek: We'll be vulnerable to Y5.138K bug. :(

michalwski: True, but we have some time to propose better solution :)

apps/ejabberd/src/mod_offline_riak.erl
fenek: OK, let's leave it like this then. :(

apps/ejabberd/src/mod_offline_riak.erl
michalwski: I added the `random:seed` in `mod_offline:init` function as this function will be called from that process.

michalwski:
>General refactoring could include replacing all now() occurences with os:timestamp() in old code.

I was thinking about this, but I'm not sure if it's safe. F.e. when using os:timestamp we may get duplicates and change order of offline messages when the time is changed due to daylight saving.

fenek:
OK, it might be a subject of another PR. It's a shame that `monotonic_time` is not available before 18.0. :(

Issue #582, opened 2015-11-12
Stop supporting R15B03 and older


michalwski:
I think we can stop supporting older Erlang/OTP versions than R16. Here are some reasons (and action points) regarding this:
1. We can merge with `paulgray/exml` and benefit from he's NIF's reductions introduced in paulgray/exml@3e9d2238c44021211c1c63745b5f3ce8b9b56c5b
2. We can remove custom cowboy dep [1].
3. We can start using `application:get_env/3` instead of `application:get_env/2` and additional lines of code holding default value.
4. One job less on travis.

Probably some other improvements can be done but I let (and encourage) the team @esl/mongooseim-developers to comment on this issue and share other ideas :) 

[1]. https://github.com/esl/MongooseIM/blob/23cb2de2ea783e9fbc52fd4c0a1adf118f4f5126/rebar.config.script#L3-L8

fenek:
:+1:

We could remove walkaround for `binary_to_integer` and `integer_to_binary` as well!

pzel:
Yes, this would allow us to clean up quite a bit of accumulated kludges!

On Thu, Nov 12, 2015 at 11:34 AM, Piotr Nosek <notifications@github.com>
wrote:

> [image: :+1:]
>
> We could remove walkaround for binary_to_integer and integer_to_binary as
> well!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/esl/MongooseIM/issues/582#issuecomment-156064911>.
>



-- 
Simon Zelazny
Erlang Solutions, Ltd.

michalwski:
TODO list after ditching support for R15

- [ ] merge esl/exml#16
- [ ] remve custom cowboy version [see reabar.config.script](https://github.com/esl/MongooseIM/blob/f784669958d792d2e10afcf1e460589e419ee766/rebar.config.script)
- [ ] use application:get_env/3 when default value is needed
- [ ] use application:ensure_all_started instead of custom mechanism
- [ ] remove workaround for `binary_to_integer` in [rebar.config](https://github.com/esl/MongooseIM/blob/f784669958d792d2e10afcf1e460589e419ee766/apps/ejabberd/rebar.config#L6) and look for all places where `no_binary_to_integer` is used. Remember to remove `ejabberd_binary:binary_to_integer` and `ejabberd_binary:integer_to_binary`
[...]

PR #613, opened 2015-12-08
Ditch support for R15B03 and older

michalwski:
This PR addresses #582

### DIFF
--- a/apps/ejabberd/rebar.config
+++ b/apps/ejabberd/rebar.config
@@ -3,7 +3,6 @@
             {d, xml_nif},
             {parse_transform, lager_transform},
             {platform_define, "^R1[56]", latin1_characters},
-            {platform_define, "^R15", no_crypto_hmac}, %% Ref: OTP commit ffef856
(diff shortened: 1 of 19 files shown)


apps/ejabberd/rebar.config
```diff
@@ -3,8 +3,6 @@
             {d, xml_nif},
             {parse_transform, lager_transform},
             {platform_define, "^R1[56]", latin1_characters},
-            {platform_define, "^R15", no_binary_to_integer},
-            {platform_define, "^R15", no_crypto_hmac}, %% Ref: OTP commit ffef856
             {platform_define, "^R1[56]", only_builtin_types}]}. %% Ref: OTP commit b66e75c
 
 {require_otp_vsn, "R?1[5678]"}.
```
legoscia: Now this regexp can be changed as well.

michalwski: True, it's a must actually. Thanks for pointing that out.

PR #616, opened 2015-12-09
Stream compression after sasl

ludwikbukowski:
- fixed compresion advertising after tls negotiation.
- it's now possible to change negotation features order ( [XEP-0170](http://xmpp.org/extensions/xep-0170.html#c2s-compress)) due to https://github.com/esl/MongooseIM/issues/242
- few more tests for tls, compression, autentication and bind order in connect_SUITE
- tests work with https://github.com/esl/escalus/pull/88


michalwski:
I think you have to change `rebar.config` in `test/ejabberd_tests` to use escalus from your branch.

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -279,12 +279,15 @@ stream_start_by_protocol_version(_Pre_1_0, #state{lang = Lang, server = Server}
...
+    SockMod = (S#state.sockmod):get_sockmod(S#state.socket),
+    case {S#state.authenticated, S#state.resource, SockMod} of
+        {false, _, _} ->
             stream_start_features_before_auth(S);
-        {_, <<>>} ->
+        {_, <<>>, ejabberd_zlib} ->
             stream_start_features_before_bind(S);
-        {_, _} ->
+        {_, <<>>, _} ->
+            stream_start_features_before_auth(S);
```
michalwski: In this branch the user is authenticated so it's misleading to use function named `*_before_auth` (even if the output is the same). Maybe the function name should be changed to reflect both cases.

ludwikbukowski: You're right it can lead to confusion

test/ejabberd_tests/tests/connect_SUITE.erl
```diff
@@ -338,3 +355,99 @@ default_context(To) ->
...
+    ?assert_equal(false, has_feature(starttls, Features4)),
+    %%
+    escalus_session:authenticate(Conn4, Props4, Features4).
+
+has_feature(Feature, FeatureList) ->
+    {_, Value} = lists:keyfind(Feature, 1, FeatureList),
+    Value.
+
+%% tests for different order feature negotiation
+tls_compression_authenticate(Config) ->
```
michalwski: According to XEP-0138 we shouldn't allow compression before authentication. It may be used by malicious and not-registered users to attack the server.

ludwikbukowski:
I will try do some code refactoring so its still WIP

test/ejabberd_tests/tests/connect_SUITE.erl
ludwikbukowski: Hence order of feature negotiation should be changed in 'escalus_connection':'start' function:
https://github.com/esl/escalus/blob/master/src/escalus_connection.erl#L70

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -772,7 +756,7 @@ wait_for_session_or_sm({xmlstreamelement, El}, StateData0) ->
         #iq{type = set, xmlns = ?NS_SESSION} ->
             maybe_open_session(El, StateData);
         _ ->
-            fsm_next_state(wait_for_session_or_sm, StateData)
+            maybe_do_compress(El, wait_for_session_or_sm, StateData)
```
michalwski: I think we should end up in the same state after enabling the compression, but this time compression shouldn't be allowed. How else the binding process will happen?

michalwski: Ah, sorry, forgot that you have to open session after enabling a feature... no comment then...

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -734,7 +718,7 @@ wait_for_bind_or_resume({xmlstreamelement, El}, StateData) ->
                                    StateData#state{resource = R, jid = JID})
             end;
         _ ->
-            fsm_next_state(wait_for_bind_or_resume, StateData)
+            maybe_do_compress(El, wait_for_bind_or_resume, StateData)
     end;
 
 wait_for_bind_or_resume(timeout, StateData) ->
```
michalwski: I think this name is not valid any more. Now we can do `bind` or `resume` or `compression` here. How about naming it `wait_for_feature_after_auth`?

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -790,6 +774,51 @@ wait_for_session_or_sm({xmlstreamerror, _}, StateData) ->
...
+        false ->
+            send_element(StateData, compress_setup_failed()),
+            fsm_next_state(NextState, StateData);
+        _ ->
+            case xml:get_subtag(El, <<"method">>) of
+                false ->
+                    send_element(StateData, compress_setup_failed()),
+                    fsm_next_state(NextState, StateData);
+                Method ->
+                    case xml:get_tag_cdata(Method) of
```
michalwski: This case is too deep. I think new function here instead of case would be more readable.

ludwikbukowski:
Why tests don't pass?
- Escalus dependency is updated to https://github.com/esl/escalus/pull/97 but before there was released another PR (https://github.com/esl/MongooseIM/pull/651) which breaks backward compatibility in `escalus:create_users/2` function. 
- The fix for the issue on the Mongoose side is here https://github.com/esl/MongooseIM/pull/651 and it is now merged to `rel-1.7` branch
- To make this PR working we should merge branch `rel-1.7` into master and then rebase this PR

michalwski:
Alternatively, we can rebase this branch on top of `rel-1.7`. The amount of changes and possible conflicts (while rebasing) will be comparable.

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -790,6 +774,51 @@ wait_for_session_or_sm({xmlstreamerror, _}, StateData) ->
 wait_for_session_or_sm(closed, StateData) ->
     {stop, normal, StateData}.
 
+maybe_do_compress(El, NextState, StateData) ->
+    SockMod = (StateData#state.sockmod):get_sockmod(StateData#state.socket),
+    {Zlib, _} = StateData#state.zlib,
+    #xmlel{name = Name, attrs = Attrs, children = _} = El,
```
michalwski: How about moving this matching to the function's head?
Also this part:

    children = _

is redundant

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -790,6 +774,51 @@ wait_for_session_or_sm({xmlstreamerror, _}, StateData) ->
...
+      check_compression_method(_El, NextState, StateData)
+  end.
+
+check_compression_method(El, NextState, StateData) ->
+  case xml:get_subtag(El, <<"method">>) of
+    false ->
+      send_element(StateData, compress_setup_failed()),
+      fsm_next_state(NextState, StateData);
+    Method ->
+      case xml:get_tag_cdata(Method) of
```
michalwski: I think we can use `exml_query:path/2` function here to look for the method in line `803`. Doing that, we'll have only one case clause.

test/ejabberd_tests/tests/connect_SUITE.erl
```diff
@@ -57,6 +60,13 @@ groups() ->
                      invalid_stream_namespace]},
      {pre_xmpp_1_0, [], [pre_xmpp_1_0_stream]},
      {starttls, test_cases()},
+     {feature_order, [stream_features_test,
```
michalwski: As we are in process of adopting parallel tests I strongly recommend using the `parallel` property for this group and `escalus_fresh` module. You can take a look at this SUITE [1] for inspiration how to use it (also outside story - test case `server_announces_csi`).

[1]. https://github.com/esl/MongooseIM/blob/684bc0fbcab4c6e298961bcb7a0147c2b0678c0a/test/ejabberd_tests/tests/xep_0352_csi_SUITE.erl

test/ejabberd_tests/tests/connect_SUITE.erl
```diff
@@ -428,5 +554,9 @@ default_context(To) ->
      {to, To},
      {stream_ns, ?NS_XMPP}].
 
+node2_rpccall(Module, Function, Args) ->
```
michalwski: I suggest using function `distributed_helper:rpc/4` instead of this function.

michalwski:
Additionally I have a general comment/tip. Please set your IDE to use 4 spaces for tab and re-indent your changes. Currently you have 2 spaces.

Issue #585, opened 2015-11-16
Riak backend for mod_privacy


michalwski:

michalwski:
Many thanks @gbour!

PR #618, opened 2015-12-13
Support riak backend in mod_privacy module

gbour:
see #585

### DIFF
--- a/apps/ejabberd/src/mod_privacy_riak.erl
+++ b/apps/ejabberd/src/mod_privacy_riak.erl
@@ -17,8 +17,8 @@
 %
 % NOTES:
 %   user default privacy is stored in bucket {<<"privacy_defaults">>, LServer}, key LUser
-%   user privacy lists are stored in bucket {<<"privacy_lists">>, <<"LServer,$/,LUser">>}, key ListName
-%
+%   user privacy lists names are stored in set {<<"privacy_lists_names">>, <<"LServer">>}, key LUser
+%   user privace lists content are stored in bucket {<<"privacy_lists">>, LServer}, key <<LUser,$/,ListName>>
 
 -module(mod_privacy_riak).
 -author('guillaume@bour.cc').
@@ -40,8 +40,9 @@
 
 -include_lib("riakc/include/riakc.hrl").
 
--define(BKT_DFTS(Server), {<<"privacy_defaults">>, Server}).
--define(BKT_LSTS(Server, User), {<<"privacy_lists">>, <<Server/binary, $/, User/binary>>}).
+-define(BKT_DEFAULT_LIST(Server), {<<"privacy_defaults">>, Server}).
+-define(BKT_LISTS_NAMES(Server), {<<"privacy_lists_names">>, Server}).
+-define(BKT_LISTS(Server), {<<"privacy_lists">>, Server}).
 
 init(_Host, _Opts) ->
     ok.
@@ -65,7 +66,7 @@ get_list_names(LUser, LServer) ->
     {ok, {Default, Names}}.
 
 get_default_list_name(LUser, LServer) ->
-    case mongoose_riak:get(?BKT_DFTS(LServer), LUser) of
+    case mongoose_riak:get(?BKT_DEFAULT_LIST(LServer), LUser) of
         {ok, Obj} ->
             riakc_obj:get_value(Obj);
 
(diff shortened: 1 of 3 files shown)


michalwski:
Hi @gbour,

That's an excellent job you did. The idea with bucket for every user is interesting but also it maybe inefficient for millions of users, such situation would need to be load-tested properly.
In my head I was thinking about using 2i to get list of all available privacy lists for given user and then have buckets per server (XMPP host) and not per user. Other possible solutions is to have one set (Riak 2.0 datatype) with all privacy list names for given user so then it's also possible to keep all lists in the same bucket (per XMPP host).

gbour:
I did not found any information on riak documentation about performance impact when creating millions of buckets (just that buckets with non-default configuration are gossiped).
You"re right it should be stress-tested to evaluate the impact on the database.
I thought about storing a list of user privacy-lists, but it would require some kind of lock to avoid race conditions.

secondary index is maybe the best tradeoff

michalwski:
> I did not found any information on riak documentation about performance impact when creating millions of buckets (just that buckets with non-default configuration are gossiped).

Correct, and we have non-default configuration which will have to be gossiped.

> I thought about storing a list of user privacy-lists, but it would require some kind of lock to avoid race conditions.

Using Riak 2.0 Set datatype [1] you don't have to think about conflicts that much, they will be resolved automatically. Also in most cases only one user will modifying his own privacy list.

[1]. https://docs.basho.com/riak/latest/dev/using/data-types/

gbour:
Hi @michalslaski 
I rewrote my patch to store user's privacy_list lists in a Set as you suggested

apps/ejabberd/src/mod_privacy_riak.erl
```diff
@@ -0,0 +1,166 @@
...
+get_list_names_only(LUser, LServer) ->
+    case mongoose_riak:fetch_type(?BKT_LISTS_NAMES(LServer), LUser) of
+        {ok, Set} ->
+            riakc_set:value(Set);
+
+        {error, {notfound, set}} ->
+            [];
+
+        Err -> 
+            lager:error("~p", [Err]),
```
michalwski: Please use macro `?ERROR_MSG` defined in `ejabberd.hrl` file to be more consistent.

apps/ejabberd/src/mod_privacy_riak.erl
```diff
@@ -0,0 +1,166 @@
...
+            []
+    end.
+
+get_privacy_list(LUser, LServer, Name) ->
+    case mongoose_riak:get(?BKT_LISTS(LServer), <<LUser/binary, $/, Name/binary>>) of
+        {ok, Obj} ->
+            Val = binary_to_term(riakc_obj:get_value(Obj)),
+            {ok, Val};
+
```
michalwski: I think we can remove such empty lines separating different case clauses. This will improve code readability IMO.

apps/ejabberd/src/mod_privacy_riak.erl
```diff
@@ -0,0 +1,166 @@
...
+            mongoose_riak:update_type(?BKT_LISTS_NAMES(LServer), LUser, riakc_set:to_op(S2));
+
+        Err -> 
+            lager:error("~p", [Err]),
+            Err 
+    end.
+
+replace_privacy_list(LUser, LServer, Name, List) ->
+    % store privacy-list content
+    %NOTE: List is automatically serialized
```
michalwski: That's interesting note. Do you know how this list will look like if read by other client then Erlang?

michalwski:
Thanks for your changes. I think we are good to merge this into master. Some function specs would be more then welcome, though :)

apps/ejabberd/src/mod_privacy_riak.erl
```diff
@@ -0,0 +1,166 @@
...
+
+forget_default_list(LUser, LServer) ->
+    mongoose_riak:delete(?BKT_DEFAULT_LIST(LServer), LUser).
+
+set_default_list(LUser, LServer, Name) ->
+    case mongoose_riak:get(?BKT_LISTS(LServer), <<LUser/binary, $/, Name/binary>>) of
+        {ok, _} ->
+            case mongoose_riak:get(?BKT_DEFAULT_LIST(LServer), LUser) of
+                % update entry
+                {ok, Obj} ->
```
michalwski: There is no need to distinguish between updating and creating new entry as in the end last write wins IMO.

apps/ejabberd/src/mod_privacy_riak.erl
gbour: well, tastes and colors :)

apps/ejabberd/src/mod_privacy_riak.erl
gbour: It is serialized using BERT format. See L89 how I decode read value.
Other clients would require to decode it also.
i.e in python, you can use https://github.com/samuel/python-bert: 

    > http http://localhost:8098/types/privacy_lists/buckets/localhost/keys/jdoe |python3 -u  -c "import os;import sys;import bert; sys.stdin = os.fdopen(sys.stdin.fileno(), 'rb', 0); print(bert.codec.BERTDecoder().decode(sys.stdin.read()))"
    (Atom('foo'), Atom('bar'))

michalwski: I see, that's what I expected. Thanks for explaining this.

gbour:
I think specs would be redundant with mod_privacy callbacks definitions mod_privacy_riak implements.
I added specs for internal functions though

PR #622, opened 2015-12-18
Riak MAM for muc

michalwski:
This PR introduces simplest MAM MUC archive in Riak. MUC archives are kept in the same buckets as regular one-to-one archives. The archive is "owned" by the room only - this means:
* the message is stored only once despite the number of occupants
* everyone can read the history, even if joined after a message was archived


test/ejabberd_tests/tests/mam_SUITE.erl
```diff
@@ -476,6 +474,7 @@ init_modules(odbc_async, _, Config) ->
 init_modules(riak_timed_yz_buckets, _, Config) ->
     init_module(host(), mod_mam_riak_timed_arch_yz, [pm]),
```
arcusfelis: Should be if you want to archive MUC messages
```erlang
init_module(host(), mod_mam_riak_timed_arch_yz, [pm, muc]),
```

Issue #404, opened 2015-03-16
XEP-0045: muc#roomconfig_getmemberlist


gmodarelli:
I've noticed that it is not possible to set this value when configuring a Multi-user chat. In the XMPP-0045 extension documentation you can see that the configuration form shows this value, and you can pick the roles and affiliations that may retrieve the members list of a muc.

AFAIK in MongooseIM only Owners and Admins can retrieve the members list

https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mod_muc_room.erl#L2560

My question is: are you planning to add this feature? If not, would you accept a pull request about it?

Thank you!

michalwski:
Thanks for finding this issue. We can add this feature in future but can't promise any dates.
If you send a pull request will review and most probably accept it.

vickyrappier:
Hello Team,

I too have faced this issue, Can you please advice on this. I want the members to show the settings page like whatsapp to show who is admin and members in that group. What would be the current way to achieve this.

http://cocoadocs.org/docsets/XMPPFramework/3.6.4/Protocols/XMPPRoomDelegate.html
Referring to the above link I do see the requirement functions given as below. But how should I use them in IOS Robbiehanson framework

– xmppRoom:didFetchMembersList:
– xmppRoom:didNotFetchMembersList:

michalwski:
Fixed by #623.

@gmodarelli and @vickyrappier could you please verify if this works as you expect.

Please reopen the issue if needed.

rappier:
this is how I queried and got the response rather doing getmemberlist.

Remember Affiliation remains constant and ownership is temporary.

SEND: <iq to="mucgrp@muc.chat.com" id="memberlist" type="get"><query xmlns="http://jabber.org/protocol/muc#admin"><item affiliation="member"/></query></iq>

RECV: <iq xmlns="jabber:client" from="mucgrp@muc.chat.com" to="+61xxxxxx@chat.com/iOS" id="memberlist" type="result"><query xmlns="http://jabber.org/protocol/muc#admin">
<item affiliation="member" jid="+6123467@chat.com"><reason/></item>
<item affiliation="member" jid="+1345678@chat.com"><reason/></item>
<item affiliation="member" jid="+1456789@chat.com"><reason/></item>
</query></iq>

PR #624, opened 2015-12-21
Implementation and tests for issue #404 (XEP-00045)

bartekgorny:


test/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -1493,9 +1521,155 @@ admin_member_list(Config) ->
...
+
+        %% Now Kate requests member list, and doesn't get it because she is neither owner, admin or moderator
+        %% and the room is configured without the roomconfig_getmemberlist setting
+        escalus:send(Kate, stanza_affiliation_list_request(
+            ?config(room, Config), <<"member">>)),
+        Error = escalus:wait_for_stanza(Kate),
+        escalus:assert(is_error, [<<"auth">>, <<"forbidden">>], Error)
   end).
 
+check_memberlist(Login, yes, Config) ->
```
michalwski: The `Login` var name can be misleading. In fact the variable contains record describing client's connection.

michalwski:
This looks good. Thanks a lot!

apps/ejabberd/src/mod_muc_room.erl
```diff
@@ -3542,6 +3586,13 @@ set_xoption([{<<"muc#roomconfig_maxusers">>, [Val]} | Opts], Config) ->
     _ ->
         ?SET_NAT_XOPT(max_users, Val)
     end;
+set_xoption([{<<"muc#roomconfig_getmemberlist">>, Val} | Opts], Config) ->
+    case Val of
+        <<"none">> ->
+            ?SET_XOPT(maygetmemberlist, []);
+        _ ->
+            ?SET_XOPT(maygetmemberlist, [binary_to_atom(V, latin1) || V <- Val])
```
michalwski: `binary_to_atom` is very dangerous if we convert any input data from any user to an atom. It would be safer to use `binary_to_existing_atom` provided all acceptable atoms are created (for by typing them in source code). There is good chance they are already created.

bartekgorny: You're right, I'll fix it.

michalwski:
Thanks a lot for these changes!

PR #627, opened 2015-12-26
mod_http_notification and mod_http_notification_SUITE for sending htt…

vooolll:
…p messages, modifications in ejabberd.cfg,node1_vars.config, node2_vars.config to this add module

### DIFF
--- a/apps/ejabberd/src/mod_http_notification.erl
+++ b/apps/ejabberd/src/mod_http_notification.erl
@@ -1,0 +1,70 @@
+%%%----------------------------------------------------------------------
+%%% File    : mod_notify.erl
+%%% Author  : Baibossynv Valery <baibossynov.valery@gmail.com>
+%%% Purpose : Message passing via http
+%%% Created : 16 Dec 2015 by Baibossynv Valery <baibossynov.valery@gmail.com>
+%%%----------------------------------------------------------------------
+
+-module(mod_http_notification).
+-author("baibossynov.valery@gmail.com").
+-behaviour(gen_mod).
+
+%% API
+-export([start/2, stop/1, on_user_send_packet/3]).
+
+-include("ejabberd.hrl").
+-include("jlib.hrl").
+
+-define(DEFAULT_HTTP_POOL_SIZE, 5).
+-define(DEFAULT_HTTP_HOST, "http://localhost").
+-define(DEFAULT_PREFIX_PATH, <<"/">>).
+
+start(Host, _Opts) ->
+  HttpHost = gen_mod:get_module_opt(Host, ?MODULE, host, ?DEFAULT_HTTP_HOST),
+  PoolSize = gen_mod:get_module_opt(Host, ?MODULE, pool_size, ?DEFAULT_HTTP_POOL_SIZE),
+  ChildMods = [fusco],
+  ChildMF = {fusco, start_link},
+  ChildArgs = {for_all, [HttpHost, []]},
+  {ok, _} = supervisor:start_child(ejabberd_sup,
+    {{http_notification_sup, Host},
+      {cuesport, start_link,
+        [pool_name(Host), PoolSize, ChildMods, ChildMF, ChildArgs]},
+      transient, 2000, supervisor, [cuesport | ChildMods]}),
+  ejabberd_hooks:add(user_send_packet, Host, ?MODULE, on_user_send_packet, 100),
+  ok.
+
+stop(Host) ->
+  %%TODO: delete children of supervisor
+  ejabberd_hooks:delete(user_send_packet, Host, ?MODULE, on_user_send_packet, 100),
+  ok.
+
+on_user_send_packet(From, To, Packet) ->
+  Type = xml:get_tag_attr_s(list_to_binary("type"), Packet),
+  Body = xml:get_path_s(Packet, [{elem, list_to_binary("body")}, cdata]),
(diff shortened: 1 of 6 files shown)


apps/ejabberd/src/mod_http_notification.erl
```diff
@@ -0,0 +1,70 @@
...
+  {ok, _} = supervisor:start_child(ejabberd_sup,
+    {{http_notification_sup, Host},
+      {cuesport, start_link,
+        [pool_name(Host), PoolSize, ChildMods, ChildMF, ChildArgs]},
+      transient, 2000, supervisor, [cuesport | ChildMods]}),
+  ejabberd_hooks:add(user_send_packet, Host, ?MODULE, on_user_send_packet, 100),
+  ok.
+
+stop(Host) ->
+  %%TODO: delete children of supervisor
```
michalwski: Yes, this has to be done :)

apps/ejabberd/src/mod_http_notification.erl
```diff
@@ -0,0 +1,70 @@
...
+  ejabberd_hooks:delete(user_send_packet, Host, ?MODULE, on_user_send_packet, 100),
+  ok.
+
+on_user_send_packet(From, To, Packet) ->
+  Type = xml:get_tag_attr_s(list_to_binary("type"), Packet),
```
michalwski: I suggest using just `<<"type">>` here.

apps/ejabberd/src/mod_http_notification.erl
```diff
@@ -0,0 +1,70 @@
...
+  Body = xml:get_path_s(Packet, [{elem, list_to_binary("body")}, cdata]),
```
michalwski: As we are going towards `exml` application instead of the `xml` module I suggest using `exml_query` [1] interface to get the element.

[1]. https://github.com/esl/exml/blob/master/src/exml_query.erl

apps/ejabberd/src/mod_http_notification.erl
```diff
@@ -0,0 +1,70 @@
...
+  if (Type == <<"chat">>) and (Body /= <<"">>) ->
```
michalwski: `if`s are not very readable in Erlang. I think new function like the following:

```erlang
maybe_make_req(<<"chat">>, Body, OtherArg1, OtherArg2, ...) when Body /= <<"">> ->
   make_req(post, ...);
maybe_make_req(_, _, ...) ->
   ok.
```

test/ejabberd_tests/http.specs
```diff
@@ -0,0 +1,7 @@
+%% do not remove below SUITE if testing mongoose
+{suites, "tests", mod_http_notification_SUITE}.
```
michalwski: Please add the `mod_http_notification_SUITE` to `default.spec` so travis can test it.

rel/files/ejabberd.cfg
```diff
@@ -626,10 +626,11 @@
   {{mod_offline}}
   {{mod_privacy}}
   {{mod_private}}
+  {{mod_http_notification}}
 % {mod_private, [{backend, mnesia}]},
 % {mod_private, [{backend, odbc}]},
   {mod_register, [
-		  %%
+		  %%s
```
michalwski: The new `s` here is probably a typo.

apps/ejabberd/src/mod_http_notification.erl
```diff
@@ -0,0 +1,70 @@
...
+  end.
+
+make_req(Method, Host, Sender, Receiver, Message) ->
+  PathPrefix = gen_mod:get_module_opt(Host, ?MODULE, prefix_path, ?DEFAULT_PREFIX_PATH),
+  Connection = cuesport:get_worker(existing_pool_name(Host)),
+  Query = <<"author=", Sender/binary, "&server=", Host/binary, "&reciever=", Receiver/binary, "&message=", Message/binary>>,
+  ?INFO_MSG("Making request '~s' for user ~s@~s...", [PathPrefix, Sender, Host]),
+  Header = [{<<"Content-Type">>, <<"application/x-www-form-urlencoded">>}],
+  {ok, {{Code, _Reason}, _RespHeaders, RespBody, _, _}} = case Method of
+    get -> fusco:request(Connection, <<PathPrefix/binary, "?", Query/binary>>,
```
michalwski: Looks like the only possible value of `Method` var is `post`. I suggest simply removing the variable and not considering `get` value.

vooolll:
May I return something from on_user_send_packet function (in test purpose)?

apps/ejabberd/src/mod_http_notification.erl
vooolll: It should be also done in ejabberd_http_auth I guess.

michalwski:
Hi @vooolll,

Any progress on this?

Re the `on_user_send_packet` function, yes, you can return sth from this function as long as it is not atom `stop`.

apps/ejabberd/src/mod_http_notification.erl
michalwski: More over I'd like to see this test configurable. The default behaviour can be exactly the same as here, but it would be great if we could override it be specifying some different callback module. An inspiration how to do it can be found here: https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mod_mam.erl#L947

michalwski:
This feature has been moved to #684. I'm closing this one.

PR #628, opened 2016-01-04
FIPS mode support

michalwski:
This PR allows to enable FIPS mode provided a custom version of Erlang/OTP is used. More information can be found in OpenSSL-and-FIPS.md file.


apps/ejabberd/rebar.config.script
```diff
@@ -21,15 +11,49 @@ GetFullVer = fun() ->
...
+           Config
+    end
+end,
+
+MaybeChangeOpenSSL = fun(Config) ->
+    case os:getenv("OPENSSL_LIB") of
+        false ->
+            Config;
+        OpenSSL ->
+            io:format("using OpenSSL ~p~n", [OpenSSL]),
```
lavrin: This clause body is on a different level of abstraction, so might use exporting to a well named function.

PR #631, opened 2016-01-05
Ldap vcard search operator and binary fields

michalwski:
This PR:
* allows to configure ldap vcard search operator (default is 'and'),
* allows to specify which filelds should be base64 encoded before sending to a client as a search result


lavrin:
It seems `apps/ejabberd/src/mod_vcard_ldap.erl.orig` accidentally got checked in.

PR #633, opened 2016-01-08
OAuth-like token support

erszcz:
This PR implements a token authentication feature similar to [OAuth][1] which allows for secure access to the system not requiring a user to provide password on each login or storage of username / password or any other vulnerable data on a user's device.

The main simplification in comparison with OAuth is that MongooseIM is both the authorization (it generates tokens) and resource (it allows for access to the XMPP network) server.

Changes mainly affect `cyrsasl` and introduce new modules: `mod_auth_token` responsible for token generation and verification, and `mod_keystore`, which provides access to on-disk as a well as RAM-only keys. There's no key format expected for persistent keys - arbitrary binary data will be accepted as of now. RAM keys are generated on cluster startup and are lost on shutdown of the last cluster member. All tokens become invalid after their expiry date is reached. `refresh` tokens can also be revoked through the command line interface. `access` tokens can't be revoked, as the server doesn't track the `access` tokens it generates.

A new CYRSASL mechanism, X-OAUTH, is introduced. When a user connecting to the system chooses this mechanism, then the only challenge required by the system is a Base64 encoded token of type `access`, `refresh` or `provision`. Token types are further described in `mod_auth_token` documentation. In order to obtain an `access` or `refresh` token the user has to login with another CYRSASL mechanism at least once - requesting a token from the server is only possible once an XMPP session is established; `provision` tokens ought to be obtained from an external authorization server. Once the user is in possession of a token, (s)he can use X-OAUTH. Successful authentication results in a new set of tokens being returned with the `<success/>` element. During an XMPP session, a token is obtained with a new kind of IQ (see `mod_auth_token` doc).

**Caveat:** when a user is logged into the system and his/her token gets compromised, the operator can revoke the `refresh` token. However, **the `access` token can't be revoked and will be valid until expiration**, therefore will allow the user to login again to the system even if he/she is kicked by the operator. 

[1]: http://tools.ietf.org/html/rfc6749
[2]: https://developers.google.com/talk/jep_extensions/oauth


test/ejabberd_tests/full.spec
```diff
@@ -16,6 +16,7 @@
 {suites, "tests", login_SUITE}.
 {suites, "tests", muc_SUITE}.
 {suites, "tests", mam_SUITE}.
+{suites, "tests", oauth_SUITE}.
```
michalwski: In order to use this SUITE in travis build, `default.spec` file has to be modified as well.

apps/ejabberd/cover.spec
```diff
@@ -1,4 +1,4 @@
-{export, "/tmp/mognoose_ct.coverdata"}.
+{export, "mognoose_ct.coverdata"}.
```
michalwski: This will break current coverage reporting. See [run_common_test.erl#L221](https://github.com/esl/MongooseIM/blob/db630b7385aac8376d7c8b538f6ee632a585789e/test/ejabberd_tests/run_common_test.erl#L221)

michalwski:
I see token storage is possible only with Postgres, how about MySQL. To me it looks like there is no Pgsql specific features.

lavrin:
It's true there should be no problems with MySQL. However, I'd leave it for a separate PR as my _best effort_ pool of resources is drying up ;)

lavrin:
Food for thought. If we continue using this feature then:
- token format description should be extracted from `mod_auth_token` to a callback module (or 2/3 separate modules depending on token type - there are 3 types now: `access`, `refresh`, `provision`),
- a custom serialization / on-the-wire format is used now; we should probably use / add support for [JSON Web Tokens](http://jwt.io/) as it seems to gain traction (though there's a [security caveat to be aware of](https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/)).

erszcz:
Now rebased onto #651.

emnvn:
Hello,

I have tried to use access token and refresh token to authenticate to XMPP server. After login by user/pass I request this pair of token by seeing this xml stanza
`<iq from='test01@dtsc.com.vn' to='test01@dtsc.com.vn/87B0EA7F1BEE07C01469778484314618' id='55850c43-e9e7-4508-b4e6-fc10a40f4c8b:sendIQ' type='result' xmlns='jabber:client'><items xmlns='erlang-solutions.com:xmpp:token-auth:0'><access_token xmlns='erlang-solutions.com:xmpp:token-auth:0'>YWNjZXNzAHRlc3QwMUBkdHNjLmNvbS52bi84N0IwRUE3RjFCRUUwN0MwMTQ2OTc3ODQ4NDMxNDYxOAA2MzYzNzAwMzQ1MQAwMmRlYzM0MTQyNzczOTY5ZTFlMjQzOGM2OGNhZDA4MGVjZjMzMjViNTIzOTZkMmE0YTU0ZjMyNDJlNmJhZjRlZTUzMTI2NDc3NWM5MWM0MmViMTNmMTRjOTQ3ZDhlOTM=</access_token><refresh_token xmlns='erlang-solutions.com:xmpp:token-auth:0'>cmVmcmVzaAB0ZXN0MDFAZHRzYy5jb20udm4vODdCMEVBN0YxQkVFMDdDMDE0Njk3Nzg0ODQzMTQ2MTgANjM2MzcwODYyNTEAMQBiOTE2NGUwNDc1MGExNDFlZTVlZjVhNWJjNGFmZmU3ODc2Mzc4MzFiZDY1MmZiYjZjMjE4NWU4ZGM5OWU0NjI2ZDdhNzJkODA2YWNlYjU4OGY0NTY5ZTE4OWVkZmY3Y2U=</refresh_token></items></iq>`

Then I use the access token to authenticate with xmpp server, I sent this xml stanza

`<auth xmlns='urn:ietf:params:xml:ns:xmpp-sasl' mechanism='X-OAUTH'>YWNjZXNzAHRlc3QwMUBkdHNjLmNvbS52bi84N0IwRUE3RjFCRUUwN0MwMTQ2OTc3ODQ4NDMxNDYxOAA2MzYzNzAwMzQ1MQAwMmRlYzM0MTQyNzczOTY5ZTFlMjQzOGM2OGNhZDA4MGVjZjMzMjViNTIzOTZkMmE0YTU0ZjMyNDJlNmJhZjRlZTUzMTI2NDc3NWM5MWM0MmViMTNmMTRjOTQ3ZDhlOTM=</auth>`

But I can't receive any response from xmpp server. I saw this message which is received by xmpp server. I don't know what is my mistake ?

Hope that someone can help me.

Thanks.

michalwski:
Hi @emnvn ,

Thanks for contacting us. Could you please open a new issue with your problem? Also please show us your `ejabberd.cfg` file (remove all passwords before sending :) ) and server logs - `ejabberd.log` file.

emnvn:
Hello @michalslaski ,

I have created this issue https://github.com/esl/MongooseIM/issues/915 

Thanks for helping me.

emnvn:
Hi @michalwski 

I have created this issue https://github.com/esl/MongooseIM/issues/915

Hope that you can check it.

Thanks

PR #640, opened 2016-01-15
External component adjustments

studzien:
This PR:
* changes the routing mechanism - the new `external_component` table is checked firstly
* allows to use external_component registered on different node in the cluster
* allows to expose `ejabberd_service` over WebSockets
* includes changes to tests:
    * new dev node for federation
    * old node2 is now meant to be used only for clustering with node2
    * new `component_SUITE` testing ejabberd_service interface
    * `reload_helper` module to be used whenever a node reload is required

Todo:
- [x] fix tests :smile: 
- [x] fix warnings :smile:
- [ ] make ``is_subdomain`` feature configurable (default disabled)
- [ ] make the component routing via additional routing table (``external_component``) optional
- [ ] modify the component routing so that first it's checked if the component is registered locally, route to an externally registered service as a last resort (allow to register the same component on different nodes)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
 before_script:
         - tools/travis-setup-db.sh
         - if [ $PRESET = 'ldap_mnesia' ]; then sudo tools/travis-setup-ldap.sh; fi
-script: tools/travis-test.sh
+script: tools/travis-test.sh $PRESET
 
 after_success:
         - make cover_report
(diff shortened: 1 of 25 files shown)


erszcz:
Closing and reopening to trigger Travis.

michalwski:
Great job @erszcz and @studzien! I'm finally merging this!

PR #643, opened 2016-01-18
Export CT logs from Travis

arcusfelis:
Git example: https://github.com/arcusfelis/mongooseim-ct-reports
HTML example: http://arcusfelis.github.io/mongooseim-ct-reports/1186.1/

Basically, we upload failed reports on github pages.
We asked Travis to encrypt SSH deploy key for arcusfelis/mongooseim-ct-reports (repo for pages).
SSH key is only available for trusted builds (local branches under esl account).
After this we checkout pages, add ct_report directory, commit.
Using github for the pages should not cause problems because of low number of failed builds. And we do nothing if build passes.

I will write a blog post how to do this.


lavrin:
This looks pretty nice! Can we place the whole code used to publish the reports in a script? E.g. `tools/travis-publish-ct-reports.sh`?

michalwski:
Thanks for this @arcusfelis, that's really useful. Could you please create the same repo under esl's github account? I think it would be better to publish results to esl/mongooseim-ct-reports

PR #647, opened 2016-01-21
Remove unneeded timer:sleep/1 calls from tests

arcusfelis:


arcusfelis:
It will not decrease test execution time because the longest part of each test is connecting clients.
But we can remove strange `sleep(1500)` with magic numbers.

For example, if we delete from archive we don't really need to wait 1.5 seconds. We can just check archive size several times until it's empty.

erszcz:
Well, it doesn't guarantee shortening test execution times by a huge amount, but it shaves a few seconds here and there.
Good thinking and a step in the right direction!
Is there still anything more to be added to this PR or is it OK to merge?

michalwski:
This was merged along with #668

PR #654, opened 2016-01-26
Fix bugs in mod_mam_odbc_prefs

arcusfelis:
Add tests for MAM preferences
Bug 1: JIDs are not normalized before stored in preferences
Bug 2: Result from SQL query never matches in get_behaviour


arcusfelis:
Progress so far:
- More tests for MAM prefs.
- Fixed mam_*_prefs modules. None of them worked.
- Reduced mam_SUITE execution time

arcusfelis:
Ready for review.

erszcz:
Let's not forget to remove `apps/ejabberd/src/mod_mam_mnesia_prefs.erl` and the relevant tests once this PR is merged. To make things simpler, let it be another PR.

The reason is that it duplicates functionality of `mod_mam_mnesia_dirty_prefs` while also probably being less efficient. All in all, MAM seems complex enough right now, we don't need more modules doing the same thing (as it was with `mod_mam_user` and `mod_mam_server_user`) and hyper-configurability to the degree of madness ;)

PR #657, opened 2016-02-01
Store messages in XML

arcusfelis:
mod_mam_odbc_arch and mod_mam_muc_odbc_arch gets a new parameter `simple`.
Messages are stored in XML and JIDs are stored as `user@server/resource`.

By default `simple` parameter is `false` to remain compatible with previous configurations.

+ `hand_made_partitions` is not an macro anymore. This option that allows to store messages in 16 separate tables to improve scalability. By default disabled.


apps/ejabberd/src/mod_mam_muc_odbc_arch.erl
```diff
@@ -662,3 +650,78 @@ maybe_encode_compact_uuid(undefined, _) ->
     undefined;
 maybe_encode_compact_uuid(Microseconds, NodeID) ->
     encode_compact_uuid(Microseconds, NodeID).
+
+
+%% ----------------------------------------------------------------------
+%% Optimizations
+
+packet_to_stored_binary(Packet) ->
+    case odbc_message_format() of
```
michalwski: I think it would be better to implement this case as callback. That way it will be easier to add new formats, for example encrypted message.

michalwski: The same comment applies to all other places where there is similar case like this one.

arcusfelis: I suggest add another option `odbc_message_format() -> {callback, fun .....}`.
So, it will be:
- compressed_term
- term
- xml
- `{callback, fun my_mod:packet_to_stored_binary/1}`

arcusfelis: Something like:

```erlang
packet_to_stored_binary(Packet) ->                                                   
    case odbc_message_format() of                                                    
        compressed_term ->                                                           
            term_to_binary(Packet, [compressed]);                                    
        term ->                                                                      
            term_to_binary(Packet, []);                                              
        xml ->                                                                       
            exml:to_binary(Packet);                                                  
        {callback, Encoder, _Decoder} ->                                             
            Encoder(Packet)                                                          
    end.                                                                             
                                                                                     
stored_binary_to_packet(Bin) ->                                                      
    case odbc_message_format() of                                                    
        compressed_term ->                                                           
            binary_to_term(Bin);                                                     
        term ->                                                                      
            binary_to_term(Bin);                                                     
        xml ->                                                                       
            {ok, Packet} = exml:parse(Bin),                                          
            Packet;                                                                  
        {callback, _Encoder, Decoder} ->                                             
            Decoder(Packet)                                                          
    end. 
```

michalwski: I'd even go with this:

```Erlang
packet_to_stored_binary(Packet, CallBackModule) ->
    CallBackModule:encode(Packet).

sotred_binary_to_packet(Bin, CallBackModule) ->
   CallBackModule:decode(Packet)
```

michalwski:
Also this encoding should apply to other backends like Riak and Cassandra.

erszcz:
:+1: for:

```erlang
packet_to_stored_binary(Packet, Format) ->
    Format:encode(Packet).

stored_binary_to_packet(Bin, Format) ->
    Format:decode(Packet)
```

It makes even more sense to use this approach when taking the later Michal's comment into account (i.e. serialization format should be configurable for all backends, not select few). Then we have a clean call graph with no duplicated "special cases" in different backend modules:

```
                     mod_mam
          |           /   \
calls     |          /     \
          v         /       \
                  odbc     riak
          |         \       /
calls     |          \     /
          v           \   /
                   serialization
                      format
```

apps/ejabberd/priv/mysql.sql
```diff
@@ -161,6 +161,30 @@ CREATE TABLE roster_version (
 -- ALTER TABLE rosterusers ADD COLUMN askmessage text AFTER ask;
 -- UPDATE rosterusers SET askmessage = '';
 -- ALTER TABLE rosterusers ALTER COLUMN askmessage SET NOT NULL;
+--
+-- NOTE: "minified" JID is specially encoded form of JID storing only
+--       between user's JID and the real JID.
```
erszcz: > storing only >the difference< between

Seems _the difference_ somehow got omitted, didn't it?

erszcz:
Is it possible to include changes to Riak backend in the same PR?
If there's some reason not to, :+1: from me and ready to merge.

arcusfelis:
I added encoding of messages for Riak.
I didn't add encoding of JIDs because some of JIDs are parts of keys. So, some JIDs will be minified and some not in Riak.

apps/ejabberd/priv/mysql.sql
erszcz: ping

PR #658, opened 2016-02-01
Replace mod_mam_mnesia_prefs with mod_mam_mnesia_dirty_prefs

arcusfelis:
Remove one of mnesia backends for MAM preferences. Because we don't need two.

+ And "revert" deleting users from mam_user table automatically. It breaks tests. It was merged in the previous PR for MAM. I will investigate why there the tests fail later. It's good to go to the prev state for now. I added auto_remove option for the module (it's good to have it for modules that archive messages too because some of our users would like to keep messages, maybe should add it after discussion).


apps/ejabberd/src/mod_mam_mnesia_prefs.erl
```diff
@@ -117,172 +120,145 @@ stop_muc(Host) ->
 %% ----------------------------------------------------------------------
 %% Internal functions and callbacks
 
--spec get_behaviour(DefaultBvr :: mod_mam:archive_behaviour(),
-        _Host :: ejabberd:server(), ArcId :: mod_mam:archive_id(),
-        LocJID :: ejabberd:jid(), RemJID :: ejabberd:jid())
-            -> mod_mam:archive_behaviour().
-get_behaviour(DefaultBehaviour, _Host, _ArcID,
+-spec get_behaviour(Default :: behaviour(), Host :: ejabberd:server(),
```
erszcz: Not a big deal, but regarding style, wouldn't it be better to format it like this, if we have a spec so long that it spans multiple lines?

```erlang
-spec get_behaviour(DefaultBh, Host, ArcId, LocJID, RemJID) -> Result when
      DefaultBh :: mod_mam:archive_behaviour(),
      Host :: ejabberd:server(),
      ArcId :: mod_mam:archive_id(),
      LocJID :: ejabberd:jid(),
      RemJID :: ejabberd:jid(),
      Result :: mod_mam:archive_behaviour().
```

On one hand _de gustibus non disputandum est_, but on the other it seems much more readable to me that way. For specs that make it into a single line _inline_ types are OK from my point of view.

erszcz:
Any idea at what number of users the DETS 2GiB file size limit caps us? Probably "enough", but it never hurts to know.

apps/ejabberd/src/mod_mam_mnesia_prefs.erl
arcusfelis: I like your way more.

arcusfelis:
erszcz Some users can have large block lists. 2GiB problem does not concern me a lot because we can always use partitions. But starting time can be high (a real issue). Anyway, on big installations this data probably will go to redis.

erszcz:
No vetos, Travis is OK, so I'm merging. Good stuff!

PR #659, opened 2016-02-02
Added integration with Buildtimetrend

astachurski:
Added basic integration with buildtimetrend service for easy and nice inspection of various build parameters.
More info on the service is available at: http://buildtimetrend.herokuapp.com
Other projects are available in dashboard/projects section. Not much can be said about quality of this service as it's a new thing. But looks as nice complement to travis and coveralls. More badges and better resolution (inter-test-suite statistics) will be added over time as we find more about this service.


michalwski:
I suggest squashing some of the commits with rebase and leaving at most 1 or 2.

PR #665, opened 2016-02-03
Fix security issue CVE-2016-1232

michalwski:
This PR was inspired by security issue CVE-2016-1232:
* implements XEP-0185
* changes random string generator to make it harder to guess the string


arcusfelis:
Looks fine.

PR #666, opened 2016-02-03
preserve resume timeout when a session is resumed

jfjalburquerque:
In a resumption, resume timeout is not merged, so it will get the default resume timeout instead of the timeout of the configuration file.


michalwski:
Good catch! Many thanks for your patch!

PR #668, opened 2016-02-04
Support for new versions of XEP-0313 MAM

arcusfelis:
All changes:
- Support for MAM v0.3, v0.4.1, v0.5. There are no major changes in v0.5.
- A client should specify appropriate namespace for IQ's he sends. After that a server decides which version should be used.
- Override is_complete_packet/3 (for cases when you want to choose what to archive).
- Make `<archived/>` element configurable.
- Fix access rights for MAM MUC (anyone was able to read the archive).
- no-store hint (client can add it to disable storing message).
- mam_SUITE was extended to support three MAM versions.
- Prefs backends were updated to support not-implemented error.
- ejabberd.cfg config change: `{access, mam_lookup_messages, [{default, all}]}.` becomes optional. The old variant is supported too (i.e. no need for config update).
- filter_room_packet was extended with more arguments.

| namespace | XEP version | Result stanzas | Query using forms |
|-------------|---------------|-------------|-----|
| urn:xmpp:mam:tmp | 0.2 | messages, iq with rsm | no |
| urn:xmpp:mam:0 | 0.3 | iq, messages, fin message with rsm | yes |
| urn:xmpp:mam:1 | 0.4.1 and 0.5 | messages, fin iq with rsm | yes |

Spec diffs

- http://xmpp.org/extensions/diff/api/xep/0313/diff/0.2/vs/0.3
- http://xmpp.org/extensions/diff/api/xep/0313/diff/0.3/vs/0.4.1
- http://xmpp.org/extensions/diff/api/xep/0313/diff/0.4.1/vs/0.5


arcusfelis:
TODO next. Archive this for mod_mam_muc based on XEP v0.5. We can patch packet adding this element before storing. After extracting we can remove it depending on access rights.

```erlang
        <x xmlns='http://jabber.org/protocol/muc#user'>
          <item affiliation='none'
                jid='witch2@shakespeare.lit'
                role='participant' />
```

arcusfelis:
Cherry-picked https://github.com/esl/MongooseIM/pull/647 into this branch (because we reached 48 minutes limit)

michalwski:
I see that `<archived>` is used only in version 0.2. If I understand @arcusfelis's changes correctly, this element will be added for other clients using newer version if we configure the server to do it. @arcusfelis , correct me if I'm wrong.

arcusfelis:
By default it's enabled for all clients.
But it's possible to disable it using config for all clients.

michalwski:
I think this should be at least disabled by default or determined in some other way wether to add this tag or not.

arcusfelis:
Fine. We can disable it by default.

apps/ejabberd/src/mod_muc_room.erl
```diff
@@ -714,6 +742,39 @@ can_send_to_conference(From, StateData) ->
     orelse
     is_allowed_nonparticipant(From, StateData).
 
+can_read_conference(UserJID,
```
arcusfelis: Looks scary

arcusfelis:
It works for Kaiwa (stanza.io 7.3.4). It uses MAM v0.3.

PR #681, opened 2016-02-22
Remove mam index duplicates

arcusfelis:
Use this query under the root mysql user to check which indexes are created:

```sql
select database_name, table_name, index_name, stat_value*@@innodb_page_size
from mysql.innodb_index_stats where stat_name='size';
```

Basically,

```sql
... PRIMARY KEY (my_key) ... // create BTREE with data in leafs
CREATE INDEX another_index (my_key); // create BTREE with pointers on data in leafs
```

creates two indexes.
This second index can affect performance.

So, the pull request removes unwanted index duplicates.


michalwski:
How about pgsql and mssql? I don't see any changes in these schema files.

arcusfelis:
@michalwski OK, I will check them too. Maybe they are "smart enough" not to create duplicates physically.

PR #683, opened 2016-02-23
Join/leave cluster completely at the Erlang level

erszcz:
Apart from the rationale in 20d40e8 description, there's one more reason: using just RPC (instead of a shell-out) to cluster/disband nodes makes it possible to keep cover-compiled modules loaded through the whole cluster life cycle. This is required to make rel-1.7-service tests finally pass.

### DIFF
--- a/rel/files/mongooseim
+++ b/rel/files/mongooseim
@@ -118,7 +118,7 @@ case "$1" in
         if [ "$RES2" = "pong" ]; then
             RES3=`$ADD_TO_CLUSTER_CMD`
             if [ "$RES3" = "ok" ]; then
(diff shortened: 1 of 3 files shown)


rel/files/mongooseim
```diff
@@ -118,7 +118,7 @@ case "$1" in
         if [ "$RES2" = "pong" ]; then
             RES3=`$ADD_TO_CLUSTER_CMD`
             if [ "$RES3" = "ok" ]; then
-                echo "Node added to cluster. Run mongooseimctl live or mongooseimclt start."
+                echo "Node added to cluster. Run mongooseimctl live or mongooseimctl start."
```
ludwikbukowski: You win a golden magnifying glass !

lupa

PR #684, opened 2016-02-23
http notification plugin

bartekgorny:
This is the vooolll's mod_http_notification, with fixes. Also, maybe it would be worthwhile to install some safety valve in case there's lot of messages and http service does not respond - we have to discuss it.

### DIFF
--- a/apps/ejabberd/src/mod_http_notification.erl
+++ b/apps/ejabberd/src/mod_http_notification.erl
@@ -66,7 +66,7 @@ get_callback_module() ->
 make_req(Host, Sender, Receiver, Message) ->
   PathPrefix = gen_mod:get_module_opt(Host, ?MODULE, prefix_path, ?DEFAULT_PREFIX_PATH),
   Connection = cuesport:get_worker(existing_pool_name(Host)),
-  Query = <<"author=", Sender/binary, "&server=", Host/binary, "&reciever=", Receiver/binary, "&message=", Message/binary>>,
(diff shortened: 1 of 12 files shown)


apps/ejabberd/src/mod_http_notification.erl
```diff
@@ -0,0 +1,81 @@
...
+  supervisor:terminate_child(ejabberd_sup, Ch),
+  supervisor:delete_child(ejabberd_sup, Ch),
+  ejabberd_hooks:delete(user_send_packet, Host, ?MODULE, on_user_send_packet, 100),
+  ok.
+
+on_user_send_packet(From, To, Packet) ->
+  Type = exml_query:attr(Packet, <<"type">>, <<>>),
+  Body = exml_query:path(Packet, [{element, <<"body">>}, cdata], <<>>),
+  Mod = get_callback_module(),
+  case Mod:should_make_req(Type, Body, From, To) of
```
michalwski: How about passing `Packet` instead of `Type` and `Body` to the callback function and doing whatever is need in the implementation? This would allow us to write more accurate checkers.

test/ejabberd_tests/http.specs
```diff
@@ -0,0 +1,7 @@
+%% do not remove below SUITE if testing mongoose
```
michalwski: The `http.specs` file is not needed in repo, I think.

vooolll:
@bartekgorny Thanks a lot for your work, have a problem with finishing my pr's.

PR #700, opened 2016-03-07
Carbons in sessions

pzel:
This PR addresses #[not-assigned]

Per-session carbon copy status (enabled/disabled/version) is stored as part of `#session.info` in the `session` table. This eliminates the need for a separate `carboncopy` table, and reduces the number of lookups made by `mod_carboncopy`, per message route event, from 2 to 1. Additionally, no cleanup phase is necessary when a node goes down, as carbons get cleaned up as part of the `session` table cleanup.

To allow this, the `ejabberd_sm_*` backends must be capable of merging `info` proplists when overwriting a user session with `create_session`. This is conveniently achieved via `ejabberd_session:merge_info(#session{}, #session{}) -> #session{}`

To store information for a given xmpp full jid, use `ejabberd_sm:store_info(user(), server(), resource(), {any(), any()}) -> {'ok', {any(), any()}} | {error, offline}`. 

If a given extension module furnishes a user session with extra info that is only valid for that session, this interface is preferred to holding a separate ets table with *de facto* foreign keys to `session`, and risking synchronization issues. `mod_carboncopy` can be used as a template for this pattern.

The unit test suite has been extended to make sure that `ejabberd_sm:store_info` and `ejabberd_sm:create_session` uphold the properties stated above.

The `rebar` noise suppression mechanism in the `Makefile` has been extracted as a Make macro

### DIFF
--- a/apps/ejabberd/src/ejabberd_session.erl
+++ b/apps/ejabberd/src/ejabberd_session.erl
@@ -1,0 +1,15 @@
+-module(ejabberd_session).
+-export([merge_info/2]).
+
+-include("ejabberd.hrl").
+
+-spec merge_info(#session{}, #session{}) -> #session{}.
+merge_info(New, Old) ->
+    NewInfo = orddict:from_list(New#session.info),
+    OldInfo = orddict:from_list(Old#session.info),
+    New#session{info =
+                    orddict:to_list(orddict:merge(fun merger/3,
+                                                  NewInfo, OldInfo))}.
+
+merger(_Key, NewVal, _OldVal) ->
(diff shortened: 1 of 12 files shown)


apps/ejabberd/src/ejabberd_session.erl
```diff
@@ -0,0 +1,15 @@
+-module(ejabberd_session).
```
michalwski: How about naming this module `mongoose_session` as this is completely new module?

pzel: Will do!

Issue #680, opened 2016-02-22
Implement XEP-0352: Client State Indication


michalwski:
The XEP-0352: Client State Indication [1] can be used by mobile application when the app goes to background. Good example of an app using this XEP is Conversation for Android.

The implementation and especially tests have to take Stream Management [2] into account. This 2 extensions have to work correctly together.

[1]. http://xmpp.org/extensions/xep-0352.html
[2]. http://xmpp.org/extensions/xep-0198.html

PR #703, opened 2016-03-11
Xep 0352 csi

michalwski:
This PR addresses #680

Proposed changes include:
* changes to ejabberd_c2s handling `inactive` and `active` states
* new `xep_0352_csi_SUITE`

TODO:

- [x] documentation (also how this works with sm enabled)
- [x] tests with stream management enabled


apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -857,13 +859,25 @@ session_established({xmlstreamelement,
     stream_mgmt_handle_ack(session_established, El, StateData);
 
 session_established({xmlstreamelement,
-                     #xmlel{name = <<"r">>} = El}, StateData) ->
+                     #xmlel{name = <<"r">>} = El},
+                    #state{csi_state = active} = StateData) ->
```
studzien: Missing space

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -2272,7 +2297,7 @@ is_ip_blacklisted({IP,_Port}) ->
 %% @doc Check from attributes.
 -spec check_from(El, FromJID) -> Result when
       El :: jlib:xmlel(), FromJID :: ejabberd:jid(),
-      Result :: 'invalid-from'  | jlib:xmlel().
+                Result :: 'invalid-from'  | jlib:xmlel().
```
studzien: Too many spaces? :)

studzien:
Does it make sense to buffer the incoming stanzas as well?
The XEP leaves freedom of choice for the server implementors, but what if...

- the client sends a message right after sending the <inactive> nonza. Do we still want to deliver and archive this message ASAP or when the client becomes active again (possibly in a month)?
- the client sends an IQ right after sending the same nonza. Do we want to send the result at time of sending or becoming active again?

I think that the main problem is that the XEP specifies almost nothing, especially not what should with stanzas sent by the client when it is inactive.

studzien:
Also, I believe that buffer size should be configurable from the config file.

michalwski:
I've sent a message to xmpp standards mailing list. Here's the thread: http://mail.jabber.org/pipermail/standards/2016-March/031012.html.

My personal opinion after re-thinking is that the server should discard the stanza and reply with an error to the client.

michalwski:
Based on the conversation on the standrads mailing list I decided not to buffer incoming stanzas, but pass as usual.

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -870,6 +870,7 @@ do_open_session_common(JID, #state{user = U, resource = R} = NewStateData0) ->
                       privacy_list = PrivList},
                     fsm_next_state_pack(session_established,
                         NewStateData).
+-define(xmlel(Name), #xmlel{name = Name}).
```
chrzaszcz: unused macro

test/ejabberd_tests/tests/xep_0352_csi_SUITE.erl
```diff
@@ -0,0 +1,251 @@
...
+    <<AliceUsername/binary, "@", AliceServer/binary>>.
+
+mk_resume_stream(SMID, PrevH) ->
+    fun (Conn, Props, Features) ->
+            escalus_connection:send(Conn, escalus_stanza:resume(SMID, PrevH)),
+            Resumed = escalus_connection:get_stanza(Conn, get_resumed),
+            true = escalus_pred:is_sm_resumed(SMID, Resumed),
+            {Conn, [{smid, SMID} | Props], Features}
+    end.
+alice_gets_message_after_buffer_overflow(Config) ->
```
chrzaszcz: Minor: fix empty lines here...

test/ejabberd_tests/tests/xep_0352_csi_SUITE.erl
```diff
@@ -0,0 +1,251 @@
...
+
+        then_client_receives_message(Alice, Flushed),
+        %% and no other stanza
+        escalus_assert:has_no_stanzas(Alice),
+        %% Alice activates
+        escalus:send(Alice, csi_stanza(<<"active">>)),
+        %% ands gets remaining stanzas
+        then_client_receives_message(Alice, Awaiting)
+    end).
+
```
chrzaszcz: ...and here. There are also some other double empty lines that seems quite random...

test/ejabberd_tests/tests/xep_0352_csi_SUITE.erl
```diff
@@ -0,0 +1,251 @@
...
+    %%When
+    Msgs = gen_msgs(<<"Hi, Alice">>, N),
+    send_msgs(Bob, Alice, Msgs),
+    timer:sleep(timer:seconds(1)),
+    Msgs.
+
+send_msgs(From, To, Msgs) ->
+    [escalus:send(From, escalus_stanza:chat_to(To, Msg)) ||
+     Msg <- Msgs].
+
```
chrzaszcz: ...like this one :-)

PR #707, opened 2016-03-16
upped exometer to 1.2

bartekgorny:
Change to a newer version of exometer. No real change.

### DIFF
--- a/apps/ejabberd/src/mongoose_metrics.erl
+++ b/apps/ejabberd/src/mongoose_metrics.erl
@@ -281,7 +281,7 @@ ensure_metric(Metric, Type) when is_tuple(Type) ->
 ensure_metric(Metric, Type) ->
     ensure_metric(Metric, Type, Type).
 
-ensure_metric(Metric, ShortType, Type) when is_list(Metric) ->
+ensure_metric(Metric, Type, ShortType) when is_list(Metric) ->
     %% the split into ShortType and Type is needed because function metrics are
     %% defined as tuples (that is Type), while exometer:info returns only 'function'
     case exometer:info(Metric, type) of
--- a/apps/ejabberd/src/mongoose_metrics.erl
+++ b/apps/ejabberd/src/mongoose_metrics.erl
@@ -276,6 +276,8 @@ create_metrics(Host) ->
     lists:foreach(fun(Name) -> ensure_metric(Name, counter) end,
                   get_total_counters(Host)).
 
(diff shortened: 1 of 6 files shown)


rebar.config
```diff
@@ -18,7 +18,7 @@
   {exml, ".*", {git, "git://github.com/esl/exml.git", "2.2.0"}},
   {lager, "2.0.3", {git, "git://github.com/basho/lager.git", "b7984d4"}},
   {cowboy, ".*", {git, "git://github.com/ninenines/cowboy.git", "1.0.4"}},
-  {exometer, ".*", {git, "git://github.com/esl/exometer.git", {branch, "1.1-patched"}}},
+  {exometer, ".*", {git, "git://github.com/esl/exometer.git", {tag, "1.2"}}},
```
michalwski: Let's use the main exometer repo which is Feuerlabs/exometer. And the newest version there is 1.2.1

michalwski:
Also please run `make ct`, this command will run tests from `apps/ejabberd/tests`. Currently they are failing on travis.

rebar.config
michalwski: I'd even try this commit Feuerlabs/exometer@7a7bd8d

apps/ejabberd/src/mongoose_metrics.erl
```diff
@@ -276,9 +276,16 @@ create_metrics(Host) ->
     lists:foreach(fun(Name) -> ensure_metric(Name, counter) end,
                   get_total_counters(Host)).
 
-ensure_metric(Metric, Type) when is_list(Metric) ->
+ensure_metric(Metric, Type) when is_tuple(Type) ->
```
michalwski: Is this `Type` tuple well defined? I mean, do we know if it's always of form `{Type, Sth}`. If so then, I think it would be better to place this check as pattern matching in function's head.

```erlang
ensure_metric(Metric, {El1, _El2} = Type) ->
    ensure_metric(Metric, Type, El1);
```

Of course `El1` should have more meaningful name.

michalwski:
Also, could you please update relevant doc page(s)?

apps/ejabberd/src/mongoose_metrics.erl
bartekgorny: I'm not sure the length is always the same, what I observed was 6 elements but I don't know if this is always like that. This is why I just extract the first element.

PR #708, opened 2016-03-17
TCP backlog improvements

michalwski:
Proposed changes include:
* increased tcp backlog for ejabberd_listeners
* fix passing tcp opts from config file
* test for default and passed backlog parameter

Also this PR is important to allow running parallel tests, introduced in #706, on MacOS X. Together with @pzel we discovered (mainly based on blogpost [1]) that the backlog queue implementation on MacOS X is less efficient than on linux. Too low backlog on MongooseIM side for port 5222 manifests by strange `{error,econnreset}`

[1]. http://veithen.github.io/2014/01/01/how-tcp-backlog-works-in-linux.html

### DIFF
--- a/apps/ejabberd/src/ejabberd_listener.erl
+++ b/apps/ejabberd/src/ejabberd_listener.erl
@@ -173,15 +173,18 @@ listen_tcp(PortIPProto, Module, SockOpts, Port, IPS) ->
             ets:delete(listen_sockets, PortIP),
             ListenSocket;
         _ ->
-            Res = gen_tcp:listen(Port, [binary,
-                                        {backlog, 100},
-                                        {packet, 0},
-                                        {active, false},
-                                        {reuseaddr, true},
-                                        {nodelay, true},
-                                        {send_timeout, ?TCP_SEND_TIMEOUT},
-                                        {keepalive, true},
-                                        {send_timeout_close, true} | SockOpts]),
+            DefaultSockOpts = [binary,
+                               {backlog, 100},
+                               {packet, 0},
+                               {active, false},
+                               {reuseaddr, true},
+                               {nodelay, true},
+                               {send_timeout, 120},
+                               {keepalive, true},
+                               {send_timeout_close, true}],
+
+            FinalSockOpts = override_sock_opts(SockOpts, DefaultSockOpts),
+            Res = gen_tcp:listen(Port, FinalSockOpts),
             case Res of
                 {ok, ListenSocket} ->
                     ListenSocket;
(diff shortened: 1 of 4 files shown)


apps/ejabberd/src/ejabberd_listener.erl
```diff
@@ -173,13 +174,14 @@ listen_tcp(PortIPProto, Module, SockOpts, Port, IPS) ->
...
             Res = gen_tcp:listen(Port, [binary,
+                                        {backlog, 100},
                                         {packet, 0},
                                         {active, false},
                                         {reuseaddr, true},
                                         {nodelay, true},
                                         {send_timeout, ?TCP_SEND_TIMEOUT},
                                         {keepalive, true},
-                                        {send_timeout_close, true}]),
+                                        {send_timeout_close, true} | SockOpts]),
```
pzel: Correct me if I'm wrong, but will this prevent our overrides from ever being picked up by `gen_tcp:listen`? The 'updated' values should be in front of the proplist.

apps/ejabberd/test/ejabberd_listener_SUITE.erl
```diff
@@ -0,0 +1,31 @@
...
+                                         meck:passthrough([Port, Opts])
+                                 end),
+
+    OverrideBacklog = {backlog, 50},
+    {ok, _Pid} = listener_started([OverrideBacklog]),
+
+    [{_Pid, {gen_tcp, listen, [_, Opts]}, _Result}] =  meck:history(gen_tcp),
+
+    DefaultBacklog = {backlog, 100},
+    [DefaultBacklog, OverrideBacklog] = proplists:lookup_all(backlog, Opts),
```
pzel: Same as above: If the listening socket is started with `[{opt_a, "a"}, {opt_a, "b"}]`, then my guess is that it will take `"a"` as the value for `opt_a`. This test shows that the values are passed, but we can't be certain that the right value is used. I propose to do this the right way and **replace** the values in the original proplist if there are overrides.

apps/ejabberd/src/ejabberd_listener.erl
michalwski: No, they have to be at the end, I looked into the `gen_tcp`, and `inet` code. There is a recursive function iterating on this opts and construing a record with opts.

apps/ejabberd/test/ejabberd_listener_SUITE.erl
michalwski: See this https://github.com/esl/MongooseIM/pull/708#discussion_r56475918 for explanation.

apps/ejabberd/src/ejabberd_listener.erl
pzel: I see -- so that's very non-obvious, and someone may come along and change it later. Let's do it the right way so no "alternative" interpretation is possible.

apps/ejabberd/src/ejabberd_listener.erl
```diff
@@ -190,16 +193,24 @@ listen_tcp(PortIPProto, Module, SockOpts, Port, IPS) ->
...
+    override_sock_opts(OverrideOpts, NewOpts).
+
+do_override({ip, _} = IP, Opts) ->
+    lists:keystore(ip, 1, Opts, IP);
+do_override({backlog, _} = Backlog, Opts) ->
+    lists:keystore(backlog, 1, Opts, Backlog);
+do_override(inet6, Opts) ->
+    [inet6 | lists:delete(inet6, Opts)];
+do_override(inet, Opts) ->
+    [inet | lists:delete(inet, Opts)];
```
pzel: I like this explicit solution! One minor nitpick:

Isn't
```[inet6 | lists:delete(inet6, Opts)]```
the same thing as
```Opts```?

Also, it's not technically possible to **delete** one of the `inet|inet6` options now. Do we need those clauses? Maybe just erroring out with `error({cannot_override_option, O, Opts})` would be more straightforward for the user?

michalwski: No, it's not the same. What I wanted to achieve is to have only one `inet6` or `inet` value in the list.

michalwski: I mean `inet6` may not be in the `Opts` list but if it is, the I want only one such element in the list.

pzel: Yes, got it now :)

PR #709, opened 2016-03-17
Rel 1.7 mongoose cluster

ludwikbukowski:
Once @erszcz created PR and said:

" Apart from the rationale in 20d40e8 description, there's one more reason: using just RPC (instead of a shell-out) to cluster/disband nodes makes it possible to keep cover-compiled modules loaded through the whole cluster life cycle. This is required to make rel-1.7-service tests finally pass."

this PR was targeted to `rel-1.7` branch
we've disscuessed that it also should point to `master` branch


michalwski:
I agree this should be done on master as well. The easiest way (IMO) is to do the change on top of master branch, send PR to master branch, then merge to `rel-1.7` after successful merge to master.

I'm closing this PR for now.

michalwski:
After thinking about it more, it will rather be easier to open the PR to `rel-1.7` branch and finally merge `rel-1.7` to `master` which I'm going to do as soon as I merge #704 (hopefully today).

PR #710, opened 2016-03-18
DockerHub and Github's releases integration

michalwski:
Proposed changes include:
* dockerhub integration:
   * build and upload docker image to https://hub.docker.com/r/mongooseim/mongooseim/
   * used Dockerfile from michalwski/mongooseim-docker
* upload tar.gz file to Github's release (when building a tag)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,13 @@ after_failure:
         - tail -100 dev/mongooseim_node1/log/ejabberd.log
         - tail -100 dev/mongooseim_node2/log/ejabberd.log
 
+after_success:
+        - make cover_report
+        - if [ $PRESET = 'internal_mnesia' ]; then tools/travis-build-and-push-docker.sh; fi
+
 services:
         - redis-server
+        - docker
 branches:
         only:
                 - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,12 +38,12 @@ after_success:
 services:
         - redis-server
         - docker
-        - riak
+
 branches:
         only:
-                - docker-image-building
                 - master
                 - stable
+                - /^rel\-\d+\.\d+$/
                 - /\d+\.\d+\.\d+(\-[a-z,0-9])*/
 
 otp_release:
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,7 +85,7 @@ deploy:
   provider: releases
   api-key: $GITHUB_TOKEN
   skip_cleanup: true
-  file: mongooseim-$TRAVIS_BRANCH.$TRAVIS_OTP_RELEASE.$(lsb_release -is | tr "A-Z" "a-z").$(lsb_release -rs).$(uname -m).tar.bz2
+  file: mongooseim-$TRAVIS_BRANCH.OTP-$TRAVIS_OTP_RELEASE.$(lsb_release -is | tr "A-Z" "a-z").$(lsb_release -rs).$(uname -m).tar.bz2
   on:
     tags: true
     condition: "$PRESET.$TRAVIS_OTP_RELEASE = internal_mnesia.17.5"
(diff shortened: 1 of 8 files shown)


Nyco:
Can we do a esl/mongooseim-docker like a potential esl/mongooseim-ami?

michalwski:
Yes we can! Firstly, though, I'd like to hear some comments on this PR.

tools/docker/riak/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM hectcastro/riak
```
erszcz: Can we add a specific version here? If the version behind this tag changes and our image has to be rebuilt after the fact, we might run into inconsistencies.

michalwski: Unfortunately not, the only available tag for this image is `latest`...

The Riak image is used only to test MongooseIM with Riak. The infrastructure enabling docker doesn't have Riak as a regular service, as it was before, that's why Riak is started in container.

erszcz:
I think we should not leave the compressed tarball in the container image. That is, if we want to upload each resulting image to DockerHub, we should not do:

```
COPY ./member/mongooseim.tar.gz mongooseim.tar.gz
```

in `Dockerfile.member`. It doesn't matter a lot when the image just sits on a laptop, but it starts when they add up _with every PR_ we post to GitHub (if I understand the scripts correctly, that's what will happen).

The archive has to be taken from the mounted volume and unpacked into the container image by a script before executing ENTRYPOINT - that way we'll probably cut down on the size of the final container.

michalwski:
But then, how can we provide the tarball  for some who downloads the image from docker hub?

michalwski:
And yes, you are correct, it pushes the image for every PR. Initially I did it only for master branch but then I decided to simplify running MongooseIM from docker for different PR, f.e. in our load tests infrastructure.

michalwski:
The image has been pushed. You can find it here: https://hub.docker.com/r/mongooseim/mongooseim/tags/

michalwski:
And here is the example of pushed tarball to Github's release: https://github.com/michalwski/MongooseIM/releases/tag/1.7.0-pre-docker

.travis.yml
```diff
@@ -74,3 +81,11 @@ cache:
 before_cache:
         - rm -f dialyzer/apps.*
 
+deploy:
+  provider: releases
+  api-key: $GITHUB_TOKEN
+  skip_cleanup: true
+  file: mongooseim-$TRAVIS_BRANCH.$TRAVIS_OTP_RELEASE.$(lsb_release -is | tr "A-Z" "a-z").$(lsb_release -rs).$(uname -m).tar.bz2
```
studzien: How about describing that OTP version is an OTP version?

tools/wait_for_riak.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+# We cannot just connect to 127.0.0.1:8098 because Docker is very "smart" and
+# exposes ports before the service is ready
+IP=$(/usr/bin/docker inspect -f {{.NetworkSettings.IPAddress}} riak)
+echo "riak IP is $IP"
+erl -eval 'try lists:map(fun(N) -> case gen_tcp:connect("'"$IP"'", 8098, [], 1000) of {error,econnrefused} -> io:format("Retry, n=~p~n", [N]), timer:sleep(1000); Other -> io:format("~p", [Other]), init:stop(0) end end, lists:seq(1,60)) after init:stop(1) end.'
```
studzien: How about using script like https://github.com/vishnubob/wait-for-it for this kind of synchronization?
I haven't used it, though.

.travis.yml
michalwski: I'll do that.

Issue #696, opened 2016-03-03
ejabberd_service - optional external_component table


michalwski:
The component routing via additional routing table (external_component) should be optional.

erszcz:
To give some more context, there's already an `xmpp_router` behaviour present. We'd like to have a config option (dynamically compiled on load) to pick which implementation of the behaviour is to be used for routing.

This way we'll have freedom over a routing algorithm - we'll be able to choose one algorithm routing based on the `external_component` table (the _current_ behaviour as of the merge of #640) or a different one which would be indifferent to the table's existence (the behaviour pre-#640). This would also give us leeway to more easily experiment with riak_core / DHT / whatnot based routing in the future.

PR #716, opened 2016-03-22
Configurable routing mech

bartekgorny:
This PR addresses issue #696, and goes a bit beyond that.
The routing system is reworked, so that there is a chain of modules, each implementing xmpp_router behaviour. Each of those modules can filter out a message, route it or pass it on. The ejabberd_router module is sort of a "routing controller". There is a default chain, and config file can specify a different one.
The engine is tested by xmpp_router_SUITE.

### DIFF
--- a/apps/ejabberd/src/ejabberd_local_delivery.erl
+++ b/apps/ejabberd/src/ejabberd_local_delivery.erl
@@ -1,0 +1,34 @@
+%%%-------------------------------------------------------------------
+%%% @author bartek
+%%% @copyright (C) 2016, <COMPANY>
+%%% @doc
+%%%
+%%% @end
+%%% Created : 21. Mar 2016 12:24
+%%%-------------------------------------------------------------------
+-module(ejabberd_local_delivery).
+-author("bartek").
+
+-include("jlib.hrl").
+
+%% API
+-export([do_local_route/5]).
+
+
+do_local_route(OrigFrom, OrigTo, OrigPacket, LDstDomain, Handler) ->
+    %% Filter locally
+    case ejabberd_hooks:run_fold(filter_local_packet, LDstDomain,
+        {OrigFrom, OrigTo, OrigPacket}, []) of
+        {From, To, Packet} ->
+            case Handler of
+                {apply_fun, Fun} ->
+                    Fun(From, To, Packet);
+                {apply, Module, Function} ->
+                    Module:Function(From, To, Packet)
+            end;
+        drop ->
+            ejabberd_hooks:run(xmpp_stanza_dropped,
+                OrigFrom#jid.lserver,
+                [OrigFrom, OrigTo, OrigPacket]),
+            ok
+    end.
--- a/apps/ejabberd/src/ejabberd_local_delivery.erl
+++ b/apps/ejabberd/src/ejabberd_local_delivery.erl
@@ -12,10 +12,10 @@
 -include("jlib.hrl").
 
 %% API
--export([do_local_route/5]).
+-export([do_route/5]).
 
 
-do_local_route(OrigFrom, OrigTo, OrigPacket, LDstDomain, Handler) ->
+do_route(OrigFrom, OrigTo, OrigPacket, LDstDomain, Handler) ->
     %% Filter locally
     case ejabberd_hooks:run_fold(filter_local_packet, LDstDomain,
         {OrigFrom, OrigTo, OrigPacket}, []) of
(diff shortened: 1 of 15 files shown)


michalwski:
Could you please change the PR's name to sth more descriptive?

apps/ejabberd/src/ejabberd_local_delivery.erl
```diff
@@ -0,0 +1,34 @@
...
+-module(ejabberd_local_delivery).
+-author("bartek").
+
+-include("jlib.hrl").
+
+%% API
+-export([do_local_route/5]).
+
+
+do_local_route(OrigFrom, OrigTo, OrigPacket, LDstDomain, Handler) ->
```
michalwski: I suggest changing the function name to just `do_route`. I think this invocation will be informative enough:
```erlang
ejabberd_local_deliver:do_route(From, To, Packet, Domain, Handler)
```

apps/ejabberd/src/ejabberd_local_delivery.erl
```diff
@@ -0,0 +1,34 @@
+%%%-------------------------------------------------------------------
+%%% @author bartek
+%%% @copyright (C) 2016, <COMPANY>
+%%% @doc
+%%%
+%%% @end
+%%% Created : 21. Mar 2016 12:24
+%%%-------------------------------------------------------------------
+-module(ejabberd_local_delivery).
```
michalwski: I'm tempted to name this and other new modules here with `mongoose_` prefix and not `ejabberd_` as this is totally MongooseIM specific implementation. But I'd like to here from others @esl/mongooseim-developers

apps/ejabberd/src/ejabberd_local_delivery.erl
```diff
@@ -0,0 +1,34 @@
+%%%-------------------------------------------------------------------
```
erszcz: These comments aren't very useful right now. Please fill in the correct details (at least author email, possibly mongooseim-im@... if you don't want to provide your own email, company name and maybe one line description of the module) or drop these info comments completely.

apps/ejabberd/src/ejabberd_router.erl
```diff
@@ -295,50 +310,62 @@ code_change(_OldVsn, State, _Extra) ->
...
+    make_routing_module_source(ModList);
+make_routing_module_source(Mods) ->
+    binary_to_list(iolist_to_binary(io_lib:format(
+        "-module(mod_routing_machine).~n"
+        "-compile(export_all).~n"
+        "get_routing_module_list() -> ~p.~n",
+        [Mods]))).
+
+route(_, _, _, []) ->
+    ok; %% shouldn't we raise error here? If we get here means we don't know
```
michalwski: This also means that we iterated over all possible items in the list. I see this as correct. If we want to fail because of empty list, the we should do this earlier, when starting the mechanism and detecting the list is empty (if I understand everything correctly).

bartekgorny: Not exactly. When we start routing a message we have in the default setup four modules - global, external, localdomain and s2s, and each of them has a chance to drop or route the message. If we are here means no module knew what to do, every one passed it on. So, we have no idea what to do with this message, which doesn't seem good. Either we signal a problem, or install some 'catch-all' stuff, or just don't care. That's the choice.

apps/ejabberd/src/ejabberd_local_delivery.erl
bartekgorny: Yes I know, I already asked MP what should be put here - what license etc. Files in source tree have various headers. Thaks for the reminder.

apps/ejabberd/src/ejabberd_router.erl
michalwski: But from what I understand the `route/4` is a recursive function and the case where list is empty is a stop condition for the recursion. How else the recursion will be stopped?

bartekgorny: If any Mod:filter returns 'drop', or Mod:route returns 'done', then the recursion is stopped. route/4 is called again on Tail only if both functions return the message, which means the module didn't handle the message but choose to pass it further on.

michalwski: Now I understand. Having that it mind, I think we should write and error log message, and maybe update a metric counting routing errors.

bartekgorny: How would you name it?

michalwski: `[global, routingErrors]`?

michalwski:
Replaced by #729

Issue #685, opened 2016-02-24
Unify `add_to_cluster` and `remove_from_cluster` commands.


michalwski:
Since there is module `mongoose_cluster` [1] taking care of joining and leaving the cluster, the mongooseimctl commands `add_to_cluster` [2] and `remove_from_cluster` [3] should be rewritten to use this new module.

[1]. https://github.com/esl/MongooseIM/blob/20d40e87ab492013b77c32153fbce392890efae1/apps/ejabberd/src/mongoose_cluster.erl
[2]. https://github.com/esl/MongooseIM/blob/20d40e87ab492013b77c32153fbce392890efae1/rel/files/mongooseim#L98
[3]. https://github.com/esl/MongooseIM/blob/20d40e87ab492013b77c32153fbce392890efae1/rel/files/mongooseim#L131

ludwikbukowski:
Would it be a good practise to leave the node also from distributed erlang cluster in `mongoose_cluster:leave/0` function? I mean use `erlang:disconnect_node/1`so that fun `nodes/0` would not list any connected nodes.
here:
https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mongoose_cluster.erl#L39

erszcz:
@ludwikbukowski I don't have a strong opinion.
Given that connecting to distributed nodes is transparent (i.e. when you do `rpc:call(some_node, ...)` it _Just Knows to Connect_ in the background), there's not a huge difference between leaving the distributed Erlang cluster and not doing so.
The important difference is how to check for membership: `mnesia:system_info(db_nodes)` versus `nodes()`. But even if we leave the Erlang cluster on leaving a MongooseIM cluster, `nodes()` is still not sufficient to get _all_ (offline or not) MongooseIM cluster members.

PR #717, opened 2016-03-24
Rel 1.7 mongoose cluster

ludwikbukowski:
This PR addresses https://github.com/esl/MongooseIM/issues/685

Proposed changes include:
* Replacing commands `add_to_cluster` and `remove_from_cluster` with brand-new
`join_cluster` and `leave_cluster` in _ejabberd_admin_ module. The commands use module _mongoose_cluster_ for the cluster management hence it's done completely at the Erlang level
* updated tests in _cluster_commands_SUITE_

### DIFF
--- a/apps/ejabberd/src/ejabberd_admin.erl
+++ b/apps/ejabberd/src/ejabberd_admin.erl
@@ -44,7 +44,8 @@
          dump_to_textfile/1, dump_to_textfile/2,
          mnesia_change_nodename/4,
          restore/1, % Still used by some modules%%
-         get_loglevel/0]).
+         get_loglevel/0,
+         join_cluster/1, leave_cluster/0]).
 
 -include("ejabberd.hrl").
 -include("ejabberd_commands.hrl").
@@ -147,13 +148,74 @@ commands() ->
      #ejabberd_commands{name = reload_cluster, tags = [server],
                         desc = "Reload configuration file in the cluster",
                         module = ejabberd_config, function = reload_cluster,
-                        args = [], result = {res, restuple}}
+                        args = [], result = {res, restuple}},
+     #ejabberd_commands{name = join_cluster, tags = [server],
+                        desc = "Join the node to the cluster",
+                        module = ?MODULE, function = join_cluster,
+                        args = [{node, string}],
+                        result = {res, restuple}},
+     #ejabberd_commands{name = leave_cluster, tags = [server],
+                        desc = "Leave a node from the cluster",
+                        module = ?MODULE, function = leave_cluster,
+                        args = [],
+                        result = {res, restuple}}
     ].
 
 
 %%%
 %%% Server management
 %%%
+
+-spec join_cluster(string()) -> {ok, string()} | {pang, string()} | {mnesia_error, string()} | {error, string()}.
+join_cluster(NodeString) ->
+  NodeAtom = list_to_atom(NodeString),
+  NodeList = mnesia:system_info(running_db_nodes),
+  case lists:member(NodeAtom, NodeList) of
+    true ->
+      String = io_lib:format("The node ~s has already joined the cluster~n", [NodeString]),
+      {ok, String};
+    _ ->
+      do_join_cluster(NodeAtom)
+  end.
+
+do_join_cluster(Node) ->
+    try mongoose_cluster:join(Node) of
+        ok ->
+            String = io_lib:format("You have successfully joined the node ~p to the cluster with node member ~p~n", [node(), Node]),
+            {ok, String}
+    catch
+        error:pang ->
+            String = io_lib:format("Timeout while attempting to connect to node ~s~n", [Node]),
+            {pang, String};
+        error:{cant_get_storage_type, {T, E, R}} ->
+            String = io_lib:format("Cannot get storage type for table ~p~n. Reason: ~p:~p", [T, E, R]),
+            {mnesia_error, String};
+        E:R ->
+            {error, {E, R}}
+    end.
+
+-spec leave_cluster() -> {ok, string()} | {error, term()} | {not_in_cluster, string()}.
+leave_cluster() ->
+  NodeList = mnesia:system_info(running_db_nodes),
+  case NodeList of
+    [] ->
+      String = io_lib:format("The node ~p is not in the cluster~n", [node()]),
+      {not_in_cluster, String};
+    _ ->
+      do_leave_cluster()
+  end.
+
+do_leave_cluster() ->
+  try mongoose_cluster:leave() of
+    ok ->
+      String = io_lib:format("You have successfully left the node ~p from
+      the cluster~n", [node()]),
+      {ok, String}
+  catch
+    E:R ->
+      {error, {E, R}}
+  end.
+
 
 -spec status() -> {'ejabberd_not_running', io_lib:chars()} | {'ok', io_lib:chars()}.
 status() ->
--- a/apps/ejabberd/src/ejabberd_admin.erl
+++ b/apps/ejabberd/src/ejabberd_admin.erl
@@ -166,17 +166,18 @@ commands() ->
 %%% Server management
 %%%
 
--spec join_cluster(string()) -> {ok, string()} | {pang, string()} | {mnesia_error, string()} | {error, string()}.
+-spec join_cluster(string()) -> {ok, string()} | {pang, string()} | {alread_joined, string()} |
+                                {mnesia_error, string()} | {error, string()}.
 join_cluster(NodeString) ->
-  NodeAtom = list_to_atom(NodeString),
-  NodeList = mnesia:system_info(running_db_nodes),
-  case lists:member(NodeAtom, NodeList) of
-    true ->
-      String = io_lib:format("The node ~s has already joined the cluster~n", [NodeString]),
-      {ok, String};
-    _ ->
-      do_join_cluster(NodeAtom)
-  end.
+    NodeAtom = list_to_atom(NodeString),
+    NodeList = mnesia:system_info(running_db_nodes),
+    case lists:member(NodeAtom, NodeList) of
+        true ->
+            String = io_lib:format("The node ~s has already joined the cluster~n", [NodeString]),
+            {alread_joined, String};
+        _ ->
+            do_join_cluster(NodeAtom)
+    end.
 
 do_join_cluster(Node) ->
     try mongoose_cluster:join(Node) of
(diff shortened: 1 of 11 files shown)


apps/ejabberd/src/ejabberd_admin.erl
```diff
@@ -44,7 +44,8 @@
          dump_to_textfile/1, dump_to_textfile/2,
          mnesia_change_nodename/4,
          restore/1, % Still used by some modules%%
-         get_loglevel/0]).
+         get_loglevel/0,
+         maybe_join_cluster/1, maybe_leave_cluster/0]).
```
erszcz: Should we export `maybe_*` functions as public API? In case of public interface I'd be for making the module API calls shorter. If after doing the checks (the "maybe"s hint at) the actual implementation needs to be put into deeper functions, we can call them `do_*` (`do_join_cluster`...).

apps/ejabberd/src/ejabberd_admin.erl
```diff
@@ -147,7 +148,17 @@ commands() ->
      #ejabberd_commands{name = reload_cluster, tags = [server],
                         desc = "Reload configuration file in the cluster",
                         module = ejabberd_config, function = reload_cluster,
-                        args = [], result = {res, restuple}}
+                        args = [], result = {res, restuple}},
+      #ejabberd_commands{name = join_cluster, tags = [server],
```
erszcz: Indentation is inconsistent with the former command (line 147/148). Tabs vs spaces issue? BTW, in the following lines - why 2 spaces, when everything around is 4 spaces?

ludwikbukowski: it's an "Easter egg"

apps/ejabberd/src/ejabberd_admin.erl
```diff
@@ -169,7 +169,7 @@ commands() ->
 -spec maybe_join_cluster(string()) -> {ok, string()} | {pang, string()} | {mnesia_error, string()} | {error, string()}.
 maybe_join_cluster(NodeString) ->
   NodeAtom = list_to_atom(NodeString),
-  NodeList = nodes(),
+  NodeList = mnesia:system_info(running_db_nodes),
   case lists:any(fun(El) -> El == NodeAtom end, NodeList) of
```
erszcz: `lists:member`?

apps/ejabberd/src/ejabberd_admin.erl
```diff
@@ -197,7 +197,7 @@ join_cluster(Node) ->
 
 -spec maybe_leave_cluster() -> {ok, string()} | {error, term()} | {not_in_cluster, string()}.
 maybe_leave_cluster() ->
-  NodeList = nodes(),
+  NodeList = mnesia:system_info(running_db_nodes),
```
erszcz: I'm not sure this can happen or not, but what if due to some inconsistency (network issue) we're not in `running_db_nodes`, but are in `db_nodes`? Again, maybe it's impossible, but some food for thought.

rel/files/mongooseimctl
```diff
@@ -70,8 +90,8 @@ help ()
     echo "  debug  Attach an interactive Erlang shell to a running MongooseIM node"
     echo "  live   Start MongooseIM node in live (interactive) mode"
     echo "MongooseIM cluster management commads:"
-    echo "  add_to_cluster other_node_name       Add current node to cluster"
-    echo "  remove_from_cluster dead_node_name   Remove other node from cluster"
+    echo "  join_cluster other_node_name       Add current node to cluster"
+    echo "  leave_cluster   Leave current node from the cluster"
```
erszcz: Which looks better?

```
$ git --help
usage: git [--version] [--help] [-C <path>] [-c name=value]
...

The most commonly used git commands are:
   add        Add file contents to the index
   bisect     Find by binary search the change that introduced a bug
   branch     List, create, or delete branches
   checkout   Checkout a branch or paths to the working tree
   clone      Clone a repository into a new directory
   commit     Record changes to the repository
   diff       Show changes between commits, commit and working tree, etc
   fetch      Download objects and refs from another repository
```

Or?

```
$ git --help
usage: git [--version] [--help] [-C <path>] [-c name=value]
...

The most commonly used git commands are:
   add  Add file contents to the index
   bisect  Find by binary search the change that introduced a bug
   branch  List, create, or delete branches
   checkout  Checkout a branch or paths to the working tree
   clone  Clone a repository into a new directory
   commit  Record changes to the repository
   diff  Show changes between commits, commit and working tree, etc
   fetch  Download objects and refs from another repository
```

test/ejabberd_tests/tests/cluster_commands_SUITE.erl
```diff
@@ -79,18 +77,38 @@ init_per_group(Group, Config) when Group == clustered orelse Group == ejabberdct
             remove_node_from_cluster(Config1),
             {skip, nondistributed_sm}
     end;
+
+init_per_group(clustering, _Config) ->
+    case is_sm_distributed() of
+        true ->
+            % escalus:create_users(Config, escalus:get_users([alice, clusterguy]));
```
ludwikbukowski: Ops. I will remove this line before merge

apps/ejabberd/src/ejabberd_admin.erl
```diff
@@ -155,6 +166,59 @@ commands() ->
...
+            {ok, String};
+        _ ->
+            do_join_cluster(NodeAtom)
+    end.
+
+do_join_cluster(Node) ->
+    try mongoose_cluster:join(Node) of
+        ok ->
+            String = io_lib:format("You have successfully joined the node ~p to
+      the cluster with node member ~p~n", [node(), Node]),
```
michalwski: Is it intentional to add a new line here? If so I'd rather add `~n` and not split the string into many lines. The indentation is unintentional here.

michalwski:
I'd invest some more time in testing the negative test cases like node tries to join twice or node is removed twice. Basically consult coveralls.io [1] for your build and see what new lines added by you are not covered yet.

[1]. https://coveralls.io/builds/5617628

apps/ejabberd/src/ejabberd_admin.erl
```diff
@@ -155,6 +166,59 @@ commands() ->
...
+        E:R ->
+            {error, {E, R}}
+    end.
+
+-spec leave_cluster() -> {ok, string()} | {error, term()} | {not_in_cluster, string()}.
+leave_cluster() ->
+    NodeList = mnesia:system_info(running_db_nodes),
+    ThisNode = node(),
+    case NodeList of
+        [ThisNode] ->
```
michalwski: This matching doesn't look good to me. What will happen if we have more then 1 node in running_db_nodes? A test for that would be more than welcome!

rel/files/mongooseimctl
```diff
@@ -219,8 +239,8 @@ done
 
 case $1 in
     'start') start;;
-    'add_to_cluster') add_to_cluster;;
-    'remove_from_cluster') remove_from_cluster;;
+    'join_cluster') join_cluster;;
```
michalwski: Why this name change? I understand its better but I see no updated in the doc about this change.

rel/files/mongooseimctl
```diff
@@ -219,8 +239,8 @@ done
...
+    'leave_cluster') leave_cluster;;
```
michalwski: As far as I understand leave_cluster should be run remove a node (different one) from a cluster. If we have 3 nodes A, B, C and we want to remove C from the cluster. Then we need to run this command on A or B. Am I right? If so the `leave_cluster` name is not intuitive to me.

apps/ejabberd/src/ejabberd_admin.erl
ludwikbukowski: If there is more than one node the case would match here [1] and it would mean that node is in any cluster and can leave it. Otherwise [2] there is no cluster to leave because `mnesia:system_info(running_db_nodes)` returned just the local node
**However you're right about the lacking tests!**

[1]https://github.com/esl/MongooseIM/blob/rel-1.7-mongoose-cluster/apps/ejabberd/src/ejabberd_admin.erl#L206
[2]https://github.com/esl/MongooseIM/blob/rel-1.7-mongoose-cluster/apps/ejabberd/src/ejabberd_admin.erl#L203

rel/files/mongooseimctl
ludwikbukowski: Ok, I will update the docs

rel/files/mongooseimctl
ludwikbukowski: Command `join_cluster` takes as a parameter some cluster member and it's called from joining node ("the alone one")
Command 'leave_cluster' takes no parameters and it's called from  the node which is going to leave the cluster.
In your scenario, if we have clustered nodes A, B and C and we want to remove C we have to call `leave_cluster` from node C
What I've done in this PR was just a pure command line interface for this module [1] and maybe I didn't understand the task correctly?
[1]https://github.com/esl/MongooseIM/blob/20d40e87ab492013b77c32153fbce392890efae1/apps/ejabberd/src/mongoose_cluster.erl

apps/ejabberd/src/ejabberd_admin.erl
ludwikbukowski: I must have split it because the line was too long but not having in mind that it's text to display!
I will fix it

rel/files/mongooseimctl
michalwski: Thanks for the explanation. Now I understand :)

apps/ejabberd/src/ejabberd_admin.erl
michalwski: Right. It was before my morning coffee. Thanks for explaining anyway :)

test/ejabberd_tests/tests/cluster_commands_SUITE.erl
```diff
@@ -134,3 +152,53 @@ set_master_test(ConfigIn) ->
     ejabberdctl("set_master", ["self"], ConfigIn),
     [MasterNode] = rpc_call(mnesia, table_info, [TableName, master_nodes]).
 
+join_successful(Config) ->
+    Node2 = ct:get_config(ejabberd2_node),
```
ludwikbukowski: I'm going also to stop using config values like `ejabberd2_node` and `ejabberd_node` and convert suite to use `hosts` block defined here [1]

[1]https://github.com/esl/MongooseIM/blob/master/test/ejabberd_tests/test.config#L26

test/ejabberd_tests/tests/cluster_commands_SUITE.erl
```diff
@@ -17,60 +17,86 @@
...
+        {group, ejabberdctl},
+        {group, clustering_two},
+        {group, clustering_theree}].
 groups() ->
     [{clustered, [], [one_to_one_message]},
-     {ejabberdctl, [], [set_master_test]}].
-
+        {clustering_two, [],
+            [join_successful,
+            leave_successful,
```
michalwski: Indentation :) please align the whole list to line 42.

test/ejabberd_tests/tests/cluster_commands_SUITE.erl
```diff
@@ -17,60 +17,86 @@
...
+            [cluster_of_theree,
+             leave_the_theree]},
+        {ejabberdctl, [], [set_master_test]}].
 suite() ->
+    require_all_nodes() ++
     escalus:suite().
 
+require_all_nodes() ->
+    [{require, mim_node, {hosts, mim, node}},
+        {require, mim_node2, {hosts, mim2, node}},
```
michalwski: Indentation again :)

test/ejabberd_tests/tests/cluster_commands_SUITE.erl
```diff
@@ -103,20 +158,20 @@ one_to_one_message(ConfigIn) ->
...
+        Stanza1 = escalus:wait_for_stanza(ClusterGuy, 5000),
+        escalus:assert(is_chat_message, [<<"Hi!">>], Stanza1),
+
+        %% When ClusterGuy sends a response
+        Msg2 = escalus_stanza:chat_to(Alice, <<"Oh hi!">>),
+        escalus:send(ClusterGuy, Msg2),
+        %% Then Alice also receives it
+        Stanza2 = escalus:wait_for_stanza(Alice, 5000),
+        escalus:assert(is_chat_message, [<<"Oh hi!">>], Stanza2)
+                                                         end).
```
michalwski: It'd be more readable to have `end` at the same level as `escalus:story`.

test/ejabberd_tests/tests/distributed_helper.erl
```diff
@@ -43,42 +42,58 @@ wait_until_started(Cmd, Retries) ->
...
     escalus_ct:rpc_call(Node, M, F, A, TimeOut, Cookie).
+
+get_or_fail(Key) ->
+    Val = ct:get_config(Key),
+    Val == undefined andalso error({undefined, Key}),
+    Val.
+
+
+%% MongooseIM node names
+mim() ->
```
michalwski: I'd move this function  (also mim2 and fed) to `ejabberd_node_utils`. IMO they better suite there. Ideally would be to rename the `ejabberd_node_utils` to  `mongoose_node_utils`, or just `node_utils`. But this may introduce too many changes for this PR so maybe it's a subject of a separate PR.

PR #719, opened 2016-03-28
Cassandra backend for MAM

arcusfelis:
Added cassandra backend.

Proposed changes include:
* privacy, private, roster cassandra backends
* bindings to seestrar
* MAM cassandra backend for MUC, for PM and prefs backend (all fully functional)
* Corresponding docs
* Travis uses Ubuntu 14.04 and docker

Configuring Cassandra connector pool is similar to riak, but you can configure several pools.


apps/ejabberd/src/mongoose_cassandra.erl
```diff
@@ -0,0 +1,35 @@
...
+-spec stop() -> _.
+stop() ->
+    mongoose_cassandra_sup:stop().
+
+%% @doc Return timestamp in nanoseconds
+now_timestamp() ->
+    now_to_usec(os:timestamp()).
+
+-spec now_to_usec(erlang:timestamp()) -> non_neg_integer().
+now_to_usec({MSec, Sec, USec}) ->
```
michalwski: I suggest using `esl/usec` instead of writing this function again.

apps/ejabberd/src/mongoose_lib.erl
```diff
@@ -0,0 +1,11 @@
+-module(mongoose_lib).
```
michalwski: What is the purpose of this module? Do we really have to parse the argument to get an int value from it? Why simple binary_to_integer won't work here?

michalwski:
That's great work @arcusfelis, thanks a lot! I have a general suggestion at this point. Split this PR it several smaller. The first would be with the the cassandra job on travis and MAM. And subsequent PRs would be for every feature, on top of the first PR.

apps/ejabberd/src/mongoose_lib.erl
arcusfelis: The binary looks like "123rest". And I need to parse "rest" as a next step. I don't know where to put this function, but last time I added functions to mod_mam_utils, other developers wanted to use the functions inside other modules too.

apps/ejabberd/src/mongoose_cassandra.erl
arcusfelis: ok, I still want to keep `mongoose_cassandra:now_timestamp()` to keep it's opaque to other code (and for mocking too).

arcusfelis:
@michalwski np, I think about next PRs:
- Update travis with cassandra, riak and mysql inside docker;
- Add cassandra worker code.

On top of this:
- MAM modules;
- private;
- privacy;
- roster.
But it will block travis with tasks for some time :)

michalwski:
Sounds good. Let's proceed that way.

arcusfelis:
Too old

PR #729, opened 2016-04-06
Configurable routing

bartekgorny:
This PR addresses issue #696, and goes a bit beyond that.
The routing system is reworked, so that there is a chain of modules, each implementing xmpp_router behaviour. Each of those modules can filter out a message, route it or pass it on. The ejabberd_router module is sort of a "routing controller". There is a default chain, and config file can specify a different one.
The engine is tested by xmpp_router_SUITE.


michalwski:
A general comment from me (which made before): how about renaming all the new `ejabberd_router_*` module to `mongoose_router_*` and similar with `ejabberd_local_delivery`. These modules are only MongooseIM specific.

apps/ejabberd/src/ejabberd_router.erl
```diff
@@ -295,50 +311,65 @@ code_change(_OldVsn, State, _Extra) ->
...
-                    Fun(From, To, Packet);
-                {apply, Module, Function} ->
-                    Module:Function(From, To, Packet)
-            end;
+routing_modules_list() ->
+    %% this is going to be compiled on startup from settings
+    mod_routing_machine:get_routing_module_list().
+
+compile_routing_module() ->
+    Mods = ejabberd_config:get_local_option(routing_modules),
```
erszcz: I think it's a good practice to handle `undefined` values on the system border as early as we hit them instead of allowing them to be passed around (they're the Erlang equivalent of nullable types eschewed in general). Ideally, `get_local_option` would take the default as another parameter to return in case it doesn't find the option ([it does in ejabberd](https://github.com/processone/ejabberd/blob/a9e50468b613f8b988e91e65b79ad155b65c622f/src/ejabberd_config.erl#L759), though `Default` was introduced after we forked). In this case, I'd do:

```erlang
case ejabberd_config:get_local_option(routing_modules) of
    undefined -> default_routing_modules();
    Defined -> Defined
end
```

Later on, we can make `make_routing_module_source/1` take a list and actually verify that it's a list with `when is_list/1`.

apps/ejabberd/src/ejabberd_router.erl
```diff
@@ -57,19 +53,18 @@
 -include("ejabberd.hrl").
 -include("jlib.hrl").
 
+-record(state, {}).
+
 -type handler() :: 'undefined'
-                | {'apply_fun',fun((_,_,_) -> any())}
```
erszcz: Why such a change?

apps/ejabberd/test/xmpp_route_SUITE.erl
```diff
@@ -2,64 +2,92 @@
...
 all() ->
-    [ success_with_module_implementing_behaviour,
-      fail_with_module_not_implementing_behaviour,
-      fail_when_do_route_crashes
-    ].
+    [ basic_routing ].
 
 init_per_suite(C) ->
-    application:start(stringprep),
+    application:ensure_started(stringprep),
```
erszcz: It might be a bit safer to use `ensure_all_started/1`. It does the same and even more (more: starts dependencies recursively) and doesn't require specific knowledge (does the app have any dependencies?) about the particular application passed as the argument. I'm nitpicking, yes - it's not a serious problem here.

erszcz:
Looks good to me :+1:

erszcz:
Apart from the fact that there are no changes under `docs/` ;)

apps/ejabberd/src/ejabberd_router.erl
bartekgorny: I didn't change type definitions, just moved them around.

erszcz: I mean just the indentation - (to me) it seems bizarre. Is it a preference or tooling issue?

The natural indentation for me is the Emacs style - in Vim it requires a couple more keystrokes, but it's still fine:

```erlang
-type handler() :: 'undefined'
                 | {'apply_fun', fun((_,_,_) -> any())}
                 | {'apply', M::atom(), F::atom()}.
```

There's a similar thing below:

```erlang
-type external_component() :: #external_component{domain :: domain(),
                         handler :: handler()}.
```

I can't see a rule here - it's neither "inside braces -> indent one level deeper" nor "indent to the level of opening brace".

Not a big deal, was mostly curious, but the nesting level gets lost this way.

michalwski:
I'm afraid my comment about new modules name will be omitted again. @esl/mongooseim-developers what do you think about the suggestion I made in this comment https://github.com/esl/MongooseIM/pull/729#issuecomment-206375457?

erszcz:
@michalwski @bartekgorny I also think we should use `mongoose_*` prefix, this infrastructure has nothing to do with ejabberd as it originates in Mongoose.

bartekgorny:
I don't know enough about mongoose and ejabberd to venture my own opinion, so if you guys agree on that then I'll do it.

apps/ejabberd/src/ejabberd_router.erl
arcusfelis: Curiouser and curiouser. Why was `'apply_fun'` used instead of `apply_fun`? (not related to this PR)

apps/ejabberd/src/ejabberd_router_global.erl
```diff
@@ -0,0 +1,29 @@
+%%%-------------------------------------------------------------------
+%%% @doc
+%%% Part of a routing chain; does only filtering - uses hooks to check
+%%% if the message should be routed at all. This one should be the first
+%%% module in chain.
+%%% @end
+%%%-------------------------------------------------------------------
+-module(ejabberd_router_global).
+-author("bartek").
```
arcusfelis: The better format is `Full Name <email>`

michalwski:
>I don't know enough about mongoose and ejabberd to venture my own opinion, so if you guys agree on that then I'll do it.

2 opinions (mine and @erszcz) are more than usual so I'd say we agree on this. There is still time for veto till  12:00 CEST 08.04.2016 :calendar:

apps/ejabberd/src/ejabberd_router.erl
bartekgorny: Yes, you're right - if only because it makes the source much more readable.

Issue #678, opened 2016-02-17
XEP-0092: Software Version


Nyco:
The support for [XEP-0092: Software Version](https://xmpp.org/extensions/xep-0092.html) can be easily implemented.

michalwski:
Tips:
* `gen_mod` behaviour (see mod_time [1] for reference)
* `iq_handlers` (see mod_time [1] for reference)
* version should be taken from `mongooseim` application.

[1]. https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mod_time.erl

PR #731, opened 2016-04-11
Xep 0092

zsuidakra:
This PR addresses #678 

Proposed changes include:
* response with software name and version (currently name is taken from mongoose configuration)
* tests cover service discovery and checking wheter version and name are included in a response

I would also like to return operating system info and  software name based on user's configuration.
Is there a way to keep configuration options during module lifetime, or should I register different functions for processing iqs based on different options passed to `start/1` ?

### DIFF
--- a/apps/ejabberd/src/mod_version.erl
+++ b/apps/ejabberd/src/mod_version.erl
@@ -1,0 +1,15 @@
+-module(mod_version).
+-behaviour(gen_mod).
+-export([start/2, stop/1]).
+-include("jlib.hrl").
+-include("ejabberd.hrl").
+
+start(Host, Opts) ->
+    mod_disco:register_feature(Host, ?NS_VERSION),
+    IQDisc = gen_mod:get_opt(iqdisc, Opts, one_queue),
+    gen_iq_handler:add_iq_handler(ejabberd_local, Host,
+                                  ?NS_VERSION, ?MODULE, process_local_iq, IQDisc).
+
+stop(Host) ->
+    mod_disco:unregister_feature(Host, ?NS_VERSION),
+    gen_iq_handler:remove_iq_handler(ejabberd_local, Host, ?NS_VERSION).
--- a/apps/ejabberd/src/mod_version.erl
+++ b/apps/ejabberd/src/mod_version.erl
@@ -1,6 +1,6 @@
 -module(mod_version).
 -behaviour(gen_mod).
--export([start/2, stop/1, process_local_iq/3]).
+-export([start/2, stop/1, process_iq/3, process_iq_with_os/3]).
 -include("jlib.hrl").
 -include("ejabberd.hrl").
 -xep([{xep, 92}, {version, "1.1"}]).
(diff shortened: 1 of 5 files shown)


apps/ejabberd/src/mod_version.erl
```diff
@@ -0,0 +1,47 @@
...
+mongoose_info() ->
+    {ok, Version} = application:get_key(mongooseim, vsn),
+    {ok, Name} = application:get_key(mongooseim, description),
+    {list_to_binary(Name), list_to_binary(Version)}.
+
+os_info() ->
+    {Family, Name} = os:type(),
+    {Major, Minor, Release} = os:version(),
+    list_to_binary(
+        atom_to_list(Family) ++ " " ++
```
michalwski: I suggest using `atom_to_binary` with encoding utf8 here and below.

michalwski: Or maybe, not :) I noticed just now that you are converting everything to a binary. That's ok.

michalwski:
>software name based on user's configuration.

What do you mean by software name?

zsuidakra:
Should it always response with "MongooseIM"? 
I thought about setting

zsuidakra:
I thought about setting name in configuration file, but it's just my suggestion:)

zsuidakra:
And by software name I mean the content of 'name' tag.

michalwski:
Yes, this should always response with "MongooseIM" or else with the name set in mongooseim.app - exactly how you did it now. By configuration I understand file ejabberd.cfg or app.config.

All in all your PR looks good so far. I see you have a function reading os info, but this is not added to the reply.

zsuidakra:
Considering safety, I would like to return os information only when the
application is configured to do so (in ejabberd.cfg). It is recommended in
XEP-0092 specification. But I don't know if it is possible to keep some
options through module's lifetime, or should I register different functions
for processing requests based on options passed to start/2 function.

W dniu wtorek, 12 kwietnia 2016 Michał Piotrowski <notifications@github.com>
napisał(a):

michalwski:
You can read module configuration using `gen_mod:get_module_opt/4`. You will find there parameters specified in the list for given module. F.e if you have `{mod_version, [{key1, val1}]}` in your `ejabberd.cfg` (or while starting this module dynamically in tests), you can read this with aforementioned function.

zsuidakra:
Thank you, I will try this and push changes when they're ready.

W dniu wtorek, 12 kwietnia 2016 Michał Piotrowski <notifications@github.com>
napisał(a):

zsuidakra:
What do you think about changes @michalwski? Do you think anything should be corrected? :)

michalwski:
Yes :) I'd rather have one iq handler and add the os info only if the module option is set accordingly. This will allow to dynamically change the option value without restarting the module. Also the code will be simpler :)

zsuidakra:
I thought about it, but I must admit I could't figure out one thing. To get module options, I must pass host name as an argument to `gen_mod:get_module_opt/4`, and I couldn't find a way to get host name during iq processing. I have searched source files but could't find anything suitable :/ Is there a function to get current host name?

michalwski:
Tip: take a look at `From` and `To` parameters.

michalwski:
Looks good now :) Could you also add some documentation in the `doc/modules`? If so, please remember to modify `mkdocs.yml` and test the doc by `mkdocs serve` (from repo root dir).

zsuidakra:
Sure, I'll write some docs :) Should I also add type specs?

michalwski:
Yes, please :)

zsuidakra:
Agh, I accidentaly closed PR again, sorry for that, but comment button is way to close to the 'close' button :/ 
I was going to ask if there is a way to rerun a build, because it seemed to fail not because of the changes I made, but it looks like reopening PR forces the build process.

michalwski:
All is good with your code. The failing job uses riak from Docker and this image got broken over the weekend.

zsuidakra:
Thanks :+1:  And sorry for trouble, I just got worried.

zsuidakra:
🎉

PR #732, opened 2016-04-11
Pubsub and PEP implementation

chrzaszcz:


apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -1200,6 +1208,44 @@ handle_info({force_update_presence, LUser}, StateName,
             StateData
     end,
     {next_state, StateName, NewStateData};
+handle_info({send_filtered, Feature, From, To, Packet}, StateName, StateData) ->
+    Drop = ejabberd_hooks:run_fold(c2s_filter_packet, StateData#state.server,
+				   true, [StateData#state.server, StateData,
+					  Feature, To, Packet]),
+    NewStateData = if Drop ->
```
michalwski: How about rewriting this whole big `if` with `case` statements into several short functions?

apps/ejabberd/src/gen_mod.erl
```diff
@@ -404,3 +414,17 @@ clear_opts(Module, Opts0) ->
...
+
+-spec v_db(db_type() | internal) -> db_type().
+
+v_db(odbc) -> odbc;
+v_db(internal) -> mnesia;
+v_db(mnesia) -> mnesia;
+v_db(riak) -> riak.
+
+db_type(Host, Module) when is_atom(Module) ->
+    get_module_opt(Host, Module, db_type, mnesia);
```
michalwski: I'd like to be consistent with other our modules and use param `backend` instead of `db_type`.

apps/ejabberd/src/mod_caps.erl
```diff
@@ -0,0 +1,759 @@
...
+%%%
+%%% 2009, improvements from ProcessOne to support correct PEP handling
+%%% through s2s, use less memory, and speedup global caps handling
+%%%----------------------------------------------------------------------
+
+-module(mod_caps).
+
+-author('henoch@dtek.chalmers.se').
+
+-protocol({xep, 115, '1.5'}).
```
michalwski: To be consistent with other our modules we should use `-xep` here instead of `-protocol`

apps/ejabberd/src/pubsub_migrate.erl
```diff
@@ -0,0 +1,443 @@
...
+
+-module(pubsub_migrate).
+
+-include("ejabberd.hrl").
+-include("pubsub.hrl").
+
+-export([update_node_database/2, update_state_database/2]).
+-export([update_item_database/2, update_lastitem_database/2]).
+
+update_item_database_binary() ->
```
michalwski: I guess we don't need this function since we start with binaries and there is no need to upgrade from strings.

apps/ejabberd/src/pubsub_migrate.erl
```diff
@@ -0,0 +1,443 @@
...
+    case mnesia:transaction(F) of
+	{aborted, Reason} ->
+	    ?ERROR_MSG("Failed to binarize pubsub items table: ~p", [Reason]);
+	{atomic, no_need} ->
+	    ok;
+	{atomic, Result} ->
+	    ?INFO_MSG("Pubsub items table has been binarized: ~p", [Result])
+    end.
+
+update_item_database(_Host, _ServerHost) ->
```
michalwski: Same here, as for now there is no 'old' pubsub items.

apps/ejabberd/src/pubsub_migrate.erl
```diff
@@ -0,0 +1,443 @@
+%%%----------------------------------------------------------------------
+%%% File    : pubsub_migrate.erl
```
michalwski: I think we can remove this module completely. There is nothing to migrate from as for now.

Issue #697, opened 2016-03-03
ejabberd_service - route to local components first


michalwski:
Modify the component routing so that first it's checked if the component is registered locally, route to an externally registered service as a last resort (allow to register the same component on different nodes).

PR #734, opened 2016-04-12
Route local pr

bartekgorny:
This PR addresses #697.

Proposed changes include:

* two-way registration of external components: globally and locally for a node
* another routing module which goes before mongoose_router_external and tries to route to a locally registered component first


apps/ejabberd/src/ejabberd_router.erl
```diff
@@ -326,9 +415,10 @@ compile_routing_module() ->
 
 default_routing_modules() ->
     [mongoose_router_global,
-    mongoose_router_external,
-    mongoose_router_localdomain,
-    ejabberd_s2s].
+     mongoose_router_external_localnode,
+     mongoose_router_external,
+     mongoose_router_localdomain,
```
michalwski: Can we have `mongoose_router_localdomain` before external components by default? I think this is the most most plausible case when we route stanzas to other users in the system.

apps/ejabberd/src/ejabberd_router.erl
```diff
@@ -118,51 +124,133 @@ route_error(From, To, ErrPacket, OrigPacket) ->
...
+            NDomain = {LDomain, Node},
+            Component = #external_component{domain = NDomain, handler = Handler, node = Node},
+            mnesia:write(Component);
+        _ -> mnesia:abort(route_already_exists)
+    end.
+
+check_component(LDomain, Node) ->
+    check_component(route, LDomain, Node).
+
+check_component(route, LDomain, Node) ->
```
michalwski: IMO, it would be more readable to have 3 differently named function instead of pattern matching since the atoms `external_comp_local` and `external_com_global` are used only for this.

michalwski:
Please apply suggestions and re-open to master branch.

apps/ejabberd/src/ejabberd_router.erl
michalwski: Example pseudocode just to explain what I mean:
```erlang
check_component(LDomain, Node) ->
    sth:sth() orelse check_component_local(LDomain, Node).

 check_component_local(LDomain, Node) ->
    sth2:sth() orelse check_component_global(LDomian, Node).

check_component_global(LDomain, Node) ->
   sth3:sth().
```

Issue #711, opened 2016-03-18
in file ejabberd_listener.erl:175  , when call listen_tcp sucess  , it should add ets:delete(listen_sockets, { PortIPProto, ListenSocket}),


ChelinTsien:
MongooseIM version: (put the version)
Installed from: (source or pkg)
Erlang/OTP version: (put the version)

Describe the issue.
in file ejabberd_listener.erl:171  , when call listen_tcp sucess , it should add ets:insert(listen_sockets, { PortIPProto, ListenSocket}),

michalwski:
Thanks a lot for reporting this. This is very important issue. How did you find this?

ChelinTsien:
IN file ejabberd_config:601, line 601&602 are the same with line 605&606, so i think you can delete one of them

ChelinTsien:
@michalwski IN file ejabberd_config:601, line 601&602 are the same with line 605&606, so i think you can delete one of them

bartekgorny:
I'm not sure it really should - the implementation dates back to 2010, and its purpose is to prevent mongoose crash on startup if a port is in use. It supports only xml_stream ports, and it works, though (a) there is no test to prove it, and (b) I'm not sure this is a desired behaviour. If an xml ports (e.g. c2s) is in use then mongoose starts fine and doesn't event issue warnings, it just doesn't receive any data because everything is captured by the other application using this port. One might argue that if an app can not use a port it needs then it should crash (as it indeed does if you block a port used by cowboy).

bartekgorny:
Turns out that the real reason why ejabberd can be started when something listens on its port is this:
`{reuseaddr, true},`
in ejabberd_listener. It's been there since the very beginning (entered by Alexey Shchepin in 2002). It can be removed or changed to 'false', this makes the daemon properly crash on startup if a port is in use, but it makes it very difficult to restart listeners because if the process keeps running then broken connections still hang around for a while in a transitional state making in impossible to bind to the port again (if reuseaddr is false). It could probably be solved somehow, I'm not sure it is worth the effort.

PR #738, opened 2016-04-13
Removed unused code, tests for listeners

bartekgorny:
This PR addresses #711, though in a bit different way: the code in question had an unclear purpose and since 2014 was not used at all, so it could be safely removed. I added some tests to check whether listeners are started/stopped properly.

One notable change in management api: before, stopping listeners cause them to be erased from config, so stop and start were not symmetric - starting them again required running ejabberd_config:reload_local.


apps/ejabberd/test/ejabberd_listener_SUITE.erl
```diff
@@ -2,9 +2,19 @@
...
+
+-define(DEFAULT_PORTS, [5222, 5280, 5269]).
+-define(ALT_PORTS, [5222, 5280, 5296]). %% yes it is different
+
+
 all() ->
-    [tcp_socket_is_started_with_default_backlog,
-     tcp_socket_is_started_with_options].
+    [tcp_start_stop_reload].
+%%    [tcp_socket_is_started_with_default_backlog,
```
michalwski: Why did you disable these tests?

apps/ejabberd/test/ejabberd_listener_SUITE_data/ejabberd.alt.cfg
```diff
@@ -0,0 +1,745 @@
+%%%
```
michalwski: I know we did it before for other ejabberd.cfg used in tests, but I think it'd be better to remove all the commented lines, for the sake of readability.

apps/ejabberd/test/ejabberd_listener_SUITE.erl
bartekgorny: ooops... just to make it run faster locally, and forgot to reenable them. Fortunately you are on guard :)

michalwski:
Could you please modify all new file `ejabberd.*.cfg` by removing commented lines? This lines are not necessary for tests and makes the diff needlessly big.

michalwski:
Ok, I removed these lines on my own. Also I removed ejabberd.cfg file as this one is produces from other config files and there is no need to keep it in repo.

PR #746, opened 2016-04-18
Route local pr

bartekgorny:
This PR addresses #697.

Proposed changes include:

* two-way registration of external components: globally and locally for a node
* another routing module which goes before mongoose_router_external and tries to route to a locally registered component first


apps/ejabberd/src/ejabberd_router.erl
```diff
@@ -326,9 +424,10 @@ compile_routing_module() ->
 
 default_routing_modules() ->
```
michalwski: I think my question about the default order got lost in prev PR. Is it possible to have `mongoose_router_localdomain` before all external component routers by default? Routing to local user is the most frequent case.

PR #757, opened 2016-04-26
Make calls to accept/3 and udp_recv/3 fully qualified

arkdro:
Calls to accept/3 and udp_recv/3 are local. This results in the ejabberd_listener keeping
the old version of code. The manual code reloading forces the listener process to crash.
With fully qualified calls the manual code reloading would not
crash the listener process.

### DIFF
--- a/apps/ejabberd/src/ejabberd_listener.erl
+++ b/apps/ejabberd/src/ejabberd_listener.erl
@@ -29,6 +29,8 @@
 
 -export([start_link/0, init/1, start/3,
          init/3,
+         accept/3,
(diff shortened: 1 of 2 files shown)


michalwski:
Hi @arkdro,

Many thanks for your PR. Would you mind writing a test case in [ejabberd_listener_SUITE](https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/test/ejabberd_listener_SUITE.erl)
 covering the `ejabberd_listener:init_udp/6` function?

michalwski:
Many thanks @arkdro!

Issue #739, opened 2016-04-14
It's not possible to remove dead node from cluster


studzien:
MongooseIM version: (master, 2c322ab35d7)

After removing ``remove_from_cluster`` command from ``mongooseimctl`` it's not possible to remove a node that is already dead and is not coming back.

michalwski:
In other words we need to restore `mongooseimctl remove_from_cluster` removed in #717.

ludwikbukowski:
I can restore `add_to_cluster` and `remove_from_cluster` but rewriten in the `mongoose_cluster.erl` module so the lines in the `nodetool [1]` would be removed. 
Then there will be four commands available:
- `join_cluster`
- `leave_cluster`
- `add_to_cluster`
- `remove_from_cluster`

Maybe also change the names of the new commands to `join_node` (instead of `add_to_cluster`) and `leave_node` (instead of `remove_from_cluster`)? Then the command names would be more consistent. 
However it would break backward name compability...
What do you think?

[1] https://github.com/esl/MongooseIM/blob/8112a1db3792f96e891f69120f946bb5cabe72ee/rel/files/nodetool#L55

michalwski:
The only relevant command from previous version is `remove_from_cluster` from the reason described in this issue. There is no need to restore other commands.

PR #758, opened 2016-04-27
Remove from cluster

ludwikbukowski:
This PR addresses https://github.com/esl/MongooseIM/issues/739

Proposed changes include:
* restored `remove_from_cluster` command. Use it in purpose to remove dead <strike>(and only dead)</strike> node from mongoose cluster.
* Added new node `ejabberd3@localhost` to travis architecture
* updated and added new tests in `cluster_commands_SUITE`
* updated docs

**[EDIT]**
After disscusion with _mim_ team I will modify the `remove_from_cluster` command so that it will be also possible to force remove _alive_ node.


ludwikbukowski:
Only one test doesn't pass:
`compontent_SUITE:register_in_cluster/1`
after adding `ejabberd3@localhost` to travis architecture.
I'm trying to figure out why

test/ejabberd_tests/test.config
```diff
@@ -31,6 +31,10 @@
                  {domain, <<"localhost">>},
                  {vars, "node2_vars.config"},
                  {cluster, mim}]},
+         {mim3, [{node, ejabberd3@localhost},
```
michalwski: Do we really want to name the new node with `ejabberd` prefix?

test/ejabberd_tests/tests/cluster_commands_SUITE.erl
```diff
@@ -308,14 +403,21 @@ normalize_args(Args) ->
...
-    run_interactive(Cmd, Response, timer:seconds(30)).
+    run_interactive(Cmd, Response, timer:seconds(60)).
 
 run_interactive(Cmd, Response, Timeout) ->
     Port = erlang:open_port({spawn, Cmd}, [exit_status]),
     %% respond to interactive question (yes/no)
     Port ! {self(), {command, Response}},
     loop(Port, [], Timeout).
 
+run(Cmd) ->
```
michalwski: I see similar function in `ejabberd_ctl_helper`. Can we reuse it here?

ludwikbukowski: yes we can, I will also reuse `normalize_args/1` and `loop/3` from the `ejabberdctl_helper` module

test/ejabberd_tests/test.config
ludwikbukowski: I'd love to name it with `mongooseim` prefix. I feel free to rename also `ejabberd2@localhost` node to `mongooseim2@localhost`. Is there any reason why not to do that?

michalwski: Yes, some tests may fail because of that :) so that's work for separate PR I think.

PR #796, opened 2016-05-05
Add support for <text> tag in auth failures

bernardd:
This change allows SASL authentication mechanism implementations to optionally supply a value for the `<text>` tag in `<failure>` results (as permitted by https://tools.ietf.org/html/rfc6120#section-6.5). It's backwards-compatible, so shouldn't break anything.

### DIFF
--- a/apps/ejabberd/src/cyrsasl.erl
+++ b/apps/ejabberd/src/cyrsasl.erl
@@ -74,11 +74,16 @@
                                     DigestGen :: fun()) ->
                                       'false' | {'true', ejabberd_auth:authmodule()}
                                   ).
+
+% Either a simple error tag or an error tag + <text> field
+-type sasl_error_data() :: binary() | {binary(), binary()}.
+
 -export_type([get_password_fun/0,
               check_password_fun/0,
               check_pass_digest_fun/0,
               mechanism/0,
-              password_type/0]).
+              password_type/0,
+              sasl_error_data/0]).
 
 -callback mech_new(Host :: ejabberd:server(),
                    GetPassword :: get_password_fun(),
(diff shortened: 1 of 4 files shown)


michalwski:
Hi @bernardd,

Many thanks for your PR! Could you given an example when this tag would be used? Would be perfect if you could came up with a test case for that.

Please, keep in mind that returning too much details about failed authentication may be dangerous.

bernardd:
Hi Michał,
I won't give you an example of how we're planning to use it, because it's a gross abuse of the protocol (I'd actually rather be able to add arbitrary fields for our purposes, but I figured if I was going to get a patch accepted I should stick to the bounds of the RFC :)).
However the example in the RFC is a decent one:
```xml
<failure xmlns='urn:ietf:params:xml:ns:xmpp-sasl'>
        <account-disabled/>
        <text xml:lang='en'>Call 212-555-1212 for assistance.</text>
</failure>
```
Again, it's not intended to be used all the time (as you say, too much info can be a security issue), or by any of the built-in cyrsasl modules necessarily (though nor would they be precluded from using it, of course). Rather, it's just to make the full functionality permitted by the RFC available to anyone implementing their own SASL module (as we're doing).

I'll see what I can do about putting together a test case tomorrow.

bernardd:
I've added a test case for this feature. A couple of the CI builds look to have failed for some reason, but I can't see anything in there that would point to the added test being the cause.

bernardd:
I now doubly think that the failures have nothing to do with my change, because a simple comment change triggered a different build to fail and the two previous failures to succeed :)

michalwski:
Many thanks for the tests! Regarding the failing job - you are right this has nothing to do with your code. This test case likes to fail on travis from time to time due to some timing issues. I've restarted the job and as soon as it passes I merge your PR!

PR #834, opened 2016-05-21
Lager init

arkdro:
Do not change lager loglevel when the parameter is specified in the env.

When the ejabberd app is started all lager handlers are set to the same loglevel,
either from config or default. Sometimes it is necessary to keep different handlers at
different loglevels. In this case the loglevels are set for the lager app and
a particular parameter is set in the ejabberd env so ejabberd doesn't changes
the lager loglevels on start.

### DIFF
--- a/apps/ejabberd/src/ejabberd_app.erl
+++ b/apps/ejabberd/src/ejabberd_app.erl
@@ -223,5 +223,10 @@ load_drivers([Driver | Rest]) ->
 
 init_log() ->
     ejabberd_loglevel:init(),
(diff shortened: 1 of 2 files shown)


michalwski:
Hi @arkdro,

Thank you for your contribution, this is indeed helpful. Could you please add few words about this option to our doc [Advanced-configuration.md#appconfig](https://github.com/esl/MongooseIM/blob/master/doc/Advanced-configuration.md#appconfig)?

michalwski:
Many thanks @arkdro!

PR #835, opened 2016-05-24
Muc test fix

bartekgorny:
As it was, if muc_light_legacy_SUITE was run before muc_SUITE then the latter failed. The reason was that muc was loaded statically at startup, and registered itself under muc.localhost in local routing table. muc_light was loaded dynamically, registered itself under the same domain, and then was unloaded.

### DIFF
--- a/rel/files/ejabberd.cfg
+++ b/rel/files/ejabberd.cfg
@@ -662,18 +662,15 @@
                            % default: 600 seconds
                            % {resume_timeout, 600}
                           ]},
-%  {mod_muc_light, [
-%                   {host, "muclight.@HOST@"}
-%                  ]},
-  %{mod_muc, [
-             %{host, "muc.@HOST@"},
-             %{access, muc},
-             %{access_create, muc_create}
-            %]},
-  %{mod_muc_log, [
-                 %{outdir, "/tmp/muclogs"},
-                 %{access_log, muc}
-                %]},
+  %% {mod_muc_light, [{host, "muclight.@HOST@"}]},
+  %% {mod_muc, [{host, "muc.@HOST@"},
+  %%            {access, muc},
+  %%            {access_create, muc_create}
+  %%           ]},
+  %% {mod_muc_log, [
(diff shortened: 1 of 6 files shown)


rel/files/ejabberd.cfg
```diff
@@ -665,15 +665,15 @@
...
-                ]},
+  %{mod_muc, [
+             %{host, "muc.@HOST@"},
+             %{access, muc},
+             %{access_create, muc_create}
+            %]},
+  %{mod_muc_log, [
+                 %{outdir, "/tmp/muclogs"},
+                 %{access_log, muc}
+                %]},
```
ludwikbukowski: I would move those comment characters to the beginning of each line but this is minor issue.
For me the PR is ready to merge

rel/files/ejabberd.cfg
```diff
@@ -662,18 +662,15 @@
                            % default: 600 seconds
                            % {resume_timeout, 600}
                           ]},
-%  {mod_muc_light, [
-%                   {host, "muclight.@HOST@"}
-%                  ]},
-  {mod_muc, [
```
michalwski: Do we really want to disable `mod_muc` by default? I know one can enable it in config file, but most users who just want to try it will probably never touch config file.

Nyco: Isn't MUC obsolete?

michalwski: Did XSF obsolete MUC? We can say it's obsolete, I'm fine with it but then we should enable by default `mod_muc_light`.

fenek: Wouldn't it be simpler to check if these two can peacefully coexist on separate subdomains?

michalwski: That's exactly my point.

Nyco: The market says MUC is obsolete, the XSF won't say it, or until MIX is widely deployed.

Indeed: MUC, MUC light, and MIX can co-exist.

Issue #839, opened 2016-05-29
MUC Light Change Room Replacing Config


apparition47:
MongooseIM version: `master b13807d Merge pull request #577 from esl/muc_light`
Installed from: `source`
Erlang/OTP version: `18`

I'm trying to change the room configuration as per [XMPP spec](https://xmpp.org/extensions/inbox/muc-light.html#set-room-config). Here, I changed the room name to "abc6".

```
XMPP OUT: 
<iq from='me@example.com/23276826391463878223526183' id='0a1e2cdd-22d0-40a0-af57-8b48a87cf345' to='05b5c392-cd63-4645-81d9-6a17ba2c5621@muclight.example.com' type='set' xmlns='jabber:client'>
<query xmlns='urn:xmpp:muclight:0#configuration'><roomname>abc6</roomname>
</query></iq>
```

When I attempt to get the room name (by retrieving the room config), if you look at the "XMPP IN", you'll see all the `roomname` XML tags of the previous `roomname` changes I attempted. The XMPP server is just stacking all the `roomname` tags together without replacing it.

```
XMPP OUT: 
<iq from='me@example.com/23276826391463878223526183' id='97cf3637-ec60-4aa3-98b3-8c7921f6868e' type='get' to='05b5c392-cd63-4645-81d9-6a17ba2c5621@muclight.example.com' xmlns='jabber:client'>
<query xmlns='urn:xmpp:muclight:0#configuration'/>
</iq>

XMPP IN: 
<iq from='05b5c392-cd63-4645-81d9-6a17ba2c5621@muclight.example.com' to='me@example.com/23276826391463878223526183' id='97cf3637-ec60-4aa3-98b3-8c7921f6868e' type='result' xmlns='jabber:client'>
<query xmlns='urn:xmpp:muclight:0#configuration'>
<version>1463-879661-878176</version><roomname>Untitled</roomname><roomname>abc</roomname>
<roomname>abc2</roomname>
<roomname>abc6</roomname>
<subject></subject><subject>abc4</subject>
</query></iq>
```

I thought the `roomname` tag was supposed to be replaced in this case. Any ideas? Thanks in advance.

fenek:
My bad, sorry. :( Here:
https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mod_muc_light_db_mnesia.erl#L334
It should be `lists:keymerge`, not `lists:umerge`. Can somebody with spare time fix it please?

apparition47:
[...]

Issue #838, opened 2016-05-27
Muc light disco#items error


ludwikbukowski:
MongooseIM version: b13807d90b2167f08f92257af3813c157c720552
Installed from: source
Erlang/OTP version: 17.5

When I request for room list and there is no rooms created I got error `item-not-found` response:
Request:

````
<iq to='muclight.localhost' id='lV6CW-16' type='get'>
<query xmlns='http://jabber.org/protocol/disco#items'></query>
</iq>
````
Response:

````
<iq from="muclight.localhost" type="error" xml:lang="en"
 to="user@localhost/Administrators-MacBook-Pro" id="lV6CW-16">
<error type="cancel" code="404">
<item-not-found xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/>
</error>
</iq>
````
--
It happens when the _mod_muc_light_  configuration has option `rooms_per_page` set to any Integer:
``
  {mod_muc_light, [
                   {host, "muclight.@HOST@"},
                   {rooms_per_page, 10}
                  ]},
``

When I change it to `infinity` items request results with empty list and it's expected behaviour:

Request:
````
<iq to='muclight.localhost' id='lV6CW-16' type='get'>
<query xmlns='http://jabber.org/protocol/disco#items'></query>
</iq>
````
Response:
````
<iq from="muclight.localhost" type="result"
 to="user@localhost/Administrators-MacBook-Pro" id="lV6CW-16">
<query xmlns="http://jabber.org/protocol/disco#items"/>
</iq>
````
The problem dissapears when there is at least one room
I wrote some tests for it here [1]

[1] https://github.com/esl/MongooseIM/blob/muc-light-disco/test/ejabberd_tests/tests/muc_light_SUITE.erl#L250

PR #841, opened 2016-05-30
Muc light disco#items fix + room confuguration fix

ludwikbukowski:
This PR addresses https://github.com/esl/MongooseIM/issues/838 and #839

Proposed changes include:
For #838:
 added tests with four cases:
- set `{rooms_per_page, infinity}` and empty room list
- set `{rooms_per_page, infinity}` and not empty room list
- set `{rooms_per_page, 1}` and not empty room list
- set `{rooms_per_page, 1}` and empty room list <-- this test hasn't passed

Fixed the issue in module `mod_muc_light`
 ---
For #839 :
- added test for changing room name configuration
- fixed the issue with replacing the room configuration in mnesia backend

### DIFF
--- a/test/ejabberd_tests/tests/muc_light_SUITE.erl
+++ b/test/ejabberd_tests/tests/muc_light_SUITE.erl
@@ -401,7 +401,7 @@ change_subject(Config) ->
         end).
 
 change_roomname(Config) ->
-    escalus:story(Config, [{alice, 1}, {bob, 1}, {kate, 1}], fun(Alice, Bob, Kate) ->
+    escalus:story(Config, [{alice, 1}], fun(Alice) ->
         %% change room name
         ConfigChange = [{<<"roomname">>, <<"new_test_room">>}],
         Stanza = stanza_config_set(?ROOM, ConfigChange),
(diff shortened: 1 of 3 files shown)


bartekgorny:
What can I say - this PR indeed fixes bugs it was supposed to fix, has an appropriate test coverage, by all means it is ok. We might do with some more comments and docstrings, but this applies to the whole muc_light, not to this PR. Green light from me.

test/ejabberd_tests/tests/muc_light_SUITE.erl
```diff
@@ -231,6 +248,60 @@ disco_features(Config) ->
                                                      {attr, <<"var">>}]),
             escalus:assert(is_stanza_from, [?MUCHOST], Stanza)
         end).
+%% The room list is empty
+disco_rooms_empty_page_infinity(Config) ->
```
michalwski: This and `disco_rooms_empty_page_1` looks very similar. I noticed only one different line. Can they be somehow refactored in order not to duplicate code? I think the same applies to next 2 test cases.

Issue #843, opened 2016-05-31
Old sessions keep in ?SM_BACKEND on resumed session that never will be restored


artefactop:
MongooseIM version: 1.6.2
Installed from: source
Erlang/OTP version: 18

When a client resume a session it is not removed from backend https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/ejabberd_c2s.erl#L1488 then if the same client instead restore the session make a new one, the Pid of previous connection doesn't exists so the `replaced` message never arrive https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/ejabberd_sm.erl#L829 and no one delete the entry of previous session

to fix that:

```
              fun({_, Pid} = S) when S /= MaxSID ->
                case process_info(Pid) of
                  undefined -> close_session(S, LUser, LServer, LResource, replaced);
                  _ ->
                    Pid ! replaced
                end;
```

bartekgorny:
I have a problem to reproduce it. I wrote a test whereby a user established a connection, then another connection with the same resource, and the first connection receives "conflict - replaced by a new connection", exactly as it should. Please advice.

artefactop:
Sorry, I forgot to comment, I think the problem is when the new connection goes to different mongooseim node at least in redis backend.

bartekgorny:
Sorry to say that, but I wrote a test for two nodes and ran it with redis backend, and it passed 😕

artefactop:
And old session is replaced/deleted from redis?

bartekgorny:
I ran it with redis, checking user resources and sessions on both nodes, still no result - when user disconnects the backend is clean. This:
https://gist.github.com/bartekgorny/1cf4af3fbc37d9ea4a7107e1708eb3c0
is the test I ran, both on mnesia and redis. And it passes. Anything else I can do to reproduce?

artefactop:
I checked the test and I miss one test case:
- clientA connect to node1
- node1 lose connectivity with clientA and after timeout of stream management it resume the terminate process with `StateData#state.authenticated = resumed`
- clientA connect to node2 (without disconnecting from node1 due a network error)
- now ejabberd_sm:get_user_resources(User, Server) return 2 
- clientA disconnect from node2
- now ejabberd_sm:get_user_resources(User, Server) return 1

Jeyhey:
[...]

PR #849, opened 2016-06-06
Test for log in with conflicting resources

bartekgorny:
This PR is a failed attempt to reproduce #843, and at the same time the missing test for resource conflict use case.

### DIFF
--- a/resource_SUITE.erl
+++ b/resource_SUITE.erl
@@ -1,0 +1,93 @@
+%%==============================================================================
+%% Copyright 2010 Erlang Solutions Ltd.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


artefactop:
You are checking that client receive the response `escalus:assert(is_stream_error, [<<"conflict">>, <<"Replaced by new connection">>], E).` but in don't checking if session is still saved on redis

bartekgorny:
True, because it proves that 'replaced' atom is sent, received and processed. Are you 110% sure the session remains in backend database? If so, then the problem is bad cleanup.

artefactop:
Yes, it happened for us in production, could you check if when second client goes offline is you request `ejabberd_sm:get_user_resources(User, Server)` it returns one entry (the old one)

bartekgorny:
ok, I'll see to it

michalwski:
The duplicated sessions where most probably due to the race condition fixed in #1141

Issue #852, opened 2016-06-10
Exception while trying to get list of users


kshamko:
MongooseIM version: 1.6.2
Installed from: source
Erlang/OTP version: 18 [erts-7.2]

There is an exception while trying to get users list with GET request to http://localhost:5288/api/users/host/localhost 

Trace:
(mongooseim@localhost)1> 2016-06-10 12:16:09.577 [error] emulator Error in process <0.740.0> on node mongooseim@localhost with exit value:
[...]

PR #857, opened 2016-06-13
fix for issue #852

kshamko:
This PR addresses #852


michalwski:
Thanks for the patch @kshamko. Is it possible for you to write a test case reproducing the wrong behaviour? I think such a test case could be added to [users_api_SUITE.erl](https://github.com/esl/MongooseIM/blob/master/test/ejabberd_tests/tests/users_api_SUITE.erl)

kshamko:
NP. will do that in a week or two

PR #858, opened 2016-06-14
Muc light disco#items test refactored

ludwikbukowski:
This PR addresses https://github.com/esl/MongooseIM/pull/841#discussion_r66929343

Proposed changes include:
* removed code duplication in four test cases

### DIFF
--- a/test/ejabberd_tests/tests/muc_light_SUITE.erl
+++ b/test/ejabberd_tests/tests/muc_light_SUITE.erl
@@ -266,12 +266,12 @@ disco_rooms_empty_page_1(Config) ->
 %% There is one room created. Rooms_per_page set to 1
 disco_rooms_created_page_1(Config) ->
     set_mod_config(rooms_per_page, 1),
-    escalus:story(Config, [{alice, 1}], verify_user_has_one_room()).
+    escalus:story(Config, [{alice, 1}], fun verify_user_has_one_room/1).
 
 %% There is one room created. Rooms_per_page set to `infinity`
 disco_rooms_created_page_infinity(Config) ->
     set_mod_config(rooms_per_page, infinity),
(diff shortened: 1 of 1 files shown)


test/ejabberd_tests/tests/muc_light_SUITE.erl
```diff
@@ -232,19 +249,40 @@ disco_features(Config) ->
             escalus:assert(is_stanza_from, [?MUCHOST], Stanza)
         end).
 
+%% The room list is empty. Rooms_per_page set to `infinity`
+disco_rooms_empty_page_infinity(Config) ->
+    escalus:story(Config, [{alice, 1}], fun(Alice) ->
+        set_mod_config(rooms_per_page, infinity),
+        [] = get_disco_rooms(Alice)
+                                        end).
```
michalwski: Indentation. I know all auto indentation tools do it this way but in all our test cases (or at least most of them) we have the closing `end` for `escalus:story` fun on the same level as `escalus:story`. See the test case above.

test/ejabberd_tests/tests/muc_light_SUITE.erl
```diff
@@ -965,6 +943,14 @@ info_iq_verify_fun(AffUsers, Version, ConfigKVBin) ->
             verify_config(ConfigurationEl, ConfigKVBin)
     end.
 
+-spec verify_user_has_one_room() -> fun((User ::escalus:client()) -> any()).
+verify_user_has_one_room() ->
```
michalwski: It would more readable if the function took `User` as a parameter. In this case you don't have to return `fun` but simply do what you need.

Such a function can be passed to escalus:story like this: `fun verify_user_has_one_room/1`

PR #859, opened 2016-06-14
Blocking commands - tests

bartekgorny:
Parallelised, added to default.spec


michalwski:
Please take a look at logs from latest build (all jobs not using mnesia for persistent data, f,e [mysq_mnesia](http://mongooseim-ct-results.s3.amazonaws.com/PR/859/1635/3/big/mongooseim_node1/log/ejabberd.log)). There are some crashes while executing `mod_blocking_SUITE`. Hope that helps.

Issue #862, opened 2016-06-17
crash calling is_user_exists


igors:
MongooseIM version: 1.6.2
Installed from: source (master branch)
Erlang/OTP version: (18)

auth_method set to [anonymous, ldap]. The crash is as follows:

** Reason for termination =
** {undef,[{ejabberd_auth_ldap,is_user_exists,[<<"some_user">>,<<"some_domain">>],[]},{ejabberd_auth,does_user_exist_in_other_modules_loop,3,[{file,"src/ejabberd_auth.erl"},{line,475}]},{ejabberd_auth_anonymous,check_password,5,[{file,"src/ejabberd_auth_anonymous.erl"},{line,227}]},{ejabberd_auth,do_check_password_loop,2,[{file,"src/ejabberd_auth.erl"},{line,213}]},{timer,tc,2,[{file,"timer.erl"},{line,181}]},{ejabberd_auth,timed_call,4,[{file,"src/ejabberd_auth.erl"},{line,588}]},{cyrsasl_plain,mech_step,2,[{file,"src/cyrsasl_plain.erl"},{line,57}]},{cyrsasl,server_step,2,[{file,"src/cyrsasl.erl"},{line,211}]}]}

The crash happens when calling AuthModule:is_user_exists in does_user_exist_in_other_modules_loop in ejabberd_auth.erl. Note that is_user_exists was renamed to do_user_exist in commit e4ca38c8e622be5dae3f0bb5b4c1ff682a5afecc

PR #863, opened 2016-06-17
rename is_user_exists to does_user_exist

igors:
This PR addresses #862 

* renames an instance of is_user_exists() to do_user_exist that was missed in commit e4ca38c8e622be5dae3f0bb5b4c1ff682a5afecc


michalwski:
Thanks @igors!

PR #866, opened 2016-06-21
docs on databases

ludwikbukowski:
Proposed changes include:
* `Choosing a database for MongooseIM` section in `database-backends-configuration.md`

### DIFF
--- a/doc/advanced-configuration/database-backends-configuration.md
+++ b/doc/advanced-configuration/database-backends-configuration.md
@@ -5,6 +5,13 @@ them require extra work before they can be used. For example the SQL databases r
 defining schema. MongooseIM is tested with TravisCI, so the travis scripts can be used
 as a reference.
 
+# A Brief Overview
+
+Data in Mongoose is either transient or persistent:
+
+* **transient**: session data, stream management data, and other in-memory data.
+* **persistent**: roster items, credentials, and chat archives.
+
 # Choosing a database for MongooseIM
 
 Here is some general advice on the use of databases.
--- a/doc/advanced-configuration/database-backends-configuration.md
+++ b/doc/advanced-configuration/database-backends-configuration.md
@@ -7,10 +7,10 @@ as a reference.
 
 # A Brief Overview
 
-Data in Mongoose is either transient or persistent:
+Data in MongooseIM is either transient or persistent:
 
-* **transient**: session data, stream management data, and other in-memory data.
-* **persistent**: roster items, credentials, and chat archives.
+* **transient**: volatile data changing often, such as session data, stream management data, and other in-memory data. These don't need any backup, since after a potential failure, they will naturally rebuild as clients reconnect.
+* **persistent**: long-lived data, such as roster items, credentials, and chat archives. These absolutely need regular and tested backups.
 
 # Choosing a database for MongooseIM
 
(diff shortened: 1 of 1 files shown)


Nyco:
Here is my thinking, please disagree:

* we should clearly separate live data from persistent data (whatever the name):
  * live, temporary, short-lived
  * persistent, long-lived
* mention that there is no need for any kind of backup for "live" data, as it will reconstruct over time, as clients reconnect
* Recommend Mnesia over Redis, as Redis is not widely used in production yet
* Tell the limits of Mnesia: old DB, code maintenance issues, sync all in real time (lot of internode traffic), one Mnesia node per MongooseIM node
* Tell the limits of Redis: few deployments in prod, different architecture (one cluster of MongooseIM nodes, one cluster of Redis nodes), lag bewteen the two clusters
* Tell that NOSQL DBs have each a different structure and query language, as opposed to RDBMS which have SQL in common: as a result, developing the support of another NOSQL backend is more work compared to RDBMS

A structure could be:
* Choosing a database for MongooseIM 
  * Live data
    * Mnesia (default, recommended)
    * Redis
  * Persistent data
    * RDBMS: PostgreSQL, MySQL (MS SQL Server?)
    * NOSQL

Dzol:
We could make the distinction here as being between **transient** and **persistent** data.

* **transient**: say chat room (context) and session data.
* **permanent**: say roster items, credentials, and archives.

michalwski:
I also suggest pointing out that we *strongly recommend* keeping persistent data in external DB (RDBMS or Riak) for production. Mnesia for persistent data is no good fit for production and sooner or later a migration will be need, which may be painful. Persistent data in mnesia (default config) are good for trying MongooseIM easily.

ludwikbukowski:
thanks guys! 
However, I'd like to clarify the meaning of `live data`. What do we mean by that? Is it about the `users session` only, which is just mapping between user `jid` and the corresponding process pid? 
For me, the `live data` are the `ejabberd_c2s`  erlang processes which we cannot store neither in mnesia nor in redis...

michalwski:
You are right @ludwikbukowski about the core `live data`. But I think adding sessions, stream management and other in-memory only data to the `live data` definition is good idea. I usually cal them `transient data` so maybe using this name will be more understandable.

Dzol:
I've made most of the changes suggested above. Please review. @Nyco: any further suggestions on how to get the limits of Mnesia and Redis in here?

doc/advanced-configuration/database-backends-configuration.md
```diff
@@ -1,9 +1,48 @@
...
+Persistent Data:
+
+* RDBMS/ODBC - MongooseIM has a strong backend support for relational databases. Considering the [CAP theorem](https://en.wikipedia.org/wiki/CAP_theorem)
+they usually guarantees both availability and consistency which is a great choice for regular MongooseIM use cases and features
+ like `privacy lists`, `vcards`, `roster`, `private storage`, `last activity` and `message archive`. Never loose your data.
+
+* Cassandra -  since it aims to store large amounts of data, in comparison to some ODBC databases, it's good alternative for
+ the`message archive` which is usually the biggest persistent component in MongooseIM. Archive everything.
+
+* Riak KV - If you're planning to deploy a really big cluster then consider Riak KV as a potential backend solution.
```
michalwski: I'd move `Riak KV` just after `RDBMS/ODBC`

doc/advanced-configuration/database-backends-configuration.md
```diff
@@ -1,9 +1,48 @@
...
+It offers high availability and fault tolerance which is excatly what you need for your distributed MongooseIM architecture.
+Use Riak KV with `rosters`, `private storage`, `vcards` or `last activity`. Make it big.
```
michalwski: In fact `Riak KV` can be used with the same modules as `RBMS/ODBC` why did you choose to mention only these?

michalwski:
Thanks @Dzol for updating the doc. A small hint for future. If you change only docs, pleas put `[skip ci]` in the first line of your commit message. It will save some travis runs.

Issue #785, opened 2016-05-04
Stop supporting Erlang/OTP R16


michalwski:
This will allow us to start using maps. Also will remove some workarounds:
- [x]  [latin1_characters](https://github.com/esl/MongooseIM/blob/f5068f7726b48526635feab88e17ce604a052543/apps/ejabberd/rebar.config#L5)
- [x]  [only_builtin_types](https://github.com/esl/MongooseIM/blob/f5068f7726b48526635feab88e17ce604a052543/apps/ejabberd/rebar.config#L6)
- [x] update docs wherever version R16B is used

See #582 and #613 to check how support for R15 was dropped.

PR #871, opened 2016-06-27
Ditch support for R16

michalwski:
This PR addresses #785

Proposed changes include:
* removed workaround for utf8 characters in host name
* removed workaround for prefixed types like `dict:dict()`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ branches:
                 - /\d+\.\d+\.\d+(\-[a-z,0-9])*/
 
 otp_release:
-        - 17.5
+        - 18.3
 env:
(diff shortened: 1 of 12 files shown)


ludwikbukowski:
we should get rid of **every** occurence of `dict_t` type:
* https://github.com/esl/MongooseIM/blob/8112a1db3792f96e891f69120f946bb5cabe72ee/apps/ejabberd/src/eldap.erl#L156
* https://github.com/esl/MongooseIM/blob/8112a1db3792f96e891f69120f946bb5cabe72ee/apps/ejabberd/src/shaper_srv.erl#L31

PR #872, opened 2016-06-29
Fix count and index types in rsm_out

bernardd:
Both `index` and `count` in `rsm_out` can, by my reading of the XEP-0059, have 0 as a value. The existing `pos_integer()` type, however, causes dialyzer warnings for this case.


michalwski:
Thanks!

PR #875, opened 2016-06-29
Review Makefile PHONY targets

lucafavatella:
This PR reviews the top-level makefile in order to make it more explicit and readable.

Proposed changes include:
- No functionality changes (well... if random files with same name as the targets were created in the top dir, these changes would fix some issues in the build infrastructure. But it is not very relevant);
- No new or updated tests;
- No changes to the documentation.

I coded these changes while reading and understanding the makefile, as they made the file more immediate to understand in my opinion. But most changes in this PR can be considered subjective, so please feel free to close without merge straight away.


michalwski:
The main makefile will be quite significantly changed in this #1033

PR #880, opened 2016-07-06
Swagger HTTP API doc

michalwski:
This PR adds documentation about the REST API we are building:

* client API: `client_swagger.yml`
* backend API: `backend_swagger.yml`

If you want to render this, please import a file to http://editor.swagger.io/


Nyco:
For `client_swagger.yml`:
* Not sure how to fix this one: "This API described alternate, to XMPP way of using MongooseIM."
* "Returns all groups to which the user is subscribed": I assume it is MUC light, or not. Should we say it clearly?
* "Creates new group of given name and sends invitation to specified users." If that's MUC light, that's not not an invitation: "Creates new group of given name and adds the specified users."
* "Returns messages exchanged in given group. TODO pagination": can/should we add the "`size`" parameter like in `GET /messages`?

For `swagger.yaml`:
* can we rename it to `backend-swagger.yaml` or `server-swagger.yaml`?
* Is there a need for `GET /{XMPPHost}`? I'd skip it if possible
* Is there a need for `GET /{XMPPHost}/roster`? I'd skip it if possible, as the result will be very large, and generate a lot of issues
* For `GET /{XMPPHost}/roster/{username}`? I'd clearly state no nested groups
* For `POST /{XMPPHost}/roster/{username}` and `DELETE /{XMPPHost}/roster/{username}` ? do we clearly state that there is no subsequent stanza `<presence type='subscribe|unsubscribe'/>`?
* Is there a need for `GET /{XMPPHost}/register`? I'd skip it if possible, as the result will be very large, and generate a lot of issues
* Will `DELETE /{XMPPHost}/register` remove roster items for contacts who have that user as a friend?
* How about a method for password change?
* What session is it in `GET /{XMPPHost}/session`? C2S or S2S? If C2S, it's gonna be too large?
* `DELETE /{XMPPHost}/session/{username}`? Does it send `<presence/>` to subscribed contacts? Does it properly close the session's stream?
* `GET /{XMPPHost}/session/{username}`: can/should we add ACL that user is in?

michalwski:
>"Returns all groups to which the user is subscribed": I assume it is MUC light, or not. Should we say it clearly?

We can mention that. But this should not be relevant for the client API consumer. The client has the REST API and that's all, doesn't have to know it's MUC light in the backend.

>"Creates new group of given name and sends invitation to specified users." If that's MUC light, that's not not an invitation: "Creates new group of given name and adds the specified users."

True, but if I remember correctly, they'll also get a message inviting them to the room. Or am I wrong here?

> "Returns messages exchanged in given group. TODO pagination": can/should we add the "size" parameter like in GET /messages?

Yes, `size` should be added now. Pagination is I think out of the scope for POC but has to be added in the final version.

michalwski:
>So we mix both?
>User is present in MUC1 and MUC2
>User has subscribed to MUClight3 and MUClight4

I don't understand? Why to mix? I think we should focus only on MUC light. And do not expose old MUC over REST API. It will only overcomplicate things (simulate presence of REST users as an example).

Nyco:
OK

priv/http_api/backend_swagger.yaml
```diff
@@ -0,0 +1,430 @@
...
+            title: XMPPHost
+            type: object
+            properties:
+              name:
+                type: string
+              information:
+                type: string
+        404:
+          description: Failure response because the given XMPP host does not exist
+  /{XMPPHost}/roster:
```
michalwski: I don't know if we need that method. More important (IMO) is to get list of all registered users. Heaving that we can query the API for specific user's roster.

priv/http_api/backend_swagger.yaml
```diff
@@ -0,0 +1,430 @@
...
+              username:
+                type: string
+              XMPPHostname:
+                type: string
+      responses:
+        200:
+          description: The item was successfully removed 
+        404:
+          description: Failure response because the given user does not exist under this XMPP host or the XMPP host itself does not exist
+  /{XMPPHost}/register:
```
michalwski: How about naming this `users` and not `register`? It suites better IMO.

priv/http_api/backend_swagger.yaml
```diff
@@ -0,0 +1,430 @@
...
+          required: true
+          schema:
+            title: username
+            type: string
+      responses:
+        200:
+          description: Cancelation successful
+        404:
+          description: Could not find the given user under the XMPP host
+  /{XMPPHost}/session:
```
michalwski: Probably plural would be better here - `sessions`. Basically it looks to me that all resources should be in plural form.

priv/http_api/backend_swagger.yaml
```diff
@@ -0,0 +1,430 @@
...
+                  type: integer
+                node:
+                  type: string
+                uptime:
+                  type: string
+        404:
+          description: Failure response because the given user does not exist under this XMPP host or the XMPP host itself does not exist
+  /cluster:
+    get:
+      description: Retrieve a list of clustered MongooseIM nodes.
```
michalwski: The implementation should look at mnesia info, and list `running` db nodes as well as `stopped`.

priv/http_api/backend_swagger.yaml
Dzol: https://github.com/esl/MongooseIM/commit/4e7aeb7489328f290750454a820dc79714200b40

priv/http_api/backend_swagger.yaml
Dzol: 07c1502af072c66091cb52b85ea8dacaeb100626 should resolve this.

priv/http_api/backend_swagger.yaml
Dzol: 6029e119c4dac49cddbe51f8890ec7e1ccc8fa96 makes more of the resource names plural.

Dzol:
> can we rename it to backend-swagger.yaml or server-swagger.yaml?

Done.

> Is there a need for GET /{XMPPHost}? I'd skip it if possible

The comment thread for this is [here](https://github.com/esl/MongooseIM/commit/e472f664b6ce480a6c73bc486df912707cd996ba#commitcomment-18094541).

> Is there a need for GET /{XMPPHost}/roster? I'd skip it if possible, as the result will be very large, and generate a lot of issues

This is gone. But functionality like that provided by `/{XMPPHost}/users` remains.

> For GET /{XMPPHost}/roster/{username}? I'd clearly state no nested groups

@Nyco: Is this what we want? We can change the response schema to capture nested groups.

> For POST /{XMPPHost}/roster/{username} and DELETE /{XMPPHost}/roster/{username}? do we clearly state that there is no subsequent stanza <presence type='subscribe|unsubscribe'/>?

@Nyco: Are you asking if MongooseIM will notify other users?

> Will DELETE /{XMPPHost}/register remove roster items for contacts who have that user as a friend?

Yes: we'll document this.

> How about a method for password change?

We'll put this in.

> what session is it in GET /{XMPPHost}/session? C2S or S2S? If C2S, it's gonna be too large?

C2S: it'll be paginated.

> DELETE /{XMPPHost}/session/{username}? Does it send <presence/> to subscribed contacts? Does it properly close the session's stream

Yes.

> GET /{XMPPHost}/session/{username}: can/should we add ACL that user is in?

We could include this In the future.

Nyco:
> > For GET /{XMPPHost}/roster/{username}? I'd clearly state no nested groups

> @Nyco: Is this what we want? We can change the response schema to capture nested groups.

Definitely not in the PoC, preferrably not in the MVP, at later stage when we reach maturity maybe, as nested groups are supported by few implems.

arcusfelis:
@Dzol Don't forget [skip ci] tag :)

doc/http-administration-api-documentation.md
```diff
@@ -0,0 +1,5 @@
+# MongooseIM's HTTP Administration API
+
+Find the beautiful Swagger documentation below:
+
+<iframe src="https://cdn.rawgit.com/esl/MongooseIM/swagger-http-api-doc/doc/swagger/index.html?url=https://raw.githubusercontent.com/esl/MongooseIM/swagger-http-api-doc/priv/http_api/backend_swagger.yaml#/" align="middle" height="700" width="700"></iframe>
```
Dzol: Serving the Swagger UI of off GitHub Pages is another option instead the proxy at `https://cdn.rawgit.com/esl/MongooseIM/swagger-http-api-doc/doc/swagger/index.html`.

michalwski:
Do we really need all this .js files? This looks like a source code but in the end usually there is one .js file compile and minfied from all the sources.

michalwski:
I bet, that's the one [swagger-ui.min.js](https://github.com/swagger-api/swagger-ui/blob/master/dist/swagger-ui.min.js)

michalwski:
The rendered version is here: http://mongooseim.readthedocs.io/en/swagger-http-api-doc/http-administration-api-documentation/

PR #881, opened 2016-07-08
HTTP client API for messages

michalwski:
This PR implements:
* sending messages via HTTP API
* basic archive reading via HTTP API
* rooms (muc_light) creation via HTTP API
* sending messages to rooms via HTTP API
* reading information re room via HTTP API

TODO:
- [x] optional parameter specifying the size of returned list (at most 50)


michalwski:
The updated swagger doc for this API can be found here: http://mongooseim.readthedocs.io/en/http-client-api/swagger/index.html?client=true#/default

apps/ejabberd/src/mongoose_client_api.erl
```diff
@@ -0,0 +1,61 @@
...
+rest_init(Req, _HandlerOpts) ->
+    {ok, Req, #{}}.
+
+is_authorized(Req, State) ->
+    Auth = cowboy_req:parse_header(<<"authorization">>, Req),
+    case Auth of
+        {ok, undefined, _} ->
+            make_unauthorized_response(Req, State);
+        {ok, AuthDetails, Req2} ->
+           do_authorize(AuthDetails, Req2, State)
```
elbrujohalcon: Inconsistent indentation.

apps/ejabberd/src/mongoose_client_api.erl
```diff
@@ -0,0 +1,61 @@
...
+    end.
+
+allowed_methods(Req, State) ->
+    {[<<"GET">>], Req, State}.
+
+content_types_provided(Req, State) ->
+	{[
```
elbrujohalcon: Inconsistent indentation within this whole function.

doc/http-api/client_swagger.yml
```diff
@@ -23,49 +26,55 @@ paths:
...
-          type: string
-          description: JID of the user with whom the messages were exchanged.
-          in: path
-          required: true
-        - name: limit
-          in: path
-          description: Number of recent messages to return. Lower than 500
-          required: true
+        - in: query
+          name: limi
```
elbrujohalcon: should be `limit`

doc/http-api/client_swagger.yml
```diff
@@ -76,30 +85,188 @@ paths:
...
+    post:
+      description: Creates a room.
+      parameters:
+        - in: body
+          name: room
+          required: true
+          schema:
+            $ref: '#/definitions/CreateRoomBody'
+      responses:
+        200:
```
elbrujohalcon: Should be either 201 or 204, if no content is returned in the response body.

apps/ejabberd/src/mod_mam.erl
```diff
@@ -666,7 +670,8 @@ is_interesting(Host, LocJID, RemJID, ArcID) ->
 %% ----------------------------------------------------------------------
 %% Backend wrappers
 
--spec archive_id_int(ejabberd:server(), ejabberd:jid()) -> integer().
+-spec archive_id_int(ejabberd:server(), ejabberd:jid()) ->
```
arcusfelis: Something like get_archive_integer_id/1 sounds better

doc/http-api/client_swagger.yml
```diff
@@ -76,30 +85,188 @@ paths:
...
+          name: id
+          required: true
+          type: string
+          description: The ID of a room
+      responses:
+        200:
+          description: Successful response
+          schema:
+            $ref: '#/definitions/RoomDetails'
+    put:
```
elbrujohalcon: Since this endpoint is not requiring the caller to provide the whole entity that's being modified, it doesn't match with the HTTP logic behind PUT. I would recommend 2 different approaches:

1. Using `POST /rooms/{id}/invites`. This will have the additional benefit of allowing us to provide a `GET|PUT|DELETE /rooms/{id}/invites/{invite-id}` endpoint, which is cool.
2. Using `PATCH /rooms/{id}`. This will have the additional benefit of flexibility.  `PATCH` is like a wildcard in HTTP terms, you can almost do whatever you want with it.

doc/http-api/client_swagger.yml
```diff
@@ -23,49 +26,55 @@ paths:
...
       description: |
-        Gets recent messages from the archive. For the authenticated users.
+        Gets recent messages with specified user from the archive.
       parameters:
         - name: with
           type: string
-          description: JID of the user with whom the messages were exchanged.
+          description: |
+            JID of the user with whom the messages were exchanged.
+            Example: **alice@wonderland.com**
```
arcusfelis: or **alice@wonderland.com/res1**

elbrujohalcon:
@michalwski as a side note: you should totally consider using [cowboy-swagger](https://hex.pm/packages/cowboy_swagger) for your API… or at least [cowboy-trails](https://hex.pm/packages/trails) to define your routes.
You should also consider using [mixer](https://hex.pm/packages/inaka_mixer) to reduce the number of _single line calls to base handler_ like [this one](https://github.com/esl/MongooseIM/pull/881/files#diff-6ff196eabf2f475f96beab1e550960adR24).

doc/http-api/client_swagger.yml
```diff
@@ -76,30 +85,188 @@ paths:
...
-      msg:
+      body:
         type: string
   RecvMessage:
     properties:
-      sender:
-        $ref: '#/definitions/JID'
+      to:
+        type: string
+        description: The message recipinet's bare JID.
```
arcusfelis: recipient

doc/http-api/client_swagger.yml
```diff
@@ -76,30 +85,188 @@ paths:
...
+      from:
+        type: string
+        description: The message sender's bare JID.
       timestamp:
         type: integer
-        description: Unix timestamp in seconds when the message was sent
-      message_id:
+        format: int64
+        description: Unix timestamp in milliseconds when the message was sent
```
arcusfelis: Some descriptions end with a dot, some - without.

apps/ejabberd/src/mod_mam.erl
michalwski: I agree, good catch :) I just fixed the spec here, didn't want to touch the function name in this PR.

doc/http-api/client_swagger.yml
michalwski: No, only bare JIDs in the REST API.

doc/http-api/client_swagger.yml
michalwski: Yes, that's good point. I think we'll go with the first approach, looks clearer.

PR #882, opened 2016-07-11
REST API HTTP backend [do-not-merge]

ludwikbukowski:
The Work is still in progress
Proposed changes include:
* simple backend engine for `GET` and `POST` and `DELETE` method
* integration with `Commands Engine` (introduced in `command_registry` branch
* tests for `GET`, `POST`, `DELETE`

--- 
TODO:
- [ ] use `jiffy` instead of` mochijson2`
- [ ] improve the tests (more groups, more scenarios etc)
- [ ] discuss: the compability between command engine and http connector (especially how arguments maps to query path and body)
- [ ] discuss: should `POST` method return the header `location` with the new resource created? And how could it be obtained from commands engine?
- [ ] discuss: which http status codes and when?


apps/ejabberd/src/mongoose_api_backend.erl
```diff
@@ -0,0 +1,234 @@
...
+    {halt, Req2, State};
+handle_result({error, Error}, Req, State) ->
+    error_response(Error, Req, State);
+handle_result(no_call, Req, State) ->
+    error_response(not_implemented, Req, State).
+
+serialize(Data, Serializer, Req, State) ->
+    {Serializer:serialize(Data), Req, State}.
+
+register_handler(Base, Command) ->
```
bartekgorny: Does this method really register a handler, or just returns a path? If the latter then this is not a proper name.

apps/ejabberd/src/mongoose_api_backend.erl
```diff
@@ -0,0 +1,234 @@
...
+
+get_allowed_methods() ->
+    Commands = ?COMMANDS_ENGINE:list(admin),
+    [action_to_method(?COMMANDS_ENGINE:action(Command)) || {_Name, Command} <- Commands].
+
+execute_command(Args, Command) ->
+    ?COMMANDS_ENGINE:execute(admin, ?COMMANDS_ENGINE:name(Command), Args).
+
+create_url_path(Command) ->
+    "/" ++ maybe_add_bindings(category_to_resource(?COMMANDS_ENGINE:category(Command)), Command).
```
bartekgorny: maybe_add_bindings returns category part plus some bindings if present, and only the latter is the "maybe" part. Rephrased like: "/" ++ category_to_resoure ++ maybe_bindings would make it much more readable.

apps/ejabberd/src/mongoose_api_backend.erl
```diff
@@ -0,0 +1,234 @@
...
+        update ->
+            add_bindings(Base, Args);
+        delete ->
+            add_bindings(Base, Args);
+        _ ->
+            Base
+    end.
+
+add_bindings(Base, Args) ->
+    Suffix = ["/:" ++ atom_to_list(ArgName)  || {ArgName, _} <- Args],
```
ludwikbukowski: I assumed that all methods returned by command engine are consistent with the interface defined in `mongoose_commands.hrl` and in my opinion the commands engine should verify it

bartekgorny: It does, but nothing prevents you from defining command args like: 
[{arg1, integer}, {arg2, binary}, [integer]]
or
[integer, binary, binary]
that these'd be your "Args" in call to add_bindings.

ludwikbukowski: ok, I will fix it

apps/ejabberd/src/mongoose_api_backend.erl
```diff
@@ -0,0 +1,234 @@
...
+
+extract_bindings(Bindings) ->
+    [Bind || {_BindingName, Bind} <- Bindings].
+
+category_to_resource(Category) when is_atom(Category) ->
+    atom_to_list(Category);
+category_to_resource(Category) when is_list(Category) ->
+    Category.
+
+check_and_extract_args(CommandsArgList, RequestArgList) ->
```
bartekgorny: You seem to be checking if arguments comply with command args definition - this is also done by command registry, so it is redundant here (I can export the check from command_registry module if you need to do it here). Your check also imposes a requirements on command args definition (see our discussion about add_bindings). An idea: if http frontend requires that command args comply to a particular pattern (a flat list of named args, for instance), it could check it at startup and simply do not expose those which do not comply.

ludwikbukowski: it might be a begining of discussion if the command engine should take as a arguments of `execute` function the tuples of `{arg_name, value}` because the parameters in POST method could be in different order, but still in form of json `{"key" : "value"}`. That's the reason I did it this way

bartekgorny: Yes, for MVC we should do sthg like that. And command engine should make sure argspec is a flat list of named args.

bartekgorny:
Another design choice: if the func returns a record, then we have to turn it into a proplist, otherwise mochijson crashes. The question is, who should take care of it, engine or http frontend?

ludwikbukowski:
with @michalwski we decided to use [jiffy](https://github.com/davisp/jiffy) instead of mochijson so it's yet another reason to do it

apps/ejabberd/src/mongoose_api_backend.erl
```diff
@@ -78,64 +78,74 @@ rest_terminate(_Req, _State) ->
...
 %%--------------------------------------------------------------------
 %% internal funs
 %%--------------------------------------------------------------------
 
-handle_request(Serializer, Req, #backend_state{bindings=Bindings, command_category = Name}=State) ->
+handle_get(Serializer, Req, #backend_state{bindings=Bindings, command_category = Name}=State) ->
     %% should we introduce filtering not only by category?
     Commands = ?COMMANDS_ENGINE:list(admin, Name),
-    [Command] = [C || C <- Commands, ?COMMANDS_ENGINE:action(C) =:= read],
+    [Command] = [C || C <- Commands, (?COMMANDS_ENGINE:action(C) =:= read) ],
```
Dzol: The `handle_{get,delete,post}` functions have some things in common: e.g. the line above. Perhaps we can refactor some of this.

ludwikbukowski: There is a plan to make `mongoose_commands` module implement filtering by `category` and `action` and then those repeated lines will be compressed to something one line

bartekgorny: It is already done, there is a list/3 function.

PR #883, opened 2016-07-13
HTTP API backend

ludwikbukowski:
The Work is still in progress
Proposed changes include:
* simple backend engine for `GET` and `POST` `PUT` and `DELETE` method
* integration with `Commands Engine` (introduced in `command_registry` branch
* tests for `GET`, `POST`, `DELETE` and `PUT`

--- 
TODO:
- [x] use `jiffy` instead of` mochijson2`
- [x] improve the tests (more groups, more scenarios etc)
- [x] discuss: the compability between command engine and http connector (especially how arguments maps to query path and body)
- [x] discuss: should `POST` method return the header `location` with the new resource created? And how could it be obtained from commands engine?
- [x] discuss: which http status codes and when?
- [x] return the error message with the error code
- [x] implement `PUT` method after the changes in command engine
- [x] return the path to the created resource in POST `location` header
- [x] create a hook when registering new command (in case to dynamically add HTTP handler)
- [x] add some more @docs to `mongoose_api_backend`
- [x] split the implementation to "backend" and "frontend" side. The client side will require authorization
- [x] implement all simple tests for auth side
- [ ] implement all ct tests for auth side
- [ ] make great docs about the HTTP API

### DIFF
--- a/apps/ejabberd/include/mongoose_api.hrl
+++ b/apps/ejabberd/include/mongoose_api.hrl
@@ -1,0 +1,6 @@
+%% backend state for HTTP API
+-record(backend_state, {allowed_methods, bindings, parameters, command_category, entity = admin, opts = []}).
+%% Error messages
+-define(ARGS_LEN_ERROR, "Bad parameters length.").
+-define(ARGS_SPEC_ERROR, "Bad name of the parameter.").
+-define(BODY_MALFORMED, "The request body is malformed.").
--- a/apps/ejabberd/include/mongoose_api.hrl
+++ b/apps/ejabberd/include/mongoose_api.hrl
@@ -2,9 +2,9 @@
 %% backend state for HTTP API
 -record(http_api_state, {allowed_methods, bindings, parameters, command_category, entity = admin, opts = []}).
 %% Error messages
--define(ARGS_LEN_ERROR, "Bad parameters length.").
--define(ARGS_SPEC_ERROR, "Bad name of the parameter.").
--define(BODY_MALFORMED, "The request body is malformed.").
+-define(ARGS_LEN_ERROR, <<"Bad parameters length.">>).
+-define(ARGS_SPEC_ERROR, <<"Bad name of the parameter.">>).
+-define(BODY_MALFORMED, <<"The request body is malformed.">>).
 -define(COMMANDS_ENGINE, mongoose_commands).
 
 %% Types
--- a/apps/ejabberd/include/mongoose_api.hrl
+++ b/apps/ejabberd/include/mongoose_api.hrl
@@ -1,6 +1,17 @@
+-include("mongoose_commands.hrl").
 %% backend state for HTTP API
 -record(backend_state, {allowed_methods, bindings, parameters, command_category, entity = admin, opts = []}).
 %% Error messages
 -define(ARGS_LEN_ERROR, "Bad parameters length.").
 -define(ARGS_SPEC_ERROR, "Bad name of the parameter.").
 -define(BODY_MALFORMED, "The request body is malformed.").
+-define(COMMANDS_ENGINE, mongoose_commands).
+
+%% Types
+-type method() ::binary().
+-type arg_name() :: atom().
+-type arg_value() :: any().
+-type arg_spec_list() :: list(argspec()).
+-type args_applied() :: list({arg_name(), arg_value()}).
+-type arg_values() :: list(arg_value()).
+-type mongoose_command() :: #mongoose_command{}.
--- a/apps/ejabberd/include/mongoose_api.hrl
+++ b/apps/ejabberd/include/mongoose_api.hrl
@@ -1,6 +1,6 @@
 -include("mongoose_commands.hrl").
 %% backend state for HTTP API
--record(backend_state, {allowed_methods, bindings, parameters, command_category, entity = admin, opts = []}).
+-record(http_api_state, {allowed_methods, bindings, parameters, command_category, entity = admin, opts = []}).
 %% Error messages
 -define(ARGS_LEN_ERROR, "Bad parameters length.").
 -define(ARGS_SPEC_ERROR, "Bad name of the parameter.").
@@ -10,7 +10,7 @@
 %% Types
 -type method() ::binary().
 -type arg_name() :: atom().
--type arg_value() :: any().
+-type arg_value() :: binary() | string() | integer() | float().
 -type arg_spec_list() :: list(argspec()).
 -type args_applied() :: list({arg_name(), arg_value()}).
 -type arg_values() :: list(arg_value()).
--- a/apps/ejabberd/include/mongoose_api.hrl
+++ b/apps/ejabberd/include/mongoose_api.hrl
@@ -1,4 +1,3 @@
--include("mongoose_commands.hrl").
 %% backend state for HTTP API
 -record(http_api_state, {allowed_methods, bindings, parameters, command_category, entity = admin, opts = []}).
 %% Error messages
@@ -11,7 +10,6 @@
 -type method() ::binary().
 -type arg_name() :: atom().
 -type arg_value() :: binary() | string() | integer() | float().
--type arg_spec_list() :: list(argspec()).
+-type arg_spec_list() :: list(mongoose_commands:argspec()).
 -type args_applied() :: list({arg_name(), arg_value()}).
 -type arg_values() :: list(arg_value()).
--type mongoose_command() :: #mongoose_command{}.
--- a/apps/ejabberd/src/ejabberd_cowboy.erl
+++ b/apps/ejabberd/src/ejabberd_cowboy.erl
@@ -70,22 +70,28 @@ reload_dispatch(Ref) ->
 -spec start_link(_) -> 'ignore' | {'error',_} | {'ok',pid()}.
 start_link(State) ->
     gen_server:start_link(?MODULE, State, []).
+
 init(State) ->
     process_flag(trap_exit, true),
     {ok, State}.
+
 handle_call(_Request, _From, State) ->
     {noreply, State}.
+
 handle_cast(reload_dispatch, #cowboy_state{ref = Ref, opts = Opts} = State) ->
     reload_dispatch(Ref, Opts),
     {noreply, State};
 handle_cast(_Request, State) ->
     {noreply, State}.
+
 handle_info(_Info, State) ->
     {noreply, State}.
+
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
+
 terminate(_Reason, State) ->
-    stop_cowboy(State).
+    stop_cowboy(State#cowboy_state.ref).
 
 -spec handler({integer(), inet:ip_address(), tcp}) -> list().
 handler({Port, IP, tcp}) ->
@@ -144,8 +150,7 @@ reload_dispatch(Ref, Opts) ->
     cowboy:set_env(Ref, dispatch, Dispatch).
 
 stop_cowboy(Ref) ->
-    cowboy:stop_listener(cowboy_ref(Ref)),
-    ok.
+    cowboy:stop_listener(Ref).
 
 
 cowboy_ref(Ref) ->
--- a/apps/ejabberd/src/ejabberd_cowboy.erl
+++ b/apps/ejabberd/src/ejabberd_cowboy.erl
@@ -37,7 +37,7 @@
          stop/1]).
 
 %% helper for internal use
--export([handler/1]).
+-export([handler/1, reload_dispatch/1]).
 
 -include("ejabberd.hrl").
 -type options()  :: [any()].
@@ -47,6 +47,7 @@
 -type default_result() :: list({{path(), module(), options()}}).
 -type implemented_result() :: list({paths(), handler_module(), options()}).
 -callback cowboy_router_paths(path(), options()) -> implemented_result() | default_result().
+-record(cowboy_state, {ref, opts = []}).
 %%--------------------------------------------------------------------
 %% ejabberd_listener API
 %%--------------------------------------------------------------------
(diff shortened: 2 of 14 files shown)


apps/ejabberd/include/mongoose_api.hrl
```diff
@@ -0,0 +1,17 @@
+-include("mongoose_commands.hrl").
+%% backend state for HTTP API
+-record(http_api_state, {allowed_methods, bindings, parameters, command_category, entity = admin, opts = []}).
+%% Error messages
+-define(ARGS_LEN_ERROR, "Bad parameters length.").
```
michalwski: Binaries please :)

apps/ejabberd/include/mongoose_commands.hrl
```diff
@@ -0,0 +1,32 @@
+
+-type command_action() :: [create|read|update|delete]. %% just basic CRUD; sending a mesage is 'create'
```
michalwski: All the types should, IMO, go to the `mongoose_commands` module and be exported from it.

apps/ejabberd/include/mongoose_commands.hrl
```diff
@@ -0,0 +1,32 @@
...
+                  |[typedef()]. % a list, but one element
+
+-type security() :: [admin|user]. %% later acl option will be added
+
+-type errortype() :: denied|not_implemented|type_error|internal. %% we should agree on a set of atoms so that the
+                                                                 %% frontend can map it to http codes
+
+-type failure() :: {error, errortype(), binary()}.
+
+-record(mongoose_command, {
```
michalwski: The record is used only by `mongoose_commands` module and tests. With some effort we can move it to mongoose_commands and remove the .hrl file completely. I don't think there is a need to access this record from tests.

apps/ejabberd/include/mongoose_api.hrl
```diff
@@ -0,0 +1,17 @@
...
+-define(ARGS_SPEC_ERROR, "Bad name of the parameter.").
+-define(BODY_MALFORMED, "The request body is malformed.").
+-define(COMMANDS_ENGINE, mongoose_commands).
```
michalwski: Why do we need this? I think usuing `mongoose_commands` explicitly will be more readable.

apps/ejabberd/src/mongoose_api_client.erl
```diff
@@ -0,0 +1,145 @@
+%%%-------------------------------------------------------------------
+%%% @author ludwikbukowski
+%%% @copyright (C) 2016, <COMPANY>
```
michalwski: Please fill in <COMPANY> or remove the section completely as we can read this info from git.

apps/ejabberd/src/mongoose_api_common.erl
```diff
@@ -0,0 +1,324 @@
...
+%% Read more http://ninenines.eu/docs/en/cowboy/1.0/guide/routing/
+reload_dispatches(_Command) ->
+    Listeners = supervisor:which_children(ejabberd_listeners),
+    CowboyListeners = [Child || {_Id, Child, _Type, [ejabberd_cowboy]}  <- Listeners],
+    [ejabberd_cowboy:reload_dispatch(Child) || Child <- CowboyListeners].
+
+
+-spec create_admin_url_path(mongoose_command()) -> ejabberd_cowboy:path().
+create_admin_url_path(Command) ->
+    "/" ++ category_to_resource(?COMMANDS_ENGINE:category(Command))
```
michalwski: Can `io_list` be used here instead of string catenation?

apps/ejabberd/src/mongoose_api_common.erl
```diff
@@ -0,0 +1,324 @@
...
+    end;
+process_request(_Method, Command, Req, #http_api_state{bindings = Binds, entity = Entity}=State) ->
+    BindsReversed = lists:reverse(Binds) ++ maybe_add_caller(Entity),
+    handle_request(Command, BindsReversed, Req, State).
+
+
+-spec handle_request(mongoose_command(), args_applied(), term(), #http_api_state{}) ->
+    {any(), any(), #http_api_state{}}.
+handle_request(Command, Args, Req, #http_api_state{entity = Entity} = State) ->
+    Method = action_to_method(?COMMANDS_ENGINE:action(Command)),
```
michalwski: Is the method the same as in `process_request` or can it change?

apps/ejabberd/src/mongoose_api_common.erl
```diff
@@ -0,0 +1,324 @@
...
+    Header = {<<"location">>, Path},
+    {ok, Req2} = cowboy_req:reply(201, [Header], Req),
+    Req2.
+
+
+
+-spec convert_arg(atom(), any()) -> integer() | float() | binary() | string() | {error, bad_type}.
+convert_arg(binary, Binary) when is_binary(Binary) ->
+    Binary;
+convert_arg(string, Binary) when is_binary(Binary) ->
```
michalwski: Why do we need `string` type? Can `binary` hold sth else than `string` can in the API case?
If no, I think we should have only `binary` type (mainly for simplicity).

Please correct me if I'm missing sth.

apps/ejabberd/src/mongoose_commands.erl
```diff
@@ -0,0 +1,567 @@
...
+%% Arg specification is used at call-time for control, and also for introspection
+%% (see list/1, list/2, mongoose_commands:get_command/2 and args/1)
+%%
+%% Return value is also specified, and this is a bit tricky: command definition
+%% contains spec of return value - what the target func returns should comply to it.
+%% The registry, namely execute/3, returns a tuple {ok, ValueReturnedByTheFunction}
+%% If return value is defined as 'ok' then whatever target func returns is ignored.
+%% This is mostly to make a distinction between 'create' actions which actually create something
+%% and return its identifier and those 'lame creators' which cause some action to be done and
+%% something written to dbase (exemplum: sending a message), but there is no accessible resource.
```
michalwski: I think we can remove this "lame creators". F.e if we assume, that the server will generate an `id` for message (and we, IMO, want to have ids for messages), the `id` may be returned.

apps/ejabberd/src/mongoose_commands.erl
```diff
@@ -0,0 +1,567 @@
...
+%%      result :: argspec()
+%% Type spec of return value of the function to call; execute/3 eventually returns {ok, result}
+%%      identifiers :: [atom()] (optional, required in 'update' commands)
+%%
+%% You can ignore return value of the target func by specifying return value as {result, ok}. The
+%% execute/3 will then always return just 'ok' (or error).
+%%
+%% If action is 'update' then it MUST specify which args are to be used as identifiers of object to update.
+%% It has no effect on how the engine does its job, but may be used by client code to enforce proper
+%% structure of request. (this is bad programming practice but we didn't have a better idea, we had to
```
michalwski: How about requiring a named arg `id` in case of such an action?

apps/ejabberd/src/mongoose_commands.erl
```diff
@@ -0,0 +1,567 @@
...
+            ok;
+        false ->
+            throw(permission_denied)
+    end,
+    % check caller (if it is given in args, and the engine is called by a 'real' user, then it must match
+    check_caller(Caller, Command, Args),
+    % check args
+    SpecLen = length(Command#mongoose_command.args),
+    ALen = length(Args),
+    if SpecLen =/= ALen ->
```
michalwski: This pretty `ifs` in Erlang :) I think this whole logic will be more readable if moved to a function (or `case` at least).

apps/ejabberd/src/mongoose_commands.erl
```diff
@@ -0,0 +1,567 @@
...
+    false;
+apply_policies([P|Policies], Jid) ->
+    case apply_policy(P, Jid) of
+        true ->
+            true;
+        false ->
+            apply_policies(Policies, Jid)
+    end.
+
+%% @doc This is the only policy we know so far, but there will be others (like roles/acl control)
```
michalwski: Can we add this when we need this? Let's follow `KISS` and `YAGNI` :)

apps/ejabberd/src/mongoose_commands.erl
```diff
@@ -0,0 +1,567 @@
...
+    CallerJid = jid:from_binary(Caller),
+    ACallerJid = jid:from_binary(ACaller),
+    ACal = {ACallerJid#jid.user, ACallerJid#jid.server},
+    case {CallerJid#jid.user, CallerJid#jid.server} of
+        ACal ->
+            ok;
+        _ ->
+            throw(caller_jid_mismatch)
+    end.
+
```
michalwski: I think we can save same lines by removing the unnecessary lines at the end of the file :P

apps/ejabberd/src/mongoose_api_common.erl
```diff
@@ -0,0 +1,325 @@
...
+process_request(<<"POST">>, Command, Req, #http_api_state{bindings = Binds, entity = Entity} = State) ->
+    BindsReversed = lists:reverse(Binds),
+    case parse_request_body(Req) of
+        {error, _R}->
+            error_response(bad_request, ?BODY_MALFORMED , Req, State);
+        {Params, Req2} ->
+            Params2 = BindsReversed ++ Params ++ maybe_add_caller(Entity),
+            handle_request(Command, Params2, Req2, State)
+    end;
+process_request(<<"PUT">>, Command, Req, #http_api_state{bindings = Binds, entity = Entity} = State) ->
```
ludwikbukowski: since the branch for `put` method looks actually quite the same as for `post` (we decided that we allow "identifiers" also for `post` methods), those two branches can be merged

apps/ejabberd/include/mongoose_api.hrl
ludwikbukowski: you're right, it doesn't carry a value

apps/ejabberd/src/mongoose_api_common.erl
ludwikbukowski: yes, we don't need this branch

michalwski:
I've create a separate PR only for `commands registry` #892 . These 2 PR are connected, but the functionality they implement can be separated into 2 PR. Thanks to that we'll have smaller, and better split PRs.

apps/ejabberd/src/ejabberd_cowboy.erl
```diff
@@ -50,29 +57,35 @@ socket_type() ->
...
+    reload_dispatch(Ref, Opts),
+    {noreply, State};
+handle_cast(_Request, State) ->
+    {noreply, State}.
+handle_info(_Info, State) ->
+    {noreply, State}.
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+terminate(_Reason, State) ->
+    stop_cowboy(State).
```
michalwski: This ain't gonna work :) `stop_cowboy` takes the cowboy's reference and not process state... We should start using dialyzer better...

apps/ejabberd/test/commands_backend_SUITE.erl
```diff
@@ -0,0 +1,681 @@
...
+    fusco:disconnect(Pid),
+    teardown(),
+    R;
+request(Path, Method, BodyData, {{_User, _Pass} = Auth, Authorized}) ->
+    setup(client_module()),
+    meck:expect(ejabberd_auth, check_password, fun(_, _, _) -> Authorized end),
+    Body = maybe_add_body(BodyData),
+    AuthHeader = maybe_add_auth_header(Auth),
+    AcceptHeader = maybe_add_accepted_headers(Method),
+    request(Path, Method, Body, {headers, AuthHeader ++ AcceptHeader});
```
michalwski: IMO, it would be much more readable if this function clause was named differently. 2 other clauses build `headers` and call the function again with only last arg changed.

apps/ejabberd/test/commands_backend_SUITE.erl
```diff
@@ -0,0 +1,681 @@
...
+request(Path, Method, BodyData, admin) ->
+    setup(backend_module()),
```
michalwski: Such a setup should be places in `init_per_testcase`, IMO

apps/ejabberd/test/commands_backend_SUITE.erl
```diff
@@ -0,0 +1,681 @@
...
+request(Path, "DELETE", Entity) ->
+    request(Path, "DELETE", [], Entity).
+
+-spec request(binary(), method(), list({atom(), any()}),
+              {headers, list()} | admin | {{binary(), binary()}, boolean()}) -> any.
+request(Path, Method, Body, {headers, Headers}) ->
+    {ok, Pid} = fusco:start_link("http://"++ ?HOST ++ ":" ++ integer_to_list(?PORT), []),
+    R = fusco:request(Pid, Path, Method, Headers, Body, 5000),
+    fusco:disconnect(Pid),
+    teardown(),
```
michalwski: end_per_testcase is a good candidate for the teardown

apps/ejabberd/src/mongoose_api_common.erl
```diff
@@ -0,0 +1,352 @@
...
+add_bind({caller, _}, user) ->
+    "";
+add_bind({ArgName, _}, _Entity) ->
+    ["/", atom_to_list(ArgName), "/:", atom_to_list(ArgName)];
+add_bind(Other, _) ->
+    throw({error, bad_arg_spec, Other}).
+
+-spec to_atom(binary() | list() | atom()) -> atom().
+to_atom(Bin) when is_binary(Bin) ->
+    list_to_atom(binary_to_list(Bin));
```
michalwski: Do we really want to convert binary or list to atom. Can we assume these atoms exist and we can use `*_to_existing_atom`?

ludwikbukowski: You're right, such a function can be used but requires additional catch branch here [1].
It might happen that client sends request and the body contains strange/random field name so that `badarg` in function `*_to_existing_atom` would be thrown

[1] https://github.com/esl/MongooseIM/blob/http-api-backend/apps/ejabberd/src/mongoose_api_common.erl#L165

michalwski: Ok, got it. Other question, does it have to atom? Can we use binaries. If I understand it correctly this function just coverts param name from binary to atom.

ludwikbukowski: The conversion is done because later those atoms from request body are compared to `args` names [1] to check if the client request matches to existing command, and the `args` consist of tuples where the first field is just an atom.

[1] https://github.com/esl/MongooseIM/blob/command_registry/apps/ejabberd/src/mongoose_commands.erl#L116

michalwski: I see, it may be simpler to have the field as binary and not atom.

ludwikbukowski: yes, atoms are expensige

michalwski: Depends how you use them :) If you don't have to convert from binary (list) to atom, then they are far more efficient. But if you need to convert user's input (already in binary) to atom, that's different story.

ludwikbukowski: and imagine there is some nasty client who knows about that and sends thousands of requests with unique data and the new created atoms won't be garbage collected! 
Thanks for pointing it out

PR #894, opened 2016-07-28
MUC http auth

chrzaszcz:
* generic way of managing pools of outgoing http connections
* an option to enable checking passwords for password-protected rooms by an external HTTP service


apps/ejabberd/src/mod_http_client_worker.erl
```diff
@@ -0,0 +1,25 @@
...
+-module(mod_http_client_worker).
+
+%% API
+-export([start_link/1]).
+
+%%------------------------------------------------------------------------------
+%% API
+
+start_link([Host, Opts]) ->
+    fusco:start_link(Host, Opts).
```
michalwski: Can we use `fusco` directly and not through this short module? Or is there a plan to add more functionality to this module?

test/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -261,9 +275,48 @@ init_per_group(disco_rsm, Config) ->
     [Alice | _] = ?config(escalus_users, Config1),
     start_rsm_rooms(Config1, Alice, <<"aliceonchat">>);
 
+init_per_group(TC, Config) when TC =:= http_auth_no_server;
```
michalwski: The `TC` var name is misleading here. It should probably be `Group`.

apps/ejabberd/src/mongoose_http_client.erl
```diff
@@ -70,72 +79,63 @@ post(Pool, Path, Headers, Query) ->
...
-    ets:delete(tab_name(Host)).
-
 %%------------------------------------------------------------------------------
 %% internal functions
 
-make_pool(Host, Name, Opts) ->
+make_pool(Name, Opts) ->
     #pool{name = Name,
-          host = Host,
           http_host = gen_mod:get_opt(host, Opts, "http://localhost"),
```
michalwski: I guess the `http(s)://` prefix is required here. If so I'd rather call the option `URL` instead of host.  For me `host` refers only to the part of URL after the prefix and before first `/`.

PR #896, opened 2016-07-28
REST API MVP

michalwski:
This PR implements part of the MVP for REST API.


test/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -0,0 +1,225 @@
...
+    assert_inlist(#{a => 1}, Maplst),
+    assert_inlist(#{a => 1, b => 2}, Maplst),
+    assert_inlist(#{a => 4}, Maplst),
+    assert_notinlist(#{a => 5}, Maplst),
+    assert_notinlist(#{a => 1, b => 1}, Maplst).
+
+
+basic(_Config) ->
+    % list commands
+    {?OK, Lcmds} = gett(<<"/list">>),
```
michalwski: IMO, `list` is not a good name for a resource. I know its verb and noun but to me the first thought is verb and resources should be nouns. Better name for me here would be: `commands`.

test/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -0,0 +1,225 @@
...
+    DecCmds = decode_maplist(Lcmds),
+    assert_inlist(#{name => <<"listmethods">>}, DecCmds),
+    % nonexistent command
+    {?NOT_FOUND, _} = gett(<<"/isitthereornot">>),
+    % list users
+    {?OK, Lusers} = gett(<<"/user/host/localhost">>),
```
michalwski: Couple of suggestions and questions here:
1. Better name for the resource, IMO, is `users` (following the best practices for REST interface)
2. Why do we need `host` after `user` and before `localhost`?

Applying above 2 items, the final URL looks like this:
`/users/localhost`
It's shorter and gives the same information. What do you think?

test/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -0,0 +1,225 @@
...
+    assert_inlist(<<"alice@localhost">>, Lusers),
+    % create user
+    CrUser = #{user => <<"mike">>, password => <<"nicniema">>, host => <<"localhost">>},
+    {?CREATED, _} = post(<<"/user/host/localhost">>, CrUser),
+    {?OK, Lusers1} = gett(<<"/user/host/localhost">>),
+    assert_inlist(<<"mike@localhost">>, Lusers1),
+    % try to create the same user
+    {?ERROR, _} = post(<<"/user/host/localhost">>, CrUser),
+    % delete user
+    {?OK, _} = delete(<<"/user/jid/mike@localhost">>),
```
michalwski: How about: `users/mike@localhost` here?

test/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -0,0 +1,225 @@
...
+    {?OK, Lusers2} = gett(<<"/user/host/localhost">>),
+    assert_notinlist(<<"mike@localhost">>, Lusers2),
+    ok.
+
+
+sessions(Config) ->
+    % no session
+    {?OK, Sessions} = gett("/session/host/localhost"),
```
michalwski: `/sessions/localhost`, what do you think?

test/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -0,0 +1,225 @@
...
+    escalus:story(Config, [{alice, 1}, {bob, 1}], fun(Alice, Bob) ->
+        M = #{caller => <<"bob@localhost">>, to => <<"alice@localhost">>, msg => <<"hello from Bob">>},
+        {?OK, _} = post(<<"/message">>, M),
+        Res = escalus:wait_for_stanza(Alice),
+        escalus:assert(is_chat_message, [<<"hello from Bob">>], Res),
+        M1 = #{caller => <<"alice@localhost">>, to => <<"bob@localhost">>, msg => <<"hello from Alice">>},
+        {?OK, _} = post(<<"/message">>, M1),
+        Res1 = escalus:wait_for_stanza(Bob),
+        escalus:assert(is_chat_message, [<<"hello from Alice">>], Res1),
+        {?OK, Msgs} = gett("/message/caller/alice@localhost/other/bob@localhost/limit/10"),
```
michalwski: How about `/messages/alice@localhost/bob@localhost?limit=10`

apps/ejabberd/src/mongoose_admin.erl
```diff
@@ -0,0 +1,268 @@
...
+%%% mongoose commands
+%%%
+
+
+commands() ->
+    [
+        [
+            {name, listmethods},
+            {category, list},
+            {desc, "List commands"},
```
michalwski: To be consistent with the code base the description should be binary as we are no longer use old strings to represent text in the code. This comment applies to all commands in this module.

apps/ejabberd/src/mongoose_admin.erl
```diff
@@ -0,0 +1,264 @@
...
+    J = jid:from_binary(Jid),
+    ejabberd_auth:remove_user(J#jid.luser, J#jid.lserver),
+    <<"">>.
+
+
+send_message(From, To, Body) ->
+    Packet = build_packet(message_chat, Body),
+    F = jid:from_binary(From),
+    T = jid:from_binary(To),
+    ejabberd_hooks:run(user_send_packet,
```
michalwski: I'd rather use new hook for that "rest_user_send_packet" in order to have better visibility from where the message was send.

apps/ejabberd/src/mongoose_admin.erl
```diff
@@ -0,0 +1,264 @@
...
+    MsgId = case xml:get_tag_attr(<<"id">>, Msg) of
+                {value, MId} -> MId;
+                false -> <<"">>
+            end,
+    Body = exml_query:path(Msg, [{element, <<"body">>}, cdata]),
+    #{sender => Jbin, timestamp => round(Msec / 1000000), message_id => MsgId, body => Body}.
+
+build_packet(message_chat, Body) ->
+    #xmlel{ name = <<"message">>,
+        attrs = [{<<"type">>, <<"chat">>}, {<<"id">>, list_to_binary(randoms:get_string())}],
```
michalwski: +1 for the `id` but the generator should be improved. For message id, better generator would be UUID (like here: https://github.com/esl/MongooseIM/pull/881/files#diff-6ff196eabf2f475f96beab1e550960adR58)

test/ejabberd_tests/tests/rest_SUITE.erl
michalwski: This is even better: `/users/localhost/mike`

apps/ejabberd/src/mongoose_admin.erl
```diff
@@ -0,0 +1,264 @@
...
+    lists:map(fun mk/1, Lst).
+
+
+registered_users(Host) ->
+    Users = ejabberd_auth:get_vh_registered_users(Host),
+    SUsers = lists:sort(Users),
+    lists:map(fun({U, S}) -> <<U/binary, "@", S/binary>> end, SUsers).
+
+
+register(User, Host, Password) ->
```
michalwski: In order to be consistent with the code base I suggest changing the params order so `Host` is before `User` as in other places.

apps/ejabberd/src/mongoose_admin.erl
```diff
@@ -0,0 +1,264 @@
...
+    Jbin = jid:to_binary(From),
+    {Msec, _} = mod_mam_utils:decode_compact_uuid(Id),
+    MsgId = case xml:get_tag_attr(<<"id">>, Msg) of
+                {value, MId} -> MId;
+                false -> <<"">>
+            end,
+    Body = exml_query:path(Msg, [{element, <<"body">>}, cdata]),
+    #{sender => Jbin, timestamp => round(Msec / 1000000), message_id => MsgId, body => Body}.
+
+build_packet(message_chat, Body) ->
```
michalwski: What are the other packets which could be built by this function? If we don't have them now, let's remove the first parameter and make the function simpler.

PR #900, opened 2016-08-01
Add unavailable presence after blocking contact

bernardd:
XEP-0016 section 2.11 contains the following requirement:

> When a user blocks outbound presence to a contact, the user's server MUST send unavailable presence information to the contact (but only if the contact is allowed to receive presence notifications from the user in accordance with the rules defined in RFC 6121).

However for this case no unavailable message was being sent. This patch adds this behaviour, sending an "unavailable" from A to B when all the following conditions are met:
* B is subscribed to receive presence information from A (and is therefore a contact)
* A has changed their active privacy list or modified their currently active list
* The change means that where presence notifications were previously sent from A to B, they are now blocked

This covers the unambiguous case described in the XEP. It's not clear to me whether an 'unavailable' is also required to be sent when some second-order condition changes the blocking state. For example, if A has an existing privacy list that blocks all presence information to group X, and A adds B to group X, should that trigger an unavailable message? XE-0016 isn't really explcit on the matter. For now I've left those other cases out.

### DIFF
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -1317,6 +1317,7 @@ handle_routed_broadcast({privacy_list, PrivList, PrivListName}, StateData) ->
             F = jid:to_bare(StateData#state.jid),
             T = StateData#state.jid,
             PrivPushEl = jlib:replace_from_to(F, T, jlib:iq_to_xml(PrivPushIQ)),
+            maybe_update_presence(StateData, NewPL),
             {send_new, F, T, PrivPushEl, StateData#state{privacy_list = NewPL}}
     end;
 handle_routed_broadcast({blocking, Action, JIDs}, StateData) ->
@@ -2184,6 +2185,7 @@ process_privacy_iq(From, To,
                    {error, ?ERR_FEATURE_NOT_IMPLEMENTED},
                    [From, To, IQ]) of
                 {result, R, NewPrivList} ->
+                    maybe_update_presence(StateData, NewPrivList),
                     {{result, R},
                      StateData#state{privacy_list = NewPrivList}};
                 R -> {R, StateData}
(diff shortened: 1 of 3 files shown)


apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -2412,6 +2414,38 @@ flush_messages(N, Acc) ->
               {N, Acc}
     end.
 
+
+%%%----------------------------------------------------------------------
+%%% XEP-0016
+%%%----------------------------------------------------------------------
+
+maybe_update_presence(StateData = #state{jid = JID}, NewList) ->
+    {Froms, _Tos, _Pending} = ejabberd_hooks:run_fold(
```
michalwski: This hook is not necessary here, I think. If I understand it correctly, all you need (subscription lists) are already in the `StateData` var. Take a look at fields `pres_*`, they are documented quite well in the [ejabberd_c2s.hrl](https://github.com/hippware/MongooseIM/blob/e5329f0d58963ab7875798ea01845aa6c016f99b/apps/ejabberd/src/ejabberd_c2s.hrl#L40-L55) file.

michalwski: Moreover running the `roster_get_subscription_lists` hook will touch DB every time which is not the most efficient way :)

bernardd: Ah, thanks. I hadn't noticed that they were already in the state. I've pushed a fix, as well as tweaking some of the test code a little to avoid a couple of intermittent failures I was getting.

michalwski: This was quick :) thanks @bernardd

test/ejabberd_tests/tests/privacy_SUITE.erl
```diff
@@ -711,6 +711,7 @@ block_jid_message_but_not_presence(Config) ->
         end).
 
 newly_blocked_presense_jid_by_new_list(Config) ->
+    reset_users(Config),
```
michalwski: Instead of reseting users you can use `escalus_fresh:story/3` instead of `escalus:story/3`. This will create new users for you based on the provided specs. Please remember to call `escalus_fresh:clean/0` in `end_per_suite` to remove this `fresh` users.

bernardd: The only catch with that is that it creates new usernames as well, so the ones hardcoded into the privacy_helper lists don't work. Still, it feels like a neater solution, so I'll sort it out.

bernardd: Okay, fixed that and did a couple more little cleanups too.

michalwski:
Thanks a lot @bernardd for your patch! I like your code and the tests you've written for the functionality!

PR #903, opened 2016-08-02
MUC HTTP API

Dzol:
Introduce MVP for the MUC HTTP API:

- [x] Create room
- [x] Invite to room
- [x] Send message to room
- [ ] Refactor & improve source (following a review)
- [x] Improve & expand tests


apps/ejabberd/src/muc_admin.erl
```diff
@@ -0,0 +1,100 @@
...
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% Author: Joseph Yiasemides <joseph.yiasemides@erlang-solutions.com>
+%% Description: Administration commands for Mult-user Chat (MUC)
+%%==============================================================================
+
+-module(muc_admin).
```
michalwski: All modules implementing `gen_mod` behaviour are named with `mod_` prefix. This one should also follow this convention.

apps/ejabberd/src/muc_admin.erl
```diff
@@ -0,0 +1,100 @@
...
+start(_, _) ->
+    mongoose_commands:register(commands()).
+
+stop(_) ->
+    mongoose_commands:unregister(commands()).
+
+commands() ->
+    [
+     [{name, create_muc_room},
+      {category, muc},
```
michalwski: The category is the later translated to the resource in REST path. The good practise is to use plural nouns for that. I think category name `mucs` will suite better. But I'm open to here a constructive disagreement :)

apps/ejabberd/src/muc_admin.erl
```diff
@@ -0,0 +1,100 @@
...
+      {desc, "Create a MUC room."},
+      {module, ?MODULE},
+      {function, create_room},
+      {action, create},
+      {identifiers, [domain]},
```
michalwski: I think better name would be `host` instead of `domain`.

apps/ejabberd/src/muc_admin.erl
```diff
@@ -0,0 +1,100 @@
...
+      {args,
+       %% The argument `domain' is what we normally term the XMPP
+       %% host, `name' is the room name, `owner' is the XMPP entity
+       %% that would normally request an instant MUC room.
+       [{domain, binary},
+        {name, binary},
+        {owner, binary},
```
michalwski: Can we somehow sidestep the owner parameter? By f.e reading the owner?

michalwski: Sorry, I forgot it's create method. Forget my previous comment.

Dzol:
@michalwski: Commit a233e60 should resolve the comments above.

PR #906, opened 2016-08-03
REST shorter prettier

michalwski:
This PR focuses only on the REST API for backend services. The main change is the URL generator which now creates shorter and nicer URLs.

### DIFF
--- a/apps/ejabberd/src/mongoose_admin.erl
+++ b/apps/ejabberd/src/mongoose_admin.erl
@@ -9,7 +9,7 @@
          registered_users/1,
          change_user_password/2,
          listsessions/1,
-         kick_this_session/1,
+         kick_session/3,
          get_recent_messages/3,
          send_message/3
         ]).
@@ -84,7 +84,7 @@ commands() ->
 %%        ],
         [
             {name, listsessions},
-            {category, session},
+            {category, sessions},
             {desc, "Get session list"},
             {module, ?MODULE},
             {function, listsessions},
@@ -94,12 +94,12 @@ commands() ->
         ],
         [
             {name, kickuser},
-            {category, session},
+            {category, sessions},
             {desc, "Terminate user connection"},
             {module, ?MODULE},
-            {function, kick_this_session},
+            {function, kick_session},
             {action, delete},
-            {args, [{jid, binary}]},
+            {args, [{host, binary}, {user, binary}, {res, binary}]},
             {result, {msg, binary}}
         ],
         [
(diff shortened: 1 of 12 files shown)


apps/ejabberd/src/mongoose_admin.erl
```diff
@@ -181,9 +166,8 @@ register(User, Host, Password) ->
             throw({error, String})
     end.
 
-unregister(Jid) ->
-    J = jid:from_binary(Jid),
-    ejabberd_auth:remove_user(J#jid.luser, J#jid.lserver),
+unregister(Host, User) ->
+    ejabberd_auth:remove_user(User, Host),
     <<"">>.
```
Dzol: Perhaps we could return something more meaningful than `<<"">>` (even it doesn't get into an HTTP response).

michalwski: Yes, I think simple `ok` would be enough here.

apps/ejabberd/src/mongoose_commands.erl
```diff
@@ -108,14 +108,15 @@
...
           module :: module(),                         %% module to call
           function :: atom(),                         %% function to call
-          action :: action(),                 %% so that the HTTP side can decide which verb to require
+          action :: action(),                         %% so that the HTTP side can decide which verb to require
           args = [] :: [argspec()],                   %% this is both for introspection and type check on call
           caller_pos :: integer(),                    %% internal use
-          identifiers = [] :: [atom()],               %% if action is 'update' then it must be a subset of args
+          identifiers = [] :: [atom()],               %% resource identifiers, a subset of args
           security_policy = [admin] :: security(),    %% permissions required to run this command
           result :: argspec()|ok                      %% what the called func should return; if ok then return of called
```
Dzol: We could change returning the atom `ok` to returning `ignore`.

michalwski: Why would it be better, could you explain?

Dzol: A spec like `{..., {result, ignore}]}` communicates that the return value is actually ignored whereas currently that's not so clear with `{..., {result, ok}]}`.

michalwski: In fact I personally think that all the methods should always return `ok` or `{ok, STH_MORE}`. And the decisions whether or not ignore this should be left to the caller. But that's I think a candidate for separate PR.

Dzol: That sounds better to me too.

PR #907, opened 2016-08-03
Customisable MUC Light room config schemas

fenek:
This PR allows to set custom room config schema in `mod_muc_light` options. In extreme case even all parameters may be replaced (i.e. default `roomname` and `subject`).

I will also make possible to set custom default values for these fields.

**TODOs:**
- [x] Handle a case with missing `roomname` in schema.
- [x] Setting default values for custom schema.
- [x] Tests.
- [x] Docs.

### DIFF
--- a/test/ejabberd_tests/tests/muc_light_SUITE.erl
+++ b/test/ejabberd_tests/tests/muc_light_SUITE.erl
@@ -5,6 +5,9 @@
 -include_lib("common_test/include/ct.hrl").
 -include_lib("exml/include/exml.hrl").
 
+-export([ % service
+         mismatched_default_config_is_rejected/1
+        ]).
 -export([ % entity
          disco_service/1,
          disco_features/1,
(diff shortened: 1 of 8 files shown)


test/ejabberd_tests/tests/muc_light_SUITE.erl
```diff
@@ -193,23 +207,48 @@ init_per_group(_GroupName, Config) ->
...
+       orelse RoomLessCase =:=  mismatched_default_config_is_rejected ->
+    set_default_mod_config(),
+    escalus:init_per_testcase(RoomLessCase, Config);
+init_per_testcase(custom_schema_works_with_standard_default_config = CaseName, Config) ->
+    set_default_mod_config(),
+    set_custom_config(["roomname", "subject", "background", "music"], standard_default_config()),
+    create_room(?ROOM, ?MUCHOST, alice, [bob, kate], Config, ver(1)),
+    escalus:init_per_testcase(CaseName, Config);
+init_per_testcase(CustomConfigCase, Config)
+  when        CustomConfigCase =:= set_config_with_custom_schema
```
michalwski: How about `lists:member/2` instead of guard with so many `orelse`s?

PR #908, opened 2016-08-04
Privacy unavailable notification 2

bernardd:
Edit: Just rebased this off master following the merge of #900 - thanks :)

XEP-0016 2.11 says that contacts who are newly blocked from receiving
presence information should receive an 'unavailable' presence message at
the time of blocking. The XEP is not explict about this including
cases where privacy lists have not changed, but a user's subscription or
group membership have, making them newly subject to blocking. However it
seems like a reasonable case to cover, given that presumably the idea
behind the requirement is that a user should not be left thinking a user
who has blocked them is indefinitely available.

Also I just noticed I misspelled "unavailable" in the branch names. Whoops.


michalwski:
I'm closing the PR. The source branch was removed and it's 2 years old.

PR #910, opened 2016-08-05
Log cleanup duration on start

arkdro:
This PR addresses missing logging for one of cleanup calls.

Proposed changes include:
* log duration of a cleanup on start

The duration of the cleanup is logged during the normal operations when the other node
goes down in `mongoose_cleaner:run_node_cleanup/1`. Let's log it on start as well.

### DIFF
--- a/apps/ejabberd/src/ejabberd_sm_redis.erl
+++ b/apps/ejabberd/src/ejabberd_sm_redis.erl
@@ -26,7 +26,10 @@
 start(Opts) ->
     ejabberd_redis:start_link(Opts),
     %% Clean current node's sessions from previous life
-    cleanup(node()).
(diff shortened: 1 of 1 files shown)


apps/ejabberd/src/ejabberd_sm_redis.erl
```diff
@@ -26,7 +26,10 @@
 start(Opts) ->
     ejabberd_redis:start_link(Opts),
     %% Clean current node's sessions from previous life
-    cleanup(node()).
+    {Elapsed, RetVal} = timer:tc(fun() -> cleanup(node()) end),
```
michalwski:
```
timer:tc(?MODULE, cleanup, [node()])
```

would be better here.

PR #918, opened 2016-08-16
Display test progress

arcusfelis:
Proposed changes include:
* Print + when test passes
* Print - when test fails
* Force sample_external_auth.py to use python2 (on my machine it tries to use python3 and fails)


apps/ejabberd/src/ct_progress_hook.erl
```diff
@@ -0,0 +1,67 @@
+%%% @doc Common Test Example Common Test Hook module.
+-module(ct_progress_hook).
```
michalwski: Can we have this module in `apps/ejabberd/test` instead of `src`? It better suits there.

arcusfelis: We can't. rebar is too stupid to compile modules in test.

arcusfelis:
It looks like this https://travis-ci.org/esl/MongooseIM/jobs/152645125#L406

michalwski:
Is it possible to silence log messages like the one here: https://travis-ci.org/esl/MongooseIM/jobs/152645125#L488 ?

PR #919, opened 2016-08-16
Cassandra config - part 1

arcusfelis:
This PR addresses #719 

Proposed changes include:
* Removed old cassandra modules
* Cassandra pool configuration
* Configuration documentation https://github.com/arcusfelis/MongooseIM/blob/c106354035bd474bb4901720c87bd1529ff6b413/doc/advanced-configuration/Cassandra.md


.travis.yml
```diff
@@ -48,6 +48,7 @@ branches:
 otp_release:
         - 18.3
 env:
+        - PRESET=cassandra_mnesia DB=cassandra REL_CONFIG=with-cassandra CASSANDRA_VERSION=3.1
```
michalwski: Could you move the line just after or before the `riak` job spec?

apps/ejabberd/src/mongoose_cassandra.erl
```diff
@@ -0,0 +1,121 @@
...
+    [{servers, [{"localhost", 9042, 1}]}];
+expand_servers(undefined, Servers, PoolConfig) ->
+    PoolConfig;
+expand_servers(Address, undefined, PoolConfig) ->
+    DefaultPort = proplists:get_value(port, PoolConfig, 9042),
+    PoolSize    = proplists:get_value(pool_size, PoolConfig, 5),
+    Servers = expand_address(Address, DefaultPort, PoolSize),
+    [{servers, Servers} | PoolConfig];
+expand_servers(_Address, _Servers, PoolConfig) ->
+    ?ERROR_MSG("issue=config_error, details=\"Both servers and address options "
```
michalwski: What is the reason behind `servers` and `address` options? It looks we could use only one, and make the code simpler.

apps/ejabberd/src/mongoose_cassandra_worker.erl
```diff
@@ -0,0 +1,494 @@
...
+
+%% ----------------------------------------------------------------------
+%% Imports
+
+-include_lib("ejabberd/include/ejabberd.hrl").
+-include_lib("ejabberd/include/jlib.hrl").
+-include_lib("exml/include/exml.hrl").
+
+-record(state, {
+    pool_name,
```
michalwski: Indentation could be improved here :)

apps/ejabberd/src/mongoose_cassandra_worker.erl
```diff
@@ -0,0 +1,494 @@
...
+    conn,
+    prepared_queries,
+    query_refs,
+    query_refs_count}).
+
+-record(prepared_query, {
```
michalwski: And here as well :)

apps/ejabberd/src/mongoose_cassandra_worker.erl
```diff
@@ -0,0 +1,494 @@
...
+%%====================================================================
+%% Internal functions
+%%====================================================================
+
+start_link(PoolName, Addr, Port, ClientOptions) ->
+    gen_server:start_link(?MODULE, [PoolName, Addr, Port, ClientOptions], []).
+
+%% @doc Do CQL query
+%%
+%% UserJID is used for rate-limiting, statistics and debugging
```
michalwski: Is it? I don't see you use `UserJID` at all in this function :) I suggest to remove the parameter for now and add it later when we need it.

michalwski: Ah, sorry. The function is longer then I noticed at first glance. It is used. No comment then :)

michalwski: Actually the first comment stays :) I was wrong about being wrong. The `UserJID` is used but in a different function. Let's move the comment when it belongs.

apps/ejabberd/src/mongoose_cassandra_worker.erl
```diff
@@ -0,0 +1,494 @@
...
+    end.
+
+cql_query_async(Worker, _UserJID, Module, QueryName, Params) when is_pid(Worker) ->
+    gen_server:cast(Worker, {async_cql_query, Module, QueryName, Params}).
+
+cql_query_multi_async(Worker, _UserJID, Module, QueryName, MultiParams)
+  when is_pid(Worker) ->
+    gen_server:cast(Worker, {multi_async_cql_query, Module, QueryName, MultiParams}).
+
+cql_batch(Worker, UserJID, Module, Queries, not_batch) ->
```
michalwski: Do we really need this? The function name doesn't play well with the last parameter which is `not_batch` here.
Let's remove this function clause or rename the function to `maybe_cql_batch` (or some other name).

apps/ejabberd/src/mongoose_cassandra_worker.erl
```diff
@@ -0,0 +1,494 @@
...
+    seestar_session:batch_async(Conn, BatchType, BatchQueries, one).
+
+save_query_ref(From, QueryRef,
+               State=#state{query_refs=Refs, query_refs_count=RefsCount}) ->
+    Refs2 = dict:store(QueryRef, From, Refs),
+    put(query_refs_count, RefsCount+1),
+    State#state{query_refs=Refs2, query_refs_count=RefsCount+1}.
+
+forward_query_respond(ResultF, QueryRef,
+    State=#state{query_refs=Refs, query_refs_count=RefsCount}) ->
```
michalwski: Let's put a space before and after the `=`

apps/ejabberd/src/mongoose_cassandra_worker.erl
```diff
@@ -0,0 +1,494 @@
...
+%%                                      {noreply, State} |
+%%                                      {noreply, State, Timeout} |
+%%                                      {stop, Reason, Reply, State} |
+%%                                      {stop, Reason, State}
+%% Description: Handling call messages
+%%--------------------------------------------------------------------
+handle_call(_, _From, State=#state{conn=undefined}) ->
+    {reply, no_connection, State};
+handle_call({cql_query, Module, QueryName, Params}, From,
+    State=#state{conn=Conn}) ->
```
michalwski: Again, space before and after `=` will help :)
Yes, I act as Elvis now! :)

apps/ejabberd/src/mongoose_cassandra_worker.erl
```diff
@@ -0,0 +1,494 @@
...
+    case get_prepared_query(Conn, Module, QueryName, State) of
+        {ok, PreparedQuery, State2} ->
+            QueryRef = execute_prepared_query(Conn, PreparedQuery, Params),
+            {noreply, save_query_ref(From, QueryRef, State2)};
+        {error, Reason} ->
+            {reply, {error, Reason}, State}
+    end;
+handle_call({cql_batch, Module, BatchType, Queries}, From,
+    State=#state{conn=Conn}) ->
```
michalwski: spaces again :)

apps/ejabberd/src/mongoose_cassandra_worker.erl
```diff
@@ -0,0 +1,494 @@
...
+%% Function: handle_cast(Msg, State) -> {noreply, State} |
+%%                                      {noreply, State, Timeout} |
+%%                                      {stop, Reason, State}
+%% Description: Handling cast messages
+%%--------------------------------------------------------------------
+
+handle_cast(_, State=#state{conn=undefined}) ->
+    {noreply, State};
+handle_cast({async_cql_query, Module, QueryName, Params},
+    State=#state{conn=Conn}) ->
```
michalwski: Ok, I'm not Elvis any more, but please check your code and add spaces around `=` if they are not there.

apps/ejabberd/src/mongoose_cassandra_worker.erl
```diff
@@ -0,0 +1,494 @@
...
+    {ok, State}.
+
+%% Helpers
+%%--------------------------------------------------------------------
+
+async_spawn(Addr, Port, ClientOptions) ->
+    ?INFO_MSG("issue=\"Connecting to Cassandra\", address=~p, port=~p",
+              [Addr, Port]),
+    Parent = self(),
+    proc_lib:spawn_link(fun() ->
```
michalwski: Do we really need to spawn a process to start the `seestar_session`? I understand you don't want to start the connection in the `gen_server`'s `init` function.

You could f.e send a message to itself in the `init` function and start the `seestar` connection in relevant handle info. What do you think about that @arcusfelis?

apps/ejabberd/test/cassandra_SUITE.erl
```diff
@@ -0,0 +1,93 @@
...
+%%--------------------------------------------------------------------
+%% Last tests
+%%--------------------------------------------------------------------
+check_connection(Config) ->
+    case gen_tcp:connect("127.0.0.1", 9042, []) of
+        {ok, _} ->
+            ct:pal("Cassandra is listening on port 9042", []),
+            check_connection2(Config);
+        _ ->
+            ct:pal("Cassandra is not listening on port 9042, skip the test", []),
```
michalwski: I suggest to return a comment from the test case instead of or in addition to the log message.

apps/ejabberd/src/mongoose_cassandra.erl
arcusfelis: servers option is used by driver. It's not self-descriptive and people will make mistakes trying to configure it.

```erlang
{servers, [{"localhost", 9042, 1}]}
```

But I want to use address to be similar to riak backend and make it easier to use

```erlang
{address, "localhost"}
```

So, in case someone still wants to use servers to configure number of workers per server - he can do this. But it's not documented feature.

apps/ejabberd/src/mongoose_cassandra_worker.erl
arcusfelis: Batch is a feature of the driver to run queries on the same partition.
There are several types of them, logged (it's like a transaction) and not logged.

Batch can access data from different partitions but than cassandra documentation recommends to execute each query on the client side for performance reasons. 
So, not_batch is exactly what it is "execute these queries not using server-side batch but keep the same API".

So, we can name it `maybe_logged_or_not_or_manual_batch` but it feels like java.

apps/ejabberd/src/mongoose_cassandra_worker.erl
arcusfelis:
```erlang
State = #state{query_refs=Refs, query_refs_count=RefsCount})
```
or
```erlang
State=#state{query_refs = Refs, query_refs_count = RefsCount})
```
?

apps/ejabberd/src/mongoose_cassandra_worker.erl
arcusfelis: Are there code guidelines somewhere?

apps/ejabberd/src/mongoose_cassandra.erl
arcusfelis: And if we really want to use several servers we still can do it with the address option:

```erlang
{address, "localhost|otherhost"}
```

I just don't like forcing user to use the list of servers when he wants to specify one server only (90% of setups).

apps/ejabberd/src/mongoose_cassandra_worker.erl
michalwski: Ok, I agree. Let's leave it as is.

apps/ejabberd/src/mongoose_cassandra_worker.erl
michalwski: Combined :)

apps/ejabberd/src/mongoose_cassandra_worker.erl
michalwski: They are not written (yet?). But I think adding `space` around `=` or other operators makes things more readable. That's rather general advice and not Erlang-specific.

apps/ejabberd/src/mongoose_cassandra_worker.erl
arcusfelis: UserJID is used to get Worker from a pool:
```erlang
mongoose_cassandra_sup:select_worker(PoolName, UserJID)
```

So, we already using it.
Than after selecting worker we execute a query function.
So, having JID here solves several issues:
- Logging - we can print "issue=query_failed, userid=alice@localhost"
- We need to use Host sometimes for different functions, so we can get it from the UserJID.

michalwski: Thanks for the explanation.

arcusfelis:
Merged.

PR #920, opened 2016-08-16
When unblocking all contacts response should not contain jids

bartekgorny:
This PR addresses the Trello card: "Blocking Command: Bad push received when unblock all"

As stated in XEP-0191, point 3.5, when a user sends an "unblock" stanza without specifying jids to unblock, the server should unblock all contacts and reply with an empty result (no jid list). This PR fixes this.

### DIFF
--- a/test/ejabberd_tests/tests/mod_blocking_SUITE.erl
+++ b/test/ejabberd_tests/tests/mod_blocking_SUITE.erl
@@ -136,8 +136,6 @@ get_block_list(Config) ->
         Config, [{alice, 1}],
         fun(User1) ->
             Result = get_blocklist(User1),
-%%            io:format("Received stanza is ~n~p~n",[Result]),
-%%            ct:pal("RESULT ~p", [Result]),
             escalus:assert(is_iq_result, Result),
             escalus:assert(fun is_blocklist_result_empty/1, Result)
         end).
@@ -321,7 +319,6 @@ blocking_and_relogin_many(Config) ->
             client_gets_blocking_error(User1),
             message_is_delivered(User1, [User2], <<"House of th rising sun">>),
             BlockList = get_blocklist(User1),
-%%            ct:pal("blocklist ~p", [BlockList ]),
             blocklist_contains_jid(BlockList, User3)
         end).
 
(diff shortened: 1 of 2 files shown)


michalwski:
> This PR addresses the Trello card: "Blocking Command: Bad push received when unblock all"

Could you please create issue here from the card and reference to it? Or just put the examples from the card in this PR.

michalwski:
Also dialyzer has some suggestions regarding types.

test/ejabberd_tests/tests/mod_blocking_SUITE.erl
```diff
@@ -529,10 +534,10 @@ get_blocklist_items(Items) ->
 user_blocks(Blocker, Blockees) when is_list(Blockees) ->
     BlockeeJIDs = [ escalus_utils:jid_to_lower(escalus_client:short_jid(B)) || B <- Blockees ],
     AddStanza = block_users_stanza(BlockeeJIDs),
-    ct:pal("add stanza ~p", [AddStanza]),
+%%    ct:pal("add stanza ~p", [AddStanza]),
```
michalwski: Let's remove these commented `ct:pal` calls.

bartekgorny:
According to http://xmpp.org/extensions/xep-0191.html#unblockall

When I unblock all sending:

```
<iq id='6w6t1-62' type='set'><unblock xmlns='urn:xmpp:blocking'/></iq>
```

I should receive a push like this:

```
<iq from='gardano@erlang-solutions.com' to='gardano@erlang-solutions.com/9bb69f6dacb33248' id='push' type='set'><unblock xmlns='urn:xmpp:blocking'/></iq>
```

But I receive this push:

```
<iq from='gardano@erlang-solutions.com' to='gardano@erlang-solutions.com/9bb69f6dacb33248' id='push' type='set'><unblock xmlns='urn:xmpp:blocking'><item jid='griveroa-inaka@erlang-solutions.com'/><item jid='ramabit@erlang-solutions.com'/></unblock></iq>
```

Where `griveroa-inaka@erlang-solutions.com` and `ramabit@erlang-solutions.com` are the JIDs unblocked

PR #922, opened 2016-08-17
Finalise MUC Light HTTP API

Dzol:
This PR introduces source and tests for the MUC Light HTTP Administration API MVP.

### DIFF
--- a/apps/ejabberd/src/mod_muc_light_admin.erl
+++ b/apps/ejabberd/src/mod_muc_light_admin.erl
@@ -1,0 +1,91 @@
+%%==============================================================================
+%% Copyright 2016 Erlang Solutions Ltd.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%% http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% Author: Joseph Yiasemides <joseph.yiasemides@erlang-solutions.com>
+%% Description: Administration commands for MUC Light
+%%==============================================================================
+
+-module(mod_muc_light_admin).
+
+-behaviour(gen_mod).
+-export([start/2, stop/1]).
+
+-export([create_room/3]).
+
+-include("mod_muc_light.hrl").
+-include("ejabberd.hrl").
+-include("jlib.hrl").
+
+
+%%--------------------------------------------------------------------
+%% `gen_mod' callbacks
+%%--------------------------------------------------------------------
+
+start(_, _) ->
+    mongoose_commands:register(commands()).
+
+stop(_) ->
+    mongoose_commands:unregister(commands()).
+
+
+%%--------------------------------------------------------------------
+%% Interface descriptions
+%%--------------------------------------------------------------------
+
+commands() ->
+
+    [
+     [{name, create_muc_room},
+      {category, 'muc-lights'},
+      {desc, "Create a MUC room."},
+      {module, ?MODULE},
+      {function, create_room},
+      {action, create},
+      {identifiers, [domain]},
+      {args,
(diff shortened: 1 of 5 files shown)


apps/ejabberd/src/mod_muc_light_admin.erl
```diff
@@ -119,9 +119,10 @@ create_unique_room(Domain, RoomName, Creator, Subject) ->
 
 invite_to_room(Domain, RoomName, Sender, Recipient) ->
     R = muc_light_room_name_to_jid(jid:from_binary(Sender), RoomName, Domain),
-    S = jid:from_binary(Sender),
+    S = jid:replace_resource(jid:from_binary(Sender), <<>>),
```
michalwski: Really? Did you check it? We need a test case for this.

test/ejabberd_tests/tests/muc_light_http_api_SUITE.erl
```diff
@@ -102,12 +102,16 @@ invite_to_room(Config) ->
         %% XMPP: Alice creates a room.
         escalus:send(Alice, muc_light_SUITE:stanza_create_room(undefined,
             [{<<"roomname">>, Name}], [{Kate, member}])),
+        %% XMPP: Alice recieves a affiliation message to herself and
+        %% an IQ result when creating the MUC Light room.
+        escalus:wait_for_stanza(Alice),
+        escalus:assert(is_iq_result, escalus:wait_for_stanza(Alice)),
         %% HTTP: Invite Bob (change room affiliation) on Alice's behalf.
         Body = #{ sender => escalus_utils:get_jid(Alice),
```
michalwski: The path should contain bare jids. We don't know if the user is connected or no so we should always use bare jids in REST API calls.

apps/ejabberd/src/mod_muc_light_admin.erl
Dzol: Yes, I checked. @michalwski: We could wait for some time with the assumption that if it isn't received in the given time then it's never received.

michalwski: Yes, that's what we usually do when we want to check if there is no stanza for a user. There is `escalus_assert:has_no_stanzas/1` function which you can use.

michalwski: But what worries me most is the empty resource. The resource should always be empty for the REST API. So there is no need to replace a resource with empty one.

test/ejabberd_tests/tests/muc_light_http_api_SUITE.erl
Dzol: We could safe-guard against this on the server too.

michalwski: Yes, we should.

apps/ejabberd/src/mod_muc_light_admin.erl
Dzol: > Yes, that's what we usually do when we want to check if there is no stanza for a user. There is escalus_assert:has_no_stanzas/1 function which you can use.

Commit f61ceccf9bf1d20cca99f3f9c94d4dca3652e1b4 should resolve this. I don't think a timer is necessary because the IQ result should arrive just after the messages informing participants about affiliation changes.

Dzol: > But what worries me most is the empty resource. The resource should always be empty for the REST API. So there is no need to replace a resource with empty one.  

Commit d21da83fb05964aed18e35d6c7aa8474a33f502d introduces safeguards (it copies the same `jid:binary_to_bare/1`as in PR 921).

test/ejabberd_tests/tests/muc_light_http_api_SUITE.erl
```diff
@@ -0,0 +1,181 @@
...
+    MemberJID = escalus_utils:jid_to_lower(escalus_utils:get_short_jid(User)),
+    Data = exml_query:path(Stanza, [{element, <<"x">>}, {element, <<"user">>}, cdata]),
+    MemberJID == Data.
+
+%%--------------------------------------------------------------------
+%% Constants
+%%--------------------------------------------------------------------
+
+muc_light_domain() ->
+    <<"muclight.localhost">>.
```
michalwski: I think we can avoid hardcoding the `localhost` part. Correct domain you can read with:
```erlang
ct:get_config({hosts, mim, domain})
```

PR #925, opened 2016-08-18
Pubsub doc

ludwikbukowski:
For now, our PubSub feature was not documented at all. In this PR I tried to change it.

I faced few issues related to our PubSub implementation:
* pubsub lacks odbc backend implementation
* There are some not ported plugins: [`node_buddy`, `node_dispatch`, `node_mb`, `node_mix`, `node_online`, `node_private`, `node_public`]
* when creating the pubsub node there is possibility to choose plugin which handles node creation by adding `type` attribute to creation stanza. Such a custom feature is not documented in [XEP-0060: Publish-Subscribe](http://www.xmpp.org/extensions/xep-0060.html#owner-create-and-configure). In case to specify node plugin I would rather suggest to use [node configuration](  http://www.xmpp.org/extensions/xep-0060.html#owner-create-and-configure)
* from ejabberd doc:

> (...)Using <<“virtual”>> nodetree requires to start from a clean database, it will not work if you used the default <<“tree”>> nodetree before.(...)

is the cleaning really necessary ?Or can we automatize it?
* values in pubsub configuration must be binaries which is against the Mongooseim string convention

### DIFF
--- a/doc/ejabberd-2.1.8-OLD/guide.tex
+++ b/doc/ejabberd-2.1.8-OLD/guide.tex
@@ -81,7 +81,6 @@
 \newcommand{\modoffline}{\module{mod\_offline}}
 \newcommand{\modofflineodbc}{\module{mod\_offline\_odbc}}
 \newcommand{\modping}{\module{mod\_ping}}
-\newcommand{\modpubsub}{\module{mod\_pubsub}}
 \newcommand{\modprescounter}{\module{mod\_pres\_counter}}
 \newcommand{\modprivacy}{\module{mod\_privacy}}
 \newcommand{\modprivacyodbc}{\module{mod\_privacy\_odbc}}
--- a/doc/ejabberd-2.1.8-OLD/guide.tex
+++ b/doc/ejabberd-2.1.8-OLD/guide.tex
@@ -81,6 +81,7 @@
 \newcommand{\modoffline}{\module{mod\_offline}}
 \newcommand{\modofflineodbc}{\module{mod\_offline\_odbc}}
 \newcommand{\modping}{\module{mod\_ping}}
+\newcommand{\modpubsub}{\module{mod\_pubsub}}
 \newcommand{\modprescounter}{\module{mod\_pres\_counter}}
 \newcommand{\modprivacy}{\module{mod\_privacy}}
 \newcommand{\modprivacyodbc}{\module{mod\_privacy\_odbc}}
--- a/doc/modules/mod_pubsub.md
+++ b/doc/modules/mod_pubsub.md
@@ -1,43 +1,59 @@
 ### What is PubSub?
 
-It’s a fascinating design pattern which mostly promote loose coupling. In the pubsub world we have publishers who fire events and subscribers who wish to be notified when the publisher push data.
-There might be several subscribers, several publishers and even several channels (nodes) where the events are sent.
+PubSub is a fascinating design pattern which mostly promotes loose coupling between two kinds of party (publishers and subscribers).
+Like their names suggest, in the pubsub world we have publishers who fire events, and subscribers who wish to be notified when publishers push data.
+There might be several subscribers, several publishers, and even several channels (nodes) where the events are sent.
 
 ### Module Description
+
 This module implements [XEP-0060 (Publish-Subscribe)](http://www.xmpp.org/extensions/xep-0060.html).
-Due to complexity of the protocol, PubSub engine do successive calls to `nodetree` and `node plugins` in order to check validity, perform corresponding action and return result or appropriate error.
-Such an architecture make it much easier to write custom pubsub plugins and add new storage backends. It's all about to fit your needs!
+Due to the complexity of the protocol, the PubSub engine makes successive calls to `nodetree` and `node plugins` in order to check validity of requests, perform the corresponding action and return a result or appropriate error.
+Such an architecture makes it much easier to write custom pubsub plugins and add new storage backends.
+It's all about fitting PubSub to your needs!
 
 ### Options
 
-* `host` (string, default: `"pubsub.@HOST@"`): Subdomain for Pubsub service to reside under. `@HOST@` is replaced with each served domain.
+* `host` (string, default: `"pubsub.@HOST@"`): Subdomain for Pubsub service to reside under.
+`@HOST@` is replaced with each served domain.
 * `access_create` (atom, default: `all`): Who is allowed to create pubsub nodes.
 * `max_items_node` (integer, default: `10`): Define the maximum number of items that can be stored in a node.
 * `max_subscriptions_node` (integer, default: `undefined` - no limitation): The maximum number of subscriptions managed by a node.
-* `nodetree` (binary, default: `<<"tree">>`): Specify the storage and organisation of the pubsub nodes. Called on get, create and delete node.
- Only one nodetree can be used per host, and is shared by all node plugins.
+* `nodetree` (binary, default: `<<"tree">>`): Specify the storage and organisation of the pubsub nodes.
+Called on `get`, `create` and `delete` node.
+Only one nodetree can be used per host, and is shared by all node plugins.
     * `<<"tree">>"` - store nodes in database (only mnesia supported).
-    * `<<"virtual">>` - does not store nodes in database. This saves resources on systems when the number of nodes is large. If using the `<<“virtual”>>` nodetree, you can only enable those node plugins:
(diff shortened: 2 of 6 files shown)


michalwski:
>is the cleaning really necessary ?Or can we automatize it?

Good question, needs verification :)

doc/modules/mod_pubsub.md
```diff
@@ -1,43 +1,59 @@
...
-Such an architecture make it much easier to write custom pubsub plugins and add new storage backends. It's all about to fit your needs!
+Due to the complexity of the protocol, the PubSub engine makes successive calls to `nodetree` and `node plugins` in order to check validity of requests, perform the corresponding action and return a result or appropriate error.
+Such an architecture makes it much easier to write custom pubsub plugins and add new storage backends.
+It's all about fitting PubSub to your needs!
 
 ### Options
 
-* `host` (string, default: `"pubsub.@HOST@"`): Subdomain for Pubsub service to reside under. `@HOST@` is replaced with each served domain.
+* `host` (string, default: `"pubsub.@HOST@"`): Subdomain for Pubsub service to reside under.
+`@HOST@` is replaced with each served domain.
```
Dzol: So if there are **_N_** XMPP domains will there be **_N_** pubsub domains?

doc/modules/mod_pubsub.md
```diff
@@ -1,43 +1,59 @@
...
-PubSub clients can define which plugin to use when creating a node by adding `type=’plugin-name’` attribute to the create stanza element. If such an attribute is not specified, the default plugin will be the first on the plugin list.
-    * `<<"flat">>` -  No node hierarchy. It handles standard PubSub case
-    * `<<"hometree">>` - Use exact the same features as ﬂat plugin but additionally organise nodes in a tree. Basically it follows the scheme similar to ﬁlesystem's structure.
-    Every user can create nodes in its own home root: e.g `/home/user`. Each node can contain items and/or sub-nodes.
-    * `<<"pep">>` - implementation of [XEP-0060 (Personal Eventing Protocol)](http://xmpp.org/extensions/xep-0163.html). In this case, items are not persisted but kept in memory cache.
-    When the `pep` plugin enabled, user can have his own node (exposed on his bare jid) with a common namespace. Requires module `mod_caps` to be enabled.
+    In this case you should also add `<<“dag”>>` node plugin as default, for example: plugins: `[<<"dag">>,<<"flat">>,<<"hometree">>,<<"pep">>]`
```
Dzol: What do we mean by "[...] also add `<<"dag">>` node plugin as default [...]"?

doc/ejabberd-2.1.8-OLD/guide.tex
```diff
@@ -81,6 +81,7 @@
 \newcommand{\modoffline}{\module{mod\_offline}}
 \newcommand{\modofflineodbc}{\module{mod\_offline\_odbc}}
 \newcommand{\modping}{\module{mod\_ping}}
+\newcommand{\modpubsub}{\module{mod\_pubsub}}
```
michalwski: I think don't have to care about the old doc in tex. It's probably good time to remove it completely, but that's a subject for a separate PR.

doc/modules/mod_pubsub.md
```diff
@@ -0,0 +1,67 @@
+### What is PubSub?
+
+PubSub is a fascinating design pattern which mostly promotes a loose coupling between two kinds of entity (publishers and subscribers).
+Like their names suggest, in the pubsub world we have publishers who fire events, and subscribers who wish to be notified about those events when publishers push data.
```
Nyco: "subscribers who wish to be notified [in realt-time?] about those events"

doc/modules/mod_pubsub.md
ludwikbukowski: from the plugins description:
If client does not explicit specify the default plugin when creating the stanza then:

> (...) the default plugin will be the first on the plugin list. (...)

However, I must admin that p1 made pubsub configuration quite complicated ...

doc/modules/mod_pubsub.md
michalwski: Subscribers doesn't have to be online, in this case they can still get the "unread" events from the node if they wish to.

doc/ejabberd-2.1.8-OLD/guide.tex
ludwikbukowski: ok, I removed it from this list

doc/modules/mod_pubsub.md
```diff
@@ -0,0 +1,67 @@
...
+It's all about fitting PubSub to your needs!
+
+### Options
+
+* `host` (string, default: `"pubsub.@HOST@"`): Subdomain for Pubsub service to reside under.
+`@HOST@` is replaced with each served domain.
+* `access_create` (atom, default: `all`): Who is allowed to create pubsub nodes.
+* `max_items_node` (integer, default: `10`): Define the maximum number of items that can be stored in a node.
+* `max_subscriptions_node` (integer, default: `undefined` - no limitation): The maximum number of subscriptions managed by a node.
+* `nodetree` (binary, default: `<<"tree">>`): Specify the storage and organisation of the pubsub nodes.
```
michalwski: Do we really expect binaries in the pubsub's configuration? Usually the config file contains ordinary strings.

Dzol:
**Note:** we know some of the points above aren't quite resolved yet, but since we plan to polish things in the near future we leave them open, with the intention of addressing them once this module is polished.

doc/modules/mod_pubsub.md
```diff
@@ -0,0 +1,67 @@
...
+    This saves resources on systems when the number of nodes is large.
+    If using the `<<“virtual”>>` nodetree, you can only enable these node plugins:
+    `[<<“flat”>>,<<“pep”>>]` or `[<<“flat”>>]`; any other plugin configuration will not work.
+    Also, all nodes will have the defaut configuration, and this can not be changed.
+    Using the `<<“virtual”>>` nodetree requires the pubsub module to be started with a clean database, it will not work if you used the default `<<“tree”>>` nodetree before.
+    * `<<"dag">>"` - provides experimental support for [XEP-0248 (PubSub Collection Nodes)](http://xmpp.org/extensions/xep-0248.html).
+    In this case you should also add the `<<“dag”>>` node plugin as default, for example: plugins: `[<<"dag">>,<<"flat">>,<<"hometree">>,<<"pep">>]`
+* `ignore_pep_from_offline` (boolean, default: `true`): specify whether or not we should get last published PEP items from users in our roster which are offline when we connect.
+The default option is `true` hence we will get only the last items from the online contacts.
+* `last_item_cache` (boolean, default `false`): specify whether or not pubsub should cache the last items.
```
chrzaszcz: The main reason for `last_item_cache` is the ability to send the last published item to a new subscriber.

doc/modules/mod_pubsub.md
ludwikbukowski: for now it's implemented so that those variables **must** be binaries in config

doc/modules/mod_pubsub.md
ludwikbukowski: thanks for that valuable note! :)

doc/modules/mod_pubsub.md
michalwski: I see, we have to change this while improving the `pubsub` implementation.

doc/modules/mod_pubsub.md
```diff
@@ -0,0 +1,67 @@
...
+    If using the `<<“virtual”>>` nodetree, you can only enable these node plugins:
+    `[<<“flat”>>,<<“pep”>>]` or `[<<“flat”>>]`; any other plugin configuration will not work.
+    Also, all nodes will have the defaut configuration, and this can not be changed.
+    Using the `<<“virtual”>>` nodetree requires the pubsub module to be started with a clean database, it will not work if you used the default `<<“tree”>>` nodetree before.
+    * `<<"dag">>"` - provides experimental support for [XEP-0248 (PubSub Collection Nodes)](http://xmpp.org/extensions/xep-0248.html).
+    In this case you should also add the `<<“dag”>>` node plugin as default, for example: plugins: `[<<"dag">>,<<"flat">>,<<"hometree">>,<<"pep">>]`
+* `ignore_pep_from_offline` (boolean, default: `true`): specify whether or not we should get last published PEP items from users in our roster which are offline when we connect.
+The default option is `true` hence we will get only the last items from the online contacts.
+* `last_item_cache` (boolean, default `false`): specify whether or not pubsub should cache the last items. Such an optioni provides the ability to send the last published item to a new subscriber.
```
michalwski: `optioni` -> `option`

ludwikbukowski: he he it must be `vim` leftover, thanks!:wq

Issue #924, opened 2016-08-18
mod_http_notification error


binhly:
MongooseIM version: 2 beta1
Installed from: source
Erlang/OTP version: 18.3

Describe the issue.

Whenever I enabled the prefix_path option in the configs I get this error when the module is called:

2016-08-17 21:19:58.935 [error] <0.471.0> gen_server <0.471.0> terminated with reason: bad argument in mod_http_notification_requestor:make_req/7 line 87
2016-08-17 21:19:58.937 [error] <0.471.0> CRASH REPORT Process <0.471.0> with 1 neighbours exited with reason: bad argument in mod_http_notification_requestor:make_req/7 line 87
2016-08-17 21:19:58.937 [error] <0.462.0> Supervisor {<0.462.0>,poolboy_sup} had child undefined started with {mod_http_notification_requestor,start_link,undefined} at <0.471.0> exit with reason bad argument in mod_http_notification_requestor:make_req/7 line 87 in context child_terminated

{mod_http_notification, [{pool_size, 5}, {host, "http://127.0.0.1:3000"}, {prefix_path, "/push_service"}]},

When I update the config to look like this:

{mod_http_notification, [{pool_size, 5}, {host, "http://127.0.0.1:3000"}]},

The external server gets called.

michalwski:
@binhly, thanks for reporting the issue!
@bartekgorny, while fixing the bug, please try to reuse the generic http pools introduced by @chrzaszcz in #894

michalwski:
@binhly, the fix was merged. Could you try current master branch?

binhly:
@michalslaski it worked!  Thank you so much!

PR #926, opened 2016-08-18
Http notification fix

bartekgorny:
This PR addresses #924 (fixes a bug related to sending notifications with a prefix defined, and tests it), the module is also rewritten to use generic mongoose_http_client.


rel/files/ejabberd.cfg
```diff
@@ -473,11 +473,10 @@
 %%% ==================================
 %%% POOLS OF OUTGOING HTTP CONNECTIONS
 %%% ==================================
-%%
-%% {http_connections, [{conn1, [{server, "http://server.com:8080"},
-%%                              {pool_size, 50}]}
-%%                    ]}.
 
+{http_connections, [{http_pool, [{server, "http://localhost:8000"},
```
chrzaszcz: I don't think we want to enable such a pool by default - this is just an example.

michalwski: Right, the default config should have no http pools enabled. Such a pool should be started in relevant test.

apps/ejabberd/src/mod_http_notification.erl
```diff
@@ -15,38 +15,17 @@
...
-                  {timeout, Timeout},
-                  {path_prefix, PathPrefix}],
-    {ok, _} = supervisor:start_child(ejabberd_sup,
-        {PoolName,
-            {poolboy, start_link,
-                [PoolOpts, WorkerOpts]},
-            transient, 2000, supervisor, [mod_http_notification_requestor]}),
+    Path = gen_mod:get_module_opt(Host, ?MODULE, path, ?DEFAULT_PATH),
+    PoolName = gen_mod:get_module_opt(Host, ?MODULE, pool_name, ?DEFAULT_POOL_NAME),
+    mod_http_notification_requestor:start(Host, PoolName, Path),
```
michalwski: Correct me if I'm wrong but there is only one `requestor` process for all the possible requests. Am I right?

apps/ejabberd/src/mongoose_http_client.erl
```diff
@@ -165,3 +166,8 @@ stop_supervisor() ->
     Proc = sup_proc_name(),
     ok = supervisor:terminate_child(ejabberd_sup, Proc),
     ok = supervisor:delete_child(ejabberd_sup, Proc).
+
+to_binary(P) when is_list(P) ->
+    list_to_binary(P);
+to_binary(P) when is_binary(P) ->
```
michalwski: When is it possible to have `binary` here? The default option is simple `string`, and we can expect `string`s from the config file as well.

apps/ejabberd/src/mod_http_notification_requestor.erl
```diff
@@ -79,16 +82,43 @@ code_change(_OldVsn, State, _Extra) ->
...
+    mongoose_metrics:update([Host, mod_http_notifications, failed], 1),
+    ?WARNING_MSG("Sending http notification failed: ~p", [Reason]),
+    ok.
+
+%% @doc Convert to binary, strip initial slash (it is added by mongoose_http_client)
+fix_path(P) when is_list(P) ->
+    fix_path(list_to_binary(P));
+fix_path(P) when is_binary(P) ->
+    case P of
+        <<47/integer, R/binary>> ->
```
michalwski: We all know `47` is `/` in ASCII but following syntax is probably more readable:
```erlang
<<$/, R/binary>>
```
or
```erlang
<<"/", R/binary>>
```

apps/ejabberd/src/mod_http_notification.erl
chrzaszcz: It seems that this `requestor` may be not needed at all - is there any reason to keep that extra `gen_server`?

apps/ejabberd/src/mongoose_http_client.erl
bartekgorny: Yes, unless somebody (like me) puts a binary there. Anyway, does it hurt to have it "binary-safe"?

michalwski: It doesn't but why to keep code, which is almost never used? Most people configuring MongooseIM will not try to put binary in the cfg file.

apps/ejabberd/src/mod_http_notification.erl
bartekgorny: The only reason for it is to store pool name and path somewhere at hand, without having to get config upon every request. If the notification module had its own pool it would be stored there, otherwise it has to provide it at call time.

michalwski: Currently, the single process will be bigger bottleneck than reading config at call time.

michalwski:
We need more such PRs where there are more lines removed then added!

apps/ejabberd/src/mod_http_notification.erl
```diff
@@ -115,3 +89,14 @@ record_result(Host, {error, Reason}, _) ->
     mongoose_metrics:update([Host, mod_http_notifications, failed], 1),
     ?WARNING_MSG("Sending http notification failed: ~p", [Reason]),
     ok.
+
+%% @doc Convert to binary, strip initial slash (it is added by mongoose_http_client)
+fix_path(P) when is_list(P) ->
```
michalwski: Since we should expect ordinary string from the config file, maybe it would be more readable to have only one variant of the `fix_path` function which accepts binary only. The conversion from list to binary could be done where the function is called. What do you think?

apps/ejabberd/src/mod_http_notification.erl
```diff
@@ -115,3 +89,14 @@ record_result(Host, {error, Reason}, _) ->
...
+    fix_path(list_to_binary(P));
+fix_path(P) when is_binary(P) ->
+    case P of
+        <<"/", R/binary>> ->
```
michalwski: There is no need for the `case` clause here, we can do the matching in the function head.

michalwski:
Much better :) Thanks! I'll merge as soon as travis gives green light.

PR #927, opened 2016-08-22
Mongoose commands improvements

michalwski:
Proposed changes include:
* mongoose_command's category as binary
* mongoose_command's description as binary
* updated relevant modules and tests affected by the change

### DIFF
--- a/apps/ejabberd/src/mod_muc_admin.erl
+++ b/apps/ejabberd/src/mod_muc_admin.erl
@@ -38,8 +38,8 @@ stop(_) ->
 commands() ->
     [
      [{name, create_muc_room},
-      {category, mucs},
-      {desc, "Create a MUC room."},
+      {category, <<"mucs">>},
+      {desc, <<"Create a MUC room.">>},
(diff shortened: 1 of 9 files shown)


arcusfelis:
Looks fine, but I personally like how it was before (atoms and lists).

ludwikbukowski:
but actually there were just redundant conversions from atom

michalwski:
Ok, I hear you :) why did you like it better?

bartekgorny:
For me the previous version looked better, without all the <<" fluff. But it is a matter of taste. Anyway, ok from me.

michalwski:
Ok, let me explain.

1. The reason to change `category` from atom to binary is to avoid unnecessary conversion from/to binary.
2. The reason to change `des` from ordinary string to binaries is that in the code base we use binaries to represent text. I just want to be consistent with that.

bartekgorny:
Yes I know, so I'm not opposing, just chiming in to show I have something to say ;)

PR #928, opened 2016-08-22
Mongoose admin improvements

michalwski:
This PR goes on top of #927
Proposed changes include:
* rename mongoose_admin to mod_mongoose_admin
* changed names of some commands
* refactoring reducing duplicated code

### DIFF
--- a/apps/ejabberd/src/mod_mongoose_admin.erl
+++ b/apps/ejabberd/src/mod_mongoose_admin.erl
@@ -1,6 +1,8 @@
 -module(mod_mongoose_admin).
 -author('bartlomiej.gorny@erlang-solutions.com').
 
+-behaviour(gen_mod).
+
 -export([start/0, stop/0,
          start/2, stop/1,
          register/3,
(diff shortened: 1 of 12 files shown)


apps/ejabberd/src/mod_mongoose_admin.erl
```diff
@@ -0,0 +1,234 @@
...
+         kick_session/3,
+         get_recent_messages/3,
+         send_message/3
+        ]).
+
+-include("ejabberd.hrl").
+-include("jlib.hrl").
+
+start() ->
+    mongoose_commands:register(commands()).
```
ludwikbukowski: Since module with `mod_` prefix is called for every virtual domain, are we sure that registering commands e.g twice won't end up with some exception?

michalwski: > Since module with mod_ prefix is called for every virtual domain

Any module in modules section is started for every virtual host.

Anyway, `mongoose_commands` handles this and doesn't crash if a command is registered twice. The first registration is used.

ludwikbukowski: OK thanks, I guess this branch will be called https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mongoose_commands.erl#L530

michalwski: Correct.

bartekgorny:
We probably should rename mod_mongoose_admin or add another module to host commands which are not admin-specific (e.g. send_message, get_messages, list_methods).

michalwski:
We definitely should but that's a subject for a separate PR.

apps/ejabberd/src/mod_mongoose_admin.erl
```diff
@@ -0,0 +1,234 @@
...
+      } || C <- mongoose_commands:list(admin)].
+
+get_recent_messages(Caller, Other, Limit) ->
+    Res = lookup_recent_messages(Caller, Other, Limit),
+    lists:map(fun record_to_map/1, Res).
+
+change_user_password(Host, User, Password) ->
+    ejabberd_auth:set_password(User, Host, Password).
+
+record_to_map({Id, From, Msg}) ->
```
ludwikbukowski: what exactly record is expected here?

michalwski: The name is a bit missleading and I didn't improve that when making the PR.
It is not Erlang's record it's rather an `item` returned by `mod_mam`.

PR #931, opened 2016-08-23
Introduce HTTP Administration API for Kicking MUC Users 

Dzol:
This is rebased off-of a recent `master`.

### DIFF
--- a/apps/ejabberd/src/mod_muc_admin.erl
+++ b/apps/ejabberd/src/mod_muc_admin.erl
@@ -114,7 +114,8 @@ create_instant_room(Host, Name, Owner, Nick) ->
     UserRoomJID = jid:make(Name, MUCHost, Nick),
     BareRoomJID = jid:make(Name, MUCHost, <<"">>),
     %% Send presence to create a room.
-    ejabberd_router:route(OwnerJID, UserRoomJID, presence()),
+    ejabberd_router:route(OwnerJID, UserRoomJID,
+                          presence(OwnerJID, UserRoomJID)),
     %% Send IQ set to unlock the room.
     ejabberd_router:route(OwnerJID, BareRoomJID,
                           declination(OwnerJID, BareRoomJID)),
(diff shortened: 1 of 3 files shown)


apps/ejabberd/src/mod_muc_admin.erl
```diff
@@ -101,7 +116,8 @@ create_instant_room(Host, Name, Owner, Nick) ->
     %% Send presence to create a room.
     ejabberd_router:route(OwnerJID, UserRoomJID, presence()),
     %% Send IQ set to unlock the room.
-    ejabberd_router:route(OwnerJID, BareRoomJID, declination()),
```
ludwikbukowski: I know that this question is a bit out of scope of this PR but could someone remind me why do we send presence in order to create a room instead of calling `mod_muc:create_instant_room/5` ?

apps/ejabberd/src/mod_muc_admin.erl
```diff
@@ -101,7 +116,8 @@ create_instant_room(Host, Name, Owner, Nick) ->
     %% Send presence to create a room.
     ejabberd_router:route(OwnerJID, UserRoomJID, presence()),
```
ludwikbukowski: I know that this question is a bit out of scope of this PR but could someone remind me why do we send presence in order to create a room instead of calling mod_muc:create_instant_room/5 ?

michalwski: That's avery good question @ludwikbukowski. The API call would be better as it probably gives some result if the creation succeeded or not. There are some other places like this which we should re-write to use module's API.

Dzol: The module does a lot of work in various other places. Sending the `presence` stanza gets it all done. E.g. creating an instant room through the API wouldn't go through [ACL](https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mod_muc.erl#L492) checks or check if the room [already exists](https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mod_muc.erl#L673).

PR #934, opened 2016-08-24
Commands overloading

bartekgorny:
Proposed changes include:
* introduction of overloading commands (defining multiple commands in the same category and action, but varying arity)
* getting archived messages via REST - latest or prior to a specified date

### DIFF
--- a/apps/ejabberd/src/mod_mongoose_admin.erl
+++ b/apps/ejabberd/src/mod_mongoose_admin.erl
@@ -13,6 +13,7 @@
          list_sessions/1,
          kick_session/3,
          get_recent_messages/3,
+         get_recent_messages/4,
          send_message/3
         ]).
 
@@ -106,14 +107,25 @@ commands() ->
       {result, ok}
      ],
      [
-      {name, get_messages},
+      {name, get_last_messages},
       {category, <<"messages">>},
-      {desc, <<"Get recent messages">>},
+      {desc, <<"Get n last messages">>},
       {module, ?MODULE},
       {function, get_recent_messages},
       {action, read},
       {security_policy, [user]},
       {args, [{caller, binary}, {other, binary}, {limit, integer}]},
+      {result, []}
+     ],
+     [
+      {name, get_messages},
+      {category, <<"messages">>},
+      {desc, <<"Get messages before a certain date ('before' is a unix timestamp in seconds)">>},
+      {module, ?MODULE},
+      {function, get_recent_messages},
+      {action, read},
+      {security_policy, [user]},
+      {args, [{caller, binary}, {other, binary}, {before, integer}, {limit, integer}]},
       {result, []}
      ],
      [
@@ -184,7 +196,12 @@ registered_commands() ->
       } || C <- mongoose_commands:list(admin)].
 
 get_recent_messages(Caller, Other, Limit) ->
-    Res = lookup_recent_messages(Caller, Other, Limit),
+    {MegaSecs, Secs, _} = now(),
+    Future = (MegaSecs + 1) * 1000000 + Secs, % to make sure we return all messages
+    get_recent_messages(Caller, Other, Future, Limit).
+
+get_recent_messages(Caller, Other, Before, Limit) ->
+    Res = lookup_recent_messages(Caller, Other, Before, Limit),
     lists:map(fun record_to_map/1, Res).
 
 change_user_password(Host, User, Password) ->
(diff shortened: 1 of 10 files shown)


apps/ejabberd/src/mod_mongoose_admin.erl
```diff
@@ -184,7 +196,12 @@ registered_commands() ->
       } || C <- mongoose_commands:list(admin)].
 
 get_recent_messages(Caller, Other, Limit) ->
-    Res = lookup_recent_messages(Caller, Other, Limit),
+    {MegaSecs, Secs, _} = now(),
```
michalwski: `erlang:now/0` is deprecated. It's enough to take the time from `os:timestamp/0`

apps/ejabberd/test/commands_SUITE.erl
```diff
@@ -338,14 +338,24 @@ commands_new_temp() ->
...
             {result, {msg, binary}}
         ],
         [
+            {name, command_one_arity},
+            {category, <<"user">>},
+            {desc, <<"do nothing and return">>},
+            {module, ?MODULE},
+            {function, cmd_one},
+            {action, read},
+            {args, [{msg, binary}, {whatever, integer}]}, % different arity
```
michalwski: Looks like there is better name for this arg :)

test/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -68,6 +69,7 @@ test_cases() ->
      session_can_be_kicked,
      messages_are_sent_and_received,
      messages_are_archived,
+     archive_and_paginate,
```
michalwski: Probably function name `messages_can_be_paginated` describes better what the test cases checks.

apps/ejabberd/src/mongoose_api_admin.erl
```diff
@@ -88,26 +88,39 @@ rest_terminate(_Req, _State) ->
...
 from_json(Req, #http_api_state{command_category = Category,
-                               command_subcategory = SubCategory} = State) ->
+                               command_subcategory = SubCategory,
+                               bindings = B} = State) ->
     {Method, Req2} = cowboy_req:method(Req),
-    [Command] = mongoose_commands:list(admin, Category, method_to_action(Method), SubCategory),
-    mongoose_api_common:process_request(Method, Command, Req2, State).
+    Cmds = mongoose_commands:list(admin, Category, method_to_action(Method), SubCategory),
+    {ok, Body, _} = cowboy_req:body(Req),
+    {Data} = jiffy:decode(Body),
```
ludwikbukowski: parsing body is already done in `mongoose_api_common:process_request/4` few lines below.
Maybe it is possible to move this arity checks there[1] in case to avoid double parsing and code duplication?

[1] https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mongoose_api_common.erl#L101

test/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -183,6 +185,53 @@ messages_are_archived(Config) ->
         BobJID = maps:get(sender, Previous)
     end).
 
+send_and_wait(AliceJID, BobJID, Msg) ->
+    M = #{caller => BobJID, to => AliceJID, msg => Msg},
+    {?OK, _} = post(<<"/messages">>, M),
+    M1 = #{caller => AliceJID, to => BobJID, msg => Msg},
+    {?OK, _} = post(<<"/messages">>, M1),
+    timer:sleep(2000).
```
michalwski: Why do we wait here so long?

test/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -183,6 +185,53 @@ messages_are_archived(Config) ->
...
+
+get_messages(Me, Other, Count) ->
+    GetPath = lists:flatten(["/messages/",binary_to_list(Me),
+        "/",binary_to_list(Other),"/", integer_to_list(Count)]),
+    {?OK, Msgs} = gett(GetPath),
+    Msgs.
+
+get_messages(Me, Other, Before, Count) ->
+    GetPath = lists:flatten(["/messages/",binary_to_list(Me),
+        "/",binary_to_list(Other),"/", integer_to_list(Before), "/", integer_to_list(Count)]),
```
michalwski: The indentation here could be improved.

apps/ejabberd/src/mongoose_api_admin.erl
bartekgorny: Yes, we should either move arity checks down or body parsing up, I'm not sure which is better. I'd say arity check should be done before calling process_request.

test/ejabberd_tests/tests/rest_SUITE.erl
bartekgorny: Because then we search messages by timestamp in seconds, so we need a pause to make sure timestamps are different. Maybe we could get away with one second pause, though with all those random delays it could trigger random fails.

michalwski: How about checking every 1000ms but no longer than 10s f.e.
BTW, why timestamp in sec and not in milisec?

apps/ejabberd/src/mongoose_api_admin.erl
michalwski: The one which is simpler :)

michalwski: I mean simpler to read and reason about not necessarily simpler to write :)

test/ejabberd_tests/tests/rest_SUITE.erl
bartekgorny: It is not about checking - to test search I need to build an archive, so I'm sending messages every two seconds. When after sending the last one I go back three seconds I can be pretty sure I'm going to skip the last two messages.

Timestamp is is in seconds because it is eventually going to be used by web clients, I think it is more natural for them to use seconds.

michalwski: How about bootstraping the archive in the same or similar way it was done in mam_SUITE? Take a look at [this helper](https://github.com/esl/MongooseIM/blob/735991785d1832e3273f5b3e98589bbf9d0dd66e/test/ejabberd_tests/tests/mam_helper.erl#L906) and how it was used.

test/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -253,10 +233,73 @@ password_can_be_changed(Config) ->
...
+    BArcId = make_arc_id(Bclient),
+    Msg = mam_helper:generate_msg_for_date_user(AArcId, BArcId, DateTime, Content),
+    put_msg(Msg),
+    ok.
+
+put_msg({{MsgIdOwner, MsgIdRemote},
+    {_FromBin, FromJID, FromArcID},
+    {_ToBin, ToJID, ToArcID},
+    {_, Source, _}, Packet}) ->
+    Host = escalus_ct:get_config(ejabberd_domain),
```
michalwski: Such domain reading should be slowly replaced by `ct:get_config({hosts, mim, domain})`.
You can also use the `host/0` function from this module which does exactly that.

Issue #853, opened 2016-06-10
After 2 days, mongooseim is still running, but mongooseimctl status fails with Failed RPC connection to the node


kenstir:
MongooseIM version: master 1cac77f
Installed from: source
Erlang/OTP version: esl-erlang-18.3-1.x86_64

"mongooseimctl status" failed after 2 days idle.  This node was up and I could login to it, though other nodes in the cluster were down.

`# /usr/mongooseim/bin/mongooseimctl status`
Failed RPC connection to the node 'mim@nglab-perf-5-11': {'EXIT',
                                                          {function_clause,
                                                           [{calendar,
                                                             seconds_to_time,
                                                             [173946],
                                                             [{file,
                                                               "calendar.erl"},
                                                              {line,358}]},
                                                            {ejabberd_ctl,
                                                             get_uptime,0,
                                                             [{file,
                                                               "src/ejabberd_ctl.erl"},
                                                              {line,912}]},
                                                            {ejabberd_ctl,
                                                             process,1,
                                                             [{file,
                                                               "src/ejabberd_ctl.erl"},
                                                              {line,153}]},
                                                            {rpc,
                                                             '-handle_call_call/6-fun-0-',
                                                             5,
                                                             [{file,"rpc.erl"},
                                                              {line,206}]}]}}

michalwski:
Thanks for reporting @kenstir, this was fixed in #938

PR #938, opened 2016-08-25
make status ctl working for uptime longer than one day

ludwikbukowski:
This PR addresses https://github.com/esl/MongooseIM/issues/853

Proposed changes include:
* add `days` to uptime statistics in `mongooseimctl status` command

Before, it was `seconds_to_time/1` fun used. What Erlang documentation says about it:

> Computes the time from the specified number of seconds. Seconds must be less than the number of seconds per day (86400).

so the call was crashing every time uptime was longer than one day because it was called with
seconds returned by `erlang:statistics(wall_clock)` [1]

[1] https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/ejabberd_ctl.erl#L911

### DIFF
--- a/apps/ejabberd/src/ejabberd_ctl.erl
+++ b/apps/ejabberd/src/ejabberd_ctl.erl
@@ -64,7 +64,7 @@
 
 -define(ASCII_SPACE_CHARACTER, $\s).
 -define(PRINT(Format, Args), io:format(lists:flatten(Format), Args)).
--define(TIME_HMS_FORMAT, "~B:~2.10.0B:~2.10.0B").
+-define(TIME_HMS_FORMAT, "~B days ~B hours ~B minutes ~B secs").
 -define(a2l(A), atom_to_list(A)).
 
 %%-----------------------------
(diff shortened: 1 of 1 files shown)


apps/ejabberd/src/ejabberd_ctl.erl
```diff
@@ -64,7 +64,7 @@
 
 -define(ASCII_SPACE_CHARACTER, $\s).
 -define(PRINT(Format, Args), io:format(lists:flatten(Format), Args)).
--define(TIME_HMS_FORMAT, "~B:~2.10.0B:~2.10.0B").
+-define(TIME_HMS_FORMAT, "~B days ~B hours ~B minutes ~B secs").
```
michalwski: I think output like `X days 12:30:11` would be enough. Also the formatting `~2.10` should be preserved.

ludwikbukowski: I don't have a strong opinion about it; as you wish!

Issue #890, opened 2016-07-21
mongoseimctl unregister problem


andrewvmail:
MongooseIM version: current master
Installed from: source
Erlang/OTP version: OTP 18

I have mam cassandra backend, http auth, mod_admin_extra on. Not sure why its doing this..

I think the first error is related to mongooseimctl because i get it on all mongooseimctl commands. It throws that error even though the command works. The one below I think is a seperate error. 

I'll just clean up the database manually for now..

```
[root@maux bin]# /root/MongooseIM/rel/mongooseim/bin/mongooseimctl unregister user1 tosmosmobile.com
{error_logger,{{2016,7,21},{19,51,9}},"application_controller: ~ts: ~ts~n",[["syntax error before: ","'/'"],"/root/MongooseIM/rel/mongooseim/lib/ejabberd-2.1.8+mim-1.6.0-180-g7787e31/ebin/"]}

=ERROR REPORT==== 21-Jul-2016::15:51:09 ===
application_controller: syntax error before: '/': /root/MongooseIM/rel/mongooseim/lib/ejabberd-2.1.8+mim-1.6.0-180-g7787e31/ebin/
2016-07-21 19:51:10.215 [error] <0.7391.0>@ejabberd_hooks:run1:240 {badarg,[{ets,member,['ejabberd_users_muclight.tosmosmobile.com',{<<"84b73cf1-6e8c-493e-8e3e-329be37a3baf">>,<<"muclight.tosmosmobile.com">>}],[]},{ejabberd_users,is_cached_user_exists,2,[{file,"src/ejabberd_users.erl"},{line,191}]},{ejabberd_users,is_user_exists,2,[{file,"src/ejabberd_users.erl"},{line,74}]},{mod_mam_con_ca_arch,'-remove_archive/3-lc$^0/1-0-',3,[{file,"src/mod_mam_con_ca_arch.erl"},{line,567}]},{mod_mam_con_ca_arch,remove_archive,3,[{file,"src/mod_mam_con_ca_arch.erl"},{line,565}]},{safely,apply,3,[{file,"src/safely.erl"},{line,19}]},{ejabberd_hooks,run1,3,[{file,"src/ejabberd_hooks.erl"},{line,236}]},{mod_mam,remove_archive,3,[{file,"src/mod_mam.erl"},{line,680}]}]}
[...]

PR #939, opened 2016-08-25
Fix for`is_cached_user_exists` function crash

ludwikbukowski:
This PR initially addressed https://github.com/esl/MongooseIM/issues/890
but then it turned out that cassandra backend has been temporary removed so
now, it's not possible to reproduce this particular issue. 
However, I introduce the fix for that bug in case to avoid similar problems in future.

When function `ejabberd_users:is_chached_user_exists/2` was called with an unexistent domain name the crash occurred instead of returning `false` value.

### DIFF
--- a/apps/ejabberd/src/ejabberd_users.erl
+++ b/apps/ejabberd/src/ejabberd_users.erl
@@ -4,7 +4,7 @@
 -export([start/1,
          stop/1,
          start_link/2,
-         is_user_exists/2]).
+         does_user_exist/2]).
 
 %% Hooks.
 -export([remove_user/2]).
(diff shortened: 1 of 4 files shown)


michalwski:
I believe it is possible to have a test case for that (in small test preferably). The cacheing thing looks detached from cassandra backend. F.e. `mod_mam` calls `ejabberd_users:is_user_exists` [here](https://github.com/esl/MongooseIM/blob/021978075cd2ff4affd270a07c4e847dac29e245/apps/ejabberd/src/mod_mam.erl#L290)

BTW, I suggest changing the function names in this module to correct ones:
* is_user_exists -> does_user_exist
* is_cached_user_exists -> does_cached_user_exist
* is_stored_user_exists -> does_cached_user_exist

PR #940, opened 2016-08-29
Global metrics only switch

fenek:
This PR adds `all_metrics_are_global` switch.

Proposed changes include:
* `mongoose_metrics` got some serious refactoring.
* New option, when enabled, merges all per-host metrics into global equivalents, effectively reducing CPU and memory footprint in clusters with lots of XMPP domains supported.
* A side effect is an unification of metrics names. Host part (or `global` atom) always comes first now.
* Only unit tests and metrics_api suite are updated, because remaining metrics test cases rely heavily on RPC, which doesn't offer nice and consistent way of making calls to more than one node. The new switch is enabled on dev node 2 to avoid restarts or introducing yet another test configuration
* Updated documentation, including a note about performance problem the new switch solves.

Not all metrics test suites are modified to verify both nodes (with switch enabled and disabled) because they rely on many RPC calls and this mechanism is not ready yet for flexible calls to any node.

### DIFF
--- a/test/ejabberd_tests/tests/metrics_helper.erl
+++ b/test/ejabberd_tests/tests/metrics_helper.erl
@@ -4,12 +4,9 @@
 
 -compile(export_all).
 
-
-get_counter_value({_, _} = Counter) ->
-    Result = rpc:call(ct:get_config(ejabberd_node),
-                      mongoose_metrics,
-                      get_metric_value,
-                      [Counter]),
+get_counter_value({Host, Metric}) ->
+    Result = rpc:call(ct:get_config(ejabberd_node), mongoose_metrics,
+                      get_metric_value, [make_metric_name(Host, Metric)]),
     case Result of
         {ok, [{count, Total}, {one, _}]} ->
             {value, Total};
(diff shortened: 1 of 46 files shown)


test/ejabberd_tests/tests/metrics_helper.erl
```diff
@@ -19,12 +17,57 @@ get_counter_value({_, _} = Counter) ->
...
+make_global_group_name(GN) ->
+    list_to_atom(atom_to_list(GN) ++ "_all_metrics_are_global").
+
+make_global_groups_names(GroupsNames) ->
+    [{group, make_global_group_name(GN)} || GN <- GroupsNames].
+
+make_global_groups(Groups) ->
+    [{make_global_group_name(GN), Opts, Cases} || {GN, Opts, Cases} <- Groups].
+
+userspec(User1Count, Config) ->
```
michalwski: This function should accept the same user specs as `escalus:story/3` to make things simpler.
And BTW, why exactly do we need this function?

fenek: The purpose of this function is to return `escalus:story/3`-compatible userspec, based on presence (or absence) of `all_metrics_are_global` flag in test config. When it's present, users from node 2 are used in spec. This makes this choice transparent for actual suite.

michalwski: Ok, I see it now, it's not the easiest thing to follow. From what I understand it assumes that the `escalus_users` in `Config` contains only correct users. Probably a word of explanation will make things easier for others.

fenek: Word of explanation like explaining it in a comment to this function, right? No problem.

michalwski: Yes, that's what I meant :)

Issue #911, opened 2016-08-08
Unified subdomain -> domain mapping + config access


fenek:
MongooseIM version: master (ae15bbb)
Installed from: source
Erlang/OTP version: any

I'd like to unify the path from getting a service subdomain (`muc.@HOST@`, `vjud.@HOST`...) to manipulating actual configuration. Currently only `mod_mam_muc` maintains a mapping table between subdomain and domain. MUC Lights maintains own config table instead of mappings. `mod_muc` keeps service configuration in single supervised process.

I think these should be replaced with a unified table with subdomain => domain mappings manipulated with new functions in `gen_mod` API. New functions would include such, that would allow to directly access config options by subdomain.

I dislike a bit an idea of having extra "translating" call but I guess with `read_concurrency` it will be really cheap. The alternative is to maintain extra table with a copy of config options per subdomain. Or to keep subdomain entries in the same option table, as "domain" entries (`ejabberd_module`). I'm afraid it would be too painful to keep these in sync.

The task list I have in mind is: (might be split into separate issues/PRs)
- [x] Maintain subdomain -> domain mappings in `gen_mod` and add/remove them automatically
- [x] Add API for config manipulation with automatic translation between subdomain and domain, e.g. `get_module_opt_by_subhost`
- [x] Use new logic in `mod_muc_light` and `mod_mam_muc`
- [ ] Use new logic in `mod_pubsub`, `mod_muc` and `mod_vcard` - maybe their single gen_servers/fsms with configuration in state may be removed to provide better parallelism

Thoughts?

PR #945, opened 2016-08-30
Improved subdomain handling

fenek:
This PR partially addresses #911

Proposed changes include:
- `gen_mod` maintains a table of subdomain->domain mappings for modules that register themselves under these subdomains.
- Entries are automatically added/removed
- API is provided to manipulate module options by subdomain
- When any module uses `host` option, it is mandatory now (makes it easier to update mappings table)
- Main module table is now created with `read_concurrency` enabled
- Clarified naming convention in some places (Host/ServerHost/Server vs. Host/SubHost/MUCHost/etc.)
- Updated docs about `host` becoming mandatory
- `mod_mam_muc` is now configured with MUC module name, not domain; it also uses new registry instead of custom, private table
- `mod_muc_light` uses new registry instead of own, private table

### DIFF
--- a/apps/ejabberd/src/gen_mod.erl
+++ b/apps/ejabberd/src/gen_mod.erl
@@ -53,9 +53,8 @@
 
 -include("ejabberd.hrl").
 
--record(ejabberd_module, {module_host, module_subhost, opts}).
+-record(ejabberd_module, {module_host, opts}).
 -type ejabberd_module() :: #ejabberd_module{module_host :: {module(), ejabberd:server()},
-                                            module_subhost :: {module(), ejabberd:server()},
                                             opts :: list()}.
 
 -record(ejabberd_module_subhost, {subhost, module, host}).
--- a/apps/ejabberd/src/gen_mod.erl
+++ b/apps/ejabberd/src/gen_mod.erl
@@ -37,23 +37,32 @@
          get_opt/2,
          get_opt/3,
          get_opt/4,
-         get_opt_host/3,
+         get_opt_subhost/2,
          set_opt/3,
          get_module_opt/4,
-         get_module_opt_host/3,
+         get_module_opt_by_subhost/4,
          set_module_opt/4,
+         set_module_opt_by_subhost/4,
+         get_module_subhost/2,
+         get_module_and_host_by_subhost/1,
          loaded_modules/1,
          loaded_modules_with_opts/1,
-         get_hosts/2,
          get_module_proc/2,
          backend_code/3,
          is_loaded/2]).
 
 -include("ejabberd.hrl").
 
--record(ejabberd_module, {module_host, opts}).
+-record(ejabberd_module, {module_host, module_subhost, opts}).
 -type ejabberd_module() :: #ejabberd_module{module_host :: {module(), ejabberd:server()},
+                                            module_subhost :: {module(), ejabberd:server()},
                                             opts :: list()}.
+
+-record(ejabberd_module_subhost, {subhost, module, host}).
+-type ejabberd_module_subhost() :: #ejabberd_module_subhost{
+                                      subhost :: ejabberd:server(),
+                                      module :: module(),
+                                      host :: ejabberd:server()}.
 
 %% -export([behaviour_info/1]).
 %% behaviour_info(callbacks) ->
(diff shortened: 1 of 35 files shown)


apps/ejabberd/src/gen_mod.erl
```diff
@@ -55,6 +57,12 @@
 -type ejabberd_module() :: #ejabberd_module{module_host :: {module(), ejabberd:server()},
                                             opts :: list()}.
 
+-record(ejabberd_module_subhost, {subhost, module, host}).
+-type ejabberd_module_subhost() :: #ejabberd_module_subhost{
```
michalwski: Would it be enough to name the type just `subhost`?

apps/ejabberd/src/mod_mam_muc.erl
```diff
@@ -260,7 +225,7 @@ filter_room_packet(Packet, EventData) ->
         FromJID :: ejabberd:jid(), RoomJID :: ejabberd:jid(),
         Role :: mod_muc:role(), Affiliation :: mod_muc:affiliation()) -> packet().
 archive_room_packet(Packet, FromNick, FromJID=#jid{}, RoomJID=#jid{}, Role, Affiliation) ->
-    Host = server_host(RoomJID),
+    {_, Host} = gen_mod:get_module_and_host_by_subhost(RoomJID#jid.lserver),
```
michalwski: I see there is many cases where only the `Host` is needed. Maybe it would be worth to create a function which returns just `Host`? `gen_mod:get_host_by_subhost`?

michalwski:
If I understand it correctly, the `subhost` is f.e `muclight.esl.com` for the `host` `esl.com`.
There is function returning `host` by `subhost` (`gen_mod:get_module_and_host_by_subhost`). Is it possible that for `subhost` `muclight.esl.com`, this function returns sth else than `esl.com`?

apps/ejabberd/src/gen_mod.erl
fenek: `subhost_mapping`? :)

apps/ejabberd/src/mod_mam_muc.erl
fenek: Sure, why not.

fenek:
@michalwski 
Nope, hosts and subhosts have 1:n relation, so if the function returns something else than `esl.com` for `muclight.esl.com`, it's an error. Now when I think of it, module startup should fail if it tries to register under existing subdomain. I'll add it. Currently modules can poach subhost one from the other. :)

doc/modules/mod_mam.md
```diff
@@ -22,7 +22,7 @@ If you haven't chosen any of the above, skip the next part.
 **Options:**
 
 * `mod_mam_muc`:
-    * `host` (optional, default: `"conference.@HOST@"`): MUC host that will be archived; **Warning:** If you are using MUC Light, make sure this option is set to MUC Light domain.
+    * `muc_module` (optional, default: `mod_muc`): MUC module the MAM service will attach to; **Warning:** If you are using MUC Light, make sure this option is set to `mod_muc_light`.
```
michalwski: What should we do if we use both `muc` and `muc_light`?

fenek: Not sure, is it even verified that `mod_mam_muc` can be started 2 times to support both `muc `and `muc_light`? If so, I think this change won't break anything because it just adds a translation step during init from module name to its subhost.

michalwski: This should be investigated then. A test cases where MAM works for both `mod_muc` and `mod_muc_light` would be great.

fenek: Could be a subject of another PR. Currently I won't have time probably to do any open source work. :(

arcusfelis:
You can't start MAM modules twice (as any other modules).
But, yes, it's possible to extend mod_mam_muc to support several MUC modules at the same time.

michalwski:
> You can't start MAM modules twice (as any other modules).

Actually I already started `mod_mam_muc` (the one before changes in this PR) twice on our staging server. Looks like everything works fine so far.

fenek:
Will rebase, improve and open as new PR later.

Issue #958, opened 2016-09-03
REST API Latest Documentation Parameter Name Incorrect


svarlamov:
MongooseIM version: master branch at 36b777c994da6dec40aa23323971da6a3131c27d
Installed from: source
Erlang/OTP version: 1.18.3

The api documentation to register a new user at http://mongooseim.readthedocs.io/en/latest/swagger/index.html#!/default/post_users_XMPPHost has the wrong name for the 'username' parameter. Should be called 'user' as in the test file at https://github.com/esl/MongooseIM/blob/master/test/ejabberd_tests/tests/rest_SUITE.erl#L123

PR #959, opened 2016-09-03
[skip ci] Update name of username field in request

svarlamov:
This PR addresses #958 

Proposed changes include:
1. Change name of the field `username` to `user` to work with the REST API

### DIFF
--- a/doc/http-api/backend_swagger.yml
+++ b/doc/http-api/backend_swagger.yml
@@ -65,7 +65,7 @@ paths:
             title: Credentials
             type: object
             properties:
-              username:
+              user:
                 type: string
               password:
                 type: string


michalwski:
Hi @svarlamov,

Thanks for you contribution. Could you share with us your thoughts regarding the REST API for backend services?

svarlamov:
@michalslaski Thanks for accepting my contribution! I was trying to use the HTTP admin API and actually had a lot of issues, for instance the send message to muc_light room defined here: http://mongooseim.readthedocs.io/en/latest/swagger/index.html#!/default/put_mucs_XMPPHost_roomName_messages

Example:
Request:
```curl -X PUT -H "Content-Type: application/json" -d '{
    "sender": "sv@localhost",
    "message": "Hello"
}' "http://localhost:8088/api/muc-lights/localhost/my_room/messages"```
Response:
```> PUT /api/muc-lights/localhost/nr/messages HTTP/1.1
> User-Agent: curl/7.35.0
> Host: localhost:8088
> Accept: */*
> Content-Type: application/json
> Cache-Control: no-cache
> Postman-Token: 40fa2484-c38a-9aa9-2a59-606f75b4c130
> Content-Length: 57
>
* upload completely sent off: 57 out of 57 bytes
< HTTP/1.1 404 Not Found
* Server Cowboy is not blacklisted
< server: Cowboy
< date: Tue, 06 Sep 2016 10:04:03 GMT
< content-length: 22
< content-type: application/json
<
* Connection #0 to host localhost left intact```

But I already tested that the muc_light rooms work fine with the XMPP API, and I also got all of the functionality by writing a custom module for the mod_admin_extra_* set of modules. Due to some time constraints on my end, I have just written an HTTP server that takes commands and args in POST requests, and then uses the `mongooseimctl` binary to execute them. 

All of this is even odder to me considering that I tried running the quick tests suite on my box, and everything regarding the HTTP admin API stuff passed! Do you have any ideas what might be at play here? This example request also doesn't put anything into the log files (i've got it set to the Warning level in ejabberd.ctl)

Would love to hear your guys' thoughts, and how this might be improved in the future!

Issue #955, opened 2016-09-02
MUCLight room is not destroyed when all occupants are removed


michalwski:
MongooseIM version: master

In our tests there is many MUCLight rooms left. After the suite all used users are removed from the server but MUCLight rooms remain. I noticed that while developing tests for REST API in #881.

PR #960, opened 2016-09-03
Fix broken MUC Light room destruction

fenek:
This PR addresses #955 

Proposed changes include:
* Room destruction now may be triggered by user removal.
* Room removal is now working correctly when affiliation list is empty.
* New test case for user removal.
* Added a check in `leave_room` test case, whether the room gets removed.

### DIFF
--- a/test/ejabberd_tests/tests/muc_light_SUITE.erl
+++ b/test/ejabberd_tests/tests/muc_light_SUITE.erl
@@ -6,7 +6,8 @@
 -include_lib("exml/include/exml.hrl").
 
 -export([ % service
-         mismatched_default_config_is_rejected/1
+         mismatched_default_config_is_rejected/1,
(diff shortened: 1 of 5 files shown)


test/ejabberd_tests/tests/muc_light_SUITE.erl
```diff
@@ -112,18 +113,19 @@
...
 all() ->
     [
-     {group, service},
-     {group, entity},
-     {group, occupant},
-     {group, owner},
-     {group, blocking},
-     {group, limits}
+     {group, service}
+     %{group, entity},
```
michalwski: Is this a mistake the all groups but `service` are commented out?

fenek: Of course it's a mistake. Let me fix this...

Issue #962, opened 2016-09-06
auth_http sends an empty Authorization header


igors:
MongooseIM version: current git master
Installed from: source
Erlang/OTP version: Erlang/OTP 18

When using http authentication (i.e., {auth_method, http}), mongooseIM sends an "empty" Authorizaton header even when no "basic_auth" is configured.  In other words, with the following config:
```
{auth_method, http}.
{auth_opts, [
    {host, "http://localhost:9999"},
    {path_prefix, "/auth/"}
]}.
```
The following Authorization header will go on the wire:

`Authorization: Basic `

This is a) useless b) strictly speaking, it's an invalid Authorization header (the BNF for it includes base64 encoding of at least one character, `:`)

PR #963, opened 2016-09-06
Don't send empty Authorization header

igors:
This PR addresses #962 

Proposed changes include:
* Don't send Authorization header if no basic_auth is configured in auth_http


michalwski:
Many thanks for your contribution @igors

Issue #964, opened 2016-09-08
External HTTP Authentication for MUC reports integer status code as character


kenstir:
MongooseIM version: 2.0.0beta2
Installed from: source
Erlang/OTP version: 18.3

Describe the issue.

If mod_muc_room is configured using External HTTP Authentication, and it receives the response

    {"code":122,"msg":"password is incorrect"}
    
it would respond with error with a text element

    <text xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'>z password is incorrect</text>

PR #965, opened 2016-09-08
Fix error reporting by mod_muc_room using External HTTP Authentication

kenstir:
This PR addresses #964 

Convert integer to string inside decode_http_auth_response/1


kenstir:
It looks to me like some tests may have failed due to issues fixed in PR #968 .  Is there some action I should take to cause the tests to rerun?

michalwski:
Hi @kenstir, such tests fail at random when travis is loaded. We are currently working on stabilising them. I've restarted the job and now everything is green. Thanks for your patch.

PR #971, opened 2016-09-09
Check for unsupervised links in gen_mod

arcusfelis:
This PR contains two features:
- Allow for some log messages to trigger CI failure
- Warn about links added inside start_module callback.

Bad-written modules can link to the caller process in start_module callback
It works without errors until we try to start modules in the shell or using RPC

### DIFF
--- a/apps/ejabberd/src/gen_mod.erl
+++ b/apps/ejabberd/src/gen_mod.erl
@@ -77,6 +77,7 @@ start() ->
                    Module :: module(),
                    Opts :: [any()]) -> any().
 start_module(Host, Module, Opts0) ->
+    {links, LinksBefore} = erlang:process_info(self(), links),
     Opts = clear_opts(Module, Opts0),
     set_module_opts_mnesia(Host, Module, Opts),
     ets:insert(ejabberd_modules,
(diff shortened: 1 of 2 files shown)


apps/ejabberd/src/gen_mod.erl
```diff
@@ -77,13 +77,25 @@ start() ->
...
     try
         Res = Module:start(Host, Opts),
+        {links, LinksAfter} = erlang:process_info(self(), links),
+        case lists:sort(LinksBefore) =:= lists:sort(LinksAfter) of
+            true -> ok;
+            false ->
+                %% Note for programmers:
+                %% Never call start_link directory in module:start function!
+                %% The process will be killed if we start modules remotely or in shell
+                ?ERROR_MSG("issue=unexpected_links, cause_travis_build_failure=true,"
```
michalwski: I suggest adding the `cause_travis_build_failure=true` only if the env variable `TRAVIS` (or `CI`) is set to true.
Also, the phrase could be changed to `fail_ci_build=ture`.

michalwski:
@arcusfelis Will you find some time to address the comments?

PR #972, opened 2016-09-09
Extend Makefile with qtest task

arcusfelis:
This feature adds:
- Adds "make qtest SUITE=mam_SUITE" that prepares CT config for you and runs specified feature tests only. 
- Ensures that Common Test node is down when start a new test (allows to restart tests fast on dev machine), see test_node_stopped;
- Ensures that there is no config backups. Config backups are created when we reload configuration. If the backup is still here when we run another test, than we can be sure that something is wrong with configuration and our tests will fail (so, just restart mongoose nodes by hand). Useless on CI, helps a lot on dev environment against many types of errors related with configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,11 +55,11 @@ qct:
 	else ct_run -pa apps/*/ebin -pa deps/*/ebin -pa ebin -dir apps/*/test\
         -I apps/*/include -I apps/*/src -logdir /tmp/ct_log -noshell; fi
 
-test: test_deps
+test: test_node_stopped test_deps
 	cd test/ejabberd_tests; make test
 
 # $ make qtest SUITE=mam_SUITE
-qtest:
+qtest: test_node_stopped
 	cp test/ejabberd_tests/qtest.spec.template test/ejabberd_tests/qtest.spec
 	echo "{suites, \"tests\", $(SUITE)}." >> test/ejabberd_tests/qtest.spec
 	cd test/ejabberd_tests;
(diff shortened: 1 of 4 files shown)


tools/check_cfg_backup.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+if ls dev/mongooseim_*/etc/ejabberd.cfg.bak 1> /dev/null 2>&1; then
+    echo "Unrestored config exists. Your actions:"
+    echo "1. Stop nodes"
+    echo "2. Run \"make devrel\" again to restore original configs"
```
michalwski: Sometimes it's also necessary to remove dev nodes before running `make devrel`

Makefile
```diff
@@ -55,12 +55,28 @@ qct:
...
         -I apps/*/include -I apps/*/src -logdir /tmp/ct_log -noshell; fi
 
-test: test_deps
+test: test_node_stopped test_deps
 	cd test/ejabberd_tests; make test
 
-test_preset: test_deps
+# $ make qtest SUITE=mam_SUITE
+# $ make qtest SUITE="[mam_SUITE,metrics_api_SUITE]"
+qtest: test_node_stopped check_cfg_backup
```
michalwski: Other targets taking SUITE as a param add the `_SUITE` suffix. Let's be consistent and add the suffix for this target as well.

arcusfelis: There is no consistent way:

```
make ct SUITE=cassandra
make qct SUITE=amp_resolver_SUITE
```

Also, it's possible to add prefix for each element of the list, but requires extra script to parse erlang term.

michalwski: Actually you example is wrong, it should be:
```
make ct SUITE=cassandra
make qct SUITE=cassadnra
```

Try this :)

tools/check_cfg_backup.sh
bartekgorny: +1, there might be some old junk in dev nodes, and since we rebuild then removing dev nodes brings no extra cost.

tools/check_cfg_backup.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+if ls dev/mongooseim_*/etc/ejabberd.cfg.bak 1> /dev/null 2>&1; then
+    echo "Unrestored config exists. Your actions:"
```
bartekgorny: For me it sounds confusing, as if I were to choose one of the actions. I'd suggest rephrasing it to sthg like: "You should now do the following:", or just simply: "Stop nodes and rebuild dev nodes".

arcusfelis:
Too old

PR #973, opened 2016-09-09
HTTP client API small improvements

michalwski:
Proposed changes include:
* return status 403 when non-participant tries to invite a user
* return status 403 when non-participant tries to read room's archive
* allow to enable compression (for all cowboy endpoints)


ludwikbukowski:
looks OK

bartekgorny:
OK from me

PR #977, opened 2016-09-12
JSON Web Token authentication

astro:
This patchset implements a authentication using [JWT](http://jwt.io/).

It's configurable to check the user part of the authenticating JID against a key in the JWT payload. The _jwerl_ library has been extended to check validity timestamps.


bartekgorny:
It's a good thing, but we are just about to roll out 2.0, we'll come back to it immediately after that.

michalwski:
We'd like to merge your pull request to the master branch. There is one thing that bothers me, the XMPP auth mechanism. Right now your solution works with SASL PLAIN, maybe it would be better to introduce a new, custom one for JWT? For example X-JWT? What do you think @astro?

kenstir:
Even without this enhancement, authentication via JWT can be accomplished using the HTTP Authentication Module with "custom" auth token.  That mechanism is documented to use the PLAIN SASL mechanism.  That seems entirely reasonable and a best current practice.  What would be the advantage to using a new SASL mechanism for this?  Using a nonstandard SASL mechanism would burden me, because it would mean that I would need to extend client libraries to deal with it.  Also, RFC 6648 discourages the use of "X" prefix for extensions to application protocols.

michalwski:
Thanks for the explanation @kenstir, this is reasonable. I understand that this approach makes things easier on client side. 

There are some small advantages of using new method:
1. We don't overuse PLAIN which should be used for authentication with password. As you reminded me it is already overused.
2. With new mechanism there is no need to implement `ejabberd_gen_auth` behavior. For this particular auth method it would be enough to implement just `cyrsasl_jwt`.

These are only technical details and I understand that your current approach is the easiest from client side perspective. I'll merge this PR as soon as conflicts in rebar.config are solved.

test/ejabberd_tests/tests/auth_jwt_SUITE.erl
```diff
@@ -0,0 +1,117 @@
...
+
+%%--------------------------------------------------------------------
+%% Suite configuration
+%%--------------------------------------------------------------------
+-define(USERNAME, <<"10857839">>).
+-define(JWT_KEY, <<"testtesttest">>).
+
+all() ->
+    Release = list_to_integer(erlang:system_info(otp_release)),
+    if
```
gadgetci: According to **Elvis**:
> Replace the 'if' expression on line 30 with a 'case' expression or function clauses.

test/ejabberd_tests/tests/auth_jwt_SUITE.erl
```diff
@@ -0,0 +1,117 @@
...
+    {error, _} = escalus_connection:start(ClientProps0,
+                                          [start_stream,
+                                           stream_features,
+                                           authenticate
+                                          ]),
+    ok.
+
+generate_token(NbfDelta) ->
+    Now = p1_time_compat:system_time(seconds),
+    Data = #{bookingNumber => ?USERNAME, exp => Now + 60, nbf => Now + NbfDelta, iat => Now},
```
gadgetci: According to **Elvis**:
> Line 115 is too long:     Data = #{bookingNumber => ?USERNAME, exp => Now + 60, nbf => Now + NbfDelta, iat => Now},.

test/ejabberd_tests/tests/auth_jwt_SUITE.erl
```diff
@@ -0,0 +1,117 @@
...
+end_per_testcase(CaseName, Config) ->
+    escalus:end_per_testcase(CaseName, Config).
+
+%%--------------------------------------------------------------------
+%% JSON Web Token tests
+%%--------------------------------------------------------------------
+
+auth_ok(_Config) ->
+    Password = generate_token(0),
+    ClientProps0 = [{username, ?USERNAME},
```
gadgetci: According to **Elvis**:
> The code in the following (LINE, COL) locations has the same structure: (87, 18), (101, 18).

apps/ejabberd/src/ejabberd_auth_jwt.erl
```diff
@@ -0,0 +1,214 @@
...
+
+-spec start(Host :: ejabberd:server()) -> ok.
+start(Host) ->
+    ok.
+
+-spec stop(Host :: ejabberd:server()) -> ok.
+stop(_Host) ->
+    ok.
+
+store_type(Server) ->
```
gadgetci: According to **Compiler**:
> Warning: variable 'Server' is unused

astro:
I just fixed the warnings by @gadgetci 

Regarding SASL mechanisms I completely agree with @kenstir. We're using JWT as a simple PLAIN password replacement, sparing us of any client-side changes.

astro:
More tests succeed now but a few are still failing. I need more guidance regarding the test suite modifications that came with 2.0.

michalwski:
Hi @astro,

Thanks for updating the PR. The issue with `auth_jwt_SUITE.erl` is that it's placed in wrong directory. After merging changes related to `rebar3` we moved `test` directory to `test.disabled`. These tests are not yet ready to by run by `rebar3 ct` and by default `rebar3 ct` runs all suites from all `test` directories.
In short, moving your `auth_jwt_SUITE.erl` from `test/ejabberd_tests/tests` to `test.disabled/ejabberd_tests/tests` should solve the issue.

michalwski:
The PR was closed by accident. I already reopened.

astro:
@michalslaski Thank you! I moved the file to `test.disabled/`

michalwski:
Thanks for the update @astro, we are almost there :) I can see there is a syntax error in `auth_jwt_SUITE`. More details here [1] and here [2]

[1]. https://travis-ci.org/esl/MongooseIM/jobs/204976755#L1190
[2]. https://github.com/esl/MongooseIM/pull/977/files#diff-9781b6b2f8e19d121f6a3bbe2247ed85R30

astro:
Oops, I may have done too much Haskell. :-) That one is pushed with Erlang syntax now.

michalwski:
One step closer :) There are still some issues [1], [2] with unbound `Password`

[1]. https://travis-ci.org/esl/MongooseIM/jobs/205757627#L1192
[2]. https://github.com/esl/MongooseIM/pull/977/files#diff-9781b6b2f8e19d121f6a3bbe2247ed85R110

test.disabled/ejabberd_tests/tests/auth_jwt_SUITE.erl
```diff
@@ -0,0 +1,121 @@
...
+     {ssl, false}
+    ].
+
+generate_token(NbfDelta) ->
+    Now = p1_time_compat:system_time(seconds),
+    Data = #{bookingNumber => ?USERNAME,
+             exp => Now + 60,
+             nbf => Now + NbfDelta,
+             iat => Now},
+    Token = jwerl:sign(Data, #{alg => <<"HS256">>, key => ?JWT_KEY}),
```
michalwski: There is no need to assign it to `Token` and returning. It's enough to have just
```erlang
jwerl:sign(Data, #{alg => <<"HS256">>, key => ?JWT_KEY}).
```

As the last line in this function.

astro:
Fixed. Thank you for your feedback.

astro:
I think Travis broke -_-

michalwski:
Yes, they had issues related to S3 yesterday. I've just restarted your build.

PR #982, opened 2016-09-13
HTTP client API - improve room users API

michalwski:
This PR goes on top of #973 

Proposed changes include:
* user's management move to path /room/:id/users
* id attribute to affiliation change messages initiated by the API

### DIFF
--- a/apps/ejabberd/src/mod_muc_light_admin.erl
+++ b/apps/ejabberd/src/mod_muc_light_admin.erl
@@ -208,21 +208,20 @@ is_subdomain(Child, Parent) ->
         {_, _} -> true
     end.
 
-iq(S, R, T, C) when is_binary(S),
-                    is_binary(R), is_binary(T), is_list(C) ->
+iq(To, From, Type, Children) ->
     UUID = uuid:uuid_to_string(uuid:get_v4(), binary_standard),
     #xmlel{name = <<"iq">>,
-           attrs = [{<<"from">>, S},
-                    {<<"to">>, R},
-                    {<<"type">>, T},
+           attrs = [{<<"from">>, From},
+                    {<<"to">>, To},
+                    {<<"type">>, Type},
                     {<<"id">>, UUID}],
-           children = C
+           children = Children
           }.
 
-query(NS, C) when is_binary(NS), is_list(C) ->
+query(NS, Children) when is_binary(NS), is_list(Children) ->
     #xmlel{name = <<"query">>,
            attrs = [{<<"xmlns">>, NS}],
-           children = C
+           children = Children
           }.
 
 affiliate(JID, Kind) when is_binary(JID), is_binary(Kind) ->
--- a/apps/ejabberd/src/mod_muc_light_admin.erl
+++ b/apps/ejabberd/src/mod_muc_light_admin.erl
@@ -210,10 +210,12 @@ is_subdomain(Child, Parent) ->
 
 iq(S, R, T, C) when is_binary(S),
                     is_binary(R), is_binary(T), is_list(C) ->
+    UUID = uuid:uuid_to_string(uuid:get_v4(), binary_standard),
     #xmlel{name = <<"iq">>,
            attrs = [{<<"from">>, S},
                     {<<"to">>, R},
-                    {<<"type">>, T}],
+                    {<<"type">>, T},
+                    {<<"id">>, UUID}],
            children = C
           }.
 
--- a/apps/ejabberd/src/mod_muc_light_admin.erl
+++ b/apps/ejabberd/src/mod_muc_light_admin.erl
@@ -25,7 +25,7 @@
 -export([create_unique_room/4]).
 -export([send_message/4]).
 -export([invite_to_room/4]).
--export([invite_to_room_id/4]).
+-export([change_affiliation/5]).
 
 -include("mod_muc_light.hrl").
 -include("ejabberd.hrl").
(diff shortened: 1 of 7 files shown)


test/ejabberd_tests/tests/rest_client_SUITE.erl
```diff
@@ -116,13 +119,46 @@ messages_can_be_paginated(Config) ->
...
+        Stanza = escalus:wait_for_stanza(Bob),
+        assert_aff_change_stanza(Stanza, Bob, <<"none">>)
+    end).
+
+owner_can_leave_a_room_and_auto_select_owner(Config) ->
+    escalus:fresh_story(Config, [{alice, 1}, {bob, 1}], fun(Alice, Bob) ->
+        RoomID = given_new_room_with_users({alice, Alice}, [{bob, Bob}]),
+        {{<<"204">>, _}, _} = remove_user_from_a_room({alice, Alice}, RoomID, Alice),
+        Stanza = escalus:wait_for_stanza(Bob),
+        ct:pal("~p", [Stanza]),
```
gadgetci: According to **Elvis**:
> Remove the debug call to ct:pal/2 on line 151.

test/ejabberd_tests/tests/rest_client_SUITE.erl
```diff
@@ -116,13 +119,46 @@ messages_can_be_paginated(Config) ->
...
+        RoomJID = <<RoomID/binary, "@muclight.localhost">>,
+        escalus:send(Alice, escalus_stanza:to(IQ, RoomJID)),
+        escalus:assert(is_iq_result, [IQ], escalus:wait_for_stanza(Alice))
+
+    end).
+
+user_is_removed_from_a_room(Config) ->
+    escalus:fresh_story(Config, [{alice, 1}, {bob, 1}], fun(Alice, Bob) ->
+        RoomID = given_new_room_with_users({alice, Alice}, [{bob, Bob}]),
+        {{<<"204">>, _}, _} = remove_user_from_a_room({alice, Alice}, RoomID, Bob),
```
gadgetci: According to **Elvis**:
> The code in the following (LINE, COL) locations has the same structure: (141, 29), (149, 29).

test/ejabberd_tests/tests/rest_client_SUITE.erl
```diff
@@ -116,13 +119,46 @@ messages_can_be_paginated(Config) ->
...
     escalus:fresh_story(Config, [{alice, 1}], fun(Alice) ->
         RoomID = given_new_room({alice, Alice}),
-        get_room_info({alice, Alice}, RoomID)
+        RoomInfo = get_room_info({alice, Alice}, RoomID),
+        assert_room_info(Alice, RoomInfo)
     end).
 
 user_is_invited_to_a_room(Config) ->
     escalus:fresh_story(Config, [{alice, 1}, {bob, 1}], fun(Alice, Bob) ->
         RoomID = given_new_room_with_users({alice, Alice}, [{bob, Bob}]),
```
gadgetci: According to **Elvis**:
> The code in the following (LINE, COL) locations has the same structure: (128, 16), (140, 16), (148, 16), (158, 16), (202, 16).

apps/ejabberd/src/mod_muc_light_admin.erl
```diff
@@ -127,14 +127,14 @@ invite_to_room(Domain, RoomName, Sender, Recipient0) ->
     ejabberd_router:route(S, R, iq(jid:to_binary(S), jid:to_binary(R),
```
gadgetci: According to **Elvis**:
> The code in the following (LINE, COL) locations has the same structure: (127, 20), (138, 20).

michalwski:
All the code duplicates where addressed by tuning the Elvis's DRY sensitivity.

apps/ejabberd/src/mod_muc_light_admin.erl
```diff
@@ -210,10 +210,12 @@ is_subdomain(Child, Parent) ->
 
 iq(S, R, T, C) when is_binary(S),
                     is_binary(R), is_binary(T), is_list(C) ->
+    UUID = uuid:uuid_to_string(uuid:get_v4(), binary_standard),
```
bartekgorny: I'd rather call it Uuid, to keep casing consistent.

apps/ejabberd/src/mod_muc_light_admin.erl
```diff
@@ -210,10 +210,12 @@ is_subdomain(Child, Parent) ->
 
 iq(S, R, T, C) when is_binary(S),
```
bartekgorny: Some more telling variable names would be useful.

apps/ejabberd/src/mod_muc_light_admin.erl
michalwski: I'm actually for UUID as it is not only var name but also shortcut for `Universal Unique Identifier` which in many places is referred as UUID. Also to me `UUID` is more readable then `Uuid`.

apps/ejabberd/src/mod_muc_light_admin.erl
michalwski: I agree, I forgot to refactor this function.

michalwski:
1. What do you consider code duplication here?
2. It's very easy to route multiple urls to the same module, but then we'll end up in request handlers where we have to check the path, and method and whatnot before we decide what to do. When we separate the whole rooms' functionality to smaller modules implementing only one logically small subset of functionality, the implementation is more understandable but a bit longer.

bartekgorny:
I mean modules like the `mongoose_client_api_rooms_users`, which mostly calls `mongoose_client_api_rooms` and adds little of its own logic. But it is a matter of taste. I'm a bit concerned about numer of modules, it is growing very fast, and there is already quite a lot of them.

michalwski:
This could be improved by using [inaka/mixer](https://github.com/inaka/mixer). I'd rather have more modules where the functionality is clearly split and each module is responsible for one logical subset.
In the end we can organise all the REST modules into their own subdirectory or even separate `mongooseim_rest` app in `apps` dir.

bartekgorny:
Sounds like a good idea. Anyway, as of today, that's all from me.

michalwski:
If so then let's merge the PR and move forward!

PR #984, opened 2016-09-14
Feature parallel mam tests

arcusfelis:
Refactoring MAM code and tests
- parallel MAM tests (and faster)
- more MAM test cases
- Split lookup_messages function
- Remove safe_ prefix from public callback functions
- Make some options configurable for MAM
- Prefs backend does not require integer user-id anymore.

Schema changes:
- Tweaking PRIMARY KEYs. Allow two users to have message ids with the same ID in SQL MAM backends. Because archive bootstrapping will not work otherwise in parallel version because of conflicts.

SM changes:
- Add sm_filter_offline_message to allow not to redeliver messages extracted from archive (it confuses some XMPP clients and breaks some tests).


apps/ejabberd/src/mod_mam_odbc_prefs.erl
```diff
@@ -133,20 +133,46 @@ set_prefs(_Result, Host, UserID, _ArcJID, DefaultMode, AlwaysJIDs, NeverJIDs) ->
         {error, Error}
     end.
 
+
+order_by_in_delete(Host) ->
+    case ejabberd_odbc:db_engine(Host) of
+        mysql ->
```
michalwski: 2 questions:

1. Why do we order only for mysql?
2. Wha is the benefit?

test/ejabberd_tests/tests/mam_SUITE.erl
```diff
@@ -425,166 +494,116 @@ do_init_per_group(C, ConfigIn) ->
...
+    Config;
+end_per_group(mam_metrics, Config) ->
+    Config;
+end_per_group(muc02, Config) ->
+    Config;
+end_per_group(muc03, Config) ->
+    Config;
+end_per_group(muc04, Config) ->
+    Config;
+end_per_group(muc_rsm_all, Config) ->
```
michalwski: I'd move this one just before the generic the last end_per_group function head. It will improve readability and show which groups end by returning `Config` only.

apps/ejabberd/src/mod_mam_odbc_prefs.erl
arcusfelis: Only mysql has support for this.
It orders locks to avoid dead locks in theory (in practice we still have them and should retry).

arcusfelis: At least two delete operations will not be deadlocked. But delete+insert will.

arcusfelis:
Caused by recent rebased.

```erlang
====== Test name: metric_incremented_on_archive_request
====== Reason:    {{badmatch,[]},
                   [{escalus_mongooseim,metric_type,1,
                        [{file,"src/escalus_mongooseim.erl"},{line,97}]},
                    {escalus_mongooseim,read_metric_initial_value,2,
                        [{file,"src/escalus_mongooseim.erl"},{line,57}]},
                    {lists,foldl,3,[{file,"lists.erl"},{line,1262}]},
                    {escalus_mongooseim,maybe_read_initial_metric_values,1,
                        [{file,"src/escalus_mongooseim.erl"},{line,35}]},
                    {escalus_story,story,3,
                        [{file,"src/escalus_story.erl"},{line,36}]},
                    {test_server,ts_tc,3,
                        [{file,"test_server.erl"},{line,1533}]},
                    {test_server,run_test_case_eval1,6,
                        [{file,"test_server.erl"},{line,1053}]},
                    {test_server,run_test_case_eval,9,
                        [{file,"test_server.erl"},{line,985}]}]}
```

PR #987, opened 2016-09-14
Feature parallel mam tests

arcusfelis:
Refactoring MAM code and tests
- parallel MAM tests (and faster)
- more MAM test cases
- Split lookup_messages function
- Remove safe_ prefix from public callback functions
- Make some options configurable for MAM
- Prefs backend does not require integer user-id anymore.

Schema changes:
- Tweaking PRIMARY KEYs. Allow two users to have message ids with the same ID in SQL MAM backends. Because archive bootstrapping will not work otherwise in parallel version because of conflicts.

SM changes:
- Add sm_filter_offline_message to allow not to redeliver messages extracted from archive (it confuses some XMPP clients and breaks some tests).

The RP is the same as https://github.com/esl/MongooseIM/pull/984 but pushed into ESL repo for rest reports.


arcusfelis:
http://mongooseim-ct-results.s3-website-eu-west-1.amazonaws.com/PR/987/2002/2/big/ct_run.test@testing-gce-59999286-20e3-4a10-ba1e-53e918648722.2016-09-14_16.02.21/ejabberd_tests.tests.mam_SUITE.logs/run.2016-09-14_16.08.23/mam_suite.metric_incremented_on_archive_request.html

Two ideas:
- metric is actually not incremented;
- histograms are not realtime (they are updated each 5 seconds).

apps/ejabberd/src/mod_mam_riak_timed_arch_yz.erl
```diff
@@ -174,8 +174,16 @@ lookup_messages_muc(Result, Host,
...
 
-archive_size(Size, _Host, _ArchiveID, _ArchiveJID) ->
-    Size.
+archive_size(_Size, _Host, _ArchiveID, ArchiveJID) ->
+    OwnerJID = ArchiveJID,
+    RemoteJID = undefined,
+    {MsgIdStartNoRSM, MsgIdEndNoRSM} =
+        calculate_msg_id_borders(undefined, undefined, undefined, undefined),
+    F = fun get_msg_id_key/3,
+    {TotalCount, _} = read_archive(OwnerJID, RemoteJID,
```
michalwski: What you need, here is only the parameter `Count` from [1]. This is returned from single call to Yokozuna. There is no need to further go through all results and fetch messages all we need here is only the count, which is available without reading whole messages.

[1]. https://github.com/esl/MongooseIM/blob/d1ce27d2ac9aade7bec5b079041f6975709cb7c1/apps/ejabberd/src/mod_mam_riak_timed_arch_yz.erl#L397

arcusfelis: I just implemented it using the same function, that is used to count messages in offset and index functions.
I guess we should edit them if we want to optimize something.
Also, I don't care about performance here: it's used from tests only.

michalwski:
Many thanks @arcusfelis!

PR #994, opened 2016-09-21
Improve s3 test archive

michalwski:
Proposed changes include:
* test logs are organized in directories containing PREST  
* the S3 URL is printed at the beginning of a test


michalwski:
Here you can find the improved directory structure: http://mongooseim-ct-results.s3-website-eu-west-1.amazonaws.com/?prefix=PR/994/2024/

michalwski:
This URL [1] (now printed to the log) is better as it goes to the S3 browser so it's easier to navigate and get what we really need from the archive.

[1]. http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/994/2032/internal_mnesia.18.3

PR #1003, opened 2016-09-27
REST API list my rooms

michalwski:
Proposed changes include:
* implemented REST API method to list all rooms user is connected to
* optimized swagger spec

### DIFF
--- a/apps/ejabberd/src/mongoose_client_api_rooms.erl
+++ b/apps/ejabberd/src/mongoose_client_api_rooms.erl
@@ -15,6 +15,7 @@
 
 -include("ejabberd.hrl").
 -include("jlib.hrl").
+-include("mod_muc_light.hrl").
 -include_lib("exml/include/exml.hrl").
 
 init(_Transport, _Req, _Opts) ->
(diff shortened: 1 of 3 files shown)


apps/ejabberd/src/mongoose_client_api_rooms.erl
```diff
@@ -81,7 +88,11 @@ to_json(Req, #{room := Room} = State) ->
              subject => proplists:get_value(subject, Config),
              participants => [user_to_json(U) || U <- Users]
             },
-    {jiffy:encode(Resp), Req, State}.
+    {jiffy:encode(Resp), Req, State};
+to_json(Req, #{jid := #jid{luser = User, lserver = Server}} = State) ->
+    Rooms = ?BACKEND:get_user_rooms({User, Server}),
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 93. Only modules that define callbacks should make dynamic calls.

gadgetci: According to **Elvis**:
> Don't use macros (like BACKEND on line 93) as module names.

michalwski:
Let's merge it!

PR #1004, opened 2016-09-27
[skip ci] Cleaned up and updated backend swagger doc

bartekgorny:
Factored out repeating fragments as $refs; updated return codes (204)


doc/http-api/backend_swagger.yml
```diff
@@ -1,16 +1,13 @@
-# Example YAML to get you started quickly.
-# Be aware that YAML has indentation based scoping.
-# Code completion support is available so start typing for available options.
 swagger: '2.0'
-
-# This is your document metadata
 info:
   version: "0.0.0"
```
michalwski: Looks like this is no longer version 0.0.0. This was shipped with beta2 release. It's time to upgrade the version.

PR #1005, opened 2016-09-27
Fix muclight hook call

bartekgorny:
Fixed invalid call to `filter_room_packet` hook - handlers for that hook have different arity.

The call never really makes it to handlers because it is made from muc_light, while the only handler for this hook is added by mam, but still it should be done right.

### DIFF
--- a/apps/ejabberd/include/mod_muc_light.hrl
+++ b/apps/ejabberd/include/mod_muc_light.hrl
@@ -121,3 +121,5 @@
                                   | {set, #config{}, AffUsers :: aff_users()}
                                   | {error, bad_request}.
 
+-type msg() :: #msg{}.
+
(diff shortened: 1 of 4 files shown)


apps/ejabberd/src/mod_muc_light_codec_legacy.erl
```diff
@@ -99,7 +107,7 @@ decode_message(#xmlel{ attrs = Attrs, children = Children }) ->
 -spec decode_message_by_type(Type :: {binary(), binary()} | false,
```
gadgetci: According to **Elvis**:
> The spec in line 107 uses a record, please define a type for the record and use that instead.

gadgetci:

apps/ejabberd/include/mod_muc_light.hrl
```diff
@@ -121,3 +121,6 @@
                                   | {set, #config{}, AffUsers :: aff_users()}
                                   | {error, bad_request}.
 
+-type msg() :: #msg{}.
+
+-type config() :: #config{}.
```
gadgetci: According to **Compiler**:
> type config() already defined

michalwski:
Is it possible to have a test case showing the problem? This issue happened only because we didn't have it tested.

PR #1011, opened 2016-09-30
Elvis utils

bartekgorny:
This PR addresses the need to clean the code before pushing, so that Elvis doesn't scream.

Proposed changes include:
* addidional make target "make rock", to check one file or all modified files
* a script to check all modified files


rock_changed.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
michalwski: All other helper scripts are stored in `tools` directory. Do you think this one can be moved there too?

PR #1013, opened 2016-10-03
Improve travis test script

michalwski:
This PR improves the travis-test.sh script by adding following options:
* `-p` to specify the PREST, default is `internal_mnesia`
* `-s` to specify if small tests should be run, default is `true`
* `-c` to specify is cover should be enabled on MongooseIM, default is `true`.

Example, in order to run only big tests without cover, following option can be used:

```bash
tools/travis-test.sh -p mysql_redis -s false -c false
```


ludwikbukowski:
im cool with the code
travis is cool with this preset configuration,
merging

PR #1021, opened 2016-10-07
REST API: extend the list of my rooms by name and subject

michalwski:
This PR adds name and subject to the action returning list of room to which user is subscribed.


apps/ejabberd/src/mongoose_client_api_rooms.erl
```diff
@@ -90,8 +90,18 @@ to_json(Req, #{room := Room} = State) ->
...
+    {jiffy:encode(lists:flatten(RoomsMap)), Req, State}.
+
+get_room_details({RoomID, _} = RoomUS) ->
+    case mod_muc_light_db_backend:get_config(RoomUS) of
+        {ok, Config, _} ->
+            #{id => RoomID,
+              name => proplists:get_value(roomname, Config),
+              subject => proplists:get_value(subject, Config)};
+        _ ->
+            []
```
michalwski: The empty list is correct. Thanks to the `lists:flatten` which run before encoding, such faulty rooms, if there are, will be filtered out.

ludwikbukowski: ok!

PR #1029, opened 2016-10-12
Elvis - check branch, new regexp

bartekgorny:
This PR does two things:
- option to check all files which were modified in a branch (run it before making PR)
- adds another variable name regexp, commented for now, which will be required when we install new version of elvis

### DIFF
--- a/elvis.config
+++ b/elvis.config
@@ -8,7 +8,10 @@
        ruleset => erl_files,
        rules => [{elvis_style, line_length, #{limit => 100,
                                               skip_comments => false}},
-		 {elvis_style, dont_repeat_yourself, #{min_complexity => 20}}
+                 {elvis_style,
+                  variable_naming_convention,
+                  #{regex => "^([_A-Z][0-9a-zA-Z]*)$" }},
+		         {elvis_style, dont_repeat_yourself, #{min_complexity => 20}}
 		]
       }
     ]
--- a/elvis.config
+++ b/elvis.config
@@ -10,11 +10,7 @@
                                               skip_comments => false}},
                  {elvis_style,
                   variable_naming_convention,
-                  #{regex => "^([_A-Z][0-9a-zA-Z]*)$" }},
-%%                provision for newer version of elvis_core which doesn't automagically
-%%                strip leading underscore
-%%                https://github.com/inaka/elvis_core/commit/70fd9d01fd076f4640da09fdfd53b310ec59a94c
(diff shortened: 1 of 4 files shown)


elvis.config
```diff
@@ -8,7 +8,14 @@
...
        rules => [{elvis_style, line_length, #{limit => 100,
                                               skip_comments => false}},
-		 {elvis_style, dont_repeat_yourself, #{min_complexity => 20}}
+                 {elvis_style,
+                  variable_naming_convention,
+                  #{regex => "^([_A-Z][0-9a-zA-Z]*)$" }},
+%%                provision for newer version of elvis_core which doesn't automagically
+%%                strip leading underscore
+%%                https://github.com/inaka/elvis_core/commit/70fd9d01fd076f4640da09fdfd53b310ec59a94c
+%%                  #{regex => "^_{0,2}([A-Z][0-9a-zA-Z]*)$"}},
```
bartekgorny: After some hesitation, I think the best way would be to simply state we use Elvis 0.3.2, and when 0.3.3. is out switch to it and modify config file accordingly.

michalwski: After another thought, do we know which Elvis version is used by @gadgetci?

bartekgorny: Yes We Can

michalwski:
One more question. I observed that ElvisCi is checking only the changes you've made in a PR, not whole changed files. Does the `rock` command work the same way or rather on whole files?

elvis.config
```diff
@@ -8,7 +8,10 @@
        ruleset => erl_files,
        rules => [{elvis_style, line_length, #{limit => 100,
                                               skip_comments => false}},
-		 {elvis_style, dont_repeat_yourself, #{min_complexity => 20}}
+                 {elvis_style,
+                  variable_naming_convention,
+                  #{regex => "^_{0,2}([A-Z][0-9a-zA-Z]*)$"}},
```
michalwski: Does it mean that we allow **usage** of vars starting with `_`?

bartekgorny:
`rock` checks the whole files in which you've made any changes. By doing this it forces you to fix coding style in old code too. Hopefully, over time, we'll fix coding style in the whole project.

elvis.config
bartekgorny: You mean single underscore, or double? Elvis by default allows single underscore, we allow double underscore for reasons already mentioned.

PR #1033, opened 2016-10-14
Use Rebar3

erszcz:
This PR changes the build system in use to Rebar3. **It's a work in progress!**

`rebar` is stubbed with a script that will shout that's it's not really Rebar and exit with an error, for fast failure in case of tools calling it directly. Travis integration is not touched at all - it won't work. Docker image build scripts might be broken, too.

`mongooseim` Erlang application and the corresponding `src/mongooseim.app.src` files do not exist at the moment - the `.app.src` file confused Rebar3 and made it behave as though it was operating in a single application project, not in a release-like project. Maybe it can be sidestepped, maybe it can't. Due to the lack of this app:
- it's not possible to include MongooseIM itself as a dependency in other projects
- `mongooseimctl status` code which displays the version had to be modified (version of the `ejabberd` app is taken if `mongooseim` app is not found)

Top-level `Makefile` was considerably slimmed down - only the basic rules were rewritten to use Rebar3. Honestly, I'd opt for minimizing the number of rules we have in this file and rely on directly calling `rebar3` in the development workflow - all in all, it's ought to be _the standard_ build tool.

Big tests are moved from `test/` to `test.disabled/`. `disabled` here means that they're disabled from Rebar3's point of view, i.e. `rebar3 ct` won't try to run them (while it would if they were in the `test/` subdirectory). The test runner we have there is custom-wired for our use case and the suites, unfortunately, can't just be picked up and run by Rebar3.

What works?
- `cd $REPO; make; cd _build/prod/rel/mongooseim; ./bin/mongooseimctl ( live | start | stop | status )`
- running small tests with `make ct`; some tests fail though and some are disabled
- running big tests with: `cd test.disabled/ejabberd_tests` and the same procedure as before
  (i.e. there are no Makefile rules to run big tests from the main repo directory), e.g. `make quicktest`; before running the tests, though, `rebar3 compile` has to be run at least once in the top-level of the repository - some deps are reused between small and big tests and have to be fetched and compiled properly

ToDo:
- [x] fix jiffy
- [x] reenable Riak: fix Protocol Buffers support (find the right Rebar3 plugin)
- [ ] check if `alarms` app works: does it start up properly with the correct config?
- [x] check zlib driver load path in `apps/ejabberd/test/zlib_driver_SUITE.erl` - this changed
    in master to something apparently more sensible, but does it really work with rebar3?
- [ ] readd `pubsub_load_SUITE` since it was removed due to amoc using rebar3?
- [x] switch to stringprep with merged rebar3/relx fixes: https://github.com/processone/stringprep/pull/4
- [x] readd Makefile targets which seem convenient?
- [ ] can we include the stub mongooseim app into the release again?
    having the `src/mongooseim.app.src` file in place triggers rebar3 "app layout",
    instead of the project release layout that we actually use
- [x] in Makefile reenable 3rd node needed for big tests
- [x] reenable small test suites (suffixed with `apps/ejabberd/test/*_SUITE.erl.disabled`)
- [ ] run big tests with rebar3 too?
- [x] we might need to change the dependency on mustache to bbmustache (the same as used by Rebar3) - it depends on where we actually use mustache
- [x] update relevant pages in doc
- [ ] `grep -i todo` ;)


erszcz:
Build status as of commit d9442d5 (though the commit hashes differ) is at https://travis-ci.org/erszcz/MongooseIM/builds/168232909. Tests run, but fail. Jiffy is not built properly.

michalwski:
All right, tests are passing!
There are still some disabled small tests and dialyzer is not happy yet.

apps/ejabberd/src/mod_version.erl
```diff
@@ -43,13 +43,13 @@ add_os_info(Host) ->
 		    children = [#xmlcdata{content = os_info()}]}];
 	_ ->
```
gadgetci: According to **Elvis**:
> Line 44 has a tab at column 0.

apps/ejabberd/src/mod_version.erl
```diff
@@ -43,13 +43,13 @@ add_os_info(Host) ->
...
 	    []
```
gadgetci: According to **Elvis**:
> Line 45 has a tab at column 0.

apps/ejabberd/src/mod_version.erl
```diff
@@ -26,40 +26,40 @@ process_iq(From, _To, #iq{type = get} = IQ) ->
...
+          [#xmlel{name = <<"query">>,
+                  attrs = [{<<"xmlns">>, ?NS_VERSION}],
+                  children =
+                  [#xmlel{name = <<"name">>, attrs = [],
+                          children =[#xmlcdata{content = Name}]},
+                   #xmlel{name = <<"version">>, attrs = [],
+                          children =[#xmlcdata{content = Version}]}
+                  ] ++ add_os_info(Host)}]}.
 
 -spec add_os_info(binary()) -> [#xmlel{}] | [].
```
gadgetci: According to **Elvis**:
> The spec in line 38 uses a record, please define a type for the record and use that instead.

apps/ejabberd/src/ejabberd_ctl.erl
```diff
@@ -154,7 +154,7 @@ process(["status"]) ->
                                  {logs, get_log_files()}])]),
     case MongooseStatus of
         not_running -> ?STATUS_ERROR;
-        {running, _Version} -> ?STATUS_SUCCESS
+        {running, _, _Version} -> ?STATUS_SUCCESS
```
gadgetci: According to **Elvis**:
> The variable "STATUS_SUCCESS" on line 157 does not respect the format defined by the regular expression '"^_{0,2}([A-Z][0-9a-zA-Z]*)$"'.

michalwski:
I consider this PR finished.

1. External integrations works again:
    * uploading tests results to S3
    * building docker image
    * uploading coverage report to coveralls.io
1. All small tests were fixed and they pass again.
1. All travis jobs were restored and pass. This means that things like riak client or cassandra works with rebar3 correctly.
1. Documentation was updated.

Next step could be:
* flatting the source structure as we only have ejabberd in apps now.
* rebar3-cation of test.disabled/ejabberd_tests

This should be done in a separate PR after this one is merged, IMO.

rebar.config
```diff
@@ -34,23 +29,115 @@
...
+
+                   {template, "rel/files/mongooseim", "bin/mongooseim"},
+                   {template, "rel/files/mongooseimctl", "bin/mongooseimctl"},
+                   {template, "rel/files/app.config", "etc/app.config"},
+                   {template, "rel/files/vm.args", "etc/vm.args"}
+                  ]}
+
+       ]}.
+
+{profiles, [ {default, [{relx, [ {overlay_vars, "rel/vars.config"},
```
lucafavatella: I would be surprised if rebar3 allowed specifying profile `default` in this point. I tried this in the past with rebar 3.2.0, and I aborted it; I rather moved away from profile `default` and I used an explicit profile named `dev` (as I was using `default` profile for development).

michalwski: Right, default doesn't seam to work for release. For "dev" we have more profiles: `mim1`, `mim2`, `mim3` and `fed1`.

test.disabled/ejabberd_tests/rebar.config
```diff
@@ -7,7 +7,7 @@
         {escalus, ".*", {git, "git://github.com/esl/escalus.git", "e1eca08218c21f1ac9d45431194e4463119efff6"}},
         {exml, ".*", {git, "git://github.com/esl/exml.git", "2.4.0"}},
```
bartekgorny: Could we change exml version to newer? 2.4.0 is giving compilation errors if certain flags are set in environment, newer versions are safe.

Makefile
```diff
@@ -1,104 +1,38 @@
...
-etc/ejabberd.cfg:
-	@mkdir -p $(@D)
-	tools/generate_cfg.es etc/ejabberd.cfg rel/files/ejabberd.cfg
+dev: $(DEVNODES)
 
-etc/app.config:
-	@mkdir -p $(@D)
-	tools/generate_cfg.es etc/app.config rel/files/app.config
+clean:
+	-rm -rf _build
```
bartekgorny: Previous way of doing 'clean' was much better - this one removes all deps sources and they have to be downloaded again.

Makefile
```diff
@@ -1,104 +1,38 @@
...
-	. ./configure.out && ./rebar clean
-
-quick_compile: rebar
-	./rebar $(OPTS) compile skip_deps=true > $(LOG)
-
-reload: quick_compile
-	@E=`ls ./rel/mongooseim/lib/ | grep ejabberd-2 | sort -r | head -n 1` ;\
-	rsync -uW ./apps/ejabberd/ebin/*beam ./rel/mongooseim/lib/$$E/ebin/ ;\
-
-reload_dev: quick_compile
```
michalwski: I agree, you are not the first one mentioning this. I'll try restore this.

bartekgorny: qct was useful too

bartekgorny:
I'm not sure we want to flatten the structure, maybe we should rather split it into more applications - there is 252 files in one dir, and growing, and it becomes a bit difficult to work with.

Makefile
erszcz: But you're sure that you're really starting from a clean slate as described by the version controlled files (`rebar.config` and `rebar.lock`) which is the purpose of `clean` - reproducibility! I agree, though, we could add a `distclean` which would do `rm -rf`, while clean would only remove `.beam` files.

bartekgorny:
I'd suggest adding a line "rm -rf test" somewhere in code, to make absolutely sure there are no leftlovers from previous version, they cause trouble and it is not easy to figure out what's wrong.

michalwski:
This is a thing you need to do once, so I'm not sure we should be it somewhere in the code. Rather I think there should be a doc describing what you should do if you update repo from rebar2 to rebar3.

Makefile
michalwski: qct is not possible I think. I didn't find a way to tell rebar "skip deps" but just run tests.

kzemek:
@bartekgorny 
> I'm not sure we want to flatten the structure, maybe we should rather split it into more applications

There's also an option of structuring source files into directories inside a single application - rebar doesn't mind and erlang doesn't care.

Makefile
michalwski: Regarding `reload_dev`, I'd like to hear your stories, how and what for do you use it now. Maybe we can do it better with rebar3.

bartekgorny:
> there should be a doc describing what you should do if you update repo from rebar2 to rebar3.

So here is what you have to do:
1) BEFORE you update, run:
`make devclean`
`make clean`
and remove everything from /deps/
2) update to rebar3
3) run
`git clean -fd`
then chances are it is going to work.

Makefile
erszcz: I checked this and

> dev nodes use symlinks to deps

is true, but

> and MongooseIM's beam file[s]

is not:

```
17:14:16 erszcz @ x4 : /tmp/mongooseim (rebar3)
$ ls -l _build/mim1/lib/ | head -5
total 352
lrwxr-xr-x  1 erszcz  wheel    49B Dec 21 17:12 base16@ -> /private/tmp/mongooseim/_build/default/lib/base16
lrwxr-xr-x  1 erszcz  wheel    53B Dec 21 17:12 bbmustache@ -> /private/tmp/mongooseim/_build/default/lib/bbmustache
lrwxr-xr-x  1 erszcz  wheel    47B Dec 21 17:13 bear@ -> /private/tmp/mongooseim/_build/default/lib/bear
lrwxr-xr-x  1 erszcz  wheel    52B Dec 21 17:12 cache_tab@ -> /private/tmp/mongooseim/_build/default/lib/cache_tab
17:14:20 erszcz @ x4 : /tmp/mongooseim (rebar3)
$ ls -l _build/mim1/lib/ejabberd/ebin/ | head -5
total 9200
-rw-r--r--  1 erszcz  wheel    89K Dec 21 17:13 ELDAPv3.beam
-rw-r--r--  1 erszcz  wheel   6.9K Dec 21 17:13 XmppAddr.beam
-rw-r--r--  1 erszcz  wheel    12K Dec 21 17:13 acl.beam
-rw-r--r--  1 erszcz  wheel   9.0K Dec 21 17:13 adhoc.beam
```

I agree that just rebuilding the release is enough when we work on just one of them. I've been following this workflow for some time now and it's not bad, though it's a bit slower than just recompiling and copying the files over (especially for multiple release).

However, if we are to reintroduce this target, it'd be nice if it checked which releases are actually present in `_build` and copy the files to all of them instead of being hardcoded to just `mim1` / `mongoose_node1` or all of them and failing when one is not there.

erszcz:
Actually, just:

```
git checkout rebar3
git clean -fd
```

should be enough.

Makefile
michalwski: By
> and MongooseIM's beam file[s]

I mean this:
```
ls -l _build/mim1/rel/mongooseim/lib/ejabberd-2.1.8+mim*
lrwxr-xr-x  1 michal.piotrowski  staff  87 Dec 21 16:45 _build/mim1/rel/mongooseim/lib/ejabberd-2.1.8+mim-2.0.0-322-g5af94960 -> /Users/michal.piotrowski/projects/MongooseIM-rebar3-michalwski/_build/mim1/lib/ejabberd
```

The ejabberd app inside the release is symlinked to ejabberd inside mim1 profile. I've just checked it and doing just:

```
./rebar3 as mim1 compile
```

Recompiles modules and puts them in `_build/mim1/lib/ejabberd`

```
ls -lt _build/mim1/lib/ejabberd/ebin | head -10
total 9304
-rw-r--r--  1 michal.piotrowski  staff    6860 Dec 21 17:29 ejabberd.app
-rw-r--r--  1 michal.piotrowski  staff    4952 Dec 21 17:29 ejabberd.beam
-rw-r--r--  1 michal.piotrowski  staff   55420 Dec 21 17:29 ejabberd_s2s_in.beam
-rw-r--r--  1 michal.piotrowski  staff   91928 Dec 21 17:29 ELDAPv3.beam
-rw-r--r--  1 michal.piotrowski  staff    7484 Dec 21 17:29 XmppAddr.beam
-rw-r--r--  1 michal.piotrowski  staff   50268 Dec 21 17:29 eldap.beam
-rw-r--r--  1 michal.piotrowski  staff   12176 Dec 21 16:45 acl.beam
-rw-r--r--  1 michal.piotrowski  staff    9460 Dec 21 16:45 adhoc.beam
-rw-r--r--  1 michal.piotrowski  staff   16772 Dec 21 16:45 amp.beam
```

This is still slower than reload_dev, but with rebar3 where there is no `skip_deps` (at least I'm not able to do this) we probably can't be faster than this.

I'm not against re-adding `reload_dev`, I'm just trying to find if we can achieve the same with what we have built-in into rebar3.

kzemek: I guess it now would be `./rebar3 as mim1,mim2,mim3,fed1 compile`, which is something I'd probably alias locally. I'm not sure if it's worthy of a target; on the other hand it reads like a generic `dev` target in Makefile, minus dependencies

michalwski: `./rebar3 as mim1,mim2,mim3,fed1 compile` creates build inside _build with options merged from all specified profiles. This will not work for just recompiling code and putting in relevant builds.

apps/ejabberd/rebar.config
```diff
@@ -16,11 +16,6 @@
                deprecated_function_calls,
                deprecated_functions]}.
 
-{ct_extra_params, "-pa test "
-                  "-ct_hooks ct_progress_hook "
```
kzemek: Nothing seems to use `apps/ejabberd/src/ct_progress_hook.erl` now; should we remove it? There's still a copy at `test.disabled/ejabberd_tests/src/ct_progress_hook.erl` for big tests.

apps/ejabberd/src/mod_version.erl
```diff
@@ -26,40 +26,40 @@ process_iq(From, _To, #iq{type = get} = IQ) ->
...
+            []
+    end.
 
 -spec mongoose_info() -> {binary(), binary()}.
 mongoose_info() ->
-    {ok, Version} = application:get_key(mongooseim, vsn),
-    {ok, Name} = application:get_key(mongooseim, description),
-    {list_to_binary(Name), list_to_binary(Version)}.
+    {ok, EjdMimVsn} = application:get_key(ejabberd, vsn), %returns 2.1.8+mim-*
+    Version = string:sub_string(EjdMimVsn, 11),
```
kzemek: Just a small comment, but I find `"2.1.8+mim-" ++ Version = EjdMimVsn` (or `<<"2.1.8+mim-", Version/binary>> = list_to_binary(EjdMimVsn)` as binary pattern matching is more common than string pattern matching and we need a binary Version anyway) more readable in that it doesn't require a comment and a magic `11`.

erszcz: Since both `ejabberd_ctl` and `mod_version` need this the code shouldn't be replicated - it could be extracted and go into `mongooseim:version/0` or similar. However, I also think this is not a functionally significant issue and a subject for another (tiny) PR - otherwise we'll never merge this ;)

apps/ejabberd/rebar.config
michalwski: Right, I removed it.

fenek:
https://github.com/esl/MongooseIM/blob/rebar3/doc/user-guide/Getting-started.md still has outdated release paths.

apps/ejabberd/src/ejabberd_app.erl
```diff
@@ -55,7 +55,6 @@ start(normal, _Args) ->
     gen_mod:start(),
     ejabberd_config:start(),
     ejabberd_check:config(),
-    maybe_start_alarms(),
```
fenek: Why have we gotten rid of alarms? Are they incompatible with rebar3?

apps/ejabberd/test/auth_external_SUITE.erl
```diff
@@ -24,7 +24,7 @@ all_tests() ->
     ].
 
 init_per_suite(C) ->
-    ok = application:ensure_started(stringprep),
```
fenek: Why `application:ensure_started` got removed?

apps/ejabberd/test/component_reg_SUITE.erl
```diff
@@ -10,10 +10,10 @@ all() ->
     [ registering, registering_with_local ].
 
 init_per_suite(C) ->
-    application:ensure_all_started(stringprep),
```
fenek: Why `stringprep` is being started without `application` help all of a sudden everywhere?

apps/ejabberd/src/ejabberd_app.erl
michalwski: 1. They were not used anywhere after removing "watchdog" sending warnings over XMPP.
2. Wombat does more or else the same.

michalwski:
> https://github.com/esl/MongooseIM/blob/rebar3/doc/user-guide/Getting-started.md still has outdated release paths.

good catch :D

apps/ejabberd/src/ejabberd_app.erl
fenek: :+1:

doc/developers-guide/Testing-MongooseIM.md
```diff
@@ -4,55 +4,57 @@ In shell #1:
 
 ```sh
 $ cd $MONGOOSEIM
+$ ./rebar3 compile
```
fenek: I'd prefer to have dedicated `make` target, such as obvious `make compile`

apps/ejabberd/test/auth_external_SUITE.erl
michalwski: I don't fully remember but probably because of `rebar3 ct` and it being more vulnerable. It causes test cases to fail in places where `rebar ct` didn't.
@erszcz do you remember why such change? this was introduced in this commit: https://github.com/esl/MongooseIM/commit/5ac4b9856cc0fd22dd7a0f6eec963edba12b6427

doc/developers-guide/Testing-MongooseIM.md
```diff
@@ -4,55 +4,57 @@ In shell #1:
...
 ```sh
-$ cd $MONGOOSEIM/dev/mongooseim_fed1
+$ cd $MONGOOSEIM/_build/fed1/rel
 $ ./bin/mongooseimctl live
 ```
 
 Back to shell #1:
 
 ```sh
+$ cd test.disabled/ejabberd_tests
```
fenek: What about moving `ejabberd_tests` contents to `test(.disabled)` at last?

rebar.config
```diff
@@ -34,23 +29,120 @@
   {seestar, ".*", {git, "git://github.com/arcusfelis/seestar.git", "e9700e1e7e8226173ad76d22be9d3a4e0ef77310"}},
 
   {cache_tab, ".*", {git, "git://github.com/processone/cache_tab", {tag, "1.0.4"}}},
-  {stringprep, ".*", {git, "git://github.com/processone/stringprep.git", {tag, "1.0.6"}}},
+  {stringprep, ".*", {git, "git://github.com/erszcz/stringprep.git", {branch, "list-deps"}}},
```
fenek: How is it different from P1 version? Can we expect them to merge a PR?

doc/developers-guide/Testing-MongooseIM.md
michalwski: Do you mean just `test` dir in project's root? This will confuse `rebar3`.
Ideally would be to do it as you suggest, that's probably out of the scope of this already not so small PR.

rebar.lock
```diff
@@ -0,0 +1,178 @@
+{"1.1.0",
```
fenek: Is this file supposed to be included in repo?

rebar.config
michalwski: Good catch, it was merged and I remember changing the dep to mainstream... Probably a rebase reverted it without me noticing it. I'll fix that.

doc/developers-guide/Testing-MongooseIM.md
fenek: Thou speaketh the truth.

rebar.lock
michalwski: rebar.lock? Definitely, it makes sure all deps are frozen, even if a dep of dep uses master branch.

fenek: 👍

michalwski:
> https://github.com/esl/MongooseIM/blob/rebar3/doc/user-guide/Getting-started.md still has outdated release paths.

@fenek what would you see changed in this file? Both `make` and `make rel` still works.

fenek:
@michalwski 

I.e. "From your MongooseIM directory, go to the release directory:  cd rel/mongooseim" :) Or "$REPO/rel/mongooseim/etc/ejabberd.cfg"

michalwski:
Right, thanks :)

michalwski:
The `Getting started` was updated.

PR #1027, opened 2016-10-12
add opts_for_module/2 and return opts on stop_module

sstrigler:
Admin love: 
- adds a function gen_mod:opts_for_module/2 which allows you to selectively retrieve opts passed through config to a given module
- changes stop_module_keep_config/2 and stop_module/2 to return those opts

PR #1034, opened 2016-10-14
Pubsub item publisher

sstrigler:
add boolean config option item_publisher

if set to true publishers of items will be stored with event and
broadcasted along according to

https://xmpp.org/extensions/xep-0060.html#publisher-publish-success-publisher

### DIFF
--- a/apps/ejabberd/include/pubsub.hrl
+++ b/apps/ejabberd/include/pubsub.hrl
@@ -153,6 +153,7 @@
     itemid                           ,% :: {mod_pubsub:itemId(), mod_pubsub:nodeIdx()},
     creation     = {unknown, unknown},% :: {erlang:timestamp(), jlib:ljid()},
     modification = {unknown, unknown},% :: {erlang:timestamp(), jlib:ljid()},
+    publisher    = undefined         ,% :: jlib:jid(),
     payload      = []                % :: mod_pubsub:payload()
 }).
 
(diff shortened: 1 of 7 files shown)


michalwski:
Hi @sstrigler,

Thanks for your PR! Will you have some time to modify tests? I see `pep_SUITE` and `pubsub_SUITE` are failing in all jobs on travis.

sstrigler:
Sorry, haven't been notified about your comment. Will look into this.

sstrigler:
This patch also depends on https://github.com/xsf/xeps/pull/267

So before that one is accepted and merged this here should not be merged either.

sstrigler:
rebased on current master, fixed bug and force pushed

sstrigler:
*ooops* force pushed another fix

dialyzer complained and pointed out another bug which was not covered by a test case. I decided to not go write a test for my new stuff since it would require quite some work as in require a whole new configuration to start the tests with. Or is there an easier way? Maybe just reload and start the module with a different configuration and then run the test in question?

I think in the long run it would be better to be able to configure nodes accordingly during run time. This would require even more changes to the XEP though. And I left this out for now.

michalwski:
@sstrigler I see your patch to XEP-0060 got merged and published.

Do you think you will find some time to write a simple test case? 

> Maybe just reload and start the module with a different configuration and then run the test in question?

Yes, reloading with changed config for mod_pubsub will work. You have some helper functions in `dynamic_modules` in `test/ejabberd_tests/tests`.

sstrigler:
Hu?

michalwski:
Yes, I merged this without tests. Feel free to add them at anytime.
Or you are not referring to lack of tests?

sstrigler:
We're there no more tests failing?

michalwski:
They are still failing in your other PR #1027. This one was green.

sstrigler:
Oh, lol, ok! Very well then. :p

Issue #1036, opened 2016-10-18
Make travis-test.sh usable from the command line


kenstir:
MongooseIM version: master
Installed from: source
Erlang/OTP version: 18.3

I want to call tools/travis-test.sh from the command line to test my build.  On CentOS 7, it exits non-zero (143 = terminated with SIGTERM) due to the way the `trap` command is set up.  I will submit a PR to fix this.

```
$ tools/travis-test.sh
...
CT results:
    835 tests passed
    0 tests failed
    25 tests skipped by user
    0 tests skipped automatically

All tests done.
Build succeeded
Terminated
$ echo $?
143
```

PR #1037, opened 2016-10-18
Fix for issue #1036 - Make travis-test.sh usable from the command line

kenstir:
This PR addresses #1036 

Proposed changes include:
* protect against unset environment variable
* ignore SIGTERM on EXIT trap, so that killing the process group does not kill bash itself


michalwski:
Thanks for the patch!

PR #1049, opened 2016-10-25
Re-organise Swagger spec for client REST API

michalwski:
This PR adds tags to swagger spec. This groups methods into logical parts.


michalwski:
Here's the swagger spec: http://mongooseim.readthedocs.io/en/improve-swagger-specs/swagger/index.html?client=true

Issue #1104, opened 2016-12-06
external components can't connect more than once over websocket


igors:
MongooseIM version: master 
Installed from: source
Erlang/OTP version: 18

External components can be configured to connect over websocket, i.e., using something like the following listener config:

```
{52755, ejabberd_cowboy, [
      {num_acceptors, 10},
      {max_connections, 1024},
      {modules, [
          {"_", "/ws-xmpp", mod_websockets, [{ejabberd_service, [
                                        {access, all},
                                        {password, "secret"}]}
          ]}
      ]}
  ]}
```

When such a component disconnects, it can never reconnect again since its entry in external_components table is never removed (i.e., the route is never unregistered, i.e., terminate/3 never called). Logged error message is

`[error] <0.750.0>@ejabberd_service:wait_for_handshake:222 Error in component handshake: route_already_exists`

PR #1105, opened 2016-12-07
Fix external component disconnect over websocket

igors:
This PR addresses #1104

Proposed changes include:
* ejabberd_service now monitors the (cowboy web)socket process and terminates itself when the monitor is triggered. This causes the route to be unregistered so the external component will be accepted without conflict when it connects again.

This is effectively equivalent to ejabberd_c2s behavior - it also uses the monitor to detect that web socket is closed.

### DIFF
--- a/apps/ejabberd/src/ejabberd_service.erl
+++ b/apps/ejabberd/src/ejabberd_service.erl
@@ -54,6 +54,7 @@
 
 -record(state, {socket,
                 sockmod      :: ejabberd:sockmod(),
+                socket_monitor,
                 streamid,
                 password     :: binary(),
                 host         :: binary(),
@@ -156,8 +157,10 @@ init([{SockMod, Socket}, Opts]) ->
                  _ -> true
              end,
     SockMod:change_shaper(Socket, Shaper),
+    SocketMonitor = SockMod:monitor(Socket),
     {ok, wait_for_stream, #state{socket = Socket,
                                  sockmod = SockMod,
+                                 socket_monitor = SocketMonitor,
                                  streamid = new_id(),
                                  password = Password,
                                  access = Access,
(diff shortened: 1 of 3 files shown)


michalwski:
Thanks for your contribution. It's very valuable. Any chance you can add a test case for this somewhere in [component_SUITE](https://github.com/esl/MongooseIM/blob/3ebec10b62e3b845e27da49a0f57d70447865d3d/test/ejabberd_tests/tests/component_SUITE.erl)?

A generic test case for TCP (group xep0114_tcp) and WS (group xep0014_ws) connections would be perfect. In this test case you could close the connection abnormally (by f.e killing relevant escalus process)  and reconnecting again. With your fix the reconnection should work.

igors:
So I tried to add a test case as you suggested. Unfortunately, escalus_ws:kill is not implemented (or, rather, implemented to call error/2) so calling it triggers a test failure; I tried just calling exit(SocketPid, kill) on websocket process but that's somehow caught by the test framework and reported as a test failure with "testcase_aborted_or_killed" reason.

igors:
I also tried changing implementation of `escalus_ws:kill` to just return `{stop, normal, ok, S}` - that seems to work but
a) I'm not entirely sure if this is the best way to kill an escalus websocket connection; e.g., this still sends a websocket close, which is fine for this test case but you might want `kill` to just blindly close the socket
b) you'd need to pull that change into escalus and repoint rebar.config to use the version with that fix.

igors:
Just to give you an idea of what I had in mind:

The change to escalus_ws: https://github.com/igors/escalus/commit/bd06c03402328fafe87dd22ab32e25bfeb95f580
The change to component_SUITE: https://github.com/igors/MongooseIM/commit/b3371ab64a7dc214ad8ed5deac307398f70192ca

Let me know if I am completely off the track or there is a better way to do that.

michalwski:
Thanks for the update @igors. The change in the `component_SUITE` is exactly what I meant. Regarding `escalus_ws` - I didn't remember the `kill` function is not implemented. You are right saying that it'd be better to blindly close the connection without sending `close`. This would required changing the websocket library - `wsecli`. I think we can live with this for now.

Could you open a PR to escalus? after I merge it you can add your test cases and change the reference to escalus.

igors:
Done - travis will probably keep failing until https://github.com/esl/escalus/pull/131 is pulled in. Should I just point to that particular commit in rebar.config or wait until there is a new escalus release?

michalwski:
You can point to the merge commit esl/escalus@e1eca08218c21f1ac9d45431194e4463119efff6 in rebar.config

igors:
Done - rebased the last commit to include the change to escalus version in rebar.config.

michalwski:
Thanks a lot @igors!

PR #1111, opened 2016-12-08
[skip ci] Platform schema

Nyco:
This PR adds the platform schema


doc/README.md
```diff
@@ -36,6 +36,9 @@ Check out our test results:
 
 ## MongooseIM platform components
 
+<img src="MongooseIM Platform components.png" alt="MongooseIM platform schema" />
```
michalwski: When I read the diagram from top to bottom, the natural "thinking" for me is that the "Android Mangosta" block is built from "iOS XMPPFramework" and "Android Smack". The very similar flow I have for the other block "iOS Mangosta". Maybe it'd be more natural if you put blocks related to Android only under "Android Mangosta" and the same for "iOS Mangosta".

Also I can see that block representing XMPP and XMPP libraries have the same background. Probably it'b good to have the same background for all REST-related blocks, now they have a bit different background.

Nyco: Of course! What did I do... :'(

michalwski:
Do you want to add the same to the README in `doc`?

Also I suggest using `_` (underscore) instead of ` ` (space) in file names.

michalwski:
Thanks for the file name change. How about similar change in `doc/README.md`? Do want it there?

README.md
```diff
@@ -36,6 +36,8 @@ Check out our test results:
 
 ## MongooseIM platform components
 
+<img src="MongooseIM_Platform_components.png" alt="MongooseIM platform schema" />
```
michalwski: Did you try "View" button on this file? The image has wrong path here.

michalwski: Ping, @Nyco did you get my question?

PR #1114, opened 2016-12-09
Add traffic stanza hooks

bernardd:
This change adds two hooks as close as possible to the incoming/outgoing stanza points which include the IP/port of the client. We use these for logging traffic at a low level (just above the XML encoding/decoding) to diagnose errors that may cause the packets not to make it to the higher level hooks.

There's also a fix for a related type error in ejabberd_c2s.hrl.

### DIFF
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -881,6 +881,11 @@ session_established({xmlstreamelement, El}, StateData) ->
         _NewEl ->
             NewState = maybe_increment_sm_incoming(StateData#state.stream_mgmt,
                                                    StateData),
+            ejabberd_hooks:run(stanza_sent,
+                               FromJID#jid.lserver,
+                               [StateData#state.jid,
+                                StateData#state.ip,
+                                El]),
             case mod_amp:check_packet(El, FromJID, initial_check) of
                 drop -> fsm_next_state(session_established, NewState);
                 NewEl -> process_outgoing_stanza(NewEl, NewState)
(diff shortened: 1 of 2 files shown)


michalwski:
Hi @bernardd, thanks for this PR. I understand this is important and useful in your case.

I'm a bit sceptical about it regarding mainstream MongooseIM. Introduced hooks are bound to regular TCP connection while in MongooseIM we can also have WebSockets and BOSH. They both use the `c2s` process but the IP in such cases may not be set accordingly. More generic, though less efficient solution, would be to read the IP of a user from its session (ejabberd_sm) module on `user_recevie_packt` or `user_send_packet` hooks.

bernardd:
Hi @michalslaski - no worries - thanks for the feedback. I'll take a look at using those two hooks tomorrow - a single mnesia lookup to get the IP (which, if I recall off the top of my head, is what would be required to pull it from the session manager) isn't the end of the world performance-wise. If I can achieve what I want without modifying MIM itself, so much the better.

bernardd:
Okay, I remember the reason I didn't go down that path now. For the case of `user_send_packet`, the hook isn't called for error packets and otherwise unhandled ones, both of which are very handy to have in a logged stream when you're trying to figure out something that's not working :)
For `user_receive_packet`, packets generated with the `broadcast` system were excluded.
Also, in both cases I wanted the remote IP.
In both cases I didn't want to change the behaviour of existing hooks because it would have touched far more places.

If you'd like me to look at generalising the IP field to make it more compatible with the websockets and BOSH cases I'm more than happy to do so. Otherwise, I won't be offended if you decide this isn't something that should be in the mainstream :)

PR #1118, opened 2016-12-12
XEP-0198: ack pep messages

msantos:
Events sent using PEP should be included in the last stanza handled
counts.

The problem occurs when attempting to connect to mongooseim with
mod_caps/mod_pubsub enabled using the Conversations XMPP client.
The connection fails with a "Policy Violation" error. MongooseIM logs
a policy violation stanza with "h attribute too big".

### DIFF
--- a/test/ejabberd_tests/tests/pep_SUITE.erl
+++ b/test/ejabberd_tests/tests/pep_SUITE.erl
@@ -29,7 +29,8 @@ groups() -> [
                pep_caps_test,
                publish_test,
                notify_test,
-               send_caps_after_login_test
+               send_caps_after_login_test,
+               h_ok_after_notify_test
               ]
              }
             ].
@@ -54,12 +55,12 @@ end_per_suite(Config) ->
 
 init_per_group(_GroupName, Config) ->
     dynamic_modules:ensure_modules(domain(), required_modules()),
-    Users = escalus_users:get_users([alice, bob]),
+    Users = escalus_users:get_users([alice, bob, kate]),
     escalus:create_users(Config, Users),
     escalus_story:make_everyone_friends(Config, Users).
 
 end_per_group(_GroupName, Config) ->
-    escalus:delete_users(Config, escalus_users:get_users([alice, bob])),
+    escalus:delete_users(Config, escalus_users:get_users([alice, bob, kate])),
     ok.
 
 init_per_testcase(pep_caps_test, Config) ->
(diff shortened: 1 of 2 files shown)


michalwski:
That is very valuable, thanks @msantos! Is it possible for you to invest some time in writing a test case for that? Such a test case could be added in `pep_SUITE.erl` inside `test/ejabberd_tests/tests`. If you want to see how to enable stream management on the client side (of tests) you can take a look at `sm_SUITE` (the same directory).

test/ejabberd_tests/tests/pep_SUITE.erl
```diff
@@ -128,6 +130,41 @@ send_caps_after_login_test(Config) ->
...
 
+h_ok_after_notify_test(ConfigIn) ->
+    Config = escalus_users:update_userspec(ConfigIn, alice,
+                                           stream_management, true),
+    escalus:story(
+        [{escalus_overrides, [
+            {initial_activity, {?MODULE, send_initial_presence_with_caps}}]} |
+             Config
+         ],
+        [{alice,1}, {bob,1}],
```
gadgetci: According to **Elvis**:
> Missing space right "," on line 141

gadgetci: According to **Elvis**:
> Missing space right "," on line 141

test/ejabberd_tests/tests/pep_SUITE.erl
```diff
@@ -128,6 +130,41 @@ send_caps_after_login_test(Config) ->
                 Bob, <<"item2">>, {escalus_utils:get_short_jid(Alice), ?NS_USER_TUNE}, [])
```
gadgetci: According to **Elvis**:
> Line 130 is too long:                 Bob, <<"item2">>, {escalus_utils:get_short_jid(Alice), ?NS_USER_TUNE}, []).

msantos:
Amended commit with a test case and force pushed. I frankenstein'ed the test from pep_SUITE:notify_test/1 and sm_SUITE:h_ok_after_a_chat/1, so if it can be improved, please let me know and I will try to fix it.

michalwski:
Thanks for the test case. From what I understand, the most important change is in line 1164 in ejabberd_c2s - `send_and_maybe_buffer_stanza` instead of just `send_element`. This line is not covered though by tests [1] which probably means that the test case is not testing your fix. Does your test case fail while run on MongooseIM without your fix?

[1]. https://coveralls.io/builds/9244943/source?filename=apps%2Fejabberd%2Fsrc%2Fejabberd_c2s.erl#L1164

msantos:
@michalwski you are correct. I will make another attempt at this ...

test/ejabberd_tests/tests/pep_SUITE.erl
```diff
@@ -128,6 +129,28 @@ send_caps_after_login_test(Config) ->
...
```
gadgetci: According to **Elvis**:
> Line 129 is too long:                 Bob, <<"item2">>, {escalus_utils:get_short_jid(Alice), ?NS_USER_TUNE}, []).

msantos:
New test pushed. The test was verified to fail when the patch is not applied.

About the failing checks:
* elvis is complaining about a line that was not changed by the patch. I may have accidentally reformatted the line in an earlier version of the patch. Let me know if I should fix it.

* travis and the compiler check are failing because of a problem with rebar

michalwski:
Thanks for the update @msantos, this is what we needed. I restarted the build on travis and as soon as it finishes I merge your PR.

PR #1121, opened 2016-12-13
Updated XEP list in docs

fenek:
* Reorganised XEPs table to XEP-per-row format
* Added PubSub, PEP, Blocking Commands
* Removed irrelevant (i.e. client-side) XEPs
* Updated Compliance Suites entries

### DIFF
--- a/doc/user-guide/Features-and-supported-standards.md
+++ b/doc/user-guide/Features-and-supported-standards.md
@@ -1,5 +1,4 @@
-Features and supported standards
---------------------------------
+# Features and supported standards
 
 * XMPP Core: [RFC 3920](https://tools.ietf.org/html/rfc3920),
 [RFC 6120](https://tools.ietf.org/html/rfc6120)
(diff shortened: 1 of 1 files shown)


Nyco:
We should also add all non-XSF-backed extensions:
* our own extensions, such as MUC light, Token-Based Reconnection
* third party, Instant Stream Resumption (soon)

fenek:
OK, will add MUC Light and Token-Based *Authentication*.

Nyco:
If we have PubSub, then we have PEP, then we have all the PEP-based User *, correct?

* [XEP-0080: User Location](https://xmpp.org/extensions/xep-0080.html)
* [XEP-0084: User Avatar](https://xmpp.org/extensions/xep-0084.html)
* [XEP-0107: User Mood](https://xmpp.org/extensions/xep-0107.html)
* [XEP-0108: User Activity](https://xmpp.org/extensions/xep-0108.html)
* [XEP-0118: User Tune](https://xmpp.org/extensions/xep-0118.html)
* [XEP-0172: User Nickname](https://xmpp.org/extensions/xep-0172.html)

fenek:
I don't agree we should show client-side XEPs built on PEP. They don't require any special implementation on server-side.

fenek:
Updated. :)

mnemonicflow:
@Nyco Because MongooseIM supports PubSub and PEP, does that mean that all the XEPs you mentioned above, if already implemented by an XMPP client library, should work out of the box?

fenek:
@mnemonicflow 

Yes. :) Some of them may require enabling in config file, as not all extensions are on by default.

PR #1122, opened 2016-12-13
Blocking cmds fix

bartekgorny:
This PR addresses an issue with blocking commands - according to XEP-0191:

"If the user attempts to send an outbound stanza to the JID, the user's server MUST NOT route the stanza to the JID"

therefore, we should block outgoing messages. Also, a block should be propagated to all user's resources, which is not explicitly required by the XEP, but suggested by common sense, plus required if we want to effectively block via REST API.

### DIFF
--- a/apps/ejabberd/src/mod_privacy.erl
+++ b/apps/ejabberd/src/mod_privacy.erl
@@ -337,13 +337,25 @@ check_packet(_, User, Server,
             check_packet_aux(List, PType, Type, LJID, Subscription, Groups)
     end.
 
+%% allow error messages
 check_packet_aux(_, message, <<"error">>, _JID, _Subscription, _Groups) ->
     allow;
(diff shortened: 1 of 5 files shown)


apps/ejabberd/src/mod_privacy.erl
```diff
@@ -337,14 +337,26 @@ check_packet(_, User, Server,
...
     end.
 
+%% allow error messages
 check_packet_aux(_, message, <<"error">>, _JID, _Subscription, _Groups) ->
     allow;
-check_packet_aux(_, message_out, _Type, _JID, _Subscription, _Groups) ->
-    allow;
-check_packet_aux([], _PType, _Type, _JID, _Subscription, _Groups) ->
+%% if we run of of list items then it is allowed
+check_packet_aux([], PType, MType, _JID, _Subscription, _Groups) ->
```
gadgetci: According to **Compiler**:
> Warning: variable 'MType' is unused

PR #1125, opened 2016-12-15
udate lager and lager-syslog deps

michalwski:
This PR updates `lager` (to 3.2.4) and `lager_syslog` (3.0.3) to newest versions.

Previously used `lager_syslog` (3.0.1) stopped working because used `Vagabond/erlang-syslog` which stopped working with rebar2. Because of that future builds will fail without this fix.


michalwski:
I merge this on my own. Without this it's not possible to build MongooseIM now.

PR #1133, opened 2016-12-23
Improve cassandra MAM backend performance

rslota:
This PR changes cassandra's client library to cqerl which is better supported by community then seestar. Also there are some changes in cassandra DB schema in order to improve query performance.

### DIFF
--- a/apps/ejabberd/priv/cassandra.cql
+++ b/apps/ejabberd/priv/cassandra.cql
@@ -14,7 +14,14 @@ CREATE TABLE mam_message(
   // - empty
   with_jid varchar,
   message blob,
-  PRIMARY KEY ((user_jid), with_jid, id));
+  PRIMARY KEY ((user_jid, with_jid), id)) WITH CLUSTERING ORDER BY (id DESC);
+
+CREATE TABLE mam_message_offset(
+  id bigint,
+  user_jid varchar,
+  with_jid varchar,
+  offset bigint,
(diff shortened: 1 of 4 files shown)


apps/ejabberd/src/mongoose_cassandra_worker.erl
```diff
@@ -9,34 +9,14 @@
...
--export([prepared_queries/0]).
+-export([start_link/1]).
 
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
+-export([queue_length/1]).
 
 -behaviour(gen_server).
 -behaviour(mongoose_cassandra).
```
gadgetci: According to **Compiler**:
> Warning: undefined callback function prepared_queries/0 (behaviour 'mongoose_cassandra')

.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: erlang
+dist: trusty
```
michalwski: Why do we need this?

rslota: Presets from "matrix" by default run on 12.04 which is unable to compile one of the new dependencies due to lack of C++11 support in GCC.

michalwski: I see. Sounds like a bug in travis though.

rslota:
For just the Cassandra client change, please see #1145 .

michalwski:
Thanks!

PR #1136, opened 2016-12-28
Feature/full text search

rslota:
This PR introduces MVP for MAM full text search. Currently it works only with ODBC and Riak backends and its very limited when comes to text matching. This is NOT an extension to the standard. Current MAM XEP-0313 allows for queries using "forms" that may contain any server-defined fields, therefore simple field named 'full-text-search' was introduced in order to provide this feature. 

The following stanza shows a example MAM stanza that uses full text search:
```
<iq type='set' id='query1'>
  <query xmlns='urn:xmpp:mam:1'>
    <x xmlns='jabber:x:data' type='submit'>
      <field type='hidden' var='FORM_TYPE'>
        <value>urn:xmpp:mam:1</value>
      </field>
      <field type='text-single' var='full-text-search'>
        <value>cat and dog</value>
      </field>
    </x>
  </query>
</iq>
```

Full Text Search is an optional server-side feature. To find out whether its supported, client may query server about active form fields for MAM stanza:
```
<iq type='get' id='form1'>
  <query xmlns='urn:xmpp:mam:1'/>
</iq>
```  

The response shows which form fields may be used to filter MAM messages (that includes required fields i.e. with, start and end):
```
<iq type='result' id='form1'>
  <query xmlns='urn:xmpp:mam:1'>
    <x xmlns='jabber:x:data' type='form'>
      <field type='hidden' var='FORM_TYPE'>
        <value>urn:xmpp:mam:1</value>
      </field>
      <field type='jid-single' var='with'/>
      <field type='text-single' var='start'/>
      <field type='text-single' var='end'/>
      <field type='text-single' var='full-text-search'/>
    </x>
  </query>
</iq>
```

While reading PR you may want to skip "Make Elvis happy" commit to get rid of code quality changes.


apps/ejabberd/src/mod_mam_odbc_arch.erl
```diff
@@ -133,7 +143,7 @@ stop_pm(Host) ->
 %% ----------------------------------------------------------------------
 %% Add hooks for mod_mam_muc
 
--spec start_muc(ejabberd:server(),_) -> 'ok'.
+-spec start_muc(ejabberd:server(), _) -> 'ok'.
 start_muc(Host, _Opts) ->
     case gen_mod:get_module_opt(Host, ?MODULE, no_writer, false) of
```
michalwski: This is probably not needed here, am I right?

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -141,13 +149,13 @@ microseconds_to_now(MicroSeconds) when is_integer(MicroSeconds) ->
 
 %% @doc Returns time in `now()' format.
 -spec iso8601_datetime_binary_to_timestamp(iso8601_datetime_binary())
-        -> erlang:timestamp() | undefined.
+                                          -> erlang:timestamp() | undefined.
 iso8601_datetime_binary_to_timestamp(DateTime) when is_binary(DateTime) ->
     jlib:datetime_binary_to_timestamp(DateTime).
```
michalwski: A word (or more) of an explanation will makes things easier to understand. A simple example may also help.

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -216,47 +224,44 @@ is_archived_elem_for(_, _) ->
...
+                         delete_archived_elem(By, Packet)).
 
 
 -spec replace_x_user_element(FromJID :: ejabberd:jid(), Role :: mod_muc:role(),
-                             Affiliation :: mod_muc:affiliation(), jlib:xmlel()
-                             ) -> jlib:xmlel().
+                             Affiliation :: mod_muc:affiliation(), jlib:xmlel()) -> jlib:xmlel().
 replace_x_user_element(FromJID, Role, Affiliation, Packet) ->
     append_x_user_element(FromJID, Role, Affiliation,
-    delete_x_user_element(Packet)).
```
michalwski: That's clever :) I like it!

elvis.config
```diff
@@ -27,7 +27,7 @@
                  %% Allow usage of ?assertMatch and similar macros, since after expanding they
                  %% result in use of variables like __X and __V that wouldn't be permitted by
                  %% default regex.
-                 {elvis_style, variable_naming_convention, #{regex => "^(_?[A-Z][0-9a-zA-Z_]*)$"}}
+                 {elvis_style, variable_naming_convention, #{regex => "^(_?[A-Z][0-9a-zA-Z]*)$"}}
```
michalwski: I think we do need `_` in this regexp. Otherwise Elvis may complain about macros like `?A_MACRO_WITH_WORDS`, am I right? We have such macros in tests.

rslota: I'm gonna revert it. That was an merge issue.

apps/ejabberd/src/mod_mam_utils.erl
rslota: I'm not sure what this refers to, considering that i didn't even touch this function ;)

bartekgorny: Your IDE seems to have touched it.

apps/ejabberd/src/mod_mam_odbc_arch.erl
```diff
@@ -228,29 +239,32 @@ do_archive_message(_Result, Host, MessID, UserID,
     SDir = encode_direction(Dir),
     SRemLResource = ejabberd_odbc:escape(RemLResource),
     Data = packet_to_stored_binary(Packet),
+    TextBody = packet_to_search_body(Host, Packet),
+    STextBody = ejabberd_odbc:escape(TextBody),
+    string:to_lower(STextBody),
```
michalwski: This is probably not needed here, am I right?

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -668,10 +680,28 @@ form_field(Type, VarName) ->
...
+%% Text search tokenization
+
+-spec normalize_search_text(binary() | string() | undefined) -> string() | undefined.
+normalize_search_text(Text) ->
+    normalize_search_text(Text, "%").
+
+-spec normalize_search_text(binary() | string() | undefined, string()) -> string() | undefined.
+normalize_search_text(undefined, _WordSeparator) ->
+    undefined;
+normalize_search_text(Text, WordSeparator) ->
```
michalwski: A word (or more) of an explanation will makes things easier to understand. A simple example may also help.

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -817,24 +847,20 @@ maybe_previous_id(X) ->
...
+       length(MessageRows) =:= PageSize ->
     %% Number of messages on skipped pages from the beginning plus the current page
     PagedCount = Offset + PageSize,
-    if
-        TotalCount =:= PagedCount ->
-            true; %%
-        true ->
-            false %% full page but not the last one in the result set
-    end;
+    TotalCount =:= PagedCount; %% false means full page but not the last one in the result set
```
michalwski: That's clever :) I like it!

apps/ejabberd/src/mod_mam_odbc_arch.erl
rslota: Yes, it's redundant. I'm gonna remove it.

test/ejabberd_tests/tests/mam_SUITE.erl
```diff
@@ -917,6 +978,153 @@ simple_archive_request(Config) ->
         end,
     escalus_fresh:story(Config, [{alice, 1}, {bob, 1}], F).
 
+text_search_is_not_available(Config) ->
```
gadgetci: According to **Elvis**:
> The code in the following (LINE, COL) locations has the same structure: (981, 1), (1008, 1).

apps/ejabberd/src/mod_mam.erl
```diff
@@ -544,66 +570,68 @@ handle_set_message_form(
...
+            %% Forward messages
+            {FirstMessID, LastMessID} =
+                case MessageRows of
+                    []    -> {undefined, undefined};
+                    [_|_] -> {message_row_to_ext_id(hd(MessageRows)),
+                              message_row_to_ext_id(lists:last(MessageRows))}
+                end,
+            [send_message(ArcJID, From, message_row_to_xml(MamNs,
+                                                           set_client_xmlns_for_row(M), QueryID))
+             || M <- MessageRows],
```
kzemek: This seems misformatted a little bit

apps/ejabberd/src/mod_mam.erl
```diff
@@ -544,66 +570,68 @@ handle_set_message_form(
...
+            %% Forward messages
+            {FirstMessID, LastMessID} =
+                case MessageRows of
+                    []    -> {undefined, undefined};
+                    [_|_] -> {message_row_to_ext_id(hd(MessageRows)),
+                              message_row_to_ext_id(lists:last(MessageRows))}
+                end,
+            [send_message(ArcJID, From,
+                          message_row_to_xml(MamNs, set_client_xmlns_for_row(M), QueryID))
+             || M <- MessageRows],
```
kzemek: This seems misformatted a little bit

apps/ejabberd/src/mod_mam_muc.erl
```diff
@@ -300,6 +304,19 @@ is_action_allowed(Action, From, To = #jid{lserver = Host}) ->
         default -> is_action_allowed_by_default(Action, From, To)
     end.
 
+-spec packet_to_search_body(Host :: ejabberd:server(), Packet :: xmlel()) ->
+    string().
+packet_to_search_body(Host, Packet) ->
```
kzemek: Do we need to have this new function duplicated between `mod_mam.erl` and `mod_mam_muc.erl`? It sounds like a good fit for `mod_mam_utils`.

apps/ejabberd/src/mod_mam_muc.erl
```diff
@@ -300,6 +304,19 @@ is_action_allowed(Action, From, To = #jid{lserver = Host}) ->
...
+    case has_full_text_search(Host) of
+        true ->
+            BodyValue = xml:get_tag_cdata(xml:get_subtag(Packet, <<"body">>)),
+            normalize_search_text(BodyValue, " ");
+        false -> ""
+    end.
+
+-spec has_full_text_search(Host :: ejabberd:server()) -> boolean().
+has_full_text_search(Host) ->
```
kzemek: Same, looks like a good fit for `mod_mam_utils`.

apps/ejabberd/src/mod_mam_muc.erl
```diff
@@ -845,6 +872,9 @@ form_to_end_microseconds(El) ->
 form_to_with_jid(El) ->
     maybe_jid(form_field_value_s(El, <<"with">>)).
 
+-spec form_to_text(_) -> 'undefined' | binary().
+form_to_text(El) ->
```
kzemek: Another common function that's a good candidate for `mod_mam_utils`.

apps/ejabberd/src/mod_mam_riak_timed_arch_yz.erl
```diff
@@ -18,29 +18,38 @@
...
 -export([key/3]).
 
 %% For tests only
--export([create_obj/3, read_archive/6, bucket/1,
+-export([create_obj/5, read_archive/7, bucket/1,
          list_mam_buckets/0, remove_bucket/1]).
 
+%% Text search
+-import(mod_mam_utils, [
+    normalize_search_text/2
```
kzemek: Do we need to import this?

apps/ejabberd/src/mod_mam_riak_timed_arch_yz.erl
```diff
@@ -223,22 +234,35 @@ archive_message(MessID, LocJID, RemJID, SrcJID, Packet) ->
...
 
-    Ops = [{{<<"msg_id">>, register},
+create_obj(Host, MsgId, SourceJID, Packet, Type) ->
+    ModMAM =
+        case Type of
+            pm -> mod_mam;
+            muc -> mod_mam_muc
+        end,
+    BodyValue = list_to_binary(ModMAM:packet_to_search_body(Host, Packet)),
+%%    BodyValue = xml:get_tag_cdata(xml:get_subtag(Packet, <<"body">>)),
```
kzemek: A stray comment

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -456,28 +462,28 @@ encode_jids(JIDs) ->
...
     #xmlel{
-        name = <<"fin">>,
-        attrs = [{<<"xmlns">>, MamNs}]
-                ++ [{<<"complete">>, <<"true">>} || IsComplete]
-                ++ [{<<"stable">>, <<"false">>} || not IsStable]
-                ++ [{<<"queryid">>, QueryID} || is_binary(QueryID), QueryID =/= undefined],
-        children = [ResultSetEl]}.
+       name = <<"fin">>,
+       attrs = [{<<"xmlns">>, MamNs}]
+       ++ [{<<"complete">>, <<"true">>} || IsComplete]
```
kzemek: I feel like formatting in this case makes the code less readable.

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -456,28 +462,28 @@ encode_jids(JIDs) ->
...
 make_fin_element(MamNs, IsComplete, IsStable, ResultSetEl) ->
     #xmlel{
-        name = <<"fin">>,
-        attrs = [{<<"xmlns">>, MamNs}]
-                ++ [{<<"complete">>, <<"true">>} || IsComplete]
-                ++ [{<<"stable">>, <<"false">>} || not IsStable],
-        children = [ResultSetEl]}.
+       name = <<"fin">>,
+       attrs = [{<<"xmlns">>, MamNs}]
+       ++ [{<<"complete">>, <<"true">>} || IsComplete]
```
kzemek: I feel like formatting in this case makes the code less readable.

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -641,19 +647,25 @@ find_field([], Name) ->
...
+    TextSearch =
+        case Mod:has_full_text_search(Host) of
+            true -> [form_field(<<"text-single">>, <<"full-text-search">>)];
+            false -> []
+        end,
     [form_type_field(MamNs),
      form_field(<<"jid-single">>, <<"with">>),
      form_field(<<"text-single">>, <<"start">>),
-     form_field(<<"text-single">>, <<"end">>)].
+     form_field(<<"text-single">>, <<"end">>)] ++ TextSearch.
```
kzemek: `| TextSearch]`? :)

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -668,6 +680,34 @@ form_field(Type, VarName) ->
...
+
+-spec normalize_search_text(binary() | string() | undefined, string()) -> string() | undefined.
+normalize_search_text(undefined, _WordSeparator) ->
+    undefined;
+normalize_search_text(Text, WordSeparator) ->
+    BodyString = unicode:characters_to_list(Text),
+    LowerBody = string:to_lower(BodyString),
+    ReOpts = [{return, list}, global, unicode, ucp],
+    Re0 = re:replace(LowerBody, "[, .:;-?!]+", " ", ReOpts),
+    Re1 = re:replace(Re0, "([^\\w\\d ]+)|(^\\s+)|(\\s+$)", "", ReOpts),
```
kzemek: The `\d` is redundant here as `\w` group already includes digits.

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -668,6 +680,34 @@ form_field(Type, VarName) ->
...
+%% @end
+%% -----------------------------------------------------------------------
+-spec normalize_search_text(binary() | string() | undefined) -> string() | undefined.
+normalize_search_text(Text) ->
+    normalize_search_text(Text, "%").
+
+-spec normalize_search_text(binary() | string() | undefined, string()) -> string() | undefined.
+normalize_search_text(undefined, _WordSeparator) ->
+    undefined;
+normalize_search_text(Text, WordSeparator) ->
```
kzemek: This function has to do some heavy lifting on texts, have you measured the performance impact of <full-text-search on>?

apps/ejabberd/src/mod_mam.erl
rslota: Why? There is function call in first two lines and then we have generator aligned with beginning of lists comprehension.

apps/ejabberd/src/mod_mam.erl
rslota: In my opinion it should be formatted like line 590.

apps/ejabberd/src/mod_mam_utils.erl
rslota: That's not the same :) I want lists concatenation, not insertion.

kzemek: I know it's not the same in general, but isn't it more idiomatic (and marginally faster) in this case?

apps/ejabberd/src/mod_mam.erl
kzemek: I understand why it's formatted how it is; I was comparing to the original version which looked much better. Probably the comprehension's body is just too complicated for good formatting.

rslota: Maybe it looked better, but the line was to long. Thats the only reason I've changed the formatting.

rslota: I'm gonna use `lists:foreach` instead, I think it's gonna look slightly better.

apps/ejabberd/src/mod_mam_utils.erl
rslota: It adds like 100 μs to processing time per row on my slow CPU. Simple `string:to_lower` is only 10 times faster then this whole function. Also, I've checked that precompiling regexes only speed up this function by ~5%. Removing only one of those 3 regexes could give also no more then 5-10%, but I'm not sure if its easy to do.

Issue #1082, opened 2016-11-20
Redis and MongooseIM going crazy


andrewvmail:
MongooseIM version: 1.6.2-957-gb13807d
Installed from: source
Erlang/OTP version: 18

Hello starting few days ago my lab installation of MongooseIM is running at 70% and its chatting with redis non stop for some reason. I tcpdump the traffic and just dump the pcap into vim and below is a slice of what i see in vim. Been running half year and just starting to see this.

No one is connected to it... and both redis and mongoose is running heavy.

Any ideas?

```
M<+ÿÿÿÿÿÿÿÿ-Mac OS X 10.11.5, build 15F34 (Darwin 15.5.0);Dumpcap (Wireshark) 1.99.8 (v1.99.8-0-ga0c561a from master)\ÿÿ-	-Mac OS X 10.11.5, build 15F34 (Darwin 15.5.0)\À±AFH^ðÌÂ×ÌEy@@
º
§H »¢§>8xÂÇ9Ö
¦	Òh)Ù×*2
$8
SMEMBERS
$64
s3:u17785555555:momoterramobile.com:FEA757CE3F89D301479347923582361

À±AÝJ^ðhhÌÌÂ×EZÕ^@@*ë

º »§H8xÂ¢§!v,
h)Ùò¦	Ò*1
$282
hdsessionhhbÇbOb	Ägdmongooseim@localhost+2hmu17785555555mmomoterramobile.commFEA757CE3F89D301479347923582361hmu17785555555mmomoterramobile.comd	undefinedlhdauth_modulemhdccd	undefinedhdconndunknownhdiphhaÏaÂaaÏbíÏj
À±AÓS^ðÌÂ×ÌE|¶@@^
º
` »PwçxÓÈ9Ö
¦	Õh)ÙÙ*2
$8
SMEMBERS
$64
s3:u17785555555:momoterramobile.com:FEA757CE3F89D301479347923582361

À±AU^ðhhÌÌÂ×EZh	@@@

º »`ÓPwçÓ`µ
h)Ùõ¦	Õ*1
$282
hdsessionhhbÇbOb	Ägdmongooseim@localhost+2hmu17785555555mmomoterramobile.commFEA757CE3F89D301479347923582361hmu17785555555mmomoterramobile.comd	undefinedlhdauth_modulemhdccd	undefinedhdconndunknownhdiphhaÏaÂaaÏbíÏj
À±AZ^ðÌÂ×ÌEÈ4@@8à
º
[...]

PR #1141, opened 2017-01-04
Fix race condition in ejabberd_sm

michalwski:
This PR addresses #1082

When the race condition could happen? Two processes (f.e. `c2s` and process `A` ) operates on the same `user`'s session. At the same time, process `A` updates `user`'s session info and `c2s` deletes the info because the `user` disconnected. Function `ejabberd_sm:store_info` (called by `A`) first checks if there is any record for given `user` and only then updates (sets new value). It's possible that the record is deleted by `c2s` just after the get returns to `A` but set is not yet started. In this situation the info about the `user` will remain in the table, even though the `user` disconnected.

Following PR fixes this bug by running all session updates in the context of c2s process.

It was found and fixed by @arcusfelis.

### DIFF
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -42,7 +42,7 @@
          get_subscribed/1,
          send_filtered/5,
          broadcast/4,
-         store_info/5]).
+         store_session_info/5]).
 
 %% gen_fsm callbacks
 -export([init/1,
@@ -140,8 +140,8 @@ broadcast(FsmRef, Type, From, Packet) ->
 stop(FsmRef) ->
     ?GEN_FSM:send_event(FsmRef, closed).
 
-store_info(FsmRef, User, Server, Resource, KV) ->
-    FsmRef ! {store_info, User, Server, Resource, KV, self()}.
+store_session_info(FsmRef, User, Server, Resource, KV) ->
+    FsmRef ! {store_session_info, User, Server, Resource, KV, self()}.
 
 %%%----------------------------------------------------------------------
 %%% Callback functions from gen_fsm
@@ -1203,7 +1203,7 @@ handle_info(check_buffer_full, StateName, StateData) ->
             fsm_next_state(StateName,
                            StateData#state{stream_mgmt_constraint_check_tref = undefined})
     end;
-handle_info({store_info, User, Server, Resource, KV, _FromPid}, StateName, StateData) ->
+handle_info({store_session_info, User, Server, Resource, KV, _FromPid}, StateName, StateData) ->
     ejabberd_sm:store_info(User, Server, Resource, KV),
     fsm_next_state(StateName, StateData);
 handle_info(Info, StateName, StateData) ->
@@ -2404,7 +2404,7 @@ bounce_messages() ->
         {route, From, To, El} ->
             ejabberd_router:route(From, To, El),
             bounce_messages();
-        {store_info, User, Server, Resource, KV, _FromPid} ->
+        {store_session_info, User, Server, Resource, KV, _FromPid} ->
             ejabberd_sm:store_info(User, Server, Resource, KV)
     after 0 ->
               ok
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -41,7 +41,8 @@
          get_subscription/2,
          get_subscribed/1,
          send_filtered/5,
-         broadcast/4]).
+         broadcast/4,
+         store_info/5]).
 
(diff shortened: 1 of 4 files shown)


apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -139,6 +140,9 @@ broadcast(FsmRef, Type, From, Packet) ->
 stop(FsmRef) ->
     ?GEN_FSM:send_event(FsmRef, closed).
```
> Don't use macros (like GEN_FSM on line 141) as module names.

> Remove the dynamic function call on line 141. Only modules that define callbacks should make dynamic calls.

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -139,6 +140,9 @@ broadcast(FsmRef, Type, From, Packet) ->
...
 
+store_info(FsmRef, User, Server, Resource, KV) ->
+    FsmRef ! {store_info, User, Server, Resource, KV, self()}.
```
bartekgorny: Why not use gen_fsm:send_all_state_event?

apps/ejabberd/test/ejabberd_sm_SUITE.erl
```diff
@@ -33,7 +33,8 @@ end_per_suite(C) ->
     application:stop(exometer_core).
 
 groups() ->
-    [{mnesia, [], tests()},
+    [{mnesia, [],
+      [cannot_reproduce_race_condition_in_store_info]},
```
fenek: Should be `tests()`, right? :)

apps/ejabberd/test/ejabberd_sm_SUITE.erl
```diff
@@ -203,8 +205,31 @@ kv_can_be_updated_for_session(C) ->
     when_session_info_stored(U, S, R, {key2, override}),
 
     [#session{sid = Sid, info = [{key1, val1}, {key2, override}]}]
-     = ?B(C):get_sessions(U,S).
-
+     = ?B(C):get_sessions(U, S).
+
+cannot_reproduce_race_condition_in_store_info(C) ->
```
fenek: I think it deserves being in separate group.

apps/ejabberd/test/ejabberd_sm_SUITE.erl
michalwski: Right, my mistake.

apps/ejabberd/test/ejabberd_sm_SUITE.erl
michalwski: That's good idea.

apps/ejabberd/src/ejabberd_c2s.erl
michalwski: We could use `?GEN_FSM:send_all_state_event`. C2s is not regular gen_fsm.

fenek: Right, didn't notice that Bartek wrote `gen_fsm` instead of `?GEN_FSM`.

bartekgorny: and then Elvis will say "don't use macros as module names"... maybe we better leave it as it is.

fenek: Elvis should stick to singing.

kzemek: As a somewhat relevant side note, I really don't see the value of the `GEN_FSM` macro. I'd vote for `s/?GEN_FSM/p1_fsm/g`, making the code more obvious and Elvis happy.

fenek: @kzemek It's something we have inherited from ProcessOne. You'll get used to it. :)

apps/ejabberd/test/ejabberd_sm_SUITE.erl
michalwski: I finally decided not to move this single test case to a separate group. It's just addition to `mnesia` group.

apps/ejabberd/src/ejabberd_c2s.erl
michalwski: In the end I left it as is - plain `!` instead of `?GEN_FSM:send_all_state_event`.

apps/ejabberd/test/ejabberd_sm_SUITE.erl
fenek: Logically, it's not Mnesia-exclusive issue, right?

michalwski: Finally I modified the test case to work with both backends. Now it's run in both groups and there is no need to create new one.

Issue #888, opened 2016-07-21
Expose more SSL options in cowboy


igor1:
MongooseIM version: current master
Installed from: source
Erlang/OTP version: OTP 18

There are a bunch of potentially useful SSL options that are not exposed by cowboy config. The ones that seem particularly useful/important are "cacerfile" (needed to send the full certificate chain to the client, which is pretty much required if your cert is signed by an intermediate CA), "ciphers" - allows to limit the set of supported ciphers and "versions" - limits SSL/TLS versions supported by the server.

BTW, it'd be potentially nice to upgrade to a newer ranch version - it now exposes "sni_hosts" option, which could be really useful for environments serving more than one domain.

michalwski:
> BTW, it'd be potentially nice to upgrade to a newer ranch version - it now exposes "sni_hosts" option, which could be really useful for environments serving more than one domain.

What ranch version do you have in mind?

igors:
I believe sni_hosts was added in ranch 1.2.0, meaning that it's pulled in by cowboy in master only - it'll probably make it into 2.0 but it's not in any of their betas yet...

igors:
While on this topic, would you consider a patch that treats all SSL parameters as opaque and just passes them all the way down to ranch? That's pretty much what cowboy does anyway. I guess one could still keep the currently defined set of common http/https options explicit (i.e., ip, port, etc) for backwards compatibility. I.e., something like the following:

```
{ 5285, ejabberd_cowboy, [
  {num_acceptors, 10},
  {max_connections, 1024},
  {ssl, [
      {certfile, "priv/ssl/fake_cert.pem"},
      {keyfile, "priv/ssl/fake_key.pem"},
      {password, "secret"},
      {cacertfile, "priv/ssl/fake_cert_chain.pem"},
      {dhfile, "priv/ssl/dhparam2048.pem"},
      {versions, ['tlsv1.1', 'tlsv1.2']},
      {hibernate_after, 120000}
  ]},
  {modules, [
      {"_", "/http-bind", mod_bosh},
      {"_", "/ws-xmpp", mod_websockets, [
            %% Uncomment to enable connection dropping or/and server-side pings
            %{timeout, 600000}, {ping_rate, 60000}
      ]}
      %% Uncomment to serve static files
      %{"_", "/static/[...]", cowboy_static,
      % {dir, "/var/www", [{mimetypes, cow_mimetypes, all}]}
      %},
      ]}
]}
```

michalwski:
[...]

PR #1142, opened 2017-01-06
Upgrade to ranch 1.3.0

igors:
This PR addresses #888 

The ranch version pulled in by cowboy is quite outdated and does not expose a number of useful ssl parameters (dh/dhfile being one of the more critical ones). As of 1.3.0 ranch blacklists ssl_options it doesn't like (rather than whitelisting).

I've ran some tests with the upgraded ranch and everything seems to be working fine.

Proposed changes include:
* add up-to-date ranch dependency to mongooseim's rebar.config


michalwski:
Thanks for your contribution @igors!

PR #1145, opened 2017-01-10
Use cqerl as cassandra client library

rslota:
This PR changes Cassandra backend to use cqerl client library instead seestar.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: erlang
+dist: trusty
 sudo: required
 addons:
     apt:
--- a/apps/ejabberd/src/mod_mam_cassandra_arch.erl
+++ b/apps/ejabberd/src/mod_mam_cassandra_arch.erl
@@ -193,8 +193,8 @@ message_to_params(#mam_message{
                      with_jid   = BWithJID,
                      message    = BPacket
                     }) ->
-    [{id, MessID}, {user_jid, BLocJID}, {from_jid, BSrcJID},
-     {remote_jid, BRemJID}, {with_jid, BWithJID}, {message, BPacket}].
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: erlang
+dist: trusty
```
fenek: Why Trusty specifically?

michalwski: Here's the answer: https://github.com/esl/MongooseIM/pull/1133#discussion_r94373464 :)

fenek: Merci. :)

apps/ejabberd/src/mod_mam_cassandra_arch.erl
```diff
@@ -501,13 +492,14 @@ from_id(ID, Filter = #mam_ca_filter{start_id = AfterID}) ->
...
     [row_to_uniform_format(Row) || Row <- MessageRows].
 
-row_to_uniform_format([MessID, BSrcJID, Data]) ->
-    SrcJID = unserialize_jid(BSrcJID),
-    Packet = stored_binary_to_packet(Data),
-    {MessID, SrcJID, Packet}.
+row_to_uniform_format(#{from_jid := FromJID, message := Msg, id := Id}) ->
+    SrcJID = unserialize_jid(FromJID),
+    Packet = stored_binary_to_packet(Msg),
+    MsgID = Id,
```
michalwski: Do we need `MsgID`? Can we simply return `Id` or assign `id` from the input map to `MsgID`?

rslota: No, it's just an artefact of refactoring. It'll be gone after next push.

michalwski:
Thanks!

PR #1158, opened 2017-01-18
Fix Elvis errors in ejabberd_c2s and introduce mongoose_transport

fenek:
Proposed changes include:
* All Elvis errors are fixed in `ejabberd_c2s` except for the one about module having more than 32 functions. It can't be resolved without more serious refactoring and breaking this file into several smaller ones.
* Added `mongoose_transport` behaviour. It's not final yet, because it will be improved when `ejabberd_receiver` and/or `ejabberd_listener` get rocked too.

I'm aware that `mongoose_transport` should take care of maintaining unified state with transport module name inside but it would involve touching even more modules. Again, this will get fixed while rocking other stuff.


apps/ejabberd/src/mod_websockets.erl
```diff
@@ -248,6 +252,7 @@ get_sockmod(_SocketData) ->
 close(#websocket{pid = Pid}) ->
     Pid ! close.
 
+-spec peername(#websocket{}) -> mongoose_sockmod:peername_return().
```
gadgetci: According to **Elvis**:
> The spec in line 255 uses a record, please define a type for the record and use that instead.

apps/ejabberd/src/mod_websockets.erl
```diff
@@ -224,6 +226,7 @@ reset_stream(#websocket{pid = Pid} = SocketData) ->
     Pid ! reset_stream,
     SocketData.
 
+-spec send_xml(#websocket{}, mongoose_sockmod:send_xml_input()) -> ok.
```
gadgetci: According to **Elvis**:
> The spec in line 229 uses a record, please define a type for the record and use that instead.

apps/ejabberd/src/mod_websockets.erl
```diff
@@ -239,6 +242,7 @@ send(#websocket{pid = Pid}, Data) ->
 change_shaper(SocketData, _Shaper) ->
     SocketData. %% TODO: we ignore shapers for now
 
+-spec monitor(#websocket{}) -> reference().
```
gadgetci: According to **Elvis**:
> The spec in line 245 uses a record, please define a type for the record and use that instead.

apps/ejabberd/src/mod_bosh_socket.erl
```diff
@@ -1,6 +1,7 @@
 -module(mod_bosh_socket).
```
gadgetci: According to **Elvis**:
> This module has too many functions (34). Consider breaking it into a number of modules.

apps/ejabberd/src/ejabberd_socket.erl
```diff
@@ -27,6 +27,8 @@
 -module(ejabberd_socket).
 -author('alexey@process-one.net').
 
+-behaviour(mongoose_sockmod).
+
 %% API
 -export([start/4,
 	 connect/3,
```
gadgetci: According to **Elvis**:
> Line 34 has a tab at column 0.

michalwski:
Thanks!

PR #1169, opened 2017-01-25
Post rebar3 fixes

michalwski:
This PR fix some initial bugs related to switch from rebar2 to rebar3.

### DIFF
--- a/apps/ejabberd/src/adhoc.erl
+++ b/apps/ejabberd/src/adhoc.erl
@@ -29,6 +29,7 @@
 -xep([{xep, 50}, {version, "1.2"}]).
 -export([parse_request/1,
          produce_response/2,
+         produce_response/4,
          produce_response/1]).
 
 -include("ejabberd.hrl").
@@ -87,7 +88,9 @@ find_xdata_el1([_ | Els]) ->
 %% @doc Produce a <command/> node to use as response from an adhoc_response
 %% record, filling in values for language, node and session id from
 %% the request.
--spec produce_response(request(), response()) -> #xmlel{}.
+-spec produce_response(request(), response() | atom()) -> #xmlel{}.
+produce_response(Request, Status) when is_atom(Status) ->
+    produce_response(Request, Status, <<>>, []);
 produce_response(#adhoc_request{lang = Lang,
                                 node = Node,
                                 session_id = SessionID},
@@ -95,6 +98,18 @@ produce_response(#adhoc_request{lang = Lang,
     produce_response(Response#adhoc_response{lang = Lang,
                                              node = Node,
                                              session_id = SessionID}).
+
+%% @doc Produce a <command/> node to use as response from an adhoc_response
+%% record, filling in values for language, node and session id from
+%% the request.
+-spec produce_response(request(), Status :: atom(), DefaultAction :: binary(),
+                       Elements :: [jlib:xmlel()]) -> jlib:xmlel().
+produce_response(Request, Status, DefaultAction, Elements) ->
+    #adhoc_request{lang = Lang, node = Node, session_id = SessionID} = Request,
+    produce_response(#adhoc_response{lang = Lang, node = Node, session_id = SessionID,
+                                     status = Status, default_action = DefaultAction,
+                                     elements = Elements}).
+
 
 %% @doc Produce a <command/> node to use as response from an adhoc_response
 %% record.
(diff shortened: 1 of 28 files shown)


apps/ejabberd/src/ejabberd_s2s_in.erl
```diff
@@ -63,9 +63,9 @@
                 tls_required = false  :: boolean(),
                 tls_certverify = false :: boolean(),
                 tls_options = []      :: [{_, _}],
-                server                :: ejabberd:server(),
+                server                :: ejabberd:server() | undefined,
                 authenticated = false :: boolean(),
-                auth_domain           :: binary(),
+                auth_domain           :: binary() | undefined,
                 connections = ?DICT:new(),
```
gadgetci: According to **Elvis**:
> Don't use macros (like DICT on line 69) as module names.

apps/ejabberd/src/mod_mam_muc_cassandra_arch.erl
```diff
@@ -57,7 +57,7 @@
           room_jid :: binary(),
           nick_name :: binary(),
           with_nick :: binary(),
-          message :: binary()
+          message :: binary() | undefined
```
michalwski: @rslota, do we still need to add `undefined` since you changed the way the record is created in line 175?

apps/ejabberd/src/mongoose_commands.erl
```diff
@@ -230,17 +230,17 @@ list(U) ->
     list(U, any, any, any).
 
 %% @doc List commands, available for this user, filtered by category.
--spec list(caller(), atom()) -> [t()].
+-spec list(caller(), binary() | any) -> [t()].
```
michalwski: @bartekgorny, `any` as atom? Do we expect atom `any` as value for the `C` parameter?

apps/ejabberd/src/mod_mam_muc_cassandra_arch.erl
rslota: Yes, I still haven't added 'message' field since it just isn't used there.

apps/ejabberd/src/mongoose_commands.erl
bartekgorny: yes, that is what list/1 provides

michalwski: Ok, I see. Thanks!

apps/ejabberd/src/mod_admin_extra_vcard.erl
```diff
@@ -178,7 +178,7 @@ get_module_resource(Server) ->
 get_vcard_content(User, Server, Data) ->
     [{_, Module, Function, _Opts}] = ets:lookup(sm_iqtable, {?NS_VCARD, Server}),
     JID = jid:make(User, Server, list_to_binary(get_module_resource(Server))),
-    IQ = #iq{type = get, xmlns = ?NS_VCARD},
+    IQ = #iq{type = get, xmlns = ?NS_VCARD, sub_el = []},
     %% TODO: This may benefit from better type control
     IQr = Module:Function(JID, JID, IQ),
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 183. Only modules that define callbacks should make dynamic calls.

apps/ejabberd/src/mod_admin_extra_vcard.erl
```diff
@@ -208,7 +208,7 @@ set_vcard_content(U, S, D, SomeContent) when is_binary(SomeContent) ->
 set_vcard_content(User, Server, Data, ContentList) ->
     [{_, Module, Function, _Opts}] = ets:lookup(sm_iqtable, {?NS_VCARD, Server}),
     JID = jid:make(User, Server, <<>>),
-    IQ = #iq{type = get, xmlns = ?NS_VCARD},
+    IQ = #iq{type = get, xmlns = ?NS_VCARD, sub_el = []},
     IQr = Module:Function(JID, JID, IQ),
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 212. Only modules that define callbacks should make dynamic calls.

apps/ejabberd/src/adhoc.erl
```diff
@@ -87,7 +88,9 @@ find_xdata_el1([_ | Els]) ->
 %% @doc Produce a <command/> node to use as response from an adhoc_response
 %% record, filling in values for language, node and session id from
 %% the request.
--spec produce_response(request(), response()) -> #xmlel{}.
+-spec produce_response(request(), response() | atom()) -> #xmlel{}.
```
gadgetci: According to **Elvis**:
> The spec in line 91 uses a record, please define a type for the record and use that instead.

michalwski:
@rslota or @kzemek, could you merge this PR. It's important not only because of dialyzer but also fixes to our docker image.

PR #1174, opened 2017-01-31
Accum stage one

bartekgorny:
This is a rewrite of a very small part of the whole processing chain - early stages of processing outgoing presence. The idea is to eventually create a mongoose_acc:t() from xmlel when a stanza enters the system, pass it along and terminate when it is about to leave. We start with a very small part, then we'll move initialisation and termination points further apart, untill they reach both ends.

Being an intermediate PR, it contains lots of utilities, comments, commented-out code which will eventually go away. Also, in `ejabberd_hooks` there are some useful things which show cool things we can do - those may stay or go, it depends.

### DIFF
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -921,7 +921,7 @@ session_established(closed, StateData) ->
 -spec process_outgoing_stanza(El :: jlib:xmlel(), state()) -> fsm_return().
 process_outgoing_stanza(El, StateData) ->
     % initialise accumulator, fill with data
-    Acc = ?INITIALISE(El),
+    Acc = mongoose_acc:initialise(El, ?FILE, ?LINE),
     User = StateData#state.user,
     Server = StateData#state.server,
     FromJID = StateData#state.jid,
@@ -960,7 +960,7 @@ process_outgoing_stanza(El, StateData) ->
             process_outgoing_stanza(ToJID, Name, {Attrs, Acc, FromJID, StateData, Server, User});
         _ ->
             % unpack and proceed as before
-            NewElement = ?TERMINATE(Acc2),
+            NewElement = mongoose_acc:terminate(Acc2, ?FILE, ?LINE),
             process_outgoing_stanza(ToJID, Name, {Attrs, NewElement, FromJID, StateData, Server, User})
     end,
     ejabberd_hooks:run(c2s_loop_debug, [{xmlstreamelement, El}]),
@@ -989,7 +989,7 @@ process_outgoing_stanza(ToJID, <<"presence">>, Args) ->
                                    Res,
                                    [FromJID, ToJID, El]),
     ?DUMP(Res1),
-    PresenceEl = ?TERMINATE(Res1),
+    PresenceEl = mongoose_acc:terminate(Res1, ?FILE, ?LINE),
     case ToJID of
         #jid{user = User,
              server = Server,
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -961,7 +961,9 @@ process_outgoing_stanza(El, StateData) ->
         _ ->
             % unpack and proceed as before
             NewElement = mongoose_acc:terminate(Acc2, ?FILE, ?LINE),
-            process_outgoing_stanza(ToJID, Name, {Attrs, NewElement, FromJID, StateData, Server, User})
+            process_outgoing_stanza(ToJID,
+                                    Name,
+                                    {Attrs, NewElement, FromJID, StateData, Server, User})
     end,
     ejabberd_hooks:run(c2s_loop_debug, [{xmlstreamelement, El}]),
     fsm_next_state(session_established, NState).
@@ -988,17 +990,14 @@ process_outgoing_stanza(ToJID, <<"presence">>, Args) ->
                                    Server,
                                    Res,
                                    [FromJID, ToJID, El]),
-    ?DUMP(Res1),
-    PresenceEl = mongoose_acc:terminate(Res1, ?FILE, ?LINE),
     case ToJID of
         #jid{user = User,
              server = Server,
              resource = <<>>} ->
-             ?DEBUG("presence_update(~p, ~n\t~p, ~n\t~p)",
-                 [FromJID, PresenceEl, StateData]),
-             presence_update(FromJID, PresenceEl,
+             presence_update(FromJID, Res1,
                              StateData);
         _ ->
+             PresenceEl = mongoose_acc:terminate(Res1, ?FILE, ?LINE),
              presence_track(FromJID, ToJID, PresenceEl,
                             StateData)
     end;
@@ -1169,8 +1168,9 @@ handle_info({force_update_presence, LUser}, StateName,
                            StateData#state.pres_last,
                            [LUser, LServer]),
             StateData2 = StateData#state{pres_last = PresenceEl},
+            Acc = mongoose_acc:initialise(PresenceEl, ?FILE, ?LINE),
             presence_update(StateData2#state.jid,
-                            PresenceEl,
+                            Acc,
                             StateData2),
             StateData2;
         _ ->
@@ -1487,6 +1487,7 @@ terminate(_Reason, StateName, StateData) ->
             Packet = #xmlel{name = <<"presence">>,
                             attrs = [{<<"type">>, <<"unavailable">>}],
                             children = [StatusEl]},
+            Acc = mongoose_acc:from_element(Packet),
             ejabberd_sm:close_session_unset_presence(
               StateData#state.sid,
               StateData#state.user,
@@ -1495,9 +1496,9 @@ terminate(_Reason, StateName, StateData) ->
               <<"Replaced by new connection">>,
               replaced),
             presence_broadcast(
-              StateData, From, StateData#state.pres_a, Packet),
+              StateData, From, StateData#state.pres_a, Acc),
             presence_broadcast(
-              StateData, From, StateData#state.pres_i, Packet),
+              StateData, From, StateData#state.pres_i, Acc),
             reroute_unacked_messages(StateData);
         {_, resumed} ->
             ?INFO_MSG("(~w) Stream ~p resumed for ~s",
@@ -1524,6 +1525,7 @@ terminate(_Reason, StateName, StateData) ->
                     From = StateData#state.jid,
                     Packet = #xmlel{name = <<"presence">>,
                                     attrs = [{<<"type">>, <<"unavailable">>}]},
+                    Acc = mongoose_acc:from_element(Packet),
                     ejabberd_sm:close_session_unset_presence(
                       StateData#state.sid,
                       StateData#state.user,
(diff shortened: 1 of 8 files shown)


apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -939,7 +951,18 @@ process_outgoing_stanza(El, StateData) ->
...
+    Name = mongoose_acc:get(name, Acc2),
+    NState = case Name of
+        <<"presence">> ->
+            % new-style
+            % I think it still makes sense to pass all those arguments, for performance reasons
+            process_outgoing_stanza(ToJID, Name, {Attrs, Acc, FromJID, StateData, Server, User});
+        _ ->
+            % unpack and proceed as before
+            NewElement = mongoose_acc:terminate(Acc2, ?FILE, ?LINE),
+            process_outgoing_stanza(ToJID, Name, {Attrs, NewElement, FromJID, StateData, Server, User})
```
> Line 964 is too long:             process_outgoing_stanza(ToJID, Name, {Attrs, NewElement, FromJID, StateData, Server, User}).

apps/ejabberd/src/ejabberd_hooks.erl
```diff
@@ -228,7 +240,7 @@ run_fold1([{_Seq, Module, Function} | Ls], Hook, Val, Args) ->
     Res = if is_function(Function) ->
```
gadgetci: According to **Elvis**:
> Replace the 'if' expression on line 240 with a 'case' expression or function clauses.

apps/ejabberd/src/ejabberd_sm.erl
```diff
@@ -293,31 +293,35 @@ get_raw_sessions(User, Server) ->
     clean_session_list(
       ?SM_BACKEND:get_sessions(jid:nodeprep(User), jid:nameprep(Server))).
```
gadgetci: According to **Elvis**:
> Don't use macros (like SM_BACKEND on line 294) as module names.

apps/ejabberd/src/mod_last.erl
```diff
@@ -164,15 +164,17 @@ process_sm_iq(From, To,
...
+                Acc = mongoose_acc:new(),
+                Res = ejabberd_hooks:run_fold(privacy_check_packet,
+                                              Server, Acc,
+                                              [User, Server, UserListRecord,
+                                               {To, From,
+                                               #xmlel{name = <<"presence">>,
+                                                      attrs = [],
+                                                      children = []}},
+                                               out]),
+                case mongoose_acc:get(privacy_check, Res, allow) of
```
gadgetci: According to **Elvis**:
> The expression on line 177 and column 17 is nested beyond the maximum level of 3.

apps/ejabberd/src/mongoose_acc.erl
```diff
@@ -0,0 +1,123 @@
...
+%%% might be worth reimplementing as binary
+
+%%%%% devel API %%%%%
+
+%%% Eventually, we'll call initialise when a stanza enters MongooseIM and terminate
+%%% when it leaves. During development we can call both in arbitrary places, provided that
+%%% the code which is executed between them is rewritten. We will proceed by moving
+%%% both points further apart until they reach their respective ends of processing chain.
+
+initialise(El, F, L) ->
```
gadgetci: According to **Compiler**:
> Warning: variable 'L' is unused

gadgetci: According to **Compiler**:
> Warning: variable 'F' is unused

apps/ejabberd/src/mongoose_acc.erl
```diff
@@ -0,0 +1,123 @@
...
+%%    ?ERROR_MSG("AAA initialise accumulator ~p ~p", [F, L]),
+    from_element(El).
+
+terminate(M, F, L) ->
```
gadgetci: According to **Compiler**:
> Warning: variable 'L' is unused

gadgetci: According to **Compiler**:
> Warning: variable 'F' is unused

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -939,7 +951,20 @@ process_outgoing_stanza(El, StateData) ->
                 _ ->
                     NewEl1
             end,
-    NState = process_outgoing_stanza(ToJID, Name, {Attrs, NewEl, FromJID, StateData, Server, User}),
+    Acc2 = mongoose_acc:put(element, NewEl, Acc1),
+    Name = mongoose_acc:get(name, Acc2),
+    NState = case Name of
+        <<"presence">> ->
+            % new-style
+            % I think it still makes sense to pass all those arguments, for performance reasons
```
michalwski: Also since, you already changed the implementation for `presence` stanza only, maybe it would be good to rename the function to sth like `process_outgoing_presence`?

apps/ejabberd/src/ejabberd_hooks.erl
```diff
@@ -242,3 +250,20 @@ run_fold1([{_Seq, Module, Function} | Ls], Hook, Val, Args) ->
         NewVal ->
             run_fold1(Ls, Hook, NewVal, Args)
     end.
+
+hook_apply_function(_Module, Function, _Hook, Val, Args) when is_function(Function) ->
+    safely:apply(Function, [Val | Args]);
+hook_apply_function(Module, Function, Hook, Val, Args) ->
+    record(Hook, Module, Function, safely:apply(Module, Function, [Val | Args])).
```
michalwski: Assigning result from `safe:apply` to its own var in its own line may improve readability a bit.

apps/ejabberd/src/ejabberd_sm.erl
```diff
@@ -293,31 +293,35 @@ get_raw_sessions(User, Server) ->
...
 
--spec set_presence(SID, User, Server, Resource, Prio, Presence, Info) -> ok when
+-spec set_presence(Acc, SID, User, Server, Resource, Prio, Presence, Info) -> Acc1 when
```
michalwski: What is the added value with having the `mongose_acc` data here?

apps/ejabberd/src/mod_last.erl
michalwski: I agree.

apps/ejabberd/src/mongoose_metrics_hooks.erl
```diff
@@ -271,13 +271,13 @@ privacy_list_push(Acc, _From, #jid{server = Server} = _To, _Broadcast, SessionCo
...
--spec privacy_check_packet(Acc :: allow | deny | block,
+-spec privacy_check_packet(Acc :: mongoose_acc:t(),
                           binary(),
                           Server :: ejabberd:server(),
-                          term(), term(), term()) -> allow | deny | block.
+                          term(), term(), term()) -> mongoose_acc:t().
 privacy_check_packet(Acc, _, Server, _, _, _) ->
     mongoose_metrics:update(Server, modPrivacyStanzaAll, 1),
-    case Acc of
+    case mongoose_acc:get(privacy_check, Acc) of
```
michalwski: I can imagine a situation that `mod_privacy` and `mod_blocking` are disabled. To my understanding in such a case there will be no `privacy_check` at this point which will end up with `case clause` error.

bartekgorny:
Ha, it is indeed a difficult decision. The argument in favour of merging into master is obvious - the amount of conflict which'd need to be solved, especially if there are major changes to master (like elvis-related cleanups). On the other hand, having half-baked stuff in master for a long time is indeed unseemly. An idea: maybe we could set an intermediate goal, e.g. rewriting the whole of c2s, without touching sm, and determine to do it for the next release?

apps/ejabberd/src/ejabberd_c2s.erl
bartekgorny: ad.1: Frankly speaking, I have no idea how much it costs to read a value from a map, compared to passing value through a function call. Since those values are already there, I though it'd be a bit cheaper to pass them along. If you think the cost is meaningless then yes, we can stuff everything into acc and pass only values which are required for pattern matching. Though, if we begin to rely on state data in accumulator, then we'll probably need to remember to update it when changed, and remove it before sending the acc to sm.

bartekgorny: ad.2: could be, though current pattern-matching approach seems ok to me.

apps/ejabberd/src/ejabberd_sm.erl
bartekgorny: The same as in other places: set_presence calls `set_presence_hook`, and itself is called by c2s:update_priority, which is called by presence_update, which is called by process_outgoing_stanza, so they are all parts of the main stanza processing chain, and as such should receive an accumulator and be given a change to modify it.

apps/ejabberd/src/mongoose_metrics_hooks.erl
bartekgorny: right

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -875,6 +875,9 @@ session_established({xmlstreamelement,
 session_established({xmlstreamelement, El}, StateData) ->
     FromJID = StateData#state.jid,
     % Check 'from' attribute in stanza RFC 3920 Section 9.1.2
+    % this is where we probably should initialise accumulator, or even earlier
+    % but it would require hacking mod_amp at the very beginning
+    % so to make it simpler we do it a bit later
```
bartekgorny: Yes :) Here, "later" means in the next function which is `process_outgoing_stanza`

michalwski: Let's be explicit then :)

apps/ejabberd/src/ejabberd_sm.erl
michalwski: Thanks for the explanation.

fenek:
I would avoid maintaining a parallel branch with pretty invasive changes to hook system. The one who will be supposed to rebase it after some time will probably prefer to cut him/herself with Mac keyboard. In my opinion an intermediate stages of this transformations are okay to be kept in master, as long as they don't break anything.

By c2s rewrite you mean pure refactor or something else?

apps/ejabberd/src/mod_last.erl
```diff
@@ -134,7 +134,7 @@ get_node_uptime() ->
     case ejabberd_config:get_local_option(node_start) of
         {_, _, _} = StartNow ->
             now_to_seconds(now()) - now_to_seconds(StartNow);
```
gadgetci: According to **Compiler**:
> Warning: erlang:now/0: Deprecated BIF. See the "Time and Time Correction in Erlang" chapter of the ERTS User's Guide for more information.

apps/ejabberd/src/ejabberd_sm.erl
```diff
@@ -295,31 +295,35 @@ get_raw_sessions(User, Server) ->
     clean_session_list(
       ?SM_BACKEND:get_sessions(jid:nodeprep(User), jid:nameprep(Server))).
```
gadgetci: According to **Elvis**:
> Don't use macros (like SM_BACKEND on line 296) as module names.

PR #1177, opened 2017-02-02
Add rdbms prepared queries

kzemek:
This PR adds prepared queries to RDBMS. MAM database async writers are presented as a first, example use of the queries.


apps/ejabberd/src/rdbms/rdbms_queries.erl
```diff
@@ -98,7 +98,8 @@
 	     count_offline_messages/4,
 	     remove_old_offline_messages/2,
 	     remove_expired_offline_messages/2,
-	     remove_offline_messages/3]).
+	     remove_offline_messages/3,
```
gadgetci: According to **Elvis**:
> Line 101 has a tab at column 0.

apps/ejabberd/src/mod_mam_odbc_arch.erl
```diff
@@ -780,24 +741,12 @@ stored_binary_to_packet(Bin) ->
     Module = db_message_format(),
     Module:decode(Bin).
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 742. Only modules that define callbacks should make dynamic calls.

apps/ejabberd/src/mod_mam_muc_odbc_arch.erl
```diff
@@ -659,23 +626,11 @@ stored_binary_to_packet(Bin) ->
     Module = db_message_format(),
     Module:decode(Bin).
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 627. Only modules that define callbacks should make dynamic calls.

apps/ejabberd/src/rdbms/rdbms_queries.erl
```diff
@@ -98,7 +98,8 @@
 	     count_offline_messages/4,
 	     remove_old_offline_messages/2,
```
gadgetci: According to **Elvis**:
> Line 99 has a tab at column 0.

apps/ejabberd/src/rdbms/rdbms_queries.erl
```diff
@@ -98,7 +98,8 @@
...
 	     remove_expired_offline_messages/2,
```
gadgetci: According to **Elvis**:
> Line 100 has a tab at column 0.

apps/ejabberd/src/rdbms/rdbms_queries.erl
```diff
@@ -98,7 +98,8 @@
 	     count_offline_messages/4,
```
gadgetci: According to **Elvis**:
> Line 98 has a tab at column 0.

apps/ejabberd/src/mongoose_metrics.erl
```diff
@@ -17,7 +17,7 @@
 
 -include("ejabberd.hrl").
 -include("mongoose_metrics_definitions.hrl").
-
+-compile([export_all]).
```
fenek: Kill it with fire!

apps/ejabberd/src/rdbms/mongoose_rdbms_pgsql.erl
```diff
@@ -61,28 +76,29 @@ is_error_duplicate(_Reason) -> false.
...
 pgsql_to_odbc({error, Reason}) ->
-    {pgsql_error, Details} = Reason,
-    Message = proplists:get_value(message, Details, <<"unknown error">>),
-    {error, unicode:characters_to_list(Message)};
-pgsql_to_odbc({{select, _Count}, Rows}) ->
-    {selected, [parse_row(Row) || Row <- Rows]};
-pgsql_to_odbc({{insert, _TableOID, Count}, _Rows}) ->
-    {updated, Count};
-pgsql_to_odbc({{_, Count}, _}) when is_integer(Count) ->
+    {error, unicode:characters_to_list(element(5, Reason))};
```
fenek: I don't like this `element(5, Reason)` here. The error tuple is a well-defined error record (check `epgsql.hrl`).

What is more, I think you could match for an error code for duplicate key and then just return something like `{error, duplicate_primary_key}` or similar. This would eliminate a need for awkward `is_error_duplicate` function.

apps/ejabberd/src/mod_mam_odbc_async_pool_writer.erl
```diff
@@ -316,12 +312,25 @@ are_recent_entries_required(_End, _Now) ->
...
-run_flush(State=#state{host=Host, connection_pool=Pool, number=N,
+run_flush(State=#state{host=Host, connection_pool=Pool, max_packet_size = MaxSize,
                        flush_interval_tref=TRef, acc=Acc, subscribers=Subs}) ->
     MessageCount = length(Acc),
     cancel_and_flush_timer(TRef),
     ?DEBUG("Flushed ~p entries.", [MessageCount]),
-    case mod_mam_odbc_arch:archive_messages(Pool, Acc, N) of
+
+    InsertResult =
+        case length(Acc) of
```
fenek: No need to get the length again, it's already in `MessageCount`.

fenek:
There is still a tiny possibility that someone uses "hand made partitions" for MAM so I'd actually add a check if it is enabled in config. If so, the module startup should fail and print appropriate warning message. We could keep this for like 2-3 releases so it will be lesser surprise.

apps/ejabberd/src/mod_mam_odbc_arch.erl
```diff
@@ -780,24 +749,12 @@ stored_binary_to_packet(Bin) ->
...
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 750. Only modules that define callbacks should make dynamic calls.

apps/ejabberd/src/mod_mam_muc_odbc_arch.erl
```diff
@@ -659,23 +634,11 @@ stored_binary_to_packet(Bin) ->
...
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 635. Only modules that define callbacks should make dynamic calls.

PR #1184, opened 2017-02-13
Feature/xep 357 pubsub plugin

rslota:
This PR adds pubsub `push` node support. This node type may be used to run `push_notification` hook for every published notification. Also, this PR adds `mod_push_service_mongoosepush` that handles the hook and uses [MongoosePush](https://github.com/esl/MongoosePush) to pass notification to APNS and/or FCM.


apps/ejabberd/src/mod_pubsub.erl
```diff
@@ -2130,7 +2154,12 @@ publish_item(Host, ServerHost, Node, Publisher, ItemId, Payload, Access) ->
                      PayloadMaxSize = get_option(Options, max_payload_size),
                      if not PublishFeature ->
```
gadgetci: According to **Elvis**:
> Replace the 'if' expression on line 2155 with a 'case' expression or function clauses.

apps/ejabberd/src/ejabberd.app.src
```diff
@@ -40,7 +40,8 @@
                   usec,
                   uuid,
                   xmerl,
-                  worker_pool
+                  worker_pool,
+                  hackney
```
michalwski: In many places we already use `fusco` for HTTP(s) queries to external services. What is the rational behind adding another HTTP lib?

apps/ejabberd/src/mod_pubsub.erl
```diff
@@ -310,6 +313,10 @@ init([ServerHost, Opts]) ->
                        ?MODULE, remove_user, 50),
     ejabberd_hooks:add(anonymous_purge_hook, ServerHost,
                        ?MODULE, remove_user, 50),
+
+%%    ejabberd_hooks:add(disco_sm_identity, ServerHost,
```
michalwski: We probably don't need these lines.

apps/ejabberd/src/mod_pubsub.erl
michalwski: I agree, let's try rewriting it since you added new branch to this `if`.

apps/ejabberd/src/node_push.erl
```diff
@@ -0,0 +1,226 @@
...
+         get_affiliation/2, set_affiliation/3,
+         get_entity_subscriptions/2, get_node_subscriptions/1,
+         get_subscriptions/2, set_subscriptions/4,
+         get_pending_nodes/2, get_states/1, get_state/2,
+         set_state/1, get_items/7, get_items/3, get_item/7,
+         get_item/2, set_item/1, get_item_name/3, node_to_path/1,
+         path_to_node/1]).
+
+init(Host, ServerHost, Opts) ->
+    node_flat:init(Host, ServerHost, Opts),
```
michalwski: How about we try [inaka/mixer](https://github.com/inaka/mixer) to reduce number of LOC in this module?

mkdocs.yml
```diff
@@ -54,6 +54,7 @@ pages:
       - 'mod_privacy': 'modules/mod_privacy.md'
       - 'mod_private': 'modules/mod_private.md'
       - 'mod_pubsub': 'modules/mod_pubsub.md'
+      - 'mod_push_service_mongoosepush': 'modules/mod_push_service_mongoosepush.md'
```
michalwski: Looks like `mod_push_service_mongoosepush.md` file was not committed.

test.disabled/ejabberd_tests/tests/push_pubsub_SUITE.erl
```diff
@@ -0,0 +1,415 @@
...
+-include_lib("common_test/include/ct.hrl").
+-include_lib("eunit/include/eunit.hrl").
+-include_lib("escalus/include/escalus_xmlns.hrl").
+-include_lib("exml/include/exml.hrl").
+
+-define(NS_PUSH,                <<"urn:xmpp:push:0">>).
+-define(NS_XDATA,               <<"jabber:x:data">>).
+-define(NS_PUBSUB_PUB_OPTIONS,  <<"http://jabber.org/protocol/pubsub#publish-options">>).
+-define(PUSH_FORM_TYPE,         <<"urn:xmpp:push:summary">>).
+-define(MUCHOST,                <<"muclight.localhost">>).
```
michalwski: Hardcoded `localhost`.

test.disabled/ejabberd_tests/tests/push_pubsub_SUITE.erl
```diff
@@ -0,0 +1,415 @@
+-module(push_pubsub_SUITE).
```
michalwski: The test suite is not added to `default.spec` so it'll not be run on travis.

test.disabled/ejabberd_tests/tests/push_pubsub_SUITE.erl
```diff
@@ -0,0 +1,415 @@
...
+
+%% ----------------------------------------------
+%% REST mock handler
+setup_mock_rest() ->
+    Dispatch = cowboy_router:compile([
+		{'_', [
+			{"/[:level1/[:level2/[:level3/[:level4]]]]", ?MODULE, [{pid, self()}]}
+		]}
+	]),
+	{ok, _} = cowboy:start_http(http, 100, [{port, 8080}], [
```
michalwski: Port `8080` may often be used on local machines. How about allowing `cowboy` to choose a random port and using this port later?

apps/ejabberd/src/mod_pubsub.erl
rslota: Hard to argue with that :)

apps/ejabberd/src/ejabberd.app.src
rslota: `fusco` seems to be harder to use. It gives me a connection, while `hackney` gives me pool of connections, so I don't have to make my own. 

Also from `fusco's` github: *"Fusco is still in a very early stage of development, be aware of that!*". While `hackney` is respected, good and actively maintained project.

mkdocs.yml
rslota: Looks that way. It will be added with next push.

test.disabled/ejabberd_tests/tests/push_pubsub_SUITE.erl
rslota: Thats a good idea, I'm gonna do that.

apps/ejabberd/src/ejabberd.app.src
michalwski: I know `fusco` is harder to use and not well documented. Regarding pools `fusco` has this feature as well, take a look at `fusco_cp`. Fusco was create  for and used heavily in load tests (namely  Megaload).

I'm not against `hackney`. If we decided to use it, we should rather change other places from `fusco` to `hackney` (in a separate PR). I just want to avoid multiple libs for the same purpose (if possible).

rslota: `fusco` it is. For now :)

apps/ejabberd/src/mod_pubsub.erl
rslota: Nope, removed.

apps/ejabberd/src/node_push.erl
rslota: Sure, why not.

test.disabled/ejabberd_tests/tests/push_pubsub_SUITE.erl
rslota: It is now.

rebar.lock
```diff
@@ -63,6 +64,10 @@
   {git,"https://github.com/basho/goldrush.git",
        {ref,"8f1b715d36b650ec1e1f5612c00e28af6ab0de82"}},
   1},
+ {<<"hackney">>,
```
michalwski: We probably no longer need hackney (and its deps) in rebar.lock.

michalwski:
Thanks!

PR #1188, opened 2017-02-17
Rock mod_adhoc..mod_cowboy

fenek:
This PR fixes majority of Elvis' complaints about `mod_adhoc`..`mod_cowboy` modules. Some issues will remain unresolved for now:

* `mod_bosh_socket` is considered a god module with 34 functions but actually thanks to that, it is easy to read and understand
* `mod_cowboy` makes some dynamic calls but I'd avoid blindly adding new behaviour to solve that. I'm deferring this decision for now. This module is rarely modified anyway so Elvis errors here won't bite us in PRs very often.

This is the first PR with rocking `mod_*`. I'm splitting it into smaller parts to make review easier.


apps/ejabberd/src/mod_cowboy.erl
```diff
@@ -122,7 +122,7 @@ handle_ws_init(Transport, Req, Opts) ->
         _ ->
             {ok, Req2} = cowboy_req:reply(404, Req1),
             {shutdown, Req2}
-    end. 
+    end.
 
 init_http_handler(Handler, Transport, Req, Opts) ->
     case Handler:init(Transport, Req, Opts) of
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 128. Only modules that define callbacks should make dynamic calls.

apps/ejabberd/src/mod_admin_extra_accounts.erl
```diff
@@ -192,73 +199,66 @@ delete_old_users_vhost(Host, Days) ->
...
--spec delete_old_users(Days :: integer(), Users :: [ejabberd:simple_jid()])
-            -> {'removed', non_neg_integer(), [ejabberd:simple_jid()]}.
+-spec delete_old_users(Days :: integer(), Users :: [ejabberd:simple_bare_jid()]) ->
+    {removed, non_neg_integer(), [ejabberd:simple_bare_jid()]}.
 delete_old_users(Days, Users) ->
     %% Convert older time
     SecOlder = Days*24*60*60,
 
     %% Get current time
     {MegaSecs, Secs, _MicroSecs} = now(),
```
gadgetci: According to **Compiler**:
> Warning: erlang:now/0: Deprecated BIF. See the "Time and Time Correction in Erlang" chapter of the ERTS User's Guide for more information.

fenek:
Will be split into more PRs. Created a new one (the first) as #1191

PR #1190, opened 2017-02-17
Upgrade deps

### LINKED ISSUE #1162: cuesport provides dead pids using it with redo
Version independent. The problem is with redo Redis driver. When a tcp connection is closed, the redo is stop with a non-normal reason (tcp_closed). When there is a tcp error, redo is stopped with reason normal. [old redo](https://github.com/jkvor/redo/blob/7c7eaef4cd65271e2fc4ea88587e848407cf0762/src/redo.erl#L280)

We are using redo with cuesport, and cuesport spawns transient children. So if the connection is closed, cuesport restarts that child. But if there is a tcp error, it won't. But cuesport serves worker from an ets table, and ets table still contains the dead pids. [cuesport get_worker](https://github.com/esl/cuesport/blob/d82ff25baf393eb11712721fb1b881b4b9f80176/src/cuesport.erl#L54)

So eventually all calls of `redo:cmd(cuesport:get_worker())` will end up in `noproc`. You can test it with cutting the redis connections with iptables. Also reconnect wasn't implemented in case of tcp_error.

I made a change to overcome that error.

https://github.com/Wallapop/redo/pull/1/commits/653457d27846b778a84ffe51eff5656c1b46b8ce

Have fun!

michalwski:
This PR upgrades following dependencies:
* redo - fixes #1162
* cowboy to 1.1.2
* uuid to 1.6.0
* fast_tls to 1.0.10
* cachetab to 1.06
* riak client to 2.5.2


fenek:
Are there any specific reasons to upgrade these deps? Or is it simply "because new versions are available with some bugfixes"?

michalwski:
The most important are:
* redo bug described in #1162 
* newest riak client simplifies building with rebard3 (no need for workaround)

PR #1193, opened 2017-02-22
Rock mod_* - part 2 - mod_disco..mod_mam

fenek:
This is a second part of rocking `mod_*` with Elvis.

Only one issue will become unresolved for now: `mod_mam_utils` being a god module. I don't think decomposing a module with 50+ functions into smaller ones with ~20 functions is in scope of this PR.

### DIFF
--- a/apps/ejabberd/src/mod_http_notification.erl
+++ b/apps/ejabberd/src/mod_http_notification.erl
@@ -1,5 +1,5 @@
 %%%----------------------------------------------------------------------
-%%% File    : mod_notify.erl
+%%% File    : mod_http_notification.erl
 %%% Author  : Baibossynv Valery <baibossynov.valery@gmail.com>
 %%% Purpose : Message passing via http
 %%% Created : 16 Dec 2015 by Baibossynv Valery <baibossynov.valery@gmail.com>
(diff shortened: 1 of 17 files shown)


apps/ejabberd/src/mod_http_notification.erl
```diff
@@ -7,7 +7,11 @@
 
 -module(mod_http_notification).
 -author("baibossynov.valery@gmail.com").
+
 -behaviour(gen_mod).
+-behaviour(mod_http_notification).
```
gadgetci: According to **Compiler**:
> Warning: behaviour mod_http_notification undefined

apps/ejabberd/src/mod_last.erl
```diff
@@ -145,41 +143,36 @@ now_to_seconds({MegaSecs, Secs, _MicroSecs}) ->
...
+-spec process_sm_iq(ejabberd:jid(), ejabberd:jid(), ejabberd:iq()) -> ejabberd:iq().
+process_sm_iq(_From, _To, #iq{type = set, sub_el = SubEl} = IQ) ->
+    IQ#iq{type = error, sub_el = [SubEl, ?ERR_NOT_ALLOWED]};
+process_sm_iq(From, To, #iq{type = get, sub_el = SubEl} = IQ) ->
+    User = To#jid.luser,
+    Server = To#jid.lserver,
+    {Subscription, _Groups} =
+    ejabberd_hooks:run_fold(roster_get_jid_info, Server,
+                            {none, []}, [User, Server, From]),
+    case (Subscription == both) or (Subscription == from) or
```
michalwski: To be honest, I'm not sure how should I read all these conditions to validate if this is correct or not. Would it be possible to make easier to understand by f.e introduction a function(s) or assigning results to some meaningful names?

fenek:
@michalwski 
I have to make it work first... :D

apps/ejabberd/src/mod_last.erl
fenek: Sure, not a problem.

PR #1194, opened 2017-02-22
Accum stage two

bartekgorny:
A rewrite of most (maybe all) of presence processing in c2s, with accumulator passed around and simplified function signatures (most args are passed in accumulator).


apps/ejabberd/src/mongoose_privacy.erl
```diff
@@ -0,0 +1,66 @@
...
+
+-include("ejabberd.hrl").
+-include("mod_privacy.hrl").
+
+%% API
+-export([privacy_check_packet/6, privacy_check_packet/7]).
+
+%%% API %%%
+
+-spec privacy_check_packet(Acc :: mongoose_acc:t(),
```
gadgetci: According to **Elvis**:
> The spec in line 20 uses a record, please define a type for the record and use that instead.

apps/ejabberd/src/mongoose_privacy.erl
```diff
@@ -0,0 +1,66 @@
...
+                           PrivacyList :: #userlist{},
+                           To :: ejabberd:jid(),
+                           Dir :: 'in' | 'out') -> {mongoose_acc:t(), allow|deny|block}.
+%% @doc abbreviated version - in most cases we have 'from' in Acc, but sometimes
+%% the caller wants something different (e.g. mod_last which swaps addresses)
+privacy_check_packet(Acc, Server, User, PrivacyList, To, Dir) ->
+    From = mongoose_acc:get(from_jid, Acc),
+    privacy_check_packet(Acc, Server, User, PrivacyList, From, To, Dir).
+
+-spec privacy_check_packet(Acc :: mongoose_acc:t(),
```
gadgetci: According to **Elvis**:
> The spec in line 32 uses a record, please define a type for the record and use that instead.

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -1910,101 +1904,119 @@ presence_update_to_available(StateData, From, Packet) ->
                                           NewStateData#state.server,
                                           {[], [], []},
                                           [StateData#state.user, NewStateData#state.server]),
+                      % is this processing of the same stanza, or a new chain? Means: should
+                      % we pass on the same acc or create a new one there? or just don't care?
+                      % I think it is the same, since this is the one that started the whole
```
michalwski: Right, it seems it's still the same packet. No need to create new Acc here IMO.

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -1910,101 +1904,119 @@ presence_update_to_available(StateData, From, Packet) ->
...
-                                          Packet),
+            Acc2 = presence_broadcast_to_trusted(Acc1,
+                                                 NewStateData,
+                                                 From,
+                                                 NewStateData#state.pres_f,
+                                                 NewStateData#state.pres_a,
+                                                 Packet),
             case OldPriority < 0 andalso NewPriority >= 0 of
                 true ->
+                    % same question as above
```
michalwski: The same answer as above :)

apps/ejabberd/src/mongoose_acc.erl
```diff
@@ -108,6 +109,23 @@ append(Key, Val, P) ->
     L = get(Key, P, []),
     maps:put(Key, append(Val, L), P).
 
+%% @doc Store a value in a key-value store
+-spec store(K :: atom(), Idx :: any(), Val :: any(), Acc :: t()) -> t().
+store(K, Idx, Val, Acc) ->
```
michalwski: I believe longer var names will make things easier to understand.

apps/ejabberd/src/mod_privacy.erl
```diff
@@ -336,7 +336,7 @@ check_packet(Acc, User, Server,
             Type = xml:get_attr_s(<<"type">>, Packet#xmlel.attrs),
             check_packet_aux(List, PType, Type, LJID, Subscription, Groups)
     end,
-    mongoose_acc:put(privacy_check, CheckResult, Acc).
+    mongoose_acc:store(privacy_check, jid:to_lower(To), CheckResult, Acc).
```
michalwski: Could you explain why exactly do we need to introduce a nested map here? It's not clear to me.

apps/ejabberd/src/mongoose_privacy.erl
michalwski: I agree with Elvis here.

apps/ejabberd/src/mongoose_privacy.erl
```diff
@@ -0,0 +1,66 @@
+%%%-------------------------------------------------------------------
+%%% @author bartek <bartlomiej.gorny@erlang-solutions.com>
+%%% @copyright (C) 2017, Erlang Solutions sp. z o.o.
+%%% @doc
+%%%
```
michalwski: Let's fill in the placeholder :)

apps/ejabberd/src/mongoose_privacy.erl
```diff
@@ -0,0 +1,66 @@
...
+
+%%% API %%%
+
+-spec privacy_check_packet(Acc :: mongoose_acc:t(),
+                           Server :: binary(),
+                           User :: binary(),
+                           PrivacyList :: #userlist{},
+                           To :: ejabberd:jid(),
+                           Dir :: 'in' | 'out') -> {mongoose_acc:t(), allow|deny|block}.
+%% @doc abbreviated version - in most cases we have 'from' in Acc, but sometimes
```
michalwski: Documentation before spec is probably more accurate. Do you know what doc generators expect? Not that we use them often but maybe we can follow their convention.

@erszcz how about your cool [erszcz/docsh](https://github.com/erszcz/docsh)? Do they expect the doc before or after spec? Or it doesn't really matter?

apps/ejabberd/test/acc_SUITE.erl
```diff
@@ -0,0 +1,49 @@
+%% @doc This suite tests both old ejabberd_commands module, which is slowly getting deprecated,
```
michalwski: Is it? I thought it's about testing `mongoose_acc` module.

apps/ejabberd/src/mod_privacy.erl
bartekgorny: We check whether privacy rules allow sending this message and want to store result of the privacy check. But, privacy rules differ by recipient, that's why we have to store result of the privacy check per recipient.

michalwski: And when we'll have more than one recipient?

apps/ejabberd/src/mongoose_privacy.erl
bartekgorny: I don't get it. Doesn't it use a type?

michalwski: In the function spec you used record:
```erlang
PrivacyList :: #userlist{}
```

Elvis doesn't like it and I agree with him. It's better have a type for this record and use the type.

bartekgorny: Ah, got it. I was confused because I saw line 20 "-spec privacy_check_packet(Acc :: mongoose_acc:t()," and Elvis complaining about use of a record...

michalwski: Yeah, the comment was badly placed.

apps/ejabberd/src/ejabberd_c2s.hrl
```diff
@@ -54,7 +54,7 @@
                 pres_timestamp :: calendar:datetime() | undefined,
                 %% Are we invisible?
                 pres_invis = false :: boolean(),
-                privacy_list = #userlist{} :: mod_privacy:userlist(),
```
michalwski: Actually `mod_privacy:userlist()` is better IMO. Are the definitions different? We tend to reduce use of `.hrl` files whenever it's possible.
Since you introduced module `mognoose_privacy` this type (and other related to privacy) could be move to this module if not `mod_privacy`.

apps/ejabberd/src/mod_privacy.erl
bartekgorny: When it is a broadcast, I think.

michalwski: Like presence broadcast to all your friends? This could potentially be quite big (hundreds of items).

PR #1196, opened 2017-02-24
Fix config and log print in Travis after rebar3-fication

fenek:
Travis YML file still pointed to old dev nodes paths.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,11 +23,11 @@ script: tools/travis-test.sh -p $PRESET
 
 after_failure:
         - cat `ls -1 -d -t apps/ejabberd/logs/ct_run* | head -1`/apps.ejabberd.logs/run.*/suite.log
-        - cat dev/mongooseim_node1/etc/ejabberd.cfg
-        - if [ -s dev/mongooseim_node1/log/crash.log ]; then cat dev/mongooseim_node1/log/crash.log; fi
-        - if [ -s dev/mongooseim_node2/log/crash.log ]; then cat dev/mongooseim_node2/log/crash.log; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -23,11 +23,11 @@ script: tools/travis-test.sh -p $PRESET
 
 after_failure:
         - cat `ls -1 -d -t apps/ejabberd/logs/ct_run* | head -1`/apps.ejabberd.logs/run.*/suite.log
-        - cat dev/mongooseim_node1/etc/ejabberd.cfg
-        - if [ -s dev/mongooseim_node1/log/crash.log ]; then cat dev/mongooseim_node1/log/crash.log; fi
-        - if [ -s dev/mongooseim_node2/log/crash.log ]; then cat dev/mongooseim_node2/log/crash.log; fi
-        - tail -100 dev/mongooseim_node1/log/ejabberd.log
-        - tail -100 dev/mongooseim_node2/log/ejabberd.log
+        - cat _build/mim1/rel/mongooseim/etc/ejabberd.cfg
```
michalwski: I'm wondering if we still need this since most of this is archived on S3, f.e here: http://opensource.erlang-solutions.com/mongooseim-ct-reports/s3_reports.html?prefix=PR/1196/2768/internal_mnesia.18.3/logs/2017-02-24_10.04.13/

fenek: I think we can remove the line with printing the config but I'd leave the ones with logs, because it's easier and faster to take a quick look at this output, than navigate through S3, download file etc. :)

michalwski: All right :)

PR #1201, opened 2017-02-28
Fail tests on missing suite

michalwski:
This PR addresses fails big tests if there is missing suite. This can happen mostly because of the following reasons:
* there is a syntax error in the SUITE
* the SUITE was added to `default.spec` but the file was not added to the repo

### DIFF
--- a/tools/travis-test.sh
+++ b/tools/travis-test.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euo pipefail
+set -uo pipefail
 IFS=$'\n\t'
 
 PRESET="internal_mnesia"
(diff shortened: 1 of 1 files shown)


tools/travis-test.sh
```diff
@@ -126,30 +130,30 @@ run_tests() {
...
+	SUMMARIES_DIR=$(summaries_dir ${SUMMARIES_DIRS})
 	${TOOLS}/summarise-ct-results ${SUMMARIES_DIR}
-	BIG_STATUS=$?
+	BIG_STATUS_BY_SUMMARY=$?
 
 	echo
 	echo "All tests done."
 
-	if [ $SMALL_STATUS -eq 0 -a $BIG_STATUS -eq 0 ]
+	if [[ $SMALL_STATUS -eq 0 && $BIG_STATUS -eq 0 && $BIT_STATUS_BY_SUMMARY -eq 0 ]]
```
kzemek: `$BIT_STATUS_BY_SUMMARY` -> `$BIG_STATUS_BY_SUMMARY` :)

michalwski: Right :) good catch. I'm on it!

PR #1202, opened 2017-02-28
Accum stage three

bartekgorny:
Next stage of accumulator implementation - rewrote processing outgoing errors, iq and messages, which sort of completes outgoing stuff in c2s. Next stage will be routing them out and ingesting back into c2s.


apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -2229,56 +2228,70 @@ get_priority_from_presence(PresencePacket) ->
...
+                                   StateData#state.server,
+                                   Acc,
+                                   [StateData#state.user, StateData#state.server]),
+    Rs = mongoose_acc:get(offline_messages, Acc1, []),
+    Acc2 = lists:foldl(fun({route, From, To, #xmlel{} = Packet}, A) ->
+                           check_privacy_and_route_or_ignore(A, StateData, From, To, Packet, in)
+                       end,
+                       Acc1, Rs),
+    mongoose_acc:remove(offline_messages, Acc2). % they are gone from db backend and sent
+
```
> Line 2294 has 4 trailing whitespace characters.

apps/ejabberd/src/mod_roster.erl
```diff
@@ -514,7 +514,8 @@ get_subscription_lists(Acc, User, Server) ->
     LServer = jid:nameprep(Server),
     Items = ?BACKEND:get_subscription_lists(Acc, LUser, LServer),
```
gadgetci: According to **Elvis**:
> Don't use macros (like BACKEND on line 515) as module names.

apps/ejabberd/src/mod_blocking.erl
```diff
@@ -47,16 +47,17 @@ process_iq_get(_, _From = #jid{luser = LUser, lserver = LServer},
...
-        {ok, Lst} ->
-            {result, blocking_query_response(Lst)};
-        {error, _} ->
-            {error, ?ERR_INTERNAL_SERVER_ERROR}
-    end;
+    IqRes = case Res of
+                {ok, Lst} ->
+                    {result, blocking_query_response(Lst)};
+                {error, _} ->
+                    {error, ?ERR_INTERNAL_SERVER_ERROR}
```
michalwski: I'm more and more convinced that hook implementations should return `{ok, Acc}` in success cases and sth like `{error, Reason, Acc}`. This subject of a separate PR though.

apps/ejabberd/src/mod_privacy.erl
```diff
@@ -154,25 +154,26 @@ stop(Host) ->
...
-                    ListName = xml:get_attr(<<"name">>, Attrs),
-                    process_list_get(LUser, LServer, ListName);
-                _ ->
-                    {error, ?ERR_BAD_REQUEST}
-            end;
-        _ ->
-            {error, ?ERR_BAD_REQUEST}
-    end;
+process_iq_get(Acc,
+               _From = #jid{luser = LUser, lserver = LServer},
```
michalwski: Am I right that `Acc` already contains most of the args?

apps/ejabberd/src/mongoose_acc.erl
```diff
@@ -120,3 +126,13 @@ dump(_, []) ->
 dump(Acc, [K|Tail]) ->
     ?ERROR_MSG("~p = ~p", [K, maps:get(K, Acc)]),
     dump(Acc, Tail).
+
+read_child([]) ->
+    {undefined, undefined};
+read_child([Chld]) ->
+    #xmlel{name = Name, attrs = Attrs} = Chld,
+    {xml:get_attr_s(<<"xmlns">>, Attrs), Name};
```
michalwski: I strongly encourage use of `exml_query` for things like `get_attr`.

apps/ejabberd/src/mongoose_acc.erl
```diff
@@ -120,3 +126,13 @@ dump(_, []) ->
...
+read_child(_) ->
+    % children are sometimes longer then 1 but only when it is an error, not sure yet how
```
michalwski: How about `message` stanza with `body` and other sub elements indicating for example the chat state (active, typing etc).

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -1890,44 +1888,52 @@ presence_update_to_available(Acc, StateData, From, Packet) ->
...
+%% @doc the first one is run when presence changes from unavailable to anything else
+-spec presence_update_to_available(Acc :: mongoose_acc:t(),
+                                   FromUnavailable :: boolean(),
+                                   OldPriority :: integer(),
+                                   NewPriority :: integer(),
+                                   From :: ejabberd:jid(),
+                                   Packet :: exml:element(),
+                                   StateData :: state()) -> {mongoose_acc:t(), state()}.
+presence_update_to_available(Acc, true, _, NewPriority, From, Packet, StateData) ->
```
bartekgorny: I'm looking for a convention to follow, be it (Acc, true, ..., State) or (true, ..., Acc, State) or (true, Acc, ..., State), I don't know which is the best.

michalwski: IMO `(true, Acc, ..., State)` but it all depends on the use case :) In general, I personally think that if we pattern match on a parameter in a function head, it's more readable if this is the first parameter.

apps/ejabberd/src/mod_privacy.erl
bartekgorny: Yes, this is the case with almost every handler. It might - and should - be changed, but I'm deliberately leaving it out for now. I concentrate on the main processing chain, leaving hook calls largely as they are (with exception of how they return value). The reason is to do less code changes, thus reducing risk of conflicts. Also, changing hook args is a simple, manual task, orthogonal to what we are currently doing, and as such can be done in a separate PR (maybe together with making hooks return {ok, Acc}.

apps/ejabberd/src/mod_blocking.erl
bartekgorny: I agree, see the next comment.

apps/ejabberd/src/mod_privacy.erl
michalwski: All right, could you please note the ideas for additional changes? I agree this can be done in a separate PR(s).

apps/ejabberd/src/mongoose_acc.erl
bartekgorny: Good point - I should then scan all children looking for name, right?

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -876,6 +873,8 @@ session_established({xmlstreamelement, El}, StateData) ->
     FromJID = StateData#state.jid,
     % Check 'from' attribute in stanza RFC 3920 Section 9.1.2
     % this is where we probably should initialise accumulator, or even earlier
+    % (there is an argument in favour of doing it earlier - in wait_for_session_or_sm we
```
bartekgorny: Right ho, then, let's stay here.

apps/ejabberd/test/acc_SUITE.erl
```diff
@@ -38,3 +38,19 @@ store_and_retrieve(_C) ->
     ok.
 
 
+init_from_element(_C) ->
+    Elem = {xmlel,<<"iq">>,
```
gadgetci: According to **Elvis**:
> Missing space right "," on line 42

apps/ejabberd/test/acc_SUITE.erl
```diff
@@ -38,3 +38,19 @@ store_and_retrieve(_C) ->
...
+        [{<<"xml:lang">>,<<"en">>},{<<"type">>,<<"set">>}],
+        [{xmlel,<<"block">>,
```
gadgetci: According to **Elvis**:
> Missing space right "," on line 44

apps/ejabberd/test/acc_SUITE.erl
```diff
@@ -38,3 +38,19 @@ store_and_retrieve(_C) ->
...
+            [{<<"xmlns">>,<<"urn:xmpp:blocking">>}],
+            [{xmlel,<<"item">>,
+                [{<<"jid">>,<<"bob37.814302@localhost">>}],
```
gadgetci: According to **Elvis**:
> Missing space right "," on line 47

apps/ejabberd/test/acc_SUITE.erl
```diff
@@ -38,3 +38,19 @@ store_and_retrieve(_C) ->
     ok.
 
 
+init_from_element(_C) ->
+    Elem = {xmlel,<<"iq">>,
+        [{<<"xml:lang">>,<<"en">>},{<<"type">>,<<"set">>}],
+        [{xmlel,<<"block">>,
+            [{<<"xmlns">>,<<"urn:xmpp:blocking">>}],
```
gadgetci: According to **Elvis**:
> Missing space right "," on line 45

apps/ejabberd/test/acc_SUITE.erl
```diff
@@ -38,3 +38,19 @@ store_and_retrieve(_C) ->
     ok.
 
 
+init_from_element(_C) ->
+    Elem = {xmlel,<<"iq">>,
+        [{<<"xml:lang">>,<<"en">>},{<<"type">>,<<"set">>}],
```
gadgetci: According to **Elvis**:
> Missing space right "," on line 43

apps/ejabberd/test/acc_SUITE.erl
```diff
@@ -38,3 +38,19 @@ store_and_retrieve(_C) ->
...
+        [{xmlel,<<"block">>,
+            [{<<"xmlns">>,<<"urn:xmpp:blocking">>}],
+            [{xmlel,<<"item">>,
```
gadgetci: According to **Elvis**:
> Missing space right "," on line 46

apps/ejabberd/test/acc_SUITE.erl
gadgetci: According to **Elvis**:
> Missing space right "," on line 43

gadgetci: According to **Elvis**:
> Missing space right "," on line 43

michalwski:
@bartekgorny do you monitor Travis status for your PR? Looks like all jobs are failing.

apps/ejabberd/src/mongoose_acc.erl
```diff
@@ -64,10 +64,9 @@ from_kv(K, V) ->
 -spec from_element(xmlel()) -> t().
 from_element(El) ->
     #xmlel{name = Name, attrs = Attrs, children = Children} = El,
-    Type = xml:get_attr_s(<<"type">>, Attrs),
-    {Xmlns, Cmd} = read_child(Children),
-    #{element => El, mongoose_acc => true, name => Name, attrs => Attrs, type => Type,
-      xmlns => Xmlns, command => Cmd}.
+    Type = exml_query:attr(El, <<"type">>, undefined),
+    Acc = #{element => El, mongoose_acc => true, name => Name, attrs => Attrs, type => Type},
+    read_children(Acc, Children).
```
michalwski: I'm reluctant to inspect every stanza so deeply. When exactly do we need to read the children? Can we limit it to some small number of places and do it explicitly only for stanza we know we'll need this?

apps/ejabberd/src/ln
```diff
@@ -0,0 +1,252 @@
+    3489 ELDAPv3.erl
```
michalwski: Do we really need this file here in the repo?

apps/ejabberd/src/mongoose_acc.erl
```diff
@@ -63,9 +63,10 @@ from_kv(K, V) ->
 
 -spec from_element(xmlel()) -> t().
 from_element(El) ->
-    #xmlel{name = Name, attrs = Attrs} = El,
-    Type = xml:get_attr_s(<<"type">>, Attrs),
-    #{element => El, mongoose_acc => true, name => Name, attrs => Attrs, type => Type}.
+    #xmlel{name = Name, attrs = Attrs, children = Children} = El,
+    Type = exml_query:attr(El, <<"type">>, undefined),
+    Acc = #{element => El, mongoose_acc => true, name => Name, attrs => Attrs, type => Type},
+    read_children(Acc, Children).
```
michalwski: I'm reluctant to inspect every stanza so deeply. Do we really need to do this for every stanza or maybe there is only limited number of places where it's needed?

bartekgorny: You are right, it should be lazy.

apps/ejabberd/test/acc_SUITE.erl
```diff
@@ -38,3 +38,36 @@ store_and_retrieve(_C) ->
     ok.
 
 
+init_from_element(_C) ->
+    Elem = {xmlel, <<"iq">>,
```
gadgetci: According to **Elvis**:
> The code in the following (LINE, COL) locations has the same structure: (42, 10), (58, 10).

michalwski: Thanks @gadgetci for this! @bartekgorny could you remove the duplciated code? Looks like the stanzas are exactly the same.

PR #1204, opened 2017-03-02
Add labels to docker images created on travis

michalwski:
Proposed changes include:
* build time values passed to `docker build` via `--build-args`
* removal of deprecated `--email` parameter to `docker login`
* frozen version of https://github.com/esl/mongooseim-docker/commit/843558fb15ddbde6c0fbb704295208449034b123

### DIFF
--- a/tools/travis-build-and-push-docker.sh
+++ b/tools/travis-build-and-push-docker.sh
@@ -30,9 +30,8 @@ fi
 
 IMAGE_TAG=${DOCKERHUB_USER}/mongooseim:${DOCKERHUB_TAG}
 
-git clone https://github.com/esl/mongooseim-docker.git
+git clone -b 843558f https://github.com/esl/mongooseim-docker.git
 cd mongooseim-docker
-git checkout 843558f
 
 cp ../${MONGOOSE_TGZ} member
 
@@ -44,6 +43,6 @@ docker build -f Dockerfile.member -t ${IMAGE_TAG} \
 
 docker login -u ${DOCKERHUB_USER} -p ${DOCKERHUB_PASS}
 
-docker push ${DOCKERHUB_USER}/mongooseim:${DOCKERHUB_TAG}
+docker push ${IMAGE_TAG}
 
 fi
--- a/tools/travis-build-and-push-docker.sh
+++ b/tools/travis-build-and-push-docker.sh
@@ -4,12 +4,10 @@
 
 if [ ${TRAVIS_SECURE_ENV_VARS} == 'true' ]; then
 
-
 ./tools/configure with-all
 make rel
 
 MIM_TAR_FULL_NAME=mongooseim-$TRAVIS_BRANCH.OTP-$TRAVIS_OTP_RELEASE.$(lsb_release -is | tr "A-Z" "a-z").$(lsb_release -rs).$(uname -m).tar.bz2
-
 MONGOOSE_TGZ=mongooseim.tar.gz
 
 BUILD_PATH=_build/prod/rel/mongooseim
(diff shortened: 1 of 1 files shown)


tools/travis-build-and-push-docker.sh
```diff
@@ -21,23 +19,30 @@ export BUILDS=`pwd`
...
     DOCKERHUB_TAG="PR-${TRAVIS_PULL_REQUEST}"
 elif [ ${TRAVIS_BRANCH} == 'master' ]; then
     DOCKERHUB_TAG="latest";
 fi
 
+IMAGE_TAG=${DOCKERHUB_USER}/mongooseim:${DOCKERHUB_TAG}
+
 git clone https://github.com/esl/mongooseim-docker.git
 cd mongooseim-docker
+git checkout 843558f
```
mentels: I suggest documenting the fact that you rely on a *specific* commit of the mongoose-docker

michalwski: Thanks for the suggestion @mentels. How and where would you document it? I though it'd be enough to just add

>frozen version of esl/mongooseim-docker

To the PR description.

michalwski: This fact is already reflected in travis logs, also I added link to the exact Dockerfile to our docker hub page here: https://hub.docker.com/r/mongooseim/mongooseim/

tools/travis-build-and-push-docker.sh
```diff
@@ -21,23 +19,30 @@ export BUILDS=`pwd`
...
+docker build -f Dockerfile.member -t ${IMAGE_TAG} \
+             --build-arg BUILD_DATE=`date -u +"%Y-%m-%dT%H:%M:%SZ"` \
+	     --build-arg VCS_REF=${GIT_REF} \
+	     --build-arg VERSION=${VERSION} \
+	     .
 
-docker tag mongooseim ${DOCKERHUB_USER}/mongooseim:${DOCKERHUB_TAG}
+docker login -u ${DOCKERHUB_USER} -p ${DOCKERHUB_PASS}
 
 docker push ${DOCKERHUB_USER}/mongooseim:${DOCKERHUB_TAG}
```
rslota: I guess it's better to use `${IMAGE_TAG}` here, since it's already used in command that creates this image.

tools/travis-build-and-push-docker.sh
```diff
@@ -21,23 +19,30 @@ export BUILDS=`pwd`
...
 
 if [ ${TRAVIS_PULL_REQUEST} != 'false' ]; then
     DOCKERHUB_TAG="PR-${TRAVIS_PULL_REQUEST}"
 elif [ ${TRAVIS_BRANCH} == 'master' ]; then
     DOCKERHUB_TAG="latest";
 fi
 
+IMAGE_TAG=${DOCKERHUB_USER}/mongooseim:${DOCKERHUB_TAG}
+
 git clone https://github.com/esl/mongooseim-docker.git
```
rslota: Maybe it's better to checkout the frozen version via `git clone -b 843558f ` instead of `git checkout 843558f`?

michalwski: Good idea!

tools/travis-build-and-push-docker.sh
michalwski: Good catch, I'll change that!

tools/travis-build-and-push-docker.sh
rslota: @mentels Is this what you have asked for? Please let us know, so we can merge / improve the PR :) 
For me, personally, using frozen version is a default and expected behaviour, so even without any docs about that it shouldn't come as an surprise.

PR #1208, opened 2017-03-06
fix "make ct SUITE=" call

bartekgorny:
If we want to run just one test suite we must not submit --dir parameter to rebar3, otherwise
it runs all tests present in that directory; instead we should provide --suite with appropriate
prefix, and give --dir only when we want to run all tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,8 +21,8 @@ clean:
 	-rm rel/vars.config
 
 ct:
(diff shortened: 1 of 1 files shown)


michalwski:
I don't need the fix for the `make ct SUITE=` to work as expected. How do you run it? From which directory?

bartekgorny:
True. I don't why, but for me it used to trigger all tests, now it is ok.

PR #1215, opened 2017-03-13
Allow customizing riak_setup script with env variables

bszaf:
This PR allows to customize setup_riak script with environmental variables e.q.: `RIAK_HOST=custom_riak_address ./setup_riak`.
There are still default values for variables, that will be used by Travis.

### DIFF
--- a/tools/setup_riak
+++ b/tools/setup_riak
@@ -6,14 +6,14 @@ RIAK_HOST=${RIAK_HOST:-"localhost"}
 RIAK_PORT=${RIAK_PORT:-"8098"}
 RIAK_HOST="http://${RIAK_HOST}:${RIAK_PORT}"
 
-VCARD_SCHEMA_PATH=${VCARD_SCHEMA_PATH:-"tools/vcard_search_schema.xml"}
-MAM_SCHEMA_PATH=${MAM_SCHEMA_PATH:-"tools/mam_search_schema.xml"}
+RIAK_VCARD_SCHEMA_PATH=${RIAK_VCARD_SCHEMA_PATH:-"tools/vcard_search_schema.xml"}
+RIAK_MAM_SCHEMA_PATH=${RIAK_MAM_SCHEMA_PATH:-"tools/mam_search_schema.xml"}
 
 RIAK_CONTAINER_NAME=${RIAK_CONTAINER_NAME:-"mongooseim-riak"}
 
 curl -v -XPUT $RIAK_HOST/search/schema/vcard \
     -H 'Content-Type:application/xml' \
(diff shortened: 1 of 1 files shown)


tools/setup_riak
```diff
@@ -2,11 +2,19 @@
...
 
-RIAK_HOST="http://localhost:8098"
+RIAK_HOST=${RIAK_HOST:-"localhost"}
+RIAK_PORT=${RIAK_PORT:-"8098"}
+RIAK_HOST="http://${RIAK_HOST}:${RIAK_PORT}"
+
+VCARD_SCHEMA_PATH=${VCARD_SCHEMA_PATH:-"tools/vcard_search_schema.xml"}
+MAM_SCHEMA_PATH=${MAM_SCHEMA_PATH:-"tools/mam_search_schema.xml"}
+
+export DOCKER_HOST=${DOCKER_HOST}
```
rslota: Why do we need assignment here? Seems to be redundant, since this is equivalent to `export DOCKER_HOST`. Also why do we need this export anyway?

bszaf: @rslota You are right, it is not necessary.

tools/setup_riak
```diff
@@ -2,11 +2,18 @@
 
 sleep 20
 
-RIAK_HOST="http://localhost:8098"
+RIAK_HOST=${RIAK_HOST:-"localhost"}
+RIAK_PORT=${RIAK_PORT:-"8098"}
+RIAK_HOST="http://${RIAK_HOST}:${RIAK_PORT}"
+
+VCARD_SCHEMA_PATH=${VCARD_SCHEMA_PATH:-"tools/vcard_search_schema.xml"}
```
rslota: What about naming this variable like: `RIAK_SEARCH_VCARD_SCHEMA`? Not necessary exactly like that, but I think the `RIAK_` prefix alone would be really nice to see (like in other variables). Those files are specific to `Riak Search` service so if such variable is set in different file, it would be nice to know what service it's referencing to. Especially considering that other variables have `RIAK_` prefix, so if this would also had one, it would be obvious that it's used in same file/service.

tools/setup_riak
```diff
@@ -2,11 +2,18 @@
...
+MAM_SCHEMA_PATH=${MAM_SCHEMA_PATH:-"tools/mam_search_schema.xml"}
```
rslota: Same as in my comment to `VCARD_SCHEMA_PATH `.

michalwski:
@bszaf could you address @rslota's comments regarding naming?

michalwski:
Thanks!

PR #1217, opened 2017-03-13
Shared DB pools

chrzaszcz:
Make DB connection pool configuration independent from the host config.

To declare a DB connection pool, use on of the following formats:

```
  {pool, odbc, PoolName}.
  {pool, odbc, PoolName, Options}.
```

PoolName is an atom and Options is a list of ODBC options. Options can be still specified at the top level.

1. All hosts use the ‘default’ pool by default.
  To change this, use the {odbc_pool, Option} option (top-level or per-host).

2. This means that setting ‘odbc_server’ is not enough for a DB connection to work. The default pool has to be declared as well.

3. The ODBC options specified per-host are not allowed anymore. They have been replaced with the Options element of the pool declaration.

To update the previous config with ODBC enabled to the new format, add the following line to ejabberd.cfg:

`  {pool, odbc, default}.`


apps/ejabberd/src/rdbms/mongoose_rdbms_odbc.erl
```diff
@@ -57,11 +56,11 @@ query(Connection, Query, Timeout) when is_binary(Query) ->
 query(Connection, Query, Timeout) ->
     parse(odbc:sql_query(Connection, Query, Timeout)).
 
--spec prepare(Host :: ejabberd:server(), Connection :: term(), Name :: atom(), Table :: binary(),
+-spec prepare(Pool :: mongoose_rdbms:pool(), Connection :: term(), Name :: atom(), Table :: binary(),
```
gadgetci: According to **Elvis**:
> Line 59 is too long: -spec prepare(Pool :: mongoose_rdbms:pool(), Connection :: term(), Name :: atom(), Table :: binary(),.

apps/ejabberd/src/rdbms/mongoose_rdbms_mysql.erl
```diff
@@ -51,10 +51,10 @@ disconnect(Connection) ->
 query(Connection, Query, _Timeout) ->
     mysql_to_odbc(mysql:query(Connection, Query), Connection).
 
--spec prepare(Host :: ejabberd:server(), Connection :: term(), Name :: atom(), Table :: binary(),
+-spec prepare(Pool :: mongoose_rdbms:pool(), Connection :: term(), Name :: atom(), Table :: binary(),
```
gadgetci: According to **Elvis**:
> Line 54 is too long: -spec prepare(Pool :: mongoose_rdbms:pool(), Connection :: term(), Name :: atom(), Table :: binary(),.

apps/ejabberd/src/rdbms/mongoose_rdbms_pgsql.erl
```diff
@@ -51,10 +51,10 @@ disconnect(Connection) ->
 query(Connection, Query, _Timeout) ->
     pgsql_to_odbc(epgsql:squery(Connection, Query)).
 
--spec prepare(Host :: ejabberd:server(), Connection :: term(), Name :: atom(), Table :: binary(),
+-spec prepare(Pool :: mongoose_rdbms:pool(), Connection :: term(), Name :: atom(), Table :: binary(),
```
gadgetci: According to **Elvis**:
> Line 54 is too long: -spec prepare(Pool :: mongoose_rdbms:pool(), Connection :: term(), Name :: atom(), Table :: binary(),.

doc/Advanced-configuration.md
```diff
@@ -129,7 +129,23 @@ The tuple order is important, unless the no `host_config` option is set. Retaini
...
+
+* **pool** (multi, local)
+    * **Description:** Declares a named pool of connections to the database. At least one pool is required to connect to an SQL database.
+    * **Syntax:** `{pool, odbc, PoolName}.` or `{pool, odbc, PoolName, Options}.`
+    * **Examples:** `{pool, odbc, "default"}.`
+
+* **odbc_pool** (local)
+    * **Description:** Name of the default connection pool used to connect to the database.
+    * **Syntax:** `{odbc_pool, PoolName}`
+    * **Default:** `"deafult"`
```
chenull: typo: `"default"`

apps/ejabberd/src/rdbms/mongoose_rdbms_sup.erl
```diff
@@ -30,85 +30,90 @@
...
             ?ERROR_MSG("Wrong odbc_pool_size definition '~p' "
-                       "for host ~p, default to ~p~n",
-                       [InvalidSize, Host, Size]),
+                       "for pool ~p, default to ~p~n",
+                       [InvalidSize, Pool, Size]),
             Size
     end.
+
+-spec get_option(mongoose_rdbms:pool(), atom()) -> term().
+get_option(Pool, Option) ->
```
michalwski: Would it feel more natural if this function was moved to `mongooseim_rdbms` module?

apps/ejabberd/test/mongoose_rdbms_SUITE.erl
```diff
@@ -86,10 +85,10 @@ keepalive_exit(Config) ->
 meck_config(Server, KeepaliveInterval) ->
     meck:new(ejabberd_config, [no_link]),
     meck:expect(ejabberd_config, get_local_option,
-                fun({odbc_keepalive_interval, _Host}) -> KeepaliveInterval;
-                   ({odbc_start_interval, _Host}) -> 30;
+                fun({odbc_keepalive_interval, odbc_pool, default}) -> KeepaliveInterval;
```
michalwski: I know this is a test. It's also a code we read and this 3 element tuple is not intuitive at first glance. Would it be possible to mock `mongoose_rdbms_sup:get_option` here instead of `ejabberd_config:get_local_option`?

That's OK if you don't have time for this @chrzaszcz now, I'd just like to know that.

apps/ejabberd/src/rdbms/mongoose_rdbms_sup.erl
chrzaszcz: The motivation here was to avoid cyclic dependencies between modules - I would be less reluctant to allow `mongoose_rdbms_sup` call `mongoose_rdbms` if the worker logic was extracted from the latter - which I would recommend.

apps/ejabberd/test/mongoose_rdbms_SUITE.erl
chrzaszcz: Your suggestion makes sense. I don't feel it's necessary, but it would improve the readability and unit test isolation indeed.

michalwski:
Thanks!

Issue #1218, opened 2017-03-13
mongoose_acc:read_children fails


ludwikbukowski:
MongooseIM version: 216b66d025ce5
Installed from: source
Erlang/OTP version: 18

`mongoose_acc:read_children/2` fails [1] when `Children` argument is `{xmlcdata, _}` element because of the `function_clause` of `exml_query:attr/3`.
This error causes dropping connection every time client sends stanza with some additional new line characters (e.g it happens when PSI client tries connect to the server and gets user's roster).

[1] https://github.com/esl/MongooseIM/blob/master/apps/ejabberd/src/mongoose_acc.erl#L160

michalwski:
@bartekgorny could you take care of it independently form PR #1211?

ludwikbukowski:
fixed in https://github.com/esl/MongooseIM/pull/1220

PR #1220, opened 2017-03-14
fix for #1218, with regression test

bartekgorny:
This PR addresses #1218 

Proposed changes include:
* improvement to mongoose_acc:read_children, so that it doesn't try to process cdata, plus stops when it finds xmlns
* test for the fix (failed in the old version)


michalwski:
Thanks!

PR #1222, opened 2017-03-14
mod_push fixes

astro:
Hi

These are just three little fixes. Justifications are in the commit messages.

### DIFF
--- a/apps/ejabberd/src/node_push.erl
+++ b/apps/ejabberd/src/node_push.erl
@@ -94,11 +94,12 @@ publish_item(Nidx, Publisher, Model, MaxItems, ItemId, ItemPublisher, Payload, P
                    _ -> get_state(Nidx, SubKey)
                end,
     Affiliation = SubState#pubsub_state.affiliation,
+    ElPayload = [El || #xmlel{} = El <- Payload],
 
     case is_allowed_to_publish(Model, Affiliation) of
         true ->
             do_publish_item(Nidx, Publisher, Model, MaxItems, ItemId, ItemPublisher,
-                            Payload, PublishOptions);
+                            ElPayload, PublishOptions);
         false ->
             {error, ?ERR_FORBIDDEN}
     end.


apps/ejabberd/src/mod_push_service_mongoosepush.erl
```diff
@@ -44,7 +44,8 @@
 %%--------------------------------------------------------------------
 
 -spec start(Host :: ejabberd:server(), Opts :: list()) -> any().
-start(Host, Opts) ->
+start(ServerHost, Opts) ->
```
gadgetci: According to **Compiler**:
> Warning: variable 'ServerHost' is unused

apps/ejabberd/src/mod_push_service_mongoosepush.erl
```diff
@@ -44,7 +44,8 @@
 %%--------------------------------------------------------------------
 
 -spec start(Host :: ejabberd:server(), Opts :: list()) -> any().
-start(Host, Opts) ->
+start(_ServerHost, Opts) ->
+    Host = ?MYNAME,
```
rslota: This change may lead to module crashes when the server is configured with several virtual-hosts, because the module will start multiple times with the same `Host`. Why this is an issue for you? Sure, the current code will start multiple VHosts and the pubsub will use only one of them, but "it should work". The correct fix here would be not to use `?MYNAME ` in `node_push`. I don't really remember why I did that, but it may be because it was not trivial to get the host in that context.

Anyway, if you let me know what the real issue is, we may find better solution to this problem.

apps/ejabberd/src/node_push.erl
```diff
@@ -207,20 +208,16 @@ is_allowed_to_publish(PublishModel, Affiliation) ->
 parse_form(undefined) ->
     #{};
 parse_form(Form) ->
-    IsForm = ?NS_XDATA == exml_query:attr(Form, <<"xmlns">>),
-    IsSubmit = <<"submit">> == exml_query:attr(Form, <<"type">>),
```
rslota: `type` is mandatory here. XEP-0357 specifies that this push node shall get `publish-options` form. The XEP-0060 (Publish-Subscribe) says that form inside of `publish-options` has to be the one defined in XEP-0004 (Data Forms). Finally XEP-0004 defines that `type` field of `x` element is `required` and has to be set to one of the following: `submit`, `cancel`, `result`, `form`. For reference see: [https://xmpp.org/extensions/xep-0004.html#schema](https://xmpp.org/extensions/xep-0004.html#schema)

astro: I was referring to [XEP-0357](https://xmpp.org/extensions/xep-0357.html#publishing). Multiple implementations could already exist that do not fit this requirement.

[Please be liberal in what you accept from others.](https://en.wikipedia.org/wiki/Robustness_principle)

michalwski: Hi @astro,

I understand this `type` check was an issue for you. Could you give us more details? What implementation do you have in mind?

apps/ejabberd/src/mod_push_service_mongoosepush.erl
michalwski: I agree with @rslota, there is many installations where we need multiple XMPP host served by the same MongooseIM node.

The root cause is most probably in other place: https://github.com/esl/MongooseIM/pull/1222/files#diff-64730cfd8dcd5857c353e7ed3413f59dL108 The implementation now is bound to the main domain anyway, which is not a perfect solution. I somehow missed that while doing code review. The `Host` parameter which should be used is Publisher's host if I understand it correctly. @rslota, what's your opinion on that?

@astro could you give us more details what was the main issue for you here?

rslota: @michalwski Yes, it should. The issue was that `Host` is not available in the `node_push'a` callbacks. I must have forgotten to to change `mod_pubsub` to include this variable in the `publish` callback. I'm gonna take care of that.

apps/ejabberd/src/node_push.erl
rslota: @astro @michalwski I didn't see that there is a mistake in examples from [XEP-0357](https://xmpp.org/extensions/xep-0357.html#publishing). Because of that we may skip this check but not this way in my opinion. My proposition is that we leave this check active, but we can default to `submit` form `type` in case this attribute is missing. This way, the current implementations that don't send this attribute will work and the code will reject wrong form types like `cancel`. What do you think about it @michalwski ?

michalwski: I agree with you @rslota

apps/ejabberd/src/mod_push_service_mongoosepush.erl
rslota: @michalwski Fixed in #1233 , so Now the change from this PR is not needed / applicable. Therefore, @astro, could you please revert this change after #1233 is merged?

apps/ejabberd/src/node_push.erl
rslota: @michalwski Fixed in #1234 :) @astro , could you please revert this change after #1234 is merged?

astro: Great, I'm happy with this solution and will remove the commit from this PR.

michalwski: #1234 is merged, waiting for travis to merge #1233 as well

astro:
Ok, removed the two obsolete commits. Just the approved one is left now.

I'm also happy about e7919ddf11ff76aa4d0046a0a8ea24e3544554f9 which duplicates a change I would have pushed as well. :)

michalwski:
Hi @astro,

Any chance you rebase your branch or solve conflicts soon? I'd like to merge your patch tomorrow.

michalwski:
All right, we cherry-picked your changes in #1245

michalwski:
Thanks for the patch @astro

PR #1228, opened 2017-03-15
TLS distribution

erszcz:
This PR allows to build a MongooseIM cluster using TLS for Distributed Erlang internal communication. TLS distribution is opt-in - it has to be explicitly enabled.

The changes include:

- A new config file - `vm.dist.args` - which also contains command line parameters to the Erlang VM, but is separate from `vm.args` for convenient reuse from control scripts. `vm.args` contains the `-(s)name` flag, therefore can't be reused for multiple VM instances.

- New procedure of generating certificates, which simulates signing the server certificate by a certificate authority.

- Support in big tests' Makefile for running the whole test suite against a TLS-enabled cluster. This requires the test node to also use TLS for RPC calls to the server nodes, hence reuse of the server and CA certificates.

Primer on TLS setup by @legoscia: https://www.erlang-solutions.com/blog/erlang-distribution-over-tls.html

**The remaining open issue** is whether we should have a separate Travis job testing all suites with TLS on or not.


michalwski:
> The remaining open issue is whether we should have a separate Travis job testing all suites with TLS on or not.

IMO we definitely should have it. I'm OK with reusing existing job for that, `internal_mnesia` for instance. I'd like to avoid adding one more job if possible.

michalwski:
Thanks!

PR #1233, opened 2017-03-21
Pass ServerHost to mod_pubsub's publish_item callback

rslota:
This PR fixes `node_push` behaviour when used with multiple virtual hosts. This fix required passing `ServerHost` variable to `publish_item` callback. Without It, it would be possible to fetch from `ets` pubsub's host in form of 'pubsub.somehost.com', but that would require getting the host prefix from configuration in order to strip it before running callback. Therefore passing the original `ServerHost` seemed to be better idea, although it's not as pretty as I would like it to be.

### DIFF
--- a/apps/ejabberd/src/node_push.erl
+++ b/apps/ejabberd/src/node_push.erl
@@ -20,7 +20,7 @@
 -export([init/3, terminate/2, options/0, features/0,
          create_node_permission/6, create_node/2, delete_node/1,
          purge_node/2, subscribe_node/8, unsubscribe_node/4,
-         publish_item/8, delete_item/4, remove_extra_items/3,
+         publish_item/9, delete_item/4, remove_extra_items/3,
          get_entity_affiliations/2, get_node_affiliations/1,
          get_affiliation/2, set_affiliation/3,
          get_entity_subscriptions/2, get_node_subscriptions/1,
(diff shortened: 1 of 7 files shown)


apps/ejabberd/src/node_push.erl
```diff
@@ -97,25 +98,25 @@ publish_item(Nidx, Publisher, Model, MaxItems, ItemId, ItemPublisher, Payload, P
...
         true ->
-            do_publish_item(Nidx, Publisher, Model, MaxItems, ItemId, ItemPublisher,
+            do_publish_item(ServerHost, Nidx, Publisher, Model, MaxItems, ItemId, ItemPublisher,
                             Payload, PublishOptions);
         false ->
             {error, ?ERR_FORBIDDEN}
     end.
 
-do_publish_item(_Nidx, _Publisher, _Model, _MaxItems, _ItemId, _ItemPublisher,
+do_publish_item(ServerHost, _Nidx, _Publisher, _Model, _MaxItems, _ItemId, _ItemPublisher,
```
michalwski: Can all the vars with `_` prefix be removed? We probably don't need them and without them the function head will be much more readable.

rslota: Sure, done.

apps/ejabberd/src/node_push.erl
```diff
@@ -97,26 +98,24 @@ publish_item(Nidx, Publisher, Model, MaxItems, ItemId, ItemPublisher, Payload, P
...
+            do_publish_item(ServerHost, PublishOptions, Payload);
         false ->
             {error, ?ERR_FORBIDDEN}
     end.
 
-do_publish_item(_Nidx, _Publisher, _Model, _MaxItems, _ItemId, _ItemPublisher,
-                [#xmlel{name = <<"notification">>} | _] = Notifications, PublishOptions) ->
-    Host = ?MYNAME,
+do_publish_item(ServerHost, PublishOptions,
+                [#xmlel{name = <<"notification">>} | _] = Notifications) ->
```
michalwski: Is it possible the `#xmlel{name = <<"notification">>}` will not be the first element?

PR #1242, opened 2017-03-24
Mod push tutorial

rslota:
This PR introduces the first draft of the Push Notifications tutorial. Pretty component pictures in progress.


doc/user-guide/Push-notifications.md
```diff
@@ -0,0 +1,188 @@
+## Push Notifications
```
michalwski: If you want this file to be visible on readthedocs, you need to add it to mkdocs.yml file.

rslota: Added now.

michalwski:
A preview of this tutorial can be found here: http://mongooseim.readthedocs.io/en/mod_push_tutorial/user-guide/Push-notifications/

doc/user-guide/Push-notifications.md
```diff
@@ -0,0 +1,204 @@
...
+After this step, your application shall be able to receive _FCM_ or _APNS_ token - it will be required in the next step of this tutorial.
+
+### Setting up XMPP `pubsub` node
+
+First thing that the clinet has to set up on XMPP server is the `PubSub`'s node for handling push notifications.
+Assuming that `mypubsub.com` is a domain of MongooseIM server that has [mod_pubsub][] enabled with `push` node support, the client shall send the following stanza to the server:
+
+```xml
+<iq type='set'
+    from='me@mongooseim.com'
```
michalwski: I didn't see domain `mongooseim.com` anywhere else in this article. Should it be `mychat.com` instead?

doc/user-guide/Push-notifications.md
```diff
@@ -0,0 +1,204 @@
...
+If you don't want to use either _APNS_ or _FCM_, you simply need to pass `PUSH_APNS_ENABLED=0` or `PUSH_FCM_ENABLED=0` respectively as additional env variables in your `docker run` command.
+For more advanced options and configuration please refer to _"Quick start / Configuring"_ in [MongoosePush][]'s [_README.md_][MongoosePushReadme].
+
+When your [MongoosePush][] docker is up and running, Push Notifications now can be used in your MongooseIM instance.
+
+## Using push notifications on client side
+
+The XMPP client application has very few things to do in order to receive push notifications.
+The whole process described in this section along with example notification flow we can see in the following diagram:
+ ![](push_notifications_example_message_flow.svg)
```
michalwski: A not very experienced XMPP user by looking at steps 2 and 3 in this diagram may think (s)he needs 2 different accounts on servers `mychat.com` and `mypubsub.com`. Also this means that 2 different connections needs to be established. I think this flow should look like the following:
* step 2 - client sends the pubsub node create request to `mychat.com`
* step 3 - `mychat.com` contacts `mypubsub.com` on behalf of the user
* step 4 - user enables push notifications (as in the diagram already).

doc/user-guide/Push-notifications.md
rslota: That's a good catch. It would indeed be much better to just write `mychat.com`. I'll do that.

doc/user-guide/Push-notifications.md
rslota: I totally agree. Your way is much more explicit. I'm gonna make it this way.

doc/user-guide/Push-notifications.md
michalwski: You can also omit the `from`. This should be added by the server in missing.

rslota: Even better.

doc/user-guide/Push-notifications.md
rslota: What do you think about this drawing now?

michalwski: I think that's far better, I'm not sure it's natural though. Maybe it would be more natural if the step 2 was split into 2a and 2b, and this 2 lines doesn't have common point on MongooseIM.
@Nyco, @goddammit are you able to advice here?

michalwski:
@rslota I've just noticed that links to `MongoosePush` goes nowhere.

rslota:
> @rslota I've just noticed that links to MongoosePush goes nowhere.

Thanks! I've fixed the typo.

Nyco:
Can we, should we adopt a more top-down approach, since we offer choice of technology regarding push?

That would be:
* Flexible push notifications
  * mod_http_notification
  * Amazon SNS
  * XEP-0357: Push Notifications

@rslota @michalwski @goddammit

doc/user-guide/Push-notifications.md
```diff
@@ -0,0 +1,205 @@
+## Push Notifications
+
+MongooseIM server supports push notifications using FCM (**F**irebase **C**loud **M**essaging)
+and APNS (**A**pple **P**ush **N**otification **S**ervice) providers. Server side push
+notification support is fully compliment with [XEP-0357 Push Notifications][XEP-0357], which defines several components that need to work together in order to provide clients with working push notifications.
```
Nyco: s/compliment/compliant/ ?

pwetrifork:
My remarks regarding the client perspective document (http://mongooseim.readthedocs.io/en/mod_push_tutorial/user-guide/Push-notifications/#using-push-notifications-on-client-side):

* The diagram is a bit overwhelming when you first look at it. Consider splitting into registration/messaging cases. I am also not sure if a client developer needs such detailed view of platform internals (how about merging mypubsub/mychat/push.mypubsub into some umbrella "Moongoose/XMPP" entity on the diagram?)
* The guide assumes (perhaps rightly) a decent understanding of pubsub. If we wanted to provide a more general description of the architecture in layman's terms:
  * when pubsub.mypubsum.com JID is first mentioned, we could compare this entity to an app-level XMPP<->push services gateway
  * before pubsub node creation/enabling is described in detail, introduce a node as an "entry path" to the individual device; I'd also suggest changing the example name "princely_musings" to something that indicates node per-device uniquenes
* A scheme where the client is responsible for pubsub node creation seems like a departure from 0357 (where they describe App Server doing this after being provided with a token), I think this should be stressed somewhere. It was actually the hardest part for me to understand when I was first learning about 0357 MIM implementation

michalwski:
@Nyco, regarding other push notifications we provide I think it'd be good to have it described somewhere the way you proposed. This particular tutorial is only about the XEP and I think it should remain like that.

rslota:
> The diagram is a bit overwhelming when you first look at it. Consider splitting into registration/messaging cases

This is quite reasonable. The downside is that we would have two almost identical diagrams. Right now, there is such separation in form of different arrows colour.

> I am also not sure if a client developer needs such detailed view of platform internals (how about merging mypubsub/mychat/push.mypubsub into some umbrella "Mongoose/XMPP" entity on the diagram?)

I wanted to show those components apart because this covers all use cases. The XEP assumes that the app-server with pubsub relay may and probably will be on different domain, so I think we should show this approach. Especially that this approach is more generic one. 

> when pubsub.mypubsum.com JID is first mentioned, we could compare this entity to an app-level XMPP<->push services gateway

That's true. I'm on it.

> before pubsub node creation/enabling is described in detail, introduce a node as an "entry path" to the individual device; I'd also suggest changing the example name "princely_musings" to something that indicates node per-device uniquenes

Also very good idea.

> A scheme where the client is responsible for pubsub node creation seems like a departure from 0357 (where they describe App Server doing this after being provided with a token), I think this should be stressed somewhere. It was actually the hardest part for me to understand when I was first learning about 0357 MIM implementation.

The problem with XEP-0357 is that it doesn't state how this allocation is done. In MongooseIM also the "app-server" does that for you, but the "registration" process is done via pubsub node allocation message, since MongooseIM is not dedicated XMPP Push Service. I think the authors of the XEP wanted a dedicated XMPP service that will alocate and prepare such node for it's every user. Not sure though.

rslota:
@Nyco 
> Can we, should we adopt a more top-down approach, since we offer choice of technology regarding push?

Are you suggesting describing all these in this document? I would suggest leaving the "other" push notifications out of this scope. XEP-0357 defines "the" push notifications and this tutorial is just about them. 

@michalwski what do you think about it?

rslota:
> @Nyco, regarding other push notifications we provide I think it'd be good to have it described somewhere the way you proposed. This particular tutorial is only about the XEP and I think it should remain like that.

I totally agree. We could describe the "other" push notifications somewhere, but this document is about "the" push notifications as described in XEP-0357.

doc/user-guide/Push-notifications.md
rslota: Thanks! Will do.

doc/user-guide/Push-notifications.md
rslota: Gonna do this the way you're suggesting.

doc/user-guide/Push-notifications.md
```diff
@@ -0,0 +1,205 @@
...
+Firstly, let's configure all required MongooseIM components, step by step.
+
+#### [mod_push][] a.k.a. '_XMPP Server_'
+
+The first component that we need to configure in MongooseIM is [mod_push][] module.
+This module directly communicates with XMPP clients in order to enable/disable notifications on per-client basis.
+
+The [mod_push][] module is very easy to enable - just paste in your MongooseIM configuration file:
+
+ ```Erlang
```
michalwski: This code is rendered badly on readthedocs. Should the language start with lowercase?

doc/user-guide/Push-notifications.md
```diff
@@ -0,0 +1,209 @@
...
+
+* priv/
+    * ssl/
+        * rest_cert.pem - The REST endpoint certificate
+        * rest_key.pem - private key for the REST endpoint certificate
+    * apns/
+        * prod_cert.pem - Production APNS app certificate
+        * prod_key.pem - Production APNS app certificate's private key
+        * dev_cert.pem - Development APNS app certificate
+        * dev_jey.pem - Development APNS app certificate's private key
```
michalwski: Should it be `dec_key.pem` instead of `dev_jey.pem`?

michalwski: I've made a typo commenting on a typo :laughing: Of course I suggested `dev_key.pem` :)

Nyco:
XEP-0357 is not mentionned in our "Features and supported standards":
https://mongooseim.readthedocs.io/en/mod_push_tutorial/user-guide/Features-and-supported-standards/

doc/user-guide/Push-notifications.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Registering with Push Service provider
+
+Firstly, client application has to get a device-specific token from Push Service Provider (FCM or APNS). This process is different, depending on the platform, so please consult you Push Service Provider's manual to learn about how to get this token. For example, [here](https://firebase.google.com/docs/cloud-messaging/android/client) you can lear about setting up _FCM_ on _Android_ platform and [here](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/HandlingRemoteNotifications.html#/apple_ref/doc/uid/TP40008194-CH6-SW1) you can learn about setting up _APNS_ on _iOS_ platform.
+
+After this step, your application shall be able to receive _FCM_ or _APNS_ token - it will be required in the next step of this tutorial.
+
+### Setting up XMPP `pubsub` node
+
+First thing that the clinet has to setup on XMPP server is the `PubSub`'s node for handling push notications. Assuming that `mongooseim-pubsub.com` is an domain of MongooseIM server that has [mod_pubsub][] enabled with `push` node support, the client shall send the following stanza to the server:
```
deadjdona: clinet -> client

PR #1253, opened 2017-03-31
Honesty update

Nyco:
Now that we have a clearer vision of 2.1.x because we released 2.1.0-beta1, we can add/remove items with much more confidence

### DIFF
--- a/doc/Roadmap.md
+++ b/doc/Roadmap.md
@@ -7,11 +7,9 @@ Please feel free to comment!
 * Push notifications:
     * [XEP-0357: Push Notifications](https://xmpp.org/extensions/xep-0357.html)
     * Amazon AWS SNS (Simple Notification Service)
-* Instant Stream Resumption
 * Full text search in archive
-* Bind 2.0 (experimental)
-* Contributions to Smack, XMPPFramework
-* Contributions to Jayme? Retrofit? Movim? Stanza.io? Strophe.js? Libervia?
+* Experimental pipelining
+* Contributions to ecosystem
 
 ## 3.0.x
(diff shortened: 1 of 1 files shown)


doc/Roadmap.md
```diff
@@ -7,18 +7,17 @@ Please feel free to comment!
 * Push notifications:
     * [XEP-0357: Push Notifications](https://xmpp.org/extensions/xep-0357.html)
     * Amazon AWS SNS (Simple Notification Service)
-* Instant Stream Resumption
 * Full text search in archive
-* Bind 2.0 (experimental)
```
michalwski: This is still valid. Bind 2.0 is not in beta1, it doesn't mean it will not fit into final 2.1

Nyco: Ok, once again my twisted mind that refuses to understand that. Fixed.

doc/Roadmap.md
```diff
@@ -7,18 +7,19 @@ Please feel free to comment!
 * Push notifications:
     * [XEP-0357: Push Notifications](https://xmpp.org/extensions/xep-0357.html)
     * Amazon AWS SNS (Simple Notification Service)
-* Instant Stream Resumption
 * Full text search in archive
+* Experimental pipelining
+* Instant Stream Resumption
```
michalwski: The pipelining we have doesn't equal `Instant Stream Resumption` as in the proto-XEP.

michalwski: But we can still try to deliver this, as soon as new `ISR` is accepted as draft which I see possible.

PR #1254, opened 2017-03-31
Rock mod_register..mod_websockets

fenek:
Another song written together with our King of Rock.

### DIFF
--- a/apps/ejabberd/src/mod_roster.erl
+++ b/apps/ejabberd/src/mod_roster.erl
@@ -137,8 +137,6 @@
     Result :: error | roster().
 
 
--define(BACKEND, mod_roster_backend).
-
 start(Host, Opts) ->
     IQDisc = gen_mod:get_opt(iqdisc, Opts, one_queue),
     TrackedFuns = [read_roster_version,
@@ -153,7 +151,7 @@ start(Host, Opts) ->
                    read_subscription_and_groups
                    ],
     gen_mod:start_backend_module(?MODULE, Opts, TrackedFuns),
-    ?BACKEND:init(Host, Opts),
+    mod_roster_backend:init(Host, Opts),
 
     ejabberd_hooks:add(roster_get, Host,
                        ?MODULE, get_user_roster, 50),
@@ -245,7 +243,7 @@ roster_version(LServer, LUser) ->
     end.
 
 read_roster_version(LUser, LServer) ->
-    ?BACKEND:read_roster_version(LUser, LServer).
+    mod_roster_backend:read_roster_version(LUser, LServer).
 
 write_roster_version(LUser, LServer) ->
     write_roster_version(LUser, LServer, false).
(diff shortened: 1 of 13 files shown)


apps/ejabberd/src/mod_roster.erl
```diff
@@ -781,28 +775,17 @@ send_unsubscription_to_rosteritems(LUser, LServer) ->
...
-                 both -> true;
-                 from -> true;
-                 _ -> false
-             end,
-    if IsTo ->
-           send_presence_type(jid:to_bare(From),
-                              jid:make(Item#roster.jid),
-                              <<"unsubscribe">>);
-       true -> ok
+send_unsubscribing_presence(From, #roster{ subscription = Subscription } = Item) ->
```
arcusfelis: shorted, but harder to read

apps/ejabberd/src/mod_roster_mnesia.erl
```diff
@@ -56,14 +56,10 @@ read_roster_version(LUser, LServer) ->
...
 
-write_roster_version(LUser, LServer, InTransaction, Ver) ->
-    US = {LUser, LServer},
-    if InTransaction ->
-           mnesia:write(#roster_version{us = US, version = Ver});
-       true ->
-           mnesia:dirty_write(#roster_version{us = US,
-                                              version = Ver})
-    end.
+write_roster_version(LUser, LServer, true, Ver) ->
```
arcusfelis: true what? Do at least
```erlang
write_roster_version(LUser, LServer, _InTransaction=true, Ver)
```

apps/ejabberd/src/mod_websockets.erl
```diff
@@ -197,6 +197,11 @@ do_start_fsm(FSMModule, Opts, Req, State) ->
             {shutdown, NewReq, State}
     end.
 
+call_fsm_start(ejabberd_c2s, SocketData, Opts) ->
```
arcusfelis: cool

apps/ejabberd/src/mod_roster.erl
michalwski: I partially agree with @arcusfelis here. Maybe leaving `IsTo` and `IsFrom` vars will make it easier to understand. Of course new way of deciding on the value is far better.

apps/ejabberd/src/mod_roster_mnesia.erl
michalwski: Good idea @arcusfelis :)

michalwski: Or maybe we could replace the boolean for `IsTransaction` with some meaningful atom? Like `in_transaction` and `no_transaction`?

apps/ejabberd/src/mod_vcard.erl
```diff
@@ -539,7 +541,7 @@ prepare_vcard_search_params(User, VHost, VCARD) ->
 
     US = {LUser, VHost},
 
-    if
+    case
```
michalwski: Maybe the case could be avoided completely if `b2l` throws a meaningful error when the input is sth else then a binary? Or simply expect badarg when constructing the `#vacard_search` record, this is probably to generic, though.

fenek: This function is executed always inside `try..catch` that returns `internal-server-error` to the client. I think it is good enough, as it's very difficult to trigger this error and client would have to send malformed data on purpose (or be *very* unlucky). Unfortunately XEP doesn't specify any specific error response for invalid VCard fields.

I'm going to remove this `case` and "let it crash". Do you agree?

apps/ejabberd/src/mod_roster_mnesia.erl
fenek: I'm going to add a spec to this function, so the meaning of the 3rd arg will become clear.

michalwski: All right, spec will help here. Maybe it's not worth changing the boolean to a meaningful atom for this code.

In general I think in many places where we need a boolean we can have a meaningful atom. Just an idea to think about :)

apps/ejabberd/src/mod_vcard.erl
michalwski: If only the `internal-server-error` is returned in case of `{error, badarg`} returned from this function, your suggestion will be enough.

michalwski:
Thanks!

PR #1256, opened 2017-03-31
Add `import_users` command to mongooseimctl

bszaf:
This allows to register users specified in CSV file in following
format: `user,domain,password`

- Proposed change is adding new command `import_users` to mongooseimctl.
  - It expects one parameter which is path to CSV file containing user accounts to register. They are expected to be formatted in following order - `user,host,password`. 
  - File reading is lazy, so we are not loading whole file into memory at once.
  - The `ejabberd_admin:import_users/1` function returns list of tuples `{Result, Binary}`, where `Result :: ok | exists | not_allowed | invalid_jid | null_password | bad_csv `, and `Binary :: ejabberd:user() | binary()`.
  - Library `csv` is added as dependency. It is used for parsing csv files.
- describe new or updated tests
  - add suite for this functionality.

### DIFF
--- a/apps/ejabberd/src/ejabberd_admin.erl
+++ b/apps/ejabberd/src/ejabberd_admin.erl
@@ -411,7 +411,10 @@ do_register([User, Host, Password]) ->
     end;
 
 do_register(List) ->
-    Info = list_to_binary(lists:join(<<",">>, List)),
+    JoinBinary = fun(Elem, <<"">>) -> Elem;
+                    (Elem, Acc) -> <<Elem/binary, ",", Acc/binary>>
+                 end,
+    Info = lists:foldr(JoinBinary, <<"">>, List),
     {bad_csv, Info}.
 
 get_loglevel() ->
--- a/apps/ejabberd/src/ejabberd_admin.erl
+++ b/apps/ejabberd/src/ejabberd_admin.erl
@@ -51,7 +51,7 @@
          join_cluster/1, leave_cluster/0,
          remove_from_cluster/1]).
 
--export([register_worker/1]).
+-export([registrator_proc/1]).
 
 -include("ejabberd.hrl").
 -include("ejabberd_commands.hrl").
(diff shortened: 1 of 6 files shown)


apps/ejabberd/src/ejabberd_admin.erl
```diff
@@ -325,6 +337,86 @@ registered_users(Host) ->
...
+                                             {exists, ejabberd:user()} |
+                                             {not_allowed, ejabberd:user()} |
+                                             {invalid_jid, ejabberd:user()} |
+                                             {null_password, ejabberd:user()} |
+                                             {bad_csv, binary()}].
+import_users(File) ->
+    FileStream = stdio:file(File),
+    CsvStream = csv:stream(FileStream),
+    Workers = spawn_link_workers(),
+    WorkersQueue = queue:from_list(Workers),
```
michalwski: This looks like a worker pool written from scratch. How about using https://github.com/inaka/worker_pool for that? We already have it in some places in MongooseIM.

apps/ejabberd/src/ejabberd_admin.erl
```diff
@@ -325,6 +337,86 @@ registered_users(Host) ->
...
+     {null_password, ejabberd:user()} |
+     {bad_csv, binary()}].
+do_import({}, WorkersQueue) ->
+    Workers = queue:to_list(WorkersQueue),
+    lists:flatmap(fun get_result/1, Workers);
+
+
+do_import({s, UserData, TailFun}, WorkersQueue) ->
+    {{value, Worker}, Q1} = queue:out(WorkersQueue),
+    Worker ! {proccess, UserData},
```
michalwski: I suggest hiding this message passing inside a function like `register_user(Worker, UserData)`

apps/ejabberd/src/ejabberd_admin.erl
```diff
@@ -325,6 +337,86 @@ registered_users(Host) ->
...
+                                      {null_password, ejabberd:user()} |
+                                      {bad_csv, binary()}].
+get_result(Pid) ->
+    Pid ! get_result,
+    receive
+        {result, Result} -> Result
+    end.
+
+-spec register_worker(Manager :: pid()) -> ok.
+register_worker(Manager) ->
```
michalwski: The function name is misleading. It suggest that it will register a worker in some registry, while in fact this is the main function of a worker registering users.

michalwski:
Thanks!

PR #1261, opened 2017-04-04
[skip ci] mods doc

goddammit:
mods mods mods... 
documentation


doc/modules/mod_carboncopy.md
```diff
@@ -1,37 +1,54 @@
 ### Module Description
 
 ##### Discovering Support
-The server informs in a disco query if carbons are enabled.
+The server uses a disco query to inform if carbons are enabled.
 
 ##### Enabling and disabling Carbons from the client
-Each client for a user can enable/disable carbons for itself by sending an iq stanza with a child element respectively `<enable xmlns='urn:xmpp:carbons:2'/>` or `<disable xmlns='urn:xmpp:carbons:2'/>`
+Each client can enable/disable carbons for its users by sending an iq stanza with a child element respectively `<enable xmlns='urn:xmpp:carbons:2'/>` or `<disable xmlns='urn:xmpp:carbons:2'/>`
```
michalwski: Here `client` is not `customer` is just a user's application - client of the server. Heaving that the said. The `client` will have only one `user`.  I'd put it like this: "Carbons are not enabled by default. Every client app has to enable carbons in order to get messages send to other clients of the user"

Also maybe this will help to understand the big picture. One MongooseIM's `user` can have many `client` applications.

goddammit: Enabling and disabling Carbons from the client

Carbons are not enabled by default. Every client app has to enable carbons to get messages sent to other clients of the user. Carbons are enabled and disabled with an iq stanza with a child element - <enable xmlns='urn:xmpp:carbons:2'/> or <disable xmlns='urn:xmpp:carbons:2'/>. ?

PR #1272, opened 2017-04-07
Fix node push discovery in mod_pubsub

rslota:
This PR addresses the following issues:
- node `push` disco info shall be visible under pubsub's sub domain, instead of the main server's domain
- `mod_pubsub` should allow for running on custom sub-domain

Please note that the second fix only allows to run pubsub on custom, non "pubsub.*" sub-domain. You can still run only one pubsub at a time. This is another issue which is much more complicated to fix, because the `mod_pubsub` internally holds its state on the main domain of the server. Therefore starting second `mod_pubsub` will always conflict with the first one.


michalwski:
Thanks!

PR #1278, opened 2017-04-10
Use empheral ports for http mocks in push_pubsub_SUITE

rslota:
This PR fixes possible TCP port conflict in `push_pusbsub_SUITE` that sometimes crashes the test SUITE.

### DIFF
--- a/test.disabled/ejabberd_tests/tests/push_pubsub_SUITE.erl
+++ b/test.disabled/ejabberd_tests/tests/push_pubsub_SUITE.erl
@@ -50,7 +50,6 @@ suite() ->
 
 init_per_suite(Config) ->
     application:ensure_all_started(cowboy),
-    MongoosePushMockPort = crypto:rand_uniform(20000, 50000),
 
     %% For mocking with unnamed functions
     {_Module, Binary, Filename} = code:get_object_code(?MODULE),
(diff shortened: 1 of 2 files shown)


test.disabled/ejabberd_tests/tests/push_pubsub_SUITE.erl
```diff
@@ -371,10 +369,13 @@ bare_jid(JIDOrClient) ->
...
 %% REST mock handler
-setup_mock_rest(Port) ->
+setup_mock_rest() ->
     TestPid = self(),
     HandleFun = fun(Req) -> handle(Req, TestPid) end,
-    http_helper:start(Port, "/[:level1/[:level2/[:level3/[:level4]]]]", HandleFun).
+    {ok, _} = http_helper:start(0, "/[:level1/[:level2/[:level3/[:level4]]]]",
+                                          HandleFun),
+    Info = http_helper:info(),
+    proplists:get_value(port, Info).
```
michalwski: This is correct, thanks.

Would it be simpler and shorter to just use `ranch:get_port(http_helper_listener)`. This can be encapsulated inside `http_helper` module as `info`.

At this moment I don't see usage of any other data from the `Info` list. That's why I think YAGNI :) and KISS :)

rslota: I've changed the `info/0` to `port/0`. It's even simpler :)

michalwski: Much simpler, thanks a lot!

michalwski:
Thanks!

Issue #1277, opened 2017-04-10
Vulnerability that can crash the nodes without too much work


silviucpp:
Vulnerability in ejabberd (only ESL versions).

Idea behind the vulnerability:

In order to limit the max stanza size sent by a user ESL is doing :

1. decode the entire stanza into erlang term (xmlel record)
2. transform it back in binary to see how many bytes has and if it's over the limit will send back to user an error

In ESL (https://github.com/esl/MongooseIM/blob/167e2755dea7b3708dd13925c6ee6b0d1821fb6d/apps/ejabberd/src/ejabberd_receiver.erl#L325)

In P1 version the check is taking place during parsing and not like in ESL after parsing was completed.

What's the vulnerability:

Because limit checking is not happening during parsing any malicious client can send a very big stanza in time (that exceed the memory available on the node), without hitting the other limits like bytes/second that can be sent.

Attach you can find a small script that can be used to crash the new in time. Of course can be dome more aggressive  by opening multiple sessions.

[crash_server.zip](https://github.com/esl/MongooseIM/files/910121/crash_server.zip)

Kind regards,
Silviu

kzemek:
Fixed in #1285

PR #1285, opened 2017-04-14
Update exml for low-level stanza size limiting.

kzemek:
This PR addresses #1277 

Limiting stanza size at NIF level allows us to catch overweight stanzas as early as possible, before we've processed the whole thing.


michalwski:
Thanks for the fix!

Issue #1248, opened 2017-03-29
Adopt logfmt [discussion]


arcusfelis:
Logfmt is simple log format https://brandur.org/logfmt

Example of log-message:

```javascript
msg="Success authentication" user="alice@example.com" auth_method=plain
msg="Route stanza" user="alice@example.com" dest="bob@example.com" stanza="<message>...</message>"
msg="Execute SQL query" user="alice@example.com" sql_query="SELECT * FROM users WHERE user=?"
```

We will need a document with fields and description. For example,

- msg - log event description
- user - user's JID
- sql_query - readable SQL quesy string
...

Pros:
- easy to read
- splunk/ELK friendly
- easy to grep logs

fenek:
I agree. The completely mixed and random log format in MongooseIM has been annoying me for quite some time but it's a cross-module effort to unify all log messages, probably to be done in stages (like `mongoose_acc effort`).

michalwski:
👍

arcusfelis:
@fenek I agree, we just need to decide on format first.
And then we can use the format for new messages.
No need to make to do large refactoring to just start using it.

fenek:
I think only these should be mandatory:

`event="stanza_arrived" user="user@domain.com" from="from@domain.com" to="to@domain.com"`

`user`, `from` and `to` should obviously be included only when they are present at specific points in a flow. Remaining keys will depend on context, no point in trying to standardise them (IMO).

I think using lowercase alphanumeric characters with underscores for event names (or `msg`, whatever) will be less error-prone and will discourage lengthy error descriptions

bartekgorny:
I think the first step should be writing a generic function formatting `mongoose_acc` for use in logs. There is lot of debugging data, and acc is going to be available nearly everywhere.

arcusfelis:
Discussion is closed.
Adopt it in a separate pr, for example #1255.

PR #1286, opened 2017-04-14
Rest client api roster management

bartekgorny:
This PR addresses the need to have roster management in rest api.

* user commands to add/remove roster entry and send subscription request to the contact (we call it "invite", also to accept and invitation ("accept"), see caveat below
* admin commands to manage user's rosters and to set up connection between two users at one go

There is one minor issue: if a user adds a contact to his roster, but does not send subscription, and then the contact sends "subscribe" to the user, then the contact is in {none, in} state and as such, following the RFC, is not visible to the user. The user can not then accept the invitation because the roster entry is, from his point of view, nonexistent.

### DIFF
--- a/apps/ejabberd/src/mod_commands.erl
+++ b/apps/ejabberd/src/mod_commands.erl
@@ -15,6 +15,7 @@
          add_contact/2,
          add_contact/3,
          add_contact/4,
+         subscription/3,
          kick_session/3,
          get_recent_messages/3,
          get_recent_messages/4,
@@ -119,6 +120,18 @@ commands() ->
       {action, create},
       {security_policy, [user]},
       {args, [{caller, binary}, {jid, binary}]},
+      {result, ok}
+     ],
+     [
+      {name, subscription},
+      {category, <<"contacts">>},
+      {desc, <<"Send out a subscription request">>},
+      {module, ?MODULE},
+      {function, subscription},
+      {action, update},
+      {security_policy, [user]},
+      {identifiers, [jid]},
+      {args, [{caller, binary}, {jid, binary}, {action, binary}]},
       {result, ok}
      ],
      [
@@ -321,20 +334,26 @@ create_acc(CallerJid, Name, Type, OtherJid) ->
     mongoose_acc:put(to_jid, OtherJid, A1).
 
 
-%%invite(CallerJid, OtherJid) ->
-%%    A = create_acc(CallerJid, <<"presence">>, <<"subscribe">>, OtherJid),
-%%    El = #xmlel{name = <<"presence">>, attrs = [{<<"type">>, <<"subscribe">>}]},
-%%    Acc1 = mongoose_acc:put(element, El, A),
-%%    % set subscription to
-%%    Server = CallerJid#jid.server,
-%%    LUser = CallerJid#jid.luser,
-%%    LServer= CallerJid#jid.lserver,
-%%    Acc2 = ejabberd_hooks:run_fold(roster_out_subscription,
-%%        Server,
-%%        Acc1,
-%%        [LUser, LServer, OtherJid, subscribe]),
-%%    ejabberd_router:route(CallerJid, OtherJid, mongoose_acc:get(element, Acc2)),
-%%    ok.
+subscription(Caller, Other, Action) ->
+    Act = binary_to_existing_atom(Action, latin1),
+    run_subscription(Act, jid:from_binary(Caller), jid:from_binary(Other)).
+
+-spec run_subscription(subscribe | subscribed, jid(), jid()) -> ok.
+run_subscription(Type, CallerJid, OtherJid) ->
+    StanzaType = atom_to_binary(Type, latin1),
+    A = create_acc(CallerJid, <<"presence">>, StanzaType, OtherJid),
+    El = #xmlel{name = <<"presence">>, attrs = [{<<"type">>, StanzaType}]},
+    Acc1 = mongoose_acc:put(element, El, A),
+    % set subscription to
+    Server = CallerJid#jid.server,
+    LUser = CallerJid#jid.luser,
+    LServer= CallerJid#jid.lserver,
+    Acc2 = ejabberd_hooks:run_fold(roster_out_subscription,
+        Server,
+        Acc1,
+        [LUser, LServer, OtherJid, Type]),
+    ejabberd_router:route(CallerJid, OtherJid, mongoose_acc:get(element, Acc2)),
+    ok.
 
 
 
--- a/apps/ejabberd/src/mod_commands.erl
+++ b/apps/ejabberd/src/mod_commands.erl
@@ -225,7 +225,16 @@ list_contacts(Caller) ->
 
 roster_info(M) ->
     Jid = jid:to_binary(maps:get(jid, M)),
-    #{jid => Jid}.
+    #{subscription := Sub, ask := Ask} = M,
+    case get_state(Sub, Ask) of
+        undefined ->
+            #{jid => Jid};
+        State ->
+            #{jid => Jid, state => State}
+    end.
+
+get_state(none, out) -> <<"invitee">>;
+get_state(_, _) -> undefined.
 
 add_contact(Caller, JabberID) ->
     add_contact(Caller, JabberID, <<"">>, []).
@@ -235,7 +244,9 @@ add_contact(Caller, JabberID, Name) ->
 
 add_contact(Caller, JabberID, Name, Groups) ->
     CJid = jid:from_binary(Caller),
-    mod_roster:set_roster_entry(CJid, JabberID, Name, Groups).
+    OJid = jid:from_binary(JabberID),
+    mod_roster:set_roster_entry(CJid, JabberID, Name, Groups),
+    invite(CJid, OJid).
 
 registered_commands() ->
     [#{name => mongoose_commands:name(C),
@@ -307,4 +318,38 @@ lookup_recent_messages(ArcJID, WithJID, Before, Limit) ->
                                  IsSimple]),
     {ok, {_, _, L}} = R,
     L.
+
+create_acc(CallerJid, Name, Type) ->
+    A = mongoose_acc:new(),
+    Map = #{name => Name, type => Type, from => jid:to_binary(CallerJid),
+            from_jid => CallerJid},
+    mongoose_acc:update(A, Map).
+
+create_acc(CallerJid, Name, Type, OtherJid) ->
+    A = create_acc(CallerJid, Name, Type),
+    A1 = mongoose_acc:put(to, jid:to_binary(OtherJid), A),
+    mongoose_acc:put(to_jid, OtherJid, A1).
+
+
+invite(CallerJid, OtherJid) ->
+    A = create_acc(CallerJid, <<"presence">>, <<"subscribe">>, OtherJid),
+    El = #xmlel{name = <<"presence">>, attrs = [{<<"type">>, <<"subscribe">>}]},
+    Acc1 = mongoose_acc:put(element, El, A),
+    % set subscription to
+    Server = CallerJid#jid.server,
+    LUser = CallerJid#jid.luser,
+    LServer= CallerJid#jid.lserver,
+    Acc2 = ejabberd_hooks:run_fold(roster_out_subscription,
+        Server,
+        Acc1,
+        [LUser, LServer, OtherJid, subscribe]),
+    ejabberd_router:route(CallerJid, OtherJid, mongoose_acc:get(element, Acc2)),
+    ok.
+
+
+
+
+
+
+
 
--- a/apps/ejabberd/src/mod_commands.erl
+++ b/apps/ejabberd/src/mod_commands.erl
@@ -131,7 +131,8 @@ commands() ->
       {function, subscription},
       {action, update},
       {security_policy, [user]},
-      {identifiers, [jid]},
+      {identifiers, [caller, jid]},
+      % caller has to be in identifiers, otherwise it breaks admin rest api
       {args, [{caller, binary}, {jid, binary}, {action, binary}]},
       {result, ok}
      ],
--- a/apps/ejabberd/src/mod_commands.erl
+++ b/apps/ejabberd/src/mod_commands.erl
@@ -226,15 +226,7 @@ list_contacts(Caller) ->
 roster_info(M) ->
     Jid = jid:to_binary(maps:get(jid, M)),
     #{subscription := Sub, ask := Ask} = M,
-    case get_state(Sub, Ask) of
-        undefined ->
-            #{jid => Jid};
-        State ->
-            #{jid => Jid, state => State}
-    end.
-
-get_state(none, out) -> <<"invitee">>;
-get_state(_, _) -> undefined.
+    #{jid => Jid, subscription => Sub, ask => Ask}.
 
 add_contact(Caller, JabberID) ->
     add_contact(Caller, JabberID, <<"">>, []).
@@ -244,9 +236,7 @@ add_contact(Caller, JabberID, Name) ->
 
 add_contact(Caller, JabberID, Name, Groups) ->
     CJid = jid:from_binary(Caller),
-    OJid = jid:from_binary(JabberID),
-    mod_roster:set_roster_entry(CJid, JabberID, Name, Groups),
-    invite(CJid, OJid).
+    mod_roster:set_roster_entry(CJid, JabberID, Name, Groups).
 
 registered_commands() ->
     [#{name => mongoose_commands:name(C),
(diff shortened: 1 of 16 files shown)


apps/ejabberd/src/mongoose_client_api_contacts.erl
```diff
@@ -0,0 +1,130 @@
...
+    {ok, Body, Req1} = cowboy_req:body(Req),
+    JSONData = jiffy:decode(Body, [return_maps]),
+    Jid = case maps:get(<<"jid">>, JSONData, undefined) of
+              undefined ->
+                  {J, _} = cowboy_req:binding(jid, Req1),
+                  J;
+              J -> J
+          end,
+    Action = maps:get(<<"action">>, JSONData, undefined),
+    case handle_request(Method, to_binary(Jid), Action, CJid) of
```
gadgetci: According to **Elvis**:
> The code in the following (LINE, COL) locations has the same structure: (67, 5), (82, 5).

apps/ejabberd/src/mod_commands.erl
```diff
@@ -96,6 +103,64 @@ commands() ->
...
+     ],
+     [
+      {name, subscription},
+      {category, <<"contacts">>},
+      {desc, <<"Send out a subscription request">>},
+      {module, ?MODULE},
+      {function, subscription},
+      {action, update},
+      {security_policy, [user]},
+      {identifiers, [caller, jid]},
```
gadgetci: According to **Elvis**:
> The code in the following (LINE, COL) locations has the same structure: (135, 7), (148, 7).

apps/ejabberd/src/mod_commands.erl
```diff
@@ -96,6 +103,64 @@ commands() ->
...
+      {result, []}
+     ],
+     [
+      {name, add_contact},
+      {category, <<"contacts">>},
+      {desc, <<"Add a contact to roster">>},
+      {module, ?MODULE},
+      {function, add_contact},
+      {action, create},
+      {security_policy, [user]},
```
gadgetci: According to **Elvis**:
> The code in the following (LINE, COL) locations has the same structure: (123, 7), (159, 7).

michalwski:
> then the contact is in {none, in}

I think the same issue is for regular XMPP users. What state should it be to be visible to the user?

apps/ejabberd/src/mod_roster.erl
```diff
@@ -339,6 +344,9 @@ create_sub_el(Items, Version) ->
                      {<<"ver">>, Version}],
             children = Items}].
 
+-spec get_user_roster(Acc :: [roster()],
+                     {LUser :: binary(), LServer :: binary()}) ->
+    [{roster()}].
```
michalwski: How is it possible this is acceptable by dialyzer? Would it be enough to have just `[roster()]` as the result type? Why does it need to be surrounded by  `{}`?

apps/ejabberd/src/mod_roster.erl
```diff
@@ -380,6 +388,11 @@ item_to_xml(Item) ->
 get_roster_by_jid_t(LUser, LServer, LJID) ->
     mod_roster_backend:get_roster_by_jid_t(LUser, LServer, LJID).
 
+get_roster_by_jid(LUser, LServer, LJID) ->
+    {atomic, Item} = transaction(LServer,
```
michalwski: Why exactly do we need transaction for only get if we don't modify the `Item` in this transaction?

apps/ejabberd/src/mod_roster.erl
```diff
@@ -397,10 +410,23 @@ do_process_item_set(JID1,
                     To,
                     #xmlel{attrs = Attrs, children = Els}) ->
     LJID = jid:to_lower(JID1),
+    Item = get_roster_by_jid(LUser, LServer, LJID),
```
michalwski: Why this roster `get` and `set` was moved out of transaction? In fact there 2 separate transactions now (if I read it correctly) one for `get`, the other for `set`. Things might have changed between this 2 transactions. We either need to get back to single transaction completely get rid of these transactions as in current state they doesn't make any sense to me.
I may be missing sth so please correct me where my reasoning is wrong.

apps/ejabberd/src/mod_roster.erl
```diff
@@ -415,15 +441,15 @@ do_process_item_set(JID1,
...
+        {atomic, {OldItem, NewItem}} ->
+            push_item(User, LServer, To, NewItem),
+            case NewItem#roster.subscription of
                 remove ->
                     send_unsubscribing_presence(From, OldItem), ok;
                 _ -> ok
             end;
         E ->
-            ?DEBUG("ROSTER: roster item set error: ~p~n", [E]), ok
+            ?ERROR_MSG("ROSTER: roster item set error: ~p~n", [E]), ok
```
michalwski: Since we touch log message, could we align to the new log format initiative described in: #1248

apps/ejabberd/src/mod_roster.erl
```diff
@@ -808,6 +837,55 @@ set_items(User, Server, SubEl) ->
...
+                UserJid, % To
+                Item, % Item
+                Item2 % Item2
+            )
+    end.
+
+%% @doc remove from roster
+-spec remove_from_roster(UserJid :: jid(),
+                         ContactBin :: binary()) -> ok|error.
+remove_from_roster(UserJid, ContactBin) ->
```
michalwski: This function looks very similar to `set_roster_entry`. Could we reduce the code somehow?

apps/ejabberd/src/mod_roster.erl
```diff
@@ -867,6 +945,11 @@ read_subscription_and_groups(User, Server, LJID) ->
     LServer = jid:nameprep(Server),
     mod_roster_backend:read_subscription_and_groups(LUser, LServer, LJID).
 
+-spec get_jid_info(any(),
```
michalwski: Thanks a lot for the spec!

apps/ejabberd/src/mod_roster.erl
```diff
@@ -876,6 +959,10 @@ get_jid_info(_, User, Server, JID) ->
             get_bare_jid_info(User, Server, LJID)
     end.
 
+-spec get_bare_jid_info(User :: binary(),
```
michalwski: Thanks a lot for the spec!

apps/ejabberd/src/mongoose_client_api_contacts.erl
```diff
@@ -0,0 +1,130 @@
...
+handle_contact_request(_, _, _, _) ->
+    not_implemented.
+
+to_binary(S) when is_binary(S) ->
+    S;
+to_binary(S) ->
+    list_to_binary(S).
+
+-spec jid_exists(binary(), binary()) -> boolean().
+jid_exists(CJid, Jid) ->
```
michalwski: The implementation of this function could be improved efficiency wise. There should be a more efficient way to check if a `jid` exists in the roster. Reading the whole roster and traversing the result is not optimal.

test.disabled/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -261,10 +264,113 @@ password_can_be_changed(Config) ->
...
+                ask := <<"none">>} = R1,
+            ok
+        end
+    ),
+    ok.
+
+befriend_and_alienate(Config) ->
+    escalus:fresh_story(
+        Config, [{alice, 1}, {bob, 1}],
+        fun(Alice, Bob) ->
```
michalwski: Since both `Alice` and `Bob` are online over XMPP their clients should get some roster updates from the server. It'd be good to verify this.

test.disabled/ejabberd_tests/tests/rest_client_SUITE.erl
```diff
@@ -521,3 +555,198 @@ assert_json_room_sse_message(Expected, Received) ->
...
+                             escalus_client:short_jid(Bob),
+                             <<"subscribe">>)),
+            escalus:assert(is_roster_set, escalus:wait_for_stanza(Alice)),
+            escalus:assert(is_presence_with_type, [<<"subscribe">>],
+                           escalus:wait_for_stanza(Bob, 1)),
+            % now check Bob's roster, and it is empty...
+            {?OK, R4} = gett("/contacts", BCred),
+            Result4 = decode_maplist(R4),
+            [] = Result4,
+            % because it is stated in RFC3921, 8.2.6, and implemented
```
michalwski: I think, it'd be good to ask this on XSF's standards mailing list. Could you?

test.disabled/ejabberd_tests/tests/rest_helper.erl
```diff
@@ -113,6 +113,7 @@ make_request(Method, Path, ReqBody) when not is_binary(Path) ->
     make_request(Method, list_to_binary(Path), ReqBody);
 make_request(Method, Path, ReqBody) ->
     CPath = <<?PATHPREFIX/binary, Path/binary>>,
+    ct:pal("CPath: ~p", [CPath]),
```
michalwski: Probably we don't need this line any more.

bartekgorny:
>> then the contact is in {none, in}

> I think the same issue is for regular XMPP users. What state should it be to be visible to the user?

Yes it is the same, because the RFC says it should (not must) be like that. I don't know why. Any other state is visible. It is indeed strange, because it makes your contact disappear from the roster if you receive an invitation...

apps/ejabberd/src/mod_commands.erl
```diff
@@ -252,7 +347,61 @@ lookup_recent_messages(ArcJID, WithJID, Before, Limit) ->
...
+    mongoose_acc:update(A, Map).
+
+create_acc(CallerJid, Name, Type, OtherJid) ->
+    A = create_acc(CallerJid, Name, Type),
+    A1 = mongoose_acc:put(to, jid:to_binary(OtherJid), A),
+    mongoose_acc:put(to_jid, OtherJid, A1).
+
+
+subscription(Caller, Other, Action) ->
+    _ = {invite, accept},
```
bartekgorny: Nay, it a leftover; it was here to guarantee those atoms exists, but now it is redundant, as well as a few other lines, thanks for pointing it out.

apps/ejabberd/src/mod_roster.erl
bartekgorny: My understanding was that the transaction was here mostly to protect the sequence of actions performed to update a roster (update roster, run hook, write roster version), so I did not bother much about roster get. I take your point, though, and await your recommendation - put "get" back into a transaction, or get rid of transaction altogether?

test.disabled/ejabberd_tests/tests/rest_client_SUITE.erl
bartekgorny: What exactly is the question you would like to ask? The RFC is pretty straightforward. I think the logic behind it is that the {none, in} is a temporary state meant to handle a situation whereby someone you don't know sends you a subscribe request while you are offline. The roster item is created so that the server can send you the subscription later on - you either accept the subscription or reject it, in both cases moving the roster item out of this state. If you had created the item before, then it indeed disappears from your roster, but instead you see the subscription request. Come to think of it, it makes sense, though in REST API it gets a bit counterintuitive. Anyway, the RFC says "should", so we can change it or make it configurable.

apps/ejabberd/src/mongoose_client_api_contacts.erl
bartekgorny: I guess not, the trouble is that roster backends do not provide such feature. If you query roster by jid, you always get back a roster item, whether it exists in roster or not. We can do something along the lines of `get_jid_info`, meaning try `read_subscription_and_groups` and if it returns `error` assume there is no such entry (though it sounds somewhat hackish, and at least in some backends it is not necessarily true).

apps/ejabberd/src/mod_shared_roster_ldap.erl
```diff
@@ -182,15 +182,28 @@ get_jid_info({Subscription, Groups}, User, Server, JID) ->
...
-            {stop, false};
+            {stop, mongoose_acc:put(result, false, Acc)};
         _ -> Acc
     end.
 
+-spec out_subscription(Acc:: mongoose_acc:t(),
+                      User :: binary(),
+                      Server :: binary(),
+                      JID :: jid(),
+                      Type :: sub_presence()) ->
```
gadgetci: According to **Compiler**:
> type sub_presence() undefined

gadgetci:

apps/ejabberd/src/mod_roster.erl
michalwski: I'd put the `get` back into a transaction as it was before.

apps/ejabberd/src/mongoose_client_api_contacts.erl
michalwski: > I guess not, the trouble is that roster backends do not provide such feature.

I suggest spending some time and fixing the root cause then.

test.disabled/ejabberd_tests/tests/rest_client_SUITE.erl
michalwski: How about make the roster get function configurable? In case of XMPP user it works as is, in case of REST call it will return all items if with state `{none, in}`?

michalwski:
Also, are there any chances you will try to make Elvis happy about this PR?

bartekgorny:
I thought about elvis-ising those files after all the other work is completed (the whole files, not just the modified parts). This way code review is easier, and there is fewer conflicts when rebasing. What do you think about such approach?

michalwski:
All right, make elvis happy later. For now let's focus on other comments.

bartekgorny:
Ready. I don't know why elvis is still not happy, when I check it locally everything is fine.

apps/ejabberd/src/mod_roster.erl
```diff
@@ -424,16 +428,32 @@ do_process_item_set(JID1,
...
+                      Item2 :: fun( (roster()) -> roster())) -> ok.
+set_roster_item(User, LUser, LServer, LJID, From, To, MakeItem2) ->
     F = fun () ->
-                Item = case mod_roster_backend:get_roster_entry_t(LUser, LServer, LJID) of
+                Item = case get_roster_entry(LUser, LServer, LJID) of
                            does_not_exist ->
                                #roster{usj = {LUser, LServer, LJID},
-                                       us = {LUser, LServer},
-                                       jid = LJID};
+                                   us = {LUser, LServer},
```
michalwski: Indentation got broken here and the line below. Should be indented as previously.

apps/ejabberd/src/mod_roster.erl
```diff
@@ -848,6 +885,59 @@ set_items(User, Server, SubEl) ->
...
+                       NewSubscription :: remove | unchanged) -> ok|error.
+set_roster_entry(UserJid, ContactBin, Name, Groups, NewSubscription) ->
+    LUser = UserJid#jid.luser,
+    LServer = UserJid#jid.lserver,
+    JID1 = jid:from_binary(ContactBin),
+    case JID1 of
+        error -> error;
+        _ ->
+            LJID = jid:to_lower(JID1),
+            MakeItem2 = fun(Item) ->
```
michalwski: Do we need `MakeItem2`?  Can be named just `MakeItem`?

apps/ejabberd/src/mongoose_client_api_contacts.erl
```diff
@@ -0,0 +1,126 @@
...
+
+forbidden_request(Req, State) ->
+    cowboy_req:reply(403, Req),
+    {halt, Req, State}.
+
+to_json(Req, #{jid := Caller} = State) ->
+    CJid = jid:to_binary(Caller),
+    {Method, _} = cowboy_req:method(Req),
+    {Jid, _} = cowboy_req:binding(jid, Req),
+    {ok, Res} = handle_request(Method, Jid, undefined, CJid),
```
michalwski: What will be returned here if the asked `Jid` doesn't exist?

test.disabled/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -57,6 +57,9 @@ all() ->
 
 groups() ->
     [{admin, [parallel], test_cases()},
+     {roster, [parallel], [list_contacts,
+                             befriend_and_alienate,
```
michalwski: Indentation?

test.disabled/ejabberd_tests/tests/rest_SUITE.erl
```diff
@@ -261,10 +264,147 @@ password_can_be_changed(Config) ->
...
+                    {remove, none}],
+            check_pushlist(BExp, BPushes),
+            ok
+        end
+    ),
+    ok.
+
+
+befriend_and_alienate_auto(Config) ->
+    % TOFIX - verify xmpp pushes
```
michalwski: I guess this was fixed, am I right?

test.disabled/ejabberd_tests/tests/rest_client_SUITE.erl
```diff
@@ -521,3 +555,197 @@ assert_json_room_sse_message(Expected, Received) ->
...
+                                         BCred),
+            BadPutPath = "/contacts/zorro@localhost",
+            {?NOT_FOUND, _} = putt(BadPutPath,
+                                   #{action => <<"invite">>},
+                                   BCred),
+            ok
+        end
+    ),
+    ok.
+
```
michalwski: There is probably no need for so many empty lines at the end of the file.

apps/ejabberd/src/mongoose_client_api_contacts.erl
bartekgorny: Good catch - actually, calling GET with jid should return 404.

apps/ejabberd/src/mongoose_client_api_contacts.erl
```diff
@@ -0,0 +1,132 @@
...
+    {ok, Body, Req1} = cowboy_req:body(Req),
+    JSONData = jiffy:decode(Body, [return_maps]),
+    Jid = case maps:get(<<"jid">>, JSONData, undefined) of
+              undefined ->
+                  {J, _} = cowboy_req:binding(jid, Req1),
+                  J;
+              J -> J
+          end,
+    Action = maps:get(<<"action">>, JSONData, undefined),
+    case handle_request(Method, to_binary(Jid), Action, CJid) of
```
gadgetci: According to **Elvis**:
> The code in the following (LINE, COL) locations has the same structure: (73, 5), (88, 5).

michalwski:
Thanks!

PR #1287, opened 2017-04-14
Silent push notifications

rslota:
This PR adds support for "silent" push notifications. This code depends on esl/MongoosePush#18 so it should be merged only after merging the `MongoosePush`.

TODO: 
- [x] Update docs
- [x] Revert MongoosePush version to `latest` (well, maybe `0.9.0` is better)


apps/ejabberd/src/mod_push_service_mongoosepush.erl
```diff
@@ -123,6 +123,44 @@ http_notification(Host, Method, URL, ReqHeaders, Payload) ->
 %% Helper functions
 %%--------------------------------------------------------------------
 
+%% Create notification for API v1
+make_notification(v1, _Notification, #{<<"silent">> := <<"true">>}) ->
```
michalwski: I appreciate you care about backward compatibility. Push notifications were added to MongooseIM in 2.1.0beta1 as the name suggest there may be significant changes before the final 2.1.0. IMO, it's better to remove support for the old version now. As I understand we are going to use the new API anyway.

doc/user-guide/Untitled
```diff
@@ -0,0 +1,21 @@
+defmodule Fennec.UDP.Supervisor do
```
michalwski: What's this file for?

rslota: I really, really wish I knew... Sorry,  didn't notice that. Deleting.

apps/ejabberd/src/mod_push_service_mongoosepush.erl
rslota: Do you think we should remove it only from MIM or also from MongoosePush? To be honest I've made backward compatibility also in order to showcase that it's possible, how to do it and to verify that everything needed is in place and ready to use.

michalwski:
Could fix dialyzer error and the reference to MongoosePush's docker image?

rslota:
> Could you fix dialyzer error and the reference to MongoosePush's docker image?

I have changed the reference to `mongoose_push:0.9.0` not `latest`. I think it a better idea for tests, since the `0.9.0` is actually a release candidate. We should use in production always the tested version.

michalwski:
Thanks!

PR #1289, opened 2017-04-19
Fix BOSH interleaving

fenek:
This PR addresses a problem with random BOSH fails on Travis

Proposed changes include:
* A fix for `mod_bosh_socket`. Now it maintains strict response order, according to RID.
* A cleanup of `bosh_interleave_reqs`
* New `escalus` version with improved BOSH client


michalwski:
Thanks a lot for the fix!

PR #1291, opened 2017-04-21
Use the 'clean_disconnect' branch of escalus

chrzaszcz:
See https://github.com/esl/escalus/pull/145

### DIFF
--- a/test.disabled/ejabberd_tests/tests/sm_SUITE.erl
+++ b/test.disabled/ejabberd_tests/tests/sm_SUITE.erl
@@ -119,20 +119,20 @@ end_per_testcase(CaseName, Config) ->
 
 server_announces_sm(Config) ->
     AliceSpec = given_fresh_spec(Config, alice),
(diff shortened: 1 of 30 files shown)


test.disabled/ejabberd_tests/tests/sm_SUITE.erl
```diff
@@ -791,11 +791,26 @@ assert_no_offline_msgs(Spec) ->
 assert_no_offline_msgs() ->
     0 = mongoose_helper:total_offline_messages().
 
+wait_for_c2s_state_change(C2SPid, StateName, NewStateName) ->
+    wait_for_c2s_state_change(C2SPid, StateName, NewStateName, 5000).
+
+wait_for_c2s_state_change(C2SPid, StateName, NewStateName, 0) ->
```
michalwski: This is error prone. We may end up in an infinite loop if someone wants to wait `5001` ms for instance.

Issue #1231, opened 2017-03-17
mod_mam_odbc_async_pool_writer mysql calls issue


andrewvmail:
MongooseIM version: (detached from 2.0.0) 1743443 Merge pull request #1069 from esl/rel-2.0
Installed from: source
Erlang/OTP version: 18

I use below config and getting errors when ever I am sending a message. I know the credentials for mysql is correct because privacy list function is working and Mongoose is reading and writing to the database. Just when mod_mam_odbc_async_pool_writer do the mysql calls it doesnt like it ...
```
%% https://mongooseim.readthedocs.io/en/2.0.0beta2/modules/mod_mam/

  %% STEP 1
  {mod_mam, []},
  {mod_mam_odbc_arch, [no_writer]},

  {mod_mam_muc, [{host, "muclight.@HOST@"}]},
  {mod_mam_muc_odbc_arch, [no_writer]},

  %% STEP 2
  {mod_mam_odbc_user, [muc, pm]},

  %% STEP 3
  {mod_mam_cache_user, [muc, pm]},

  %% STEP 4
  {mod_mam_odbc_prefs, [muc, pm]},

  %% STEP 5
  {mod_mam_odbc_async_pool_writer, [pm]},

  %% STEP 6
  {mod_mam_muc_odbc_async_pool_writer, [muc]}

```

```
==> /root/MongooseIM/rel/mongooseim/log/crash.log <==
2017-03-16 19:55:45 =CRASH REPORT====
  crasher:
    initial call: gen:init_it/6
    pid: <0.649.0>
    registered_name: []
    exception exit: {closed,[{p1_fsm,terminate,8,[{file,"src/p1_fsm.erl"},{line,760}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,240}]}]}
    ancestors: ['momomobile88.com_0',mod_mam_sup,ejabberd_sup,<0.217.0>]
    messages: []
    links: [<0.648.0>]
    dictionary: [{ejabberd_odbc_state,{state,<0.650.0>,mysql,30000,<<"momomobile88.com">>,<0.648.0>,true,1000,{0,{[],[]}}}},{'$internal_queue_len',0},{mim_odbc_dedicated,true},{mim_process_type,odbc_worker},{mim_host,<<"momomobile88.com">>},{ejabberd_odbc_nesting_level,0}]
    trap_exit: true
    status: running
    heap_size: 6772
    stack_size: 27
    reductions: 11999
[...]

Issue #827, opened 2016-05-17
Stop archiving incoming muc light messages


michalwski:
MongooseIM version: b13807d90b2167f08f92257af3813c157c720552

Source: @studzien's comment: https://github.com/esl/MongooseIM/pull/577#issuecomment-206951924

> Apparently, incoming MUC Light messages are stored in the one-to-one chats archive (since they are routed via `ejabberd_router:route/3`).
> I've created a test case for this here: e5c5f32611f7
> 
> The instant solution I can see is to bind to the `filter_packet` hook in `mod_muc_light` and stop running the hook just before it gets to `mod_mam`.
> Do you think this make any sense?
> 
> Obviously we want to stop the hook if and only if the message is a MUC Light message :)

vickydhas:
I enabled the default recommended configuration in ejabberd.cfg for mam_odbc... modules.

And replaced the host name like muclight.@Host@ instead of muc.@Host@, now can see my message on mysql database in mam_muc_message table.

Is this fixed now you mean??

andrewvmail:
I'm also noticing this too.

michalwski:
It's fixed now with PR #1294

PR #1294, opened 2017-04-24
Add archive_groupchats option to mod_mam

fenek:
This PR addresses #827 and #1231

Proposed changes include:
* When there is incoming groupchat message to a user, `mod_mam` stores it in individual archive only when `archive_groupchats` flag is enabled

BTW: Thanks @studzien for test case!

**TODO:**

- [x] Add positive testcase for new flag
- [x] ~Add test case for `mod_muc` integration (if it's affected as well)~ I can't see a way of starting PM+MUC `mod_mam` for MUC cases without making `init_modules` function even longer. I think I'd like to refactor `mam_SUITE` first as a separate task.

### DIFF
--- a/apps/ejabberd/src/mod_mam_utils.erl
+++ b/apps/ejabberd/src/mod_mam_utils.erl
@@ -322,12 +322,12 @@ is_archivable_message(Mod, Dir, Packet=#xmlel{name = <<"message">>}) ->
 is_archivable_message(_, _, _) ->
     false.
 
-is_valid_message_type(_, _, <<"">>)          -> true;
-is_valid_message_type(_, _, <<"normal">>)    -> true;
-is_valid_message_type(_, _, <<"chat">>)      -> true;
-is_valid_message_type(_, _, <<"groupchat">>) -> true;
-is_valid_message_type(_, _, <<"error">>)     -> false;
-is_valid_message_type(_, _, _)               -> false.
+is_valid_message_type(_, _, <<"">>) -> true;
+is_valid_message_type(_, _, <<"normal">>) -> true;
+is_valid_message_type(_, _, <<"chat">>) -> true;
+is_valid_message_type(_, incoming, <<"groupchat">>) -> true;
+is_valid_message_type(_, _, <<"error">>) -> false;
+is_valid_message_type(_, _, _) -> false.
 
 is_valid_message(_Mod, _Dir, Packet) ->
     Body     = xml:get_subtag(Packet, <<"body">>),
--- a/doc/modules/mod_mam.md
+++ b/doc/modules/mod_mam.md
@@ -35,7 +35,8 @@ For now `odbc` backend has very limited support for this feature, while `cassand
 
 #### PM-specific options
 
-* **archive_groupchats** (boolean, default: `false`) - When enabled, MAM will store groupchat messages in recipients' individual archives. **USE WITH CAUTION!** May increase archive size significantly. Disabling this option for existing installation will neither remove such messages from MAM storage, nor will filter out them from search results.
+* **archive_groupchats** (boolean, default: `true`) - When enabled, MAM will store groupchat messages in recipients' individual archives. **USE WITH CAUTION!** May increase archive size significantly. Disabling this option for existing installation will neither remove such messages from MAM storage, nor will filter out them from search results.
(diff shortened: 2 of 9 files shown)


apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -322,12 +322,12 @@ is_archivable_message(Mod, Dir, Packet=#xmlel{name = <<"message">>}) ->
...
-is_valid_message_type(_, _, <<"">>)          -> true;
-is_valid_message_type(_, _, <<"normal">>)    -> true;
-is_valid_message_type(_, _, <<"chat">>)      -> true;
-is_valid_message_type(_, _, <<"groupchat">>) -> true;
-is_valid_message_type(_, _, <<"error">>)     -> false;
-is_valid_message_type(_, _, _)               -> false.
+is_valid_message_type(_, _, <<"">>) -> true;
+is_valid_message_type(_, _, <<"normal">>) -> true;
+is_valid_message_type(_, _, <<"chat">>) -> true;
+is_valid_message_type(_, incoming, <<"groupchat">>) -> true;
```
kzemek: Will our own messages also count as incoming?

doc/modules/mod_mam.md
```diff
@@ -25,14 +25,25 @@ For now `odbc` backend has very limited support for this feature, while `cassand
...
+* **add_archived_element** (boolean, default: `false`) - Add `<archived/>` element from MAM v0.2.
 * **is_archivable_message** (module, default: `mod_mam_utils`) - Name of a module implementing [`is_archivable_message/3` callback](#is_archivable_message) that determines if the message should be archived.
-* **host** (string, default: `"conference.@HOST@"`) - MUC host that will be archived if MUC archiving is enabled. 
  **Warning**: if you are using MUC Light, make sure this option is set to the MUC Light domain.
 * **pm** (list | `false`, default: `[]`) - Override options for archivization of one-to-one messages. If the value of this option is `false`, one-to-one message archive is disabled.
 * **muc** (list | `false`, default: `false`) - Override options for archivization of group chat messages. If the value of this option is `false`, group chat message archive is disabled.
 
-All options described in this document can be overriden for a specific type of messages through `pm` and `muc` options, e.g.:
+**backend**, **add_archived_element** and **is_archivable_message** will be applied to both `pm` and `muc` (if they are enabled), unless overriden explicitly (see example below).
```
kzemek: They can still be given via top-level option instead of creating a more complicated configuration where `pm` and `muc` are given their own arguments.

e.g. this still works:
```erlang
{mod_mam_meta, [{archive_groupchats, true}]}.
```

and there's no need to write this:
```erlang
{mod_mam_meta, [{pm, [{archive_groupchats, true}]}]}.
```

apps/ejabberd/src/mod_mam_utils.erl
arcusfelis: @kzemek All groupchat messages are incoming.

fenek: There is no point to archive outgoing message because at this point we can't be sure if the message reached the room and got broadcasted.

doc/modules/mod_mam.md
fenek: This line (34) doesn't state that these params need to be set for `muc` and `pm` individually.

fenek: In case of `pm`- and `muc`-exclusive options, I think it is more consistent to expect these to be configured not in the top level but in respective "areas".

kzemek: > This line (34) doesn't state that these params need to be set for muc and pm individually.

The comment was aimed at the whole change (`All options...` -> `**backend**, **add_archived_element** and **is_archivable_message**...`).

> In case of pm- and muc-exclusive options, I think it is more consistent to expect these to be configured not in the top level but in respective "areas".

Fair enough.

kzemek:
I have one more concern, i.e. current installations not explicitly configured to filter out groupchat messages via `is_archivable_message` will suddenly stop archiving them because of the changed default.

fenek:
Did you mean `archive_groupchats`? Because the change to the default `is_archivable_message` will only stop archiving outgoing groupchat messages.

kzemek:
What I meant is that previously the only way was to use `is_archivable_message` to filter out groupchat messages; there was no way to explicitly say that you want groupchat messages stored in PM archoves. Now the default would change even if someone wanted and depended on this behaviour.

PR #1296, opened 2017-04-24
Allow to specify Cowboy options in configuration file

bszaf:
This PR addresses #

Proposed changes include:
* describe the functionality changes:
  Change way of providing configuration options in ejabberd.cfg file. Instead of providing only arbitrary options, we can provide all options now.
  Adapted one test case to new configuration.
* docs:
  Add info about cowboy endpoint cofiguration

### DIFF
--- a/apps/ejabberd/src/ejabberd_cowboy.erl
+++ b/apps/ejabberd/src/ejabberd_cowboy.erl
@@ -63,7 +63,10 @@ start_listener({Port, IP, tcp}=Listener, Opts) ->
                             [#cowboy_state{ref = Ref, opts = Opts}]},
                  transient, infinity, worker, [?MODULE]},
     {ok, Pid} = supervisor:start_child(ejabberd_listeners, ChildSpec),
-    {ok, _} = start_cowboy(Ref, [{port, Port}, {ip, IP} | Opts]),
+    TransportOpts = gen_mod:get_opt(transport_options, Opts, []),
+    TransportOpts2 = [{port, Port}, {ip, IP} | TransportOpts],
+    Opts2 = gen_mod:set_opt(transport_options, Opts, TransportOpts2),
+    {ok, _} = start_cowboy(Ref, Opts2),
     {ok, Pid}.
 
 reload_dispatch(Ref) ->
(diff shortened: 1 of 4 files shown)


apps/ejabberd/test/cowboy_SUITE.erl
```diff
@@ -206,7 +206,10 @@ mixed_requests(_Config) ->
     Responses = lists:duplicate(50, {TextPong, true, TextPong, true}).
 
 start_cowboy_returns_error_eaddrinuse(_C) ->
-    Opts = [{port, 8088}, {ip, {127, 0, 0, 1}}, {modules, []}, {retires, {2, 10}}],
+    Opts = [{transport_options, [{port, 8088},
+                                 {ip, {127, 0, 0, 1}}]},
+            {modules, []},
+            {retires, {2, 10}}],
```
fenek: Both old and new version have a typo here. Should be `retries`

apps/ejabberd/src/ejabberd_cowboy.erl
```diff
@@ -121,22 +124,13 @@ do_start_cowboy(Ref, Opts, Retries, SleepTime) ->
...
-    TransportOpts = [{port, Port},
-                     {ip, IP},
-                     {max_connections, MaxConns}
-                    ],
-    Dispatch = cowboy_router:compile(get_routes(gen_mod:get_opt(modules, Opts))),
-    ProtocolOpts = [{compress, Compress}, {env, [{dispatch, Dispatch}]} | Middlewares],
+    TransportOpts = gen_mod:get_opt(transport_options, Opts, []),
+    Modules = gen_mod:get_opt(modules, Opts),
+    Dispatch = cowboy_router:compile(get_routes(Modules)),
+    ProtocolOpts = lists:merge([{env, [{dispatch, Dispatch}]}],
```
kzemek: The `merge` is for merging two independent sorted lists into one sorted list. I believe its behaviour is undefined for unsorted list (but the implementation happens to work). This use case simply calls for `[{env, [{dispatch, Dispatch}]} | gen_mod:get_opt(protocol_options, Opts, [])]`

doc/advanced-configuration/Listener-modules.md
```diff
@@ -34,7 +34,8 @@ Manages all HTTP-based services, such as BOSH (HTTP long-polling) and WebSocket.
 
 * `ip` (IP tuple, optional, default: `{0,0,0,0}`) - IP address to bind to.
 * `num_acceptors` (positive integer, optional, default: 100) - Number of acceptors.
-* `max_connections` (positive integer, optional, default: 1024) - Maximum total number of HTTP(S) connections.
```
kzemek: It would be great to still support this option in the code (but still remove it from documentation) to preserve backwards compatibility for documented options throughout minor changes.

bszaf: @kzemek This is covered with latest commit. Old syntax has higher priority

PR #1297, opened 2017-04-25
[skip ci] stanza routing - formatting

goddammit:
looked weird


fenek:
Why? With numbering, it is emphasised, that every section is a subsequent step in routing process. Without it, the sequence might be not obvious to some readers.

PR #1301, opened 2017-04-26
Fix race condition between caps and PEP

fenek:
This PR addresses random `pep_SUITE` failures on Travis.

Before this change, a scenario was possible when a check for supported features in cache (when about to send PEP message) was performed before processing feature response from a client. This PR moves "PEP last item resend" trigger to a place when it's 100% certain we have the features stored in cache.

Unfortunately it may lead to duplicated notifications in federation but it's because `mod_caps` should subscribe not to `user_receive_packet` but to some hook in s2s. I think it is a separate matter to investigate (that requires new test case). It's out of scope of fixing failures on Travis.

It's difficult to reproduce with a dedicated integration test, because it would require ensuring that feature request is sent after PEP resend is processed. I don't like the idea of introducing crude `timer:sleep` here and I don't like the one I've added but the only way to avoid it is to introduce something like `wait_for_pubsub_loop`.

### DIFF
--- a/apps/ejabberd/src/mod_caps.erl
+++ b/apps/ejabberd/src/mod_caps.erl
@@ -98,6 +98,12 @@ stop(Host) ->
     supervisor:terminate_child(ejabberd_sup, Proc),
     supervisor:delete_child(ejabberd_sup, Proc).
 
+get_features_list(Host, Caps) ->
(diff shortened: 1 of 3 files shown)


michalwski:
Thanks for the fix!

PR #1303, opened 2017-04-27
Update of mods' docs (part 1) + minor mod_disco fix

fenek:
The fix for mod_disco patches a leftover from binary-fication.

Why page for `mod_websockets` is removed? Because it is no longer a module that can be configured in `modules` section. It is pure listener now (despite the name) and is described in listeners doc.


apps/ejabberd/src/mod_caps.erl
```diff
@@ -575,22 +575,22 @@ concat_info(Els) ->
                              Els)).
 
 concat_xdata_fields(Fields) ->
```
arcusfelis: We try to do three things in the function:
- Extract fields;
- Concat fields;
- Get FORM_TYPE field.
It would MUCH cleaner code if it was split into functions doing one and only one thing.
Yes, it would be less performant, but much more easier to read.

apps/ejabberd/src/mod_disco.erl
```diff
@@ -464,7 +464,7 @@ make_node_attr(Node) -> [{<<"node">>, Node}].
 
 -spec get_info(A :: [jlib:xmlel()], ejabberd:server(), module(), Node :: binary(),
         Lang :: ejabberd:lang()) -> [jlib:xmlel()].
-get_info(_A, Host, Mod, Node, _Lang) when Node == [] ->
+get_info(A, Host, Mod, Node, _Lang) when Node == <<>> ->
```
arcusfelis: A is two short.
Just by reading the second function clause I can understand that it's Acc.

apps/ejabberd/src/mod_disco.erl
```diff
@@ -500,23 +500,24 @@ get_fields_xml(Host, Module) ->
     fields_to_xml(FilteredFields).
 
 
--spec fields_to_xml([{_, Var :: binary(), Values :: [binary()]}]) -> [jlib:xmlel()].
+-spec fields_to_xml([{_, Var :: string(), Values :: [string()]}]) -> [jlib:xmlel()].
```
arcusfelis: What is underscore?

apps/ejabberd/src/mod_disco.erl
```diff
@@ -500,23 +500,24 @@ get_fields_xml(Host, Module) ->
...
+-spec field_to_xml({_, Var :: string(), Values :: [string()]}) -> jlib:xmlel().
 field_to_xml({_, Var, Values}) ->
     XMLValues = values_to_xml(Values),
-    #xmlel{name = <<"field">>, attrs = [{<<"var">>, Var}],
+    #xmlel{name = <<"field">>, attrs = [{<<"var">>, list_to_binary(Var)}],
            children = XMLValues}.
 
 
 -spec values_to_xml([binary()]) -> [jlib:xmlel()].
 values_to_xml(Values) ->
```
arcusfelis: we can rewrite this function the way fields_to_xml/1 is written.

doc/modules/mod_disco.md
```diff
@@ -1,10 +1,13 @@
 ### Module Description
-Implements [XEP-0030: Service Discovery](http://xmpp.org/extensions/xep-0030.html). The module itself only provides the essential disco interface, the actual capabilities announced by Service Discovery are gathered via executing a fold-type hook.
+Implements [XEP-0030: Service Discovery](http://xmpp.org/extensions/xep-0030.html). The module itself provides only the essential disco interface, the actual capabilities announced by Service Discovery are gathered via executing a fold-type hook.
 
 ### Options
-* **iqdisc**
-* **extra_domains** (list of binaries, default: `[]`): Adds domains that are not registered with other means to a local item announcement
-* **server_info** (list of tuples `{[Module] | all, Name, [URL]}`, default: `[]`): Adds extra disco information to chosen or all modules. Example: `{server_info, [{all, "abuse-address", ["admin@example.com"]}, {[mod_muc, mod_disco], "friendly-spirits", ["spirit1@localhost", "spirit2@localhost"]}]}`
+* **iqdisc** (default: `one_queue`)
```
arcusfelis: I'm wondering why defauild is `one_queue`, not parallel.

apps/ejabberd/src/mod_disco.erl
michalwski: Agree, longer var name will make things easier to understand.

apps/ejabberd/src/mod_caps.erl
```diff
@@ -575,22 +575,22 @@ concat_info(Els) ->
...
-                                               [[[Var, $<], extract_values_sorted_cdatas(Els)]
-                                                | VarFields]]
-                                      end;
-                                 (_, Acc) -> Acc
-                              end,
-                              [<<"">>, []], Fields),
+    [Form, Res] =
+    lists:foldl(fun(#xmlel{name = <<"field">>, attrs = Attrs, children = Els} = FieldEl,
+                    [FormType, VarFields] = Acc) ->
+                        case xml:get_attr_s(<<"var">>, Attrs) of
```
michalwski: I see you used `exml_query:path` instead of `xml:get_cdata(...)`, that's good, thanks! Could you also change the `xml:gett_attr_s` with equivalent from `exml_query`?

apps/ejabberd/src/mod_caps.erl
fenek: This function isn't difficult to read in my opinion and I don't think it's unclean. I've decided to only improve the style a bit, e.g. use tuple in accumulator instead of list and replace `xml` with `exml`.

doc/modules/mod_disco.md
fenek: I have no idea what was the original reason behind it but I think that revising `iqdisc` default value for all modules can be a subject of a separate ticket.

apps/ejabberd/src/mod_caps.erl
fenek: OK, but I won't change `xml:` calls in whole file, as it is docs-oriented PR above else.

apps/ejabberd/src/mod_disco.erl
fenek: Fixed.

apps/ejabberd/src/mod_disco.erl
fenek: Fixed.

apps/ejabberd/src/mod_disco.erl
fenek: Fixed

doc/advanced-configuration/Modules.md
```diff
@@ -68,6 +68,12 @@ Implements [XEP Multi-User Chat Light](https://github.com/xsf/xeps/pull/118).
 ### [mod_offline](../modules/mod_offline.md)
 Provides an offline messages storage that is compliant with [XEP-0160: Best Practices for Handling Offline Messages)](http://xmpp.org/extensions/xep-0160.html).
 
+### [mod_offline_stub](../modules/mod_offline_stub.md)
+Prevents `<service-unavailable/>` error when message recipient is offline.
```
goddammit: Prevents `<service-unavailable/>` error when the message recipient is offline.

doc/modules/mod_offline_stub.md
```diff
@@ -0,0 +1,17 @@
+### Module Description
+
+RFC 6121 requires a `<service-unavailable/>` stanza error to be sent to a user messaging an unavailable recipient if the message is not stored for delayed delivery (i.e. as an "offline message").
+If the recipient exists (i.e. auth module returns `true` from `is_user_exists`), `mod_mam` stores the message, but <service-unavailable/> is still returned. This is not compliant with the RFC.
+This module prevents returning <service-unavailable/>.
+Please note that `mod_offline_stub` is not tightly coupled with `mod_mam`. It can be used as standalone extension, if the specific application requires it.
```
goddammit: It can be used as a standalone extension, if the specific application requires it.

doc/modules/mod_stream_management.md
```diff
@@ -1,11 +1,13 @@
...
-It is implemented mostly in `ejabberd_c2s`, this module is just a "starter", to supply configuration values to new client connections.
+It is implemented mostly in `ejabberd_c2s`. 
+This module is just a "starter", to supply the configuration values to new client connections.
+It also provides a basic session table API and adds a new stream feature.
 
 ### Options
 
 * `buffer_max` (default: 100): Buffer size for messages yet to be acknowledged.
-* `ack_freq` (default: 1): Frequency of server-side acks, e.g. 1 means an ack after each stanza, 3 means an ack after each 3 stanzas.
+* `ack_freq` (default: 1): Frequency of ack requests sent from server to client, e.g. 1 means a request after each stanza, 3 means a request after each 3 stanzas.
```
goddammit: Frequency of ack requests sent from the server to the client, e.g. 1 means a request after each stanza, 3 means a request after each 3 stanzas.

PR #1305, opened 2017-04-27
uncommented registration of hook and handler calls in mongoose_acc

bartekgorny:
This PR is for load-testing - here all calls to hooks and handlers are recorded in the accumulator, for tracing and debugging. We want to see whether it has an impact on performance.


michalwski:
Looks like we no longer need this PR.

PR #1314, opened 2017-05-11
Disable DNS lookup for S2S hosts with predefined addresses.

kzemek:
When addresses are set in the configuration file via s2s_addr tuple,
it's a confusing behaviour to additionaly perform a DNS lookup of
the peer host and use its results alongside the preconfigured ones.

### DIFF
--- a/apps/ejabberd/src/ejabberd_s2s_out.erl
+++ b/apps/ejabberd/src/ejabberd_s2s_out.erl
@@ -233,13 +233,7 @@ open_socket(init, StateData) ->
                                 StateData#state.server,
                                 StateData#state.new,
                                 StateData#state.verify}]),
-    AddrList = get_predefined_addresses(StateData#state.server) ++
-               case ejabberd_s2s:domain_utf8_to_ascii(StateData#state.server) of
-                   false ->
-                       [];
-                   ASCIIAddr ->
-                       get_addr_port(ASCIIAddr)
-               end,
+    AddrList = get_addr_list(StateData#state.server),
     case lists:foldl(fun({Addr, Port}, Acc) ->
                              case Acc of
                                  {ok, Socket} ->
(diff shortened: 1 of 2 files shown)


apps/ejabberd/src/ejabberd_s2s_out.erl
```diff
@@ -233,13 +233,7 @@ open_socket(init, StateData) ->
...
-    AddrList = get_predefined_addresses(StateData#state.server) ++
-               case ejabberd_s2s:domain_utf8_to_ascii(StateData#state.server) of
-                   false ->
-                       [];
-                   ASCIIAddr ->
-                       get_addr_port(ASCIIAddr)
-               end,
+    AddrList = get_addr_list(StateData#state.server),
     case lists:foldl(fun({Addr, Port}, Acc) ->
                              case Acc of
```
gadgetci: According to **Elvis**:
> The expression on line 238 and column 30 is nested beyond the maximum level of 3.

apps/ejabberd/src/ejabberd_s2s_out.erl
```diff
@@ -233,20 +233,11 @@ open_socket(init, StateData) ->
...
-                                     open_socket1(Addr, Port)
-                             end
+    AddrList = get_addr_list(StateData#state.server),
+    case lists:foldl(fun({_Addr, _Port}, {ok, Socket}) ->
+                             {ok, Socket};
+                        ({Addr, Port}, _Acc) ->
+                             open_socket1(Addr, Port)
                      end, ?SOCKET_DEFAULT_RESULT, AddrList) of
         {ok, Socket} ->
             Version = if
```
gadgetci: According to **Elvis**:
> Replace the 'if' expression on line 243 with a 'case' expression or function clauses.

PR #1347, opened 2017-06-19
Create foreign-event IQ handler and verify requests

arkgil:
This PR installs handler for foreign-event IQ stanzas and simple verification of requests format.

### DIFF
--- a/apps/ejabberd/src/foreign_event/mod_foreign.erl
+++ b/apps/ejabberd/src/foreign_event/mod_foreign.erl
@@ -25,6 +25,9 @@
 
 -export([start/2, stop/1]).
 
+%% IQ handler
+-export([iq_handler/3]).
+
 %% Hook implementations
 -export([get_disco_items/5]).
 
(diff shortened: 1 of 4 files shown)


apps/ejabberd/src/foreign_event/mod_foreign.erl
```diff
@@ -25,6 +25,9 @@
```
arcusfelis: module description is missing.

apps/ejabberd/src/foreign_event/mod_foreign.erl
```diff
@@ -73,6 +79,20 @@ get_disco_items(empty, From, To, Node, Lang) ->
...
     Acc.
 
+-spec iq_handler(From :: jid(), To :: jid(), iq()) -> iq() | ignore.
+iq_handler(_From, _To, #iq{type = get} = IQ) ->
+    IQ#iq{type = error, sub_el = ?ERR_NOT_ALLOWED};
+iq_handler(_From, _To, #iq{type = set, sub_el = ForeignEvent} = IQ) ->
+    case parse_foreign_event(ForeignEvent) of
+        {ok, Type, Request, _Publish} ->
+            maybe_dispatch_request(Type, Request, IQ);
+        error ->
```
arcusfelis: {error, missing_request_element} | {error, missing_type_field} is more preferable.

apps/ejabberd/src/foreign_event/mod_foreign_http.erl
```diff
@@ -28,9 +28,11 @@
 %%--------------------------------------------------------------------
 
 make_request(Request, OnResponse) ->
-    {Host, Path, Method, Headers, Body} = parse(Request),
-    Response = do_make_request(Host, Path, Method, Headers, Body),
-    respond(encode(Response), OnResponse),
+    spawn(fun() ->
```
arcusfelis: I would like to receive IQ result after I've made a request.
Or an IQ error if the server cannot contact http server or any other problem.

apps/ejabberd/src/foreign_event/mod_foreign.erl
mentels: @arcusfelis what exactly do you mean?

apps/ejabberd/src/foreign_event/mod_foreign_http.erl
mentels: @arcusfelis the use case we have is that we need to make the requests asynchronously. The actual response of the HTTP request is to be published to MUC or PubSub.

rslota: If you need asynchronous request, there is one thing I need to point out: You really shouldn't do that using `spawn` directly. This way you just cannot control the number of process in the system what may be dangerous or at least "not fair" to other modules. There should be a pool for that.

arkgil: There will be a pool for that :) This PR introduced only IQ handler, `spawn` is just a placeholder. There will be aconfigurable pool later :)

PR #1372, opened 2017-06-30
Go back to old now()-like behaviour for generating MAM ids.

kzemek:
The difference is that the underlying implementation ensuring
unique microsecond timestamps is now based on atomic operations,
optimized for speed and should not be any kind of bottleneck.

### DIFF
--- a/apps/ejabberd/src/mod_mam_utils.erl
+++ b/apps/ejabberd/src/mod_mam_utils.erl
@@ -177,7 +177,7 @@ microseconds_to_datetime(MicroSeconds) when is_integer(MicroSeconds) ->
 -spec generate_message_id() -> integer().
 generate_message_id() ->
     {ok, NodeId} = ejabberd_node_id:node_id(),
-    CandidateStamp = p1_time_compat:system_time(micro_seconds),
+    CandidateStamp = p1_time_compat:os_system_time(micro_seconds),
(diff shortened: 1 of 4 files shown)


apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -177,11 +177,9 @@ microseconds_to_datetime(MicroSeconds) when is_integer(MicroSeconds) ->
 -spec generate_message_id() -> integer().
 generate_message_id() ->
     {ok, NodeId} = ejabberd_node_id:node_id(),
-    %% Unique enough
-    {T1, T2, T3} = p1_time_compat:timestamp(),
-    ImprovedTS = {T1, T2,
-                  (T3 div 1000) * 1000 + p1_time_compat:unique_integer([positive]) rem 1000 },
-    encode_compact_uuid(now_to_microseconds(ImprovedTS), NodeId).
+    CandidateStamp = p1_time_compat:system_time(micro_seconds),
```
fenek: I'd use `p1_time_compat:os_system_time(micro_seconds)` as with time warp disabled by default, some environments could move Erlang system time into the future and never become correct until VM restart.

fenek:
Looks good (except for 1 comment)! It's sad we have to replicate `now()` behaviour but it's the lesser evil. :)

Issue #1354, opened 2017-06-22
xep-0333


deadjdona:
MongooseIM version: master
Installed from: source
Erlang/OTP version: 19

actually there's smack implementation https://github.com/igniterealtime/Smack/pull/90
but is it implemented on server side?

fenek:
Chat Markers are purely client-side extension. Anything the server does with the markers, is beyond the scope of this XEP. How would you like to process them inside MongooseIM?

michalwski:
@fenek there server may do sth more with chat markers. According to [XEP-0333 section 7](https://xmpp.org/extensions/xep-0333.html#requirements) the server should archive chat markers. MongooseIM doesn't archive them at this moment.

deadjdona:
not sure if they are delivered to users who was offline at this moment.

fenek:
They are not archived. `mod_mam` should definitely have a handy switch to enable markers storage.

Still, I think they should get saved by `mod_offline` and retransmitted when the recipient becomes online again. Have you checked it?

deadjdona:
Yes, they will work, but 
> It is not well suited for applications supporting multiple user devices, because anything saved in the DB can be retrieved only once, so the message history is not synchronised between devices.

fenek:
Yes, it is true. I did not know which one you are using. :) This can by bypassed by creating custom callbacks for `mod_mam` that allow archiving markers but well... it requires some Erlang knowledge.

fenek:
Hi @deadjdona 

Current master version is now able to archive chat markers with single config option. Try it out! :)

deadjdona:
thank you, it works ^_^

PR #1377, opened 2017-07-03
Archive chat markers

arkgil:
This PR addresses #1354

Proposed changes include:
* archiving chat markers when `mod_mam_utils` is used as `is_archivable_message` callback when `{archive_chat_markers, true}` option is set

### DIFF
--- a/apps/ejabberd/src/mod_mam_meta.erl
+++ b/apps/ejabberd/src/mod_mam_meta.erl
@@ -85,7 +85,11 @@ mam_type_to_core_mod(muc) -> mod_mam_muc.
 
 -spec valid_core_mod_opts(module()) -> [atom()].
 valid_core_mod_opts(mod_mam) ->
-    [add_archived_element, is_archivable_message, full_text_search, archive_groupchats];
+    [add_archived_element,
+     is_archivable_message,
+     archive_chat_markers,
+     full_text_search,
+     archive_groupchats];
 valid_core_mod_opts(mod_mam_muc) ->
     [add_archived_element, is_archivable_message, host, full_text_search].
 
--- a/apps/ejabberd/src/mod_mam_utils.erl
+++ b/apps/ejabberd/src/mod_mam_utils.erl
@@ -318,7 +318,7 @@ get_one_of_path(_Elem, [], Def) ->
 is_archivable_message(Mod, Dir, Packet=#xmlel{name = <<"message">>}) ->
     Type = xml:get_tag_attr_s(<<"type">>, Packet),
     is_valid_message_type(Mod, Dir, Type) andalso
-        is_valid_message(Mod, Dir, Packet);
+        (is_valid_message(Mod, Dir, Packet) orelse is_chat_marker(Packet));
 is_archivable_message(_, _, _) ->
     false.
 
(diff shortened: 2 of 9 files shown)


apps/ejabberd/src/mod_mam_meta.erl
```diff
@@ -85,7 +85,7 @@ mam_type_to_core_mod(muc) -> mod_mam_muc.
 
 -spec valid_core_mod_opts(module()) -> [atom()].
 valid_core_mod_opts(mod_mam) ->
-    [add_archived_element, is_archivable_message, full_text_search, archive_groupchats];
+    [add_archived_element, is_archivable_message, archive_chat_markers, full_text_search, archive_groupchats];
```
gadgetci: According to **Elvis**:
> Line 88 is too long:     [add_archived_element, is_archivable_message, archive_chat_markers, full_text_search, archive_groupchats];.

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -346,6 +346,25 @@ is_valid_message_children(_,     false, false, false) -> true;
...
+        true ->
+            do_check_markers(Packet);
+        false ->
+            false
+    end.
+
+do_check_markers(Packet) ->
+    MarkerTags = [<<"received">>, <<"displayed">>, <<"acknowledged">>],
+    HasSubtag = fun(Tag) ->
+                    case xml:get_subtag(Packet, Tag) of
```
michalwski: 1. We tend to use `exml` and `exml_query` recently to inspect packets.
2. There is (in newest exml) `exml_query:path/2` with selector `{element_with_ns, NS}` which you can use to get tags in specified namespace.
3. You need to check the namespace after all to make sure the `<received/>` tag is really a chat marker.

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -729,7 +748,12 @@ normalize_search_text(Text, WordSeparator) ->
 packet_to_search_body(Module, Host, Packet) ->
     case has_full_text_search(Module, Host) of
         true ->
-            BodyValue = xml:get_tag_cdata(xml:get_subtag(Packet, <<"body">>)),
+            BodyValue = case xml:get_subtag(Packet, <<"body">>) of
```
michalwski: Since you touched the line, please change from `xml` to `exml`.

apps/ejabberd/src/mod_mam_utils.erl
fenek: Master is still not using `exml` with new selector. It's a post-beta2 task to change it and also update the as many parts of the code as possible to use it.

michalwski: What is the a problem with using newest `exml` with beta2?

fenek: I've planned to combine newest `exml` with some code refactoring to use it already.

michalwski: Ok, dose is prevent using newest `exml` in this PR. Will this PR be in `beta2`?

fenek: Nope. It's feature freeze Monday so I'm going to create beta2 branch a bit later, to avoid "maybe one PR more... maybe another one... and maybe this one...." syndrome. :)

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -729,7 +745,12 @@ normalize_search_text(Text, WordSeparator) ->
 packet_to_search_body(Module, Host, Packet) ->
     case has_full_text_search(Module, Host) of
         true ->
-            BodyValue = xml:get_tag_cdata(xml:get_subtag(Packet, <<"body">>)),
+            BodyValue = case exml_query:subelement(Packet, <<"body">>) of
```
michalwski: This probably could be simplified with `exml_query:path(Packet, [{element, <<"body">>}, cdata], <<"">>)`

fenek:
It's broken for Riak preset and it doesn't seem like a random failure. :)

arkgil:
Done, tests are all green now! /cc @fenek

apps/ejabberd/src/mod_mam_utils.erl
```diff
@@ -318,7 +318,7 @@ get_one_of_path(_Elem, [], Def) ->
 is_archivable_message(Mod, Dir, Packet=#xmlel{name = <<"message">>}) ->
     Type = xml:get_tag_attr_s(<<"type">>, Packet),
     is_valid_message_type(Mod, Dir, Type) andalso
-        is_valid_message(Mod, Dir, Packet);
+        (is_valid_message(Mod, Dir, Packet) orelse is_chat_marker(Packet));
```
fenek: I've just noticed that this change will cause ALL messages with markers to be archived. Including the ones with "delay" or "no-store" elements, so e.g. Stream Management retransmissions would get stored, even though they must not be.

fenek:
It's great to have this option! Thanks!

PR #1381, opened 2017-07-05
MAM refactoring

michalwski:
This PR:
* simplifies `mam_lookup_messages` and `mam_muc_lookup_messages`. They now take only 3 parameters instead of 15. All parameter describing lookup criteria are now in a map. This makes both developing new backends and debugging easier
* moves common function related to jid manipulation to `mod_mam_utils` - reduces code duplication
* moves common function for calculating `msg_id` border to `mod_mam_utils` - another duplicated code reduction
* introduces `mam_iq` for function related to parsing MAM queries and building lookup parameters


apps/ejabberd/src/mod_mam_cassandra_arch.erl
```diff
@@ -275,37 +268,18 @@ message_id_to_remote_jid(PoolName, UserJID, BUserJID, MessID) ->
...
-                PageSize, LimitPassed, MaxResultLimit,
-                IsSimple) ->
+                #{owner_jid := UserJID, rsm := RSM, borders := Borders,
+                  start_ts := Start, end_ts := End, with_jid := WithJID,
+                  search_text := undefined, page_size := PageSize,
+                  limit_passed := LimitPassed, max_result_limit := MaxResultLimit,
+                  is_simple := IsSimple}) ->
     try
         PoolName = pool_name(UserJID),
         lookup_messages2(PoolName, Host,
```
fenek: I think that `lookup_messages2` should use map as well.

fenek:
Comments made in Cassandra archiver apply to the other backends as well.

michalwski:
@fenek I agree with all you said! The RP is what I had time so far, I know it's not perfect.
I'm not sure when I'll find more time to improve it.

fenek:
If you're not going to work on this e.g. until next week or even later, then I could do it for you. ;)

apps/ejabberd/src/mod_mam_cassandra_arch.erl
```diff
@@ -16,7 +17,7 @@
 %% MAM hook handlers
 -export([archive_size/4,
          archive_message/9,
-         lookup_messages/15,
+         lookup_messages/3,
```
arcusfelis: there is still archive_message/9 :)

apps/ejabberd/src/mod_mam_iq.erl
```diff
@@ -0,0 +1,140 @@
+-module(mod_mam_iq).
```
arcusfelis: I really want to name such modules mam_iq, not mod_mam_iq.
Because it's not ejabberd module.

Same with mod_mam_utils. Should be just mam_utils.

michalwski: I like the idea. I can rename `mod_mam_iq` to `mam_iq`. Regarding `mod_mam_utils` - we should rename it as well. In other PR though, I may simply don't have enough time for that...

apps/ejabberd/src/mod_mam_cassandra_arch.erl
michalwski: I know there is `archive_message/9` it wasn't my goal to change this in this PR.

apps/ejabberd/src/mod_mam.erl
```diff
@@ -484,28 +469,13 @@ handle_lookup_messages(
   From=#jid{},
   ArcJID=#jid{},
   IQ=#iq{xmlns=MamNs, sub_el = QueryEl}) ->
-    Now = p1_time_compat:system_time(micro_seconds),
+    Namespace = IQ#iq.xmlns,
```
fenek: Duplicate of `MamNs` matched in function head.

michalwski: Good catch! How could I miss that?

fenek: Brain overload caused by refactoring too long. :)

apps/ejabberd/src/mam_iq.erl
```diff
@@ -0,0 +1,140 @@
...
+fix_rsm(RSM=#rsm_in{id = <<>>}) ->
+    RSM#rsm_in{id = undefined};
+fix_rsm(RSM=#rsm_in{id = BExtMessID}) when is_binary(BExtMessID) ->
+    MessID = mod_mam_utils:external_binary_to_mess_id(BExtMessID),
+    RSM#rsm_in{id = MessID}.
+
+
+-spec elem_to_start_microseconds(jlib:xmlel()) -> 'undefined' | non_neg_integer().
+elem_to_start_microseconds(El) ->
+    maybe_microseconds(xml:get_path_s(El, [{elem, <<"start">>}, cdata])).
```
fenek: `exml_query`, please! :)

michalwski: Of course! :) That's copy-paste left over.

apps/ejabberd/src/mod_mam.erl
michalwski: Yes, that's right. Refactoring too much is overloading.

PR #1387, opened 2017-07-07
Identifier for create room command

Dzol:
Introduce + test MUC Light room creation w/ what becomes the leading part of a room JID.

Swagger update coming.

### DIFF
--- a/apps/ejabberd/src/mod_muc_light_commands.erl
+++ b/apps/ejabberd/src/mod_muc_light_commands.erl
@@ -23,6 +23,7 @@
 -export([start/2, stop/1]).
 
 -export([create_unique_room/4]).
+-export([create_identifiable_room/5]).
 -export([send_message/4]).
 -export([invite_to_room/4]).
 -export([change_affiliation/5]).
@@ -68,6 +69,22 @@ commands() ->
        ]},
       {result, {name, binary}}],
 
+     [{name, update_muc_light_room},
+      {category, <<"muc-lights">>},
+      {desc, <<"Update a MUC Light room.">>},
+      {module, ?MODULE},
+      {function, create_identifiable_room},
+      {action, update},
+      {identifiers, [domain]},
+      {args,
+       [{domain, binary},
+        {id, binary},
+        {name, binary},
+        {owner, binary},
+        {subject, binary}
+       ]},
+      {result, {id, binary}}],
+
      [{name, invite_to_room},
       {category, <<"muc-lights">>},
       {subcategory, <<"participants">>},
@@ -107,17 +124,10 @@ commands() ->
 %%--------------------------------------------------------------------
 
 create_unique_room(Domain, RoomName, Creator, Subject) ->
-    C = jid:to_lus(jid:from_binary(Creator)),
-    MUCLightDomain = gen_mod:get_module_opt_subhost(
-                       Domain, mod_muc_light, mod_muc_light:default_host()),
-    MUCService = jid:make(<<>>, MUCLightDomain, <<>>),
-    Config = make_room_config(RoomName, Subject),
-    case mod_muc_light:try_to_create_room(C, MUCService, Config) of
-        {ok, RoomUS, _} ->
-            jid:to_binary(RoomUS);
-        {error, _Reason} = E ->
-            E
-    end.
+    create_room(Domain, <<>>, RoomName, Creator, Subject).
+
+create_identifiable_room(Domain, Identifier, RoomName, Creator, Subject) ->
+    create_room(Domain, Identifier, RoomName, Creator, Subject).
 
 invite_to_room(Domain, RoomName, Sender, Recipient0) ->
     Recipient1 = jid:binary_to_bare(Recipient0),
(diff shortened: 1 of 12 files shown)


fenek:
There are no build results actually, since last green commit has `skip ci` flag. :(

michalwski:
Yes, I know... I'll fix that later today since Joe is on holiday and I need this change pretty soon.

apps/ejabberd/src/mongoose_client_api_rooms.erl
```diff
@@ -109,10 +109,9 @@ from_json(Req, State) ->
     JSONData = jiffy:decode(Body, [return_maps]),
     handle_request(Method, JSONData, Req3, State).
 
-handle_request(<<"POST">>, JSONData, Req,
+handle_request(Meth, JSONData, Req,
```
fenek: Meth, crack, acid?

michalwski: I guess this simply stands for `Method` :)

apps/ejabberd/src/mongoose_client_api_rooms.erl
```diff
@@ -122,6 +121,16 @@ handle_request(<<"POST">>, JSONData, Req,
             {true, RespReq, State}
     end.
 
+handle_request_by_meth(<<"POST">>, JSONData, User, Server) ->
```
fenek: I think that semantically both "create room" and "create identifiable room" should use `PUT`.

Actually why aren't we reusing the same method name with `id` being optional? IMO "create identifiable room" is a bit unintuitive.

michalwski: `POST` is used if you don't know the `id` of a resource you created and let the server create one. `PUT` is used when you know the `id` and what to create an item with this exact `id` which you pass in URI. The changes made in this PR doesn't reflect that, though... I'm going to change that.

test.disabled/ejabberd_tests/tests/rest_client_SUITE.erl
```diff
@@ -439,12 +453,19 @@ get_room_messages(Client, RoomID, Count, Before) ->
...
 
-create_room({_AliceJID, _} = Creds, RoomID, Subject) ->
-    Room = #{name => RoomID,
+create_room({_AliceJID, _} = Creds, RoomName, Subject) ->
+    Room = #{name => RoomName,
              subject => Subject},
     {{<<"200">>, <<"OK">>}, {Result}} = rest_helper:post(<<"/rooms">>, Room, Creds),
     proplists:get_value(<<"id">>, Result).
 
+update_room({_AliceJID, _} = Creds, RoomName, Subject, RoomID) ->
```
fenek: Why "update"? As far as I understand, this method is for creation only, right?

michalwski: According to REST's definition for `PUT` method it can be used to created or update an item. If that's what you are asking about.

fenek: Sorry, wrong choice of word. I've meant action.

This function expects the call to return "201 Created", so naming it `update_room` is misleading.

michalwski: That's right.

apps/ejabberd/src/mongoose_api_common.erl
```diff
@@ -142,17 +142,26 @@ handle_result(Verb, ok, Req, State) ->
 %%    {halt, Req2, State};
 handle_result(<<"GET">>, {ok, Result}, Req, State) ->
     {jiffy:encode(Result), Req, State};
-handle_result(<<"POST">>, {ok, Res}, Req, State) ->
+handle_result(<<"POST">>, {ok, nocontent}, Req, State) ->
```
fenek: Where these `{ok, nocontent}` may come from? They look like soem leftover.

apps/ejabberd/src/mod_muc_light_commands.erl
```diff
@@ -68,6 +69,22 @@ commands() ->
        ]},
       {result, {name, binary}}],
 
+     [{name, create_identifiable_muc_light_room},
```
fenek: I've mentioned in other comment but: I think that creating unique room is a special case of creating room with defined username (not the other way round). This is how it's specified in XEP actually.

I think it would be cleaner to either have `create_room` and `create_unique_room` functions or even single `create_room` that will create unique group when `Id == <<>>`.

apps/ejabberd/src/mongoose_api_common.erl
michalwski: @Dzol could you help with that?

michalwski: Looks like this may come from `handle_result(Verb, ok, Req, State)`. The `nocontent` variant is most probably here to handle case when we want to reply `204` to a POST or PUT request.

PR #1388, opened 2017-07-11
Add tests for directed presence behaviour

arkgil:
This PR adds new tests to `presence_SUITE` which ensure that unavailable presences are broadcasted to all contacts which user sent directed presences to.

See https://tools.ietf.org/html/rfc6121#section-4.6


michalwski:
Thanks!

PR #1394, opened 2017-07-12
Fix backoff calculation for RDBMS reconnection.

kzemek:

### DIFF
--- a/apps/ejabberd/src/rdbms/mongoose_rdbms.erl
+++ b/apps/ejabberd/src/rdbms/mongoose_rdbms.erl
@@ -473,7 +473,7 @@ connect(Settings, Retry, RetryAfter, MaxRetryDelay) ->
                        " Retrying in ~p seconds.", [Settings, Error, SleepFor]),
             timer:sleep(timer:seconds(SleepFor)),
             NextRetryDelay = RetryAfter * RetryAfter,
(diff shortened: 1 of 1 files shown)


apps/ejabberd/src/rdbms/mongoose_rdbms.erl
```diff
@@ -473,7 +473,7 @@ connect(Settings, Retry, RetryAfter, MaxRetryDelay) ->
                        " Retrying in ~p seconds.", [Settings, Error, SleepFor]),
             timer:sleep(timer:seconds(SleepFor)),
             NextRetryDelay = RetryAfter * RetryAfter,
-            connect(Settings, Retry - 1, max(MaxRetryDelay, NextRetryDelay), MaxRetryDelay)
+            connect(Settings, Retry - 1, min(MaxRetryDelay, NextRetryDelay), MaxRetryDelay)
```
michalwski: This line is not covered by tests at all. I'm pretty sure the bug would not appear if there was test for that. A regression test for this is needed.

michalwski: The coverage you can fine here: https://coveralls.io/builds/12357642/source?filename=apps%2Fejabberd%2Fsrc%2Frdbms%2Fmongoose_rdbms.erl#L476

michalwski:
The regression test for this bug is still needed @kzemek :)

PR #1398, opened 2017-07-14
Accum stage six - is it the last?

bartekgorny:
* consistent use of accumulator as a storage (mostly append-only)
* extended reach of accumulators along processing chain, as far as it makes sense
* stanza/data to be sent passed along an accumulator
* cleaned up and refactored internal broadcasting
* cleaned up s2s_in, implemented accumulators in this part of chain

Seems this is it...


apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -1233,7 +1221,15 @@ handle_incoming_message({send_filtered, Feature, From, To, Packet}, StateName, S
             FinalPacket = jlib:replace_from_to(From, To, Packet),
             case privacy_check_packet(StateData, From, To, FinalPacket, in) of
                 allow ->
-                    send_and_maybe_buffer_stanza({From, To, FinalPacket}, StateData, StateName);
+                    {Act, _, NStateData} = send_and_maybe_buffer_stanza(Acc,
+                                                                        {From, To, FinalPacket},
+                                                                        StateData),
+                    case Act of
```
> The expression on line 1227 and column 21 is nested beyond the maximum level of 3.

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -810,11 +812,17 @@ do_open_session(El, JID, StateData) ->
     ?INFO_MSG("(~w) Opened session for ~s", [StateData#state.socket, jid:to_binary(JID)]),
     Res = jlib:make_result_iq_reply(El),
     Packet = {jid:to_bare(StateData#state.jid), StateData#state.jid, Res},
-    case send_and_maybe_buffer_stanza(Packet, StateData, wait_for_session_or_sm) of
-        {_, _, NStateData, _} ->
+    Acc = mongoose_acc:new(),
```
michalwski: From what I see function `do_open_session` is called only from `maybe_open_session`. This one is only called from `wait_for_session_or_sm`. Would it be worth to create the acc (from the stanza user sent) before `maybe_open_session` is called and pass as an argument?

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -1701,32 +1728,32 @@ send_trailer(StateData) ->
...
-    % this is a very last stage, we already terminated accumulator
+-spec send_and_maybe_buffer_stanza(mongoose_acc:t(), packet(), state()) ->
+    {ok | resume, mongoose_acc:t(), state()}.
+send_and_maybe_buffer_stanza(Acc, {J1, J2, El}, State)->
     % to be removed
     {SendResult, BufferedStateData} =
         send_and_maybe_buffer_stanza({J1, J2, mod_amp:strip_amp_el_from_request(El)}, State),
-    % if we have to check packet before sending we should do it much earlier, when it is
-    % still accumulator
     mod_amp:check_packet(El, result_to_amp_event(SendResult)),
```
michalwski: Why the `Acc` is not passed to `mod_amp:check_packet` if we have it now?

apps/ejabberd/src/ejabberd_c2s.erl
```diff
@@ -3011,6 +3040,9 @@ re_route_packets(Buffer) ->
      || {From, To, Packet} <- lists:reverse(Buffer)],
     ok.
 
+maybe_enter_resume_session(StateData) ->
```
michalwski: Why exactly do we need this helper function?

apps/ejabberd/src/ejabberd_local.erl
```diff
@@ -88,15 +86,15 @@
...
                  To :: ejabberd:jid(),
-                 Acc :: mongoose_acc:t()
+                 El :: xmlel()
                  ) -> 'nothing' | 'ok' | 'todo' | pid()
                     | {'error', 'lager_not_running'} | {'process_iq', _, _, _}.
-process_iq(From, To, Acc0) ->
+process_iq(Acc0, From, To, El) ->
     Acc = mongoose_acc:require(iq_query_info, Acc0),
     IQ = mongoose_acc:get(iq_query_info, Acc),
-    El = mongoose_acc:get(element, Acc),
```
michalwski: What's the reasoning behind using element from passed arg and from Acc as before?

apps/ejabberd/src/ejabberd_c2s.erl
bartekgorny: Because mod_amp doesn't yet know what to do with it. It certainly should - now that we have accumulators mod_amp doesn't need to use stanza to carry internal info. But rewriting mod_amp is beyond the scope of this project.

apps/ejabberd/src/ejabberd_c2s.erl
bartekgorny: good point

apps/ejabberd/src/mongoose_acc.erl
```diff
@@ -45,7 +45,7 @@ terminate(M, _F, _L) ->
 
 terminate(M, received, _F, _L) ->
 %%    ?ERROR_MSG("ZZZ terminate accumulator ~p ~p", [F, L]),
```
michalwski: Do we still need this line?

apps/ejabberd/src/ejabberd_c2s.erl
bartekgorny: It makes code cleaner, I guess. In nearly all cases stream management id is taken from state, so there is no need to pass it separately.

apps/ejabberd/src/mongoose_acc.erl
bartekgorny: no :) actually I don't think we really need mogoose_acc:terminate anymore.

apps/ejabberd/src/ejabberd_local.erl
bartekgorny: I don't understand the question. In Acc we have what came in, and we don't want to change it. When we want to send something out - which may be entirely different thing - we pass it as arg. Hence this change.

apps/ejabberd/src/mongoose_client_api_sse.erl
```diff
@@ -37,15 +37,16 @@ handle_notify(_Msg, State) ->
     {nosend, State}.
 
 handle_info({route, _From, _To, Acc}, State) ->
-    handle_msg(mongoose_acc:get(name, Acc), Acc, State);
+    handle_info({route, _From, _To, Acc, mongoose_acc:get(element, Acc)}, State);
```
gadgetci: According to **Elvis**:
> Ignored variable is being used on line 40 and column 32.

gadgetci: According to **Elvis**:
> Ignored variable is being used on line 40 and column 25.

apps/ejabberd/src/ejabberd_local.erl
michalwski: Ok, now I get it. Thanks!

apps/ejabberd/src/ejabberd_sm.erl
```diff
@@ -964,14 +968,14 @@ process_iq(From, To, Acc, Packet) ->
                     ResIQ = Module:Function(From, To, IQ),
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 968. Only modules that define callbacks should make dynamic calls.

apps/ejabberd/src/ejabberd_local.erl
```diff
@@ -106,21 +104,19 @@ process_iq(From, To, Acc0) ->
                     if
```
gadgetci: According to **Elvis**:
> Replace the 'if' expression on line 104 with a 'case' expression or function clauses.

gadgetci: According to **Elvis**:
> The expression on line 104 and column 21 is nested beyond the maximum level of 3.

apps/ejabberd/src/gen_iq_handler.erl
```diff
@@ -115,39 +115,40 @@ stop_iq_handler(_Module, _Function, Opts) ->
...
-    case catch Module:Function(From, To, IQ) of
+                 From :: ejabberd:jid(), To :: ejabberd:jid(), Acc :: mongoose_acc:t(),
+                 IQ :: ejabberd:iq()) -> mongoose_acc:t() | {'error', 'lager_not_running'}.
+process_iq(_Host, Module, Function, From, To, Acc, IQ) ->
+    case catch Module:Function(From, To, Acc, IQ) of
         {'EXIT', Reason} ->
             ?ERROR_MSG("~p", [Reason]);
-        ResIQ ->
+        {Acc1, ResIQ} ->
             if
```
gadgetci: According to **Elvis**:
> Replace the 'if' expression on line 146 with a 'case' expression or function clauses.

apps/ejabberd/src/gen_iq_handler.erl
```diff
@@ -115,39 +115,40 @@ stop_iq_handler(_Module, _Function, Opts) ->
...
 
 -spec process_iq(Host :: ejabberd:server(), Module :: atom(), Function :: atom(),
-                 From :: ejabberd:jid(), To :: ejabberd:jid(),
-                 IQ :: ejabberd:iq()) -> 'ok' | {'error', 'lager_not_running'}.
-process_iq(_Host, Module, Function, From, To, IQ) ->
-    case catch Module:Function(From, To, IQ) of
+                 From :: ejabberd:jid(), To :: ejabberd:jid(), Acc :: mongoose_acc:t(),
+                 IQ :: ejabberd:iq()) -> mongoose_acc:t() | {'error', 'lager_not_running'}.
+process_iq(_Host, Module, Function, From, To, Acc, IQ) ->
+    case catch Module:Function(From, To, Acc, IQ) of
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 142. Only modules that define callbacks should make dynamic calls.

apps/ejabberd/src/ejabberd_sm.erl
```diff
@@ -964,14 +968,14 @@ process_iq(From, To, Acc, Packet) ->
...
                     if
```
gadgetci: According to **Elvis**:
> Replace the 'if' expression on line 969 with a 'case' expression or function clauses.

gadgetci: According to **Elvis**:
> The expression on line 969 and column 21 is nested beyond the maximum level of 3.

apps/ejabberd/src/ejabberd_sm.erl
bartekgorny: Wontfix - would require a rewrite of the whole iq handling machinery, which is not within current scope of work.

apps/ejabberd/src/gen_iq_handler.erl
bartekgorny: Wontfix - would require a rewrite of the whole iq handling machinery, which is not within current scope of work.

apps/ejabberd/src/ejabberd_local.erl
```diff
@@ -88,49 +86,45 @@
...
-                    Err = jlib:make_error_reply(
-                        Acc, ?ERR_FEATURE_NOT_IMPLEMENTED),
-                    ejabberd_router:route(To, From, Err)
+    process_iq(IQ, Acc, From, To, El).
+
+process_iq(#iq{xmlns = XMLNS} = IQ, Acc, From, To, _El) ->
+    Host = To#jid.lserver,
+    case ets:lookup(?IQTABLE, {XMLNS, Host}) of
+        [{_, Module, Function}] ->
+            case Module:Function(From, To, IQ) of
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 104. Only modules that define callbacks should make dynamic calls.

apps/ejabberd/src/ejabberd_sm.erl
```diff
@@ -954,36 +957,35 @@ get_max_user_sessions(LUser, Host) ->
...
+process_iq(From, To, Acc0, Packet) ->
+    Acc = mongoose_acc:require(iq_query_info, Acc0),
+    IQ = mongoose_acc:get(iq_query_info, Acc),
+    process_iq(IQ, From, To, Acc0, Packet).
+
+process_iq(#iq{xmlns = XMLNS} = IQ, From, To, Acc, Packet) ->
+    Host = To#jid.lserver,
+    case ets:lookup(sm_iqtable, {XMLNS, Host}) of
+        [{_, Module, Function}] ->
+            case Module:Function(From, To, IQ) of
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 969. Only modules that define callbacks should make dynamic calls.

apps/ejabberd/src/mongoose_acc.erl
```diff
@@ -96,15 +90,26 @@ from_kv(K, V) ->
...
--spec from_element(xmlel(), ejabberd:jid(), ejabberd:jid()) -> t().
+        };
+from_element(#iq{type = Type} = El) ->
+    #{element => El, mongoose_acc => true, name => <<"iq">>, type => Type,
+        timestamp => os:timestamp(), ref => make_ref()
+    };
+from_element(El) when is_tuple(El) ->
+    Name = <<"broadcast">>,
+    Type = element(1, El),
+    % ref and timestamp will be filled in by strip/2
```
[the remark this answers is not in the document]
bartekgorny: This is a very specific function. Normally `from_element` is called when we are just starting stanza processing, so it creates a new accumulator and consequently have to generate ref and timestamp. This one is always called when a stanza triggers a broadcast, it is always called from `strip`, which means we already have an accumulator with ref and timestamp.

apps/ejabberd/src/mongoose_acc.erl
```diff
@@ -165,23 +192,29 @@ require(Key, Acc) ->
...
 %% @doc Convert the acc before routing it out to another c2s process - remove everything except
-%% the very bare minimum (all caches, records etc), replace element with to_send
+%% the very bare minimum
+%% We get rid of all caches, traces etc.
+%% There is a special key "persistent_properties" which is also kept, use add_prop to store
+%% persistent stuff you want to pass between processes there.
 -spec strip(t()) -> t().
 strip(Acc) ->
-    El = get(to_send, Acc),
+    strip(Acc, get(element, Acc)).
```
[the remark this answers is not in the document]
bartekgorny: There are more things that remain: 
`[ref, timestamp, from, from_jid, to, to_jid, persistent_properties]`
Should `send_result` be forwarded too? I don't know, we can start an interesting discussion.

doc/developers-guide/Hooks-and-handlers.md
```diff
@@ -28,21 +28,20 @@ I.e. if `mod_offline` was not available, the code would simply crash; if it was
...
 `From`, `To` and `Packet` are the arguments passed to the handler just as they would in case of the function being called directly;
 `LServer` is [the XMPP domain for which this hook is signalled](#sidenote-multiple-domains).
 
-
 ### Getting results from handlers
 
-Apart from being able to notify the rest of the system that some event has happened by running a hook with `ejabberd_hooks:run/3`, it's also possible to [fold](#sidenote-folds) over a sequence of handlers with `ejabberd_hooks:run_fold/4`.
-Like an ordinary `lists:foldl/3`, `ejabberd_hooks:run_fold/4` also requires an initial value to be passed to the function.
-
+Hook handlers are called by "folding", meaning iterating over the list of handlers passing them a set of arguments and some initial value which gets modified and returned by every handler.
```
goddammit: Hook handlers are called by "folding".
This means that each handler on a list is passed a set of arguments and an initial value that it then modifies, returns and hands over to the next handler in line.

doc/developers-guide/Hooks-and-handlers.md
```diff
@@ -54,13 +53,10 @@ ListOfSomething = ejabberd_hooks:run_fold(a_certain_hook,
                                            StateData#state.server])
 ```
 
-In between the XMPP domain (`StateData#state.server`) and handler arguments the initial value of the accumulator being passed through the sequence of handlers is inserted - in this case an empty list (`[]`).
-All handlers attached to the hook have to accept a list as a first argument and also return a list - possibly the same list
+In between the XMPP domain (`StateData#state.server`) and handler arguments is the initial value of the accumulator being passed through the sequence of handlers is inserted - in this case an empty list (`[]`).
```
goddammit: The initial value of the accumulator being passed through the sequence of handlers (in this case an empty list (`[]`)) is inserted between the XMPP domain (`StateData#state.server`) and handler arguments.

doc/developers-guide/Hooks-and-handlers.md
```diff
@@ -70,16 +66,9 @@ If you haven't encountered the term _fold_ before, think of it as _reduce_ (like
...
-First, usage of `ejabberd_hooks:run` is strongly discouraged and some day will not be supported.
-Second, you should not use a list as an accumulator.
-
-MongooseIM is undergoing a major architectural change: we are switching to `run_fold` everywhere, and using a dedicated data structure to
-accumulate results(see ["Accumulators"](accumulators.md)).
-This data structure is implemented in `mongoose_acc` module, which has a map-like interface: it
-has `get/2`, `get/3`, `put/3` etc.
-It is instantiated with an incoming stanza, passed along throughout the processing
-chain, supplied to and returned from hook calls, and terminated when stanza is leaving MongooseIM.
+MongooseIM uses a dedicated data structure to accumulate results(see ["Accumulators"](accumulators.md)).
```
goddammit: MongooseIM uses a dedicated data structure to accumulate results (see ["Accumulators"](accumulators.md)).

doc/developers-guide/Hooks-and-handlers.md
```diff
@@ -70,16 +66,9 @@ If you haven't encountered the term _fold_ before, think of it as _reduce_ (like
...
+This data structure is implemented in `mongoose_acc` module, which has a map-like interface: it has `get/2`, `get/3`, `put/3` etc.
```
goddammit: This data structure is implemented in the `mongoose_acc` module, which has a map-like interface: it has `get/2`, `get/3`, `put/3` etc.

doc/developers-guide/Hooks-and-handlers.md
```diff
@@ -91,9 +80,9 @@ The right way to use hooks is therefore:
 * pass the modified accumulator on
 
 Handlers should store their return values in the accumulator; there are three ways to do it:
-* if it is a one-off value which is to be discarded later use `mongoose_acc:put(result, Value, Acc)`
+* if it is a one-off value whch doesn't need to be passed on along with the accumulator (can be overwritten any time), `mongoose_acc:put(result, Value, Acc)`
```
goddammit: if it is a one-off value whch doesn't need to be passed on along with the accumulator (can be overwritten any time), use `mongoose_acc:put(result, Value, Acc)`

doc/developers-guide/Hooks-and-handlers.md
```diff
@@ -106,8 +95,19 @@ Rs = mongoose_acc:get(offline_messages, Acc1, []),
 
 ```
 
+### Sidenote: something deprecated
+
+Occassionally you may find in the MongooseIM source code some calls to `ejabberd_hooks:run/3`.
```
goddammit: Occassionally you may find some calls to `ejabberd_hooks:run/3` in the MongooseIM source code.

doc/developers-guide/Hooks-and-handlers.md
```diff
@@ -106,8 +95,19 @@ Rs = mongoose_acc:get(offline_messages, Acc1, []),
...
+Under the hood it calls the same handlers with an empty accumulator.
+This is deprecated and some day will be removed.
+
+### Error handling in hooks
+
+Hooks are meant to decouple modules; in other words, the caller signals that some event took place or that it intends to use a certain feature or a set of features, but how and if those features are implemented is beyond its interest.
+Hooks don't therefore use the "let it crash" approach. Instead it is rather like "fire-and-forget", more similar in principle to the `Pid ! signal` way.
```
goddammit: For that reason hooks don't use the "let it crash" approach. 
Instead it is rather like "fire-and-forget", more similar in principle to the `Pid ! signal` way.

doc/developers-guide/accumulators.md
```diff
@@ -1,9 +1,19 @@
 # Accumulators
 
-XMPP stanza processing starts in the `ejabberd_c2s` module, which receives the stanza from a socket, or in `ejabberd_s2s_in` which receives stanzas from other MIM clusters.
+XMPP stanza processing starts in the `ejabberd_c2s` module, which receives the stanza from a socket, or in `ejabberd_s2s_in` which receives stanzas from federated XMPP clusters.
 The stanza is processed and eventually it and/or other messages are sent out, either to the original sender, to another c2s process within the same MongooseIM installation, or to another XMPP server.
 
-At the beginning of the main processing chain an accumulator is created containing the stanza and other values.
+At the beginning of the main processing chain an accumulator is created containing the original stanza, some values extracted from stanza and additional info:
```
goddammit: At the beginning of the main processing chain an accumulator is created containing the original stanza, some values extracted from the stanza and additional info:

or 

At the beginning of the main processing chain an accumulator is created containing the original stanza, some values extracted from it and additional info:

doc/developers-guide/accumulators.md
```diff
@@ -1,9 +1,19 @@
...
+
+* ref - a unique reference of the acc, useful for tracing
+* timestamp - current time
+* element - the original stanza
+* name - <<"message">>, <<"presence">> or <<"iq">>
+* type - e.g. <<"set">> (in iq)
+* attrs - attributs of the root xml element of stanza
```
goddammit: attributes of the root xml element of a stanza

doc/developers-guide/accumulators.md
```diff
@@ -1,9 +1,19 @@
...
+* from, to - full jids of sender and recipient in binary form
```
goddammit: from, to - full jids of the sender and recipient in a binary form

doc/developers-guide/accumulators.md
```diff
@@ -61,6 +71,10 @@ Acc2 = mongoose_acc:add_prop(myprop, 123, Acc1),
 V = mongoose_acc:get_prop(myprop, Acc2).
 ```
 
+The rationale behind stripping an accumulator is that many values store in it are context-dependend.
```
goddammit: The rationale behind stripping an accumulator is that many values stored in it are context-dependend.

doc/developers-guide/accumulators.md
```diff
@@ -61,6 +71,10 @@ Acc2 = mongoose_acc:add_prop(myprop, 123, Acc1),
...
+For example, `user` and `server` refer to the owner of the c2s process, in another one user will be different (and sometimes server).
```
goddammit: For example, in a c2s process `user` and `server` refer to its owner, while elsewhere the user will be different (or sometimes even a server). ?

doc/developers-guide/accumulators.md
```diff
@@ -103,7 +117,7 @@ The notion of an exit point is a bit vague, because sending a stanza out of Mong
 A single stanza entering the system creates an acc, but then may result in multiple stanzas going out - for example changing a privacy list triggers presence updates and a roster push.
 It is in fact hard to determine when the process has really been finished.
 The approach is to pass along one acc and, when something is just about to be sent out, either send a text and return the acc or pass on a clone of the acc and return the original one.
-This gives us an option to record a fact that a stanza has been sent out.
+This gives us an option to record a fact that a stanza has been sent out: every sending function calls `mongoose_acc:record_sending`, which gives you an opportunity to consult `send_result` key to check what send attempts have been made and with what effect.
```
goddammit: This gives us an option to record a fact that a stanza has been sent out.
Every sending function calls `mongoose_acc:record_sending`, which gives you an opportunity to consult the `send_result` key to check what send attempts have been made and with what effect.

doc/developers-guide/accumulators.md
```diff
@@ -140,9 +146,10 @@ The stanza can be sent or dropped, or it may error out - whatever it does this f
 Under the hood the routing modules work by passing an original accumulator and a stanza which is to be sent, which may be the original
 xml element, same element with modifications, or a broadcast.
 
-Eventually there are three things that may happen: either the stanza is sent to another c2s process - in this case we produce
-another fresh accumulator out of the data to be sent, stripping it of all caches and the likes but preserving original ref and
-timestamp (`mongoose_acc:strip/2`), or stanza is stored as offline, or is sent out to another XMPP server via `ejabberd_s2s`.
+Eventually there are three things that may happen:
+* the accumulator is sent to another c2s process, pubsub node or muc room - in this case we produce another fresh accumulator out of the data to be sent, stripping it of all caches and the likes but preserving original ref and timestamp and those attrs that should be carried on (`mongoose_acc:strip/2`)
```
goddammit: the accumulator is sent to another c2s process, pubsub node or muc room - in this case we produce another fresh accumulator out of the data to be sent, stripping it of all caches and the likes but preserving the original ref and timestamp and those attrs that should be carried on (`mongoose_acc:strip/2`)

doc/developers-guide/Hooks-and-handlers.md
```diff
@@ -97,14 +97,14 @@ Rs = mongoose_acc:get(offline_messages, Acc1, []),
...
-Occassionally you may find in the MongooseIM source code some calls to `ejabberd_hooks:run/3`.
+Occassionally you may find some calls to `ejabberd_hooks:run/3` in the MongooseIM source code.
 Under the hood it calls the same handlers with an empty accumulator.
 This is deprecated and some day will be removed.
 
 ### Error handling in hooks
 
 Hooks are meant to decouple modules; in other words, the caller signals that some event took place or that it intends to use a certain feature or a set of features, but how and if those features are implemented is beyond its interest.
-Hooks don't therefore use the "let it crash" approach. Instead it is rather like "fire-and-forget", more similar in principle to the `Pid ! signal` way.
+Fro that reason hook don't use the "let it crash" approach. Instead it is rather like "fire-and-forget", more similar in principle to the `Pid ! signal` way.
```
deadjdona: *For* that reason..

PR #1412, opened 2017-07-22
Use $(MAKE) variable for recursive make of tools/ssl.

benkard:
The makefile used a plain `make` call for the recursive build of
tools/ssl, which breaks the build on systems where `make` is not GNU
Make (such as FreeBSD).

See
https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html
for additional rationale.


fenek:
Hi @benkard 

Thanks! Good catch! :) Could you please update `test.disabled/ejabberd_tests/Makefile` in the same manner?

PR #1415, opened 2017-07-25
Add add_acl option for mod_http_upload S3 backend.

kzemek:
Setting {add_acl, false} disables adding x-amz-acl to presigned PUT urls,
making the backend compatible with some alternative S3 implementations
(e.g. Minio).

To properly support non-AWS S3, the host header value also needed to support
non-standard ports.


Mo33n:
When this change will be merged in the master ?

michalwski:
Thanks @kzemek for the quick answer to @Mo33n need!

PR #1417, opened 2017-07-26
Allow creating rooms with specific JID via API

arkgil:
This PR allows creating rooms with specific JID via API.

Before, we could only do:

`PUT /api/rooms/some_id`

and after this PR:

`PUT /api/rooms/some_id@muclight.localhost`

### DIFF
--- a/apps/ejabberd/src/mongoose_client_api_rooms.erl
+++ b/apps/ejabberd/src/mongoose_client_api_rooms.erl
@@ -40,9 +40,9 @@ allowed_methods(Req, State) ->
     {[<<"OPTIONS">>, <<"GET">>, <<"POST">>, <<"PUT">>], Req, State}.
 
 resource_exists(Req, #{jid := #jid{lserver = Server}} = State) ->
-    {RoomID, Req2} = cowboy_req:binding(id, Req),
+    {RoomIDOrJID, Req2} = cowboy_req:binding(id, Req),
     MUCLightDomain = muc_light_domain(Server),
-    case RoomID of
+    case RoomIDOrJID of
         undefined ->
             {Method, Req3} = cowboy_req:method(Req2),
             case Method of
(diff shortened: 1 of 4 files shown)


apps/ejabberd/src/mongoose_client_api_rooms.erl
```diff
@@ -122,8 +122,16 @@ handle_request_by_method(<<"PUT">>, JSONData, Req,
                          #{user := User, jid := #jid{lserver = Server}}) ->
     #{<<"name">>    := RoomName,
       <<"subject">> := Subject} = JSONData,
-    {RoomID, _Req2} = cowboy_req:binding(id, Req),
-    mod_muc_light_commands:create_identifiable_room(Server, RoomID, RoomName, User, Subject).
+    {IDOrJID, _Req2} = cowboy_req:binding(id, Req),
+    MUCLightServer = muc_light_domain(Server),
+    case extract_room_id(IDOrJID) of
+        {ok, RoomID, <<>>} ->
+            mod_muc_light_commands:create_identifiable_room(Server, RoomID, RoomName, User, Subject);
```
gadgetci: According to **Elvis**:
> Line 129 is too long:             mod_muc_light_commands:create_identifiable_room(Server, RoomID, RoomName, User, Subject);.

apps/ejabberd/src/mongoose_client_api_rooms.erl
```diff
@@ -122,8 +122,16 @@ handle_request_by_method(<<"PUT">>, JSONData, Req,
...
+        {ok, RoomID, MUCLightServer} ->
+            mod_muc_light_commands:create_identifiable_room(Server, RoomID, RoomName, User, Subject);
```
gadgetci: According to **Elvis**:
> Line 131 is too long:             mod_muc_light_commands:create_identifiable_room(Server, RoomID, RoomName, User, Subject);.

arkgil:
@michalwski  @fenek I've changed the API so that all room actions accept both JIDs and IDs. I'd really appreciate doc review, because I am not sure if what I've written makes sense.

Also there is a little code smell - I check room ID correctness in `resource_exists` callback. IMO this should be done in `malformed_request`, but when `malformed_request` is called user is not yet  authorized, there is no user info available, so we can't retrieve the user's host.

PR #1422, opened 2017-08-03
Fix shebang of tools/install

fblackburn1:
With /bin/sh, the following comparison is unsupported:
[ x"$SYSTEM" == x"yes" ]

Tested on Debian and Ubuntu

### DIFF
--- a/tools/install
+++ b/tools/install
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


erszcz:
Hi, @fblackburn1!

Thanks for spotting the issue. Can you tweak the shebang to `#!/usr/bin/env bash`? It will then work independently of bash location.

PR #1428, opened 2017-08-04
Increase escalus timeouts in rest_client_SUITE

arkgil:
This PR increases timeouts in `rest_client_SUITE`. In some places `escalus:wait_for_stanza/2` had timout set to 1 ms, which made the suite fail often.


michalwski:
Thanks!

PR #1433, opened 2017-08-09
Fix #1313 issue

### LINKED ISSUE #1313: SASL PLAIN always returns internal-error on failure
MongooseIM version: master
Installed from: git
Erlang/OTP version: does not apply

When using plain password authentication an 'internal-error' is returned in case of an authentication failure due to a wrong password. Instead a 'not-authorized' error should be returned to the client.
Please also check comments I made to an older commit that introduced this problem.

kanes115:

### DIFF
--- a/test.disabled/ejabberd_tests/tests/login_SUITE.erl
+++ b/test.disabled/ejabberd_tests/tests/login_SUITE.erl
@@ -405,7 +405,7 @@ log_one_scram(Config) ->
 
 log_non_existent_plain(Config) ->
     {auth_failed, _, Xmlel} = log_non_existent(Config),
-    #xmlel{name = <<"failure">>, children = Data} = Xmlel,
+    #xmlel{name = <<"failure">>} = Xmlel,
     #xmlel{} = exml_query:subelement(Xmlel, <<"not-authorized">>).
 
 log_non_existent_digest(Config) ->
--- a/test.disabled/ejabberd_tests/tests/login_SUITE.erl
+++ b/test.disabled/ejabberd_tests/tests/login_SUITE.erl
@@ -404,7 +404,12 @@ log_one_scram(Config) ->
 
 
 log_non_existent_plain(Config) ->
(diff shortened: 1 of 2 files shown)


test.disabled/ejabberd_tests/tests/login_SUITE.erl
```diff
@@ -404,7 +404,12 @@ log_one_scram(Config) ->
 
 
 log_non_existent_plain(Config) ->
-    {auth_failed, _, _} = log_non_existent(Config).
+    {auth_failed, _, Xmlel} = log_non_existent(Config),
+    {xmlel, <<"failure">>, _Attrs, Data} = Xmlel,
```
fenek: Although matching on `#xmlel{}` as a tuple can still be seen sometimes in old code, in the new one we always use record syntax. So here we should have `#xmlel{ name = <<"failure">> } = `

test.disabled/ejabberd_tests/tests/login_SUITE.erl
```diff
@@ -404,7 +404,12 @@ log_one_scram(Config) ->
...
+    true = lists:any(
```
fenek: Please use `exml_query` wherever it is reasonable to do so. This call can be replaced with `#xmlel{} = exml_query:subelement(Xmlel, <<"not-authorized">>).`

test.disabled/ejabberd_tests/tests/login_SUITE.erl
```diff
@@ -404,7 +404,9 @@ log_one_scram(Config) ->
 
 
 log_non_existent_plain(Config) ->
-    {auth_failed, _, _} = log_non_existent(Config).
+    {auth_failed, _, Xmlel} = log_non_existent(Config),
+    #xmlel{name = <<"failure">>, children = Data} = Xmlel,
```
fenek: `Data` is not used.

PR #1430, opened 2017-08-08
Compile and test MongooseIM with Erlang/OTP 20

michalwski:
With this PR it's possible to compile and test MongooseIM with Erlang/OTP from 18 to 20

Proposed changes include:
* updated dependencies to be compatible with Erlang/OTP from 18 to 20
   * fix to backwards incompatible changes in `worker_pool`
* update test to be able to run with Erlang/OTP 20:
   * updated escalus as in esl/escalus/150
   * updated meck to 0.8.6
   * change remaining `crypt:rand_bytes` to `strong_rand_bytes`

TODO:
* [x] fix (or even remove) SASL-DIGEST authentication (there is one test case failing in `login_SUITE`)
* [x] fix dialyzer errors
* [x]  configure travis jobs so that most of them run with Erlang/OTP 19.3, dialyzer and one other job runs with Erlang/OTP 20 and another job for Erlang/OTP 18.3
* [x] remove possible hacks for Erlang/OTP 17.x
* [x] update documentation so that it's clear MongooseIM will work with Erlang/OTP form 18 to 20
* [x] try testing MongooseIM with FIPS mode enabled
* [x] update documentation regarding FIPS mode (doc/developers-guide/OpenSSL-and-FIPS.md)
* [x] investigate decreased coverage

PR #1439, opened 2017-08-18
add erlang 20 for test

getong:


michalwski:
MongooseIM will not work with Erlang/OTP 20 without small changes. Please take a look at #1430 where I added required changed to run and test MongooseIM with Erlang/OTP 20

getong:
@michalwski Yes. I will make some change to it.

PR #1444, opened 2017-08-21
Apply transform_module_options

varnerac:
A number of modules include a transform_module_options/1 function to reformat the readable
format of in the config file to a format easier for them to use. MongooseIM currently does not
call this function. This PR calls the function (if present) on all modules.

It needs a test. Which suite should I use?


fenek:
I'd add a new test case to `apps/ejabberd/test/ejabberd_config_SUITE.erl`. It should load a config file with `mod_vcard_ldap` enabled and then verify that final config in Mnesia has the transformed value.

PR #1445, opened 2017-08-22
[skip ci] [WIP] Add load test history chart to README

bszaf:
This PR  is intended to show new feature of Tide - generating load test history.

Questions to be answered:
 - is chart looking okay?
 - where it should be placed?
 - is it clear for both non-technical and technical users?


Nyco:
> is chart looking okay?
Great to my eyes, as a starting point.

> where it should be placed?
Where it is for now, seems perfect to me.
Late on, this may grow so that we have a dedicated page on the doc.

> is it clear for both non-technical and technical users?
It is impeccably clear to me, standing between non-tech and tech.

PR #1452, opened 2017-08-29
[MUCLight][Doc] fix documentation regarding default_config [skip ci]

michalwski:
Keys in default config has to be atoms, otherwise the key will not be found in the `config schema` list.

### DIFF
--- a/doc/modules/mod_muc_light.md
+++ b/doc/modules/mod_muc_light.md
@@ -6,20 +6,20 @@ This extension consists of several modules but only `mod_muc_light` needs to be 
 
 ### Options
 
-* **host** (string, default: `"muclight.@HOST@"`) - Domain for the MUC Light service to reside under. 
+* **host** (string, default: `"muclight.@HOST@"`) - Domain for the MUC Light service to reside under.
  `@HOST@` is replaced with each served domain.
-* **backend** (atom, default: `mnesia`) - Database backend to use. 
+* **backend** (atom, default: `mnesia`) - Database backend to use.
  `mnesia` and `odbc` are supported.
-* **equal_occupants** (boolean, default: `false`) - When enabled, MUC Light rooms won't have owners. 
- It means that every occupant will be a `member`, even the room creator. 
- **Warning:** This option does not implicitly set `all_can_invite` to `true`. 
+* **equal_occupants** (boolean, default: `false`) - When enabled, MUC Light rooms won't have owners.
+ It means that every occupant will be a `member`, even the room creator.
+ **Warning:** This option does not implicitly set `all_can_invite` to `true`.
  If that option is set to `false`, nobody will be able to join the room after the initial creation request.
-* **legacy_mode** (boolean, default: `false`) - Enables XEP-0045 compatibility mode. 
+* **legacy_mode** (boolean, default: `false`) - Enables XEP-0045 compatibility mode.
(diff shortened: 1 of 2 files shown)


doc/modules/mod_muc_light.md
```diff
@@ -64,6 +65,6 @@ Valid config field types are:
              {rooms_per_page, 5},
              {rooms_in_rosters, true},
              {config_schema, ["roomname", {"display-lines", integer}]},
-             {default_config, [{"roomname", "The Room"}, {"display-lines", 30}]}
+             {default_config, [{roomname, "The Room"}, {display-lines, 30}]}
```
deadjdona: [{erlang,list_to_atom,[roomname],[]},
 {mod_muc_light_utils,process_config_field,1,
                      [{file,"/home/ubuntu/MongooseIM/_build/prod/lib/ejabberd/src/mod_muc_light_utils.erl"},
                       {line,220}]},

deadjdona:
that's why they are converted to atoms, right? do not work now without ""

[{erlang,list_to_atom,[roomname],[]},
 {mod_muc_light_utils,process_config_field,1,
                      [{file,"/home/ubuntu/MongooseIM/_build/prod/lib/ejabberd/src/mod_muc_light_utils.erl"},
                       {line,220}]},

michalwski:
That's interesting :) For me the default config didn't work with `""`. Could you show me your MUCLight config? The one which works for me is:

```erlang
  {mod_muc_light, [{host, "muclight.@HOST@"},
                   {backend, odbc},
                   {default_config, [{roomname, ""}, {subject, ""}]}
                  ]},
```

I might have done sth wrong though. What I wanted to achieve with the above is to use default empty room name, if it's not provided. @fenek maybe you can comment on that as well?

deadjdona:
{mod_muc_light, [{host, "muclight.@HOST@"},
    {backend, odbc},
    {mam, true},
    {blocking, true},
    {config_schema, ["roomname", "viewOnly"]},
    {default_config, [{"roomname", ""}, {"viewOnly", "0"}]}
]},

michalwski:
Thanks for the config @deadjdona. Looks like there is some inconsistency, if you define `default_config` **without** `config_schema` then keys have to be **atoms**. In your case when you define the `custom_schema`, then keys have to be **strings**.

We probably need to make this consistent, or at least say it clearly in the doc that if `default_config` is provided in `mod_muc_light` configuration the, `config_schema` **has to** be provided as well.

@fenek what's your comment on that?

fenek:
I confirm that for current master the keys cannot be atoms.

```
2017-08-31 15:57:37.545 [critical] <0.230.0>@gen_mod:start_module:135 Problem starting the module mod_muc_light for host <<"localhost">>
 options: [{host,"muclight.@HOST@"},
           {default_config,[{roomname,[]},{subject,[]}]}]
 error: badarg
[{erlang,list_to_atom,[roomname],[]},
 {mod_muc_light_utils,process_config_field,1,
                      [{file,"/Users/piotrnosek/Projects/MongooseIM/MongooseIM/_build/prod/lib/ejabberd/src/mod_muc_light_utils.erl"},
                       {line,220}]},
...
```

@michalwski What error do you get when you use strings?

michalwski:
Right, that's my bad. I forgot about changes to MUClight's configuration we made. Upstream version of MongooseIM is aligned with the doc correctly. Sorry for the noise I generated. Closing the PR.

PR #1459, opened 2017-08-31
[#17433] Saving unicode messages in Mysql

ferigis:
This PR addresses #17433. It saves messages with unicode characteres in MySql. It also allows to search by them when MAM is enabled.

### DIFF
--- a/apps/ejabberd/priv/mysql.sql
+++ b/apps/ejabberd/priv/mysql.sql
@@ -212,7 +212,8 @@ CREATE TABLE mam_message(
   search_body text,
   PRIMARY KEY (user_id, id),
   INDEX i_mam_message_rem USING BTREE (user_id, remote_bare_jid, id)
-)  ENGINE=InnoDB
+)  CHARACTER SET utf8
+   ENGINE=InnoDB
    PARTITION BY HASH(user_id)
    PARTITIONS 32;
 -- Partition is selected based on MOD(user_id, 32)
@@ -300,4 +301,3 @@ CREATE TABLE muc_light_blocking(
 );
 
 CREATE INDEX i_muc_light_blocking USING HASH ON muc_light_blocking(luser, lserver);
-
--- a/apps/ejabberd/src/mod_mam_muc_odbc_arch.erl
+++ b/apps/ejabberd/src/mod_mam_muc_odbc_arch.erl
@@ -151,7 +151,8 @@ archive_message_unsafe(Host, MessID, RoomID, FromNick, Packet) ->
     SMessID = integer_to_list(MessID),
     TextBody = mod_mam_utils:packet_to_search_body(mod_mam_muc, Host, Packet),
     STextBody = mongoose_rdbms:escape(TextBody),
-    write_message(Host, SMessID, SRoomID, SFromNick, SData, STextBody).
+    BTextBody = unicode:characters_to_binary(STextBody),
+    write_message(Host, SMessID, SRoomID, SFromNick, SData, BTextBody).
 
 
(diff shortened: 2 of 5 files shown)


apps/ejabberd/priv/mysql.sql
```diff
@@ -212,7 +212,8 @@ CREATE TABLE mam_message(
   search_body text,
   PRIMARY KEY (user_id, id),
   INDEX i_mam_message_rem USING BTREE (user_id, remote_bare_jid, id)
-)  ENGINE=InnoDB
+)  CHARACTER SET utf8
```
kzemek: Should we instead use [`utf8mb4`](https://stackoverflow.com/questions/30074492/what-is-the-difference-between-utf8mb4-and-utf8-charsets-in-mysql)?

fenek: It is a separate task that will involve converting all tables to utf8mb4 and providing a script to perform migration. It will also probably require changing indices' sizes.

apps/ejabberd/src/mod_mam_muc_odbc_arch.erl
```diff
@@ -151,7 +151,8 @@ archive_message_unsafe(Host, MessID, RoomID, FromNick, Packet) ->
     SMessID = integer_to_list(MessID),
     TextBody = mod_mam_utils:packet_to_search_body(mod_mam_muc, Host, Packet),
     STextBody = mongoose_rdbms:escape(TextBody),
-    write_message(Host, SMessID, SRoomID, SFromNick, SData, STextBody).
+    BTextBody = unicode:characters_to_binary(STextBody),
```
kzemek: I'd rather make sure that escaping always returns an utf8 binary; less special cases everywhere.

apps/ejabberd/src/mod_mam_odbc_arch.erl
```diff
@@ -257,7 +257,8 @@ prepare_message(Host, MessID, UserID, LocJID = #jid{}, RemJID = #jid{lresource =
     SSrcJID = jid_to_stored_binary(Host, LocJID, SrcJID),
     SDir = encode_direction(Dir),
     Data = packet_to_stored_binary(Host, Packet),
-    TextBody = mod_mam_utils:packet_to_search_body(mod_mam, Host, Packet),
+    TextBody0 = mod_mam_utils:packet_to_search_body(mod_mam, Host, Packet),
+    TextBody = unicode:characters_to_binary(TextBody0),
```
kzemek: Similar comment as with escaping - I think it would be better to ensure that `packet_to_search_body` already returns a utf8 binary (where it would, in turn, require ensuring that `mod_mam_utils:normalize_search_text` returns utf8 binary).

apps/ejabberd/priv/mysql.sql
kzemek: Ok, I can see that other tables are `utf8` already.

apps/ejabberd/src/mod_mam_muc_odbc_arch.erl
fenek: A very good idea! Felipe, can you please implement it when you're done with ODBC issue?

apps/ejabberd/src/mod_mam_odbc_arch.erl
ferigis: in this same PR we changed mod_mam_utils:normalize_search_text in order to return utf8 binaries

deadjdona:
please merge this

kzemek:
Fixed by #1748

PR #1461, opened 2017-09-04
[#17558]: Implement the REST API to delete a MUC Light room.

epretha:
This PR addresses the feature to delete a MUC room via REST API.

Proposed changes include:
* The module ```mod_muc_light_commands.erl``` and ```mod_muc_light.erl``` have been updated with a function to delete a MUC light room.
* New tests have been added to the ```muc_light_http_api_SUITE.erl```

### DIFF
--- a/apps/ejabberd/src/mod_muc_light.erl
+++ b/apps/ejabberd/src/mod_muc_light.erl
@@ -56,6 +56,9 @@
 -export([standard_config_schema/0, standard_default_config/0, default_host/0]).
 -export([config_schema/1, default_config/1]).
 
+%% For Administration API
+-export([try_to_create_room/3, delete_room/1]).
+
 %% gen_mod callbacks
 -export([start/2, stop/1]).
 
@@ -72,11 +75,7 @@
          is_room_owner/3,
          can_access_room/3,
          can_access_identity/3,
-         muc_room_pid/2,
-         delete_room/1]).
-
-%% For Administration API
--export([try_to_create_room/3]).
+         muc_room_pid/2]).
 
 %% For propEr
 -export([apply_rsm/3]).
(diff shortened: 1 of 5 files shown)


apps/ejabberd/src/mod_muc_light.erl
```diff
@@ -72,7 +72,8 @@
          is_room_owner/3,
          can_access_room/3,
          can_access_identity/3,
-         muc_room_pid/2]).
+         muc_room_pid/2,
+         delete_room/1]).
```
fenek: It's not a hook handler. It should be a part of export for administration API.

apps/ejabberd/src/mod_muc_light.erl
```diff
@@ -373,6 +374,10 @@ can_access_identity(_Acc, _Room, _User) ->
     %% 0045 MUC element with user identity and we don't want it
     false.
 
+-spec delete_room(RoomUS :: ejabberd:simple_bare_jid()) -> ok | {error, not_exists}.
```
fenek: Shouldn't be a part of "Hook handlers" section. Maybe let's create a new one like "Admin API"?

apps/ejabberd/src/mod_muc_light_commands.erl
```diff
@@ -115,6 +116,20 @@ commands() ->
         {from, binary},
         {body, binary}
        ]},
+      {result, ok}],
+
+     [{name, delete_room},
+      {category, <<"muc-lights">>},
+      {subcategory, <<"messages">>},
```
fenek: No need for a subcategory.

apps/ejabberd/src/mod_muc_light_commands.erl
```diff
@@ -115,6 +116,20 @@ commands() ->
...
+      {desc, <<"Delete a MUC room.">>},
```
fenek: Delete a MUC **Light** room.

apps/ejabberd/src/mod_muc_light_commands.erl
```diff
@@ -169,6 +184,24 @@ send_message(Domain, RoomName, Sender, Message) ->
...
+                  Sender :: binary()) ->
+                         ok | {error, not_exists} | {error, not_allowed}.
+delete_room(DomainName, RoomName, Owner) ->
+    OwnerJID = jid:binary_to_bare(Owner),
+    case muc_light_room_name_to_jid(OwnerJID, RoomName,
+                                    DomainName) of
+        {error, _} = Error ->
+            Error;
+        RoomJID ->
+            case mod_muc_light:is_room_owner(true, RoomJID, OwnerJID) of
```
fenek: This involves additional call to DB, that can be avoided with improved `find_room_with_name`, as it could return user's affiliation in the room. Besides, it uses hook handler without actually calling a hook.

test.disabled/ejabberd_tests/tests/muc_light_http_api_SUITE.erl
```diff
@@ -32,19 +32,21 @@
...
 
 success_response() ->
-
     [create_unique_room,
      create_identifiable_room,
      invite_to_room,
-     send_message_to_room].
+     send_message_to_room,
+     delete_room_by_owner,
+     delete_room_by_non_owner,
```
fenek: `delete_room_by_non_owner`, `delete_non_existent_room` and `delete_room_without_having_a_membership` should be in dedicated test group - `negative`.

apps/ejabberd/src/mod_muc_light.erl
epretha: Sure Piotr. This now moved to the administration API.

apps/ejabberd/src/mod_muc_light.erl
epretha: We could create a new section called "Admin API" and add ```try_to_create_room/3``` and ```delete_room/1``` in it. But the bit strange thing is that, the API function ```try_to_create_room/3``` is being called by the internal function ```create_room/5```. So, if it is ok to have this situation of an internal function calling an API function, then it is fine to move them to the new section "Admin API". Whats your opinion Piotr ?

apps/ejabberd/src/mod_muc_light_commands.erl
epretha: Description message is now updated Piotr.

apps/ejabberd/src/mod_muc_light_commands.erl
epretha: The ```subcategory``` is now removed.

michalwski:
There should probably be an update to swagger API spec in `doc/http-api`

apps/ejabberd/src/mod_muc_light.erl
fenek: I think it's nothing wrong with internal function calling API function.

fenek:
Indeed @michalwski, thanks. :)

apps/ejabberd/src/mod_muc_light.erl
epretha: ok Piotr. The code is now updated as per your suggestion.

PR #1462, opened 2017-09-07
MUCLight - add read only fields to room's configuration

michalwski:
Proposed changes include:
* ability to define which config fields are read only
* updated doc
* updated tests

A read only field can only be set while creating a room. Later any attempt to modify the filed will have no effect.

### DIFF
--- a/apps/ejabberd/src/mod_muc_light_room.erl
+++ b/apps/ejabberd/src/mod_muc_light_room.erl
@@ -158,12 +158,20 @@ process_config_set(#config{ raw_config = [{<<"subject">>, _}] } = ConfigReq, Roo
 process_config_set(_ConfigReq, _RoomUS, member, _AffUsers, false) ->
     {error, not_allowed};
 process_config_set(ConfigReq, {_, RoomS} = RoomUS, _UserAff, AffUsers, _AllCanConfigure) ->
-    case mod_muc_light_utils:process_raw_config(
-           ConfigReq#config.raw_config, [], mod_muc_light:config_schema(RoomS)) of
+    Schema = mod_muc_light:config_schema(RoomS),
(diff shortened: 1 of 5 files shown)


apps/ejabberd/src/mod_muc_light_room.erl
```diff
@@ -158,12 +158,20 @@ process_config_set(#config{ raw_config = [{<<"subject">>, _}] } = ConfigReq, Roo
 process_config_set(_ConfigReq, _RoomUS, member, _AffUsers, false) ->
     {error, not_allowed};
 process_config_set(ConfigReq, {_, RoomS} = RoomUS, _UserAff, AffUsers, _AllCanConfigure) ->
-    case mod_muc_light_utils:process_raw_config(
-           ConfigReq#config.raw_config, [], mod_muc_light:config_schema(RoomS)) of
+    Schema = mod_muc_light:config_schema(RoomS),
+    case mod_muc_light_utils:process_raw_config(ConfigReq#config.raw_config, [], Schema) of
         {ok, Config} ->
+            ReadOnlyFields = mod_muc_light:read_only_config_fields(RoomS),
+            ConfigWoReadOnlyFields = mod_muc_light_utils:filter_out_read_only(Config,
```
fenek: Config update should fail with an error `bad-request` if client tries to change any of read-only fields. Or at least in case of empty change list, because it will trigger room version update and config broadcast... with actually no changes at all.

michalwski: I know that... This is what I had time for now. I'll try to improve it this week.

michalwski:
It's too old already. This functionality was not needed for 3 years so I'm closing this PR. Considering changes to MUCLight room's config management, it'll probably be easier to implement read-only fields from scratch.

PR #1465, opened 2017-09-08
Commands list argument

kanes115:
I have to remember to resolve conflicts with my other pull request - http admin auth. post function from http helper has different arguments there. Here we also added delete/3 to rest_helper.


kanes115:
@fenek, tests now pass - I found a bug making them fail.

apps/ejabberd/src/mod_commands.erl
```diff
@@ -161,6 +162,18 @@ commands() ->
       {result, ok}
      ],
      [
+      {name, delete_contacts},
+      {category, <<"contacts">>},
+      {subcategory, <<"all">>},
+      {desc, <<"Remove all contacts from roster">>},
```
fenek: It removes provided contacts, not *all* of them.

apps/ejabberd/src/mongoose_client_api_contacts.erl
```diff
@@ -83,28 +83,53 @@ from_json(Req, #{jid := Caller} = State) ->
...
-    case handle_request(<<"DELETE">>, Jid, undefined, CJid) of
+    {Jid, Req2} = cowboy_req:binding(jid, Req),
+    case kind_of_deletion(Jid, CJid) of
+        {error, not_found} ->
+            {ok, Req3} = cowboy_req:reply(404, Req2),
+            {halt, Req3, State};
+        single ->
+            handle_deletion(CJid, Jid, Req2, State);
+        multiple ->
+            handle_deletion(CJid, get_all_contacts(CJid), Req2, State)
```
fenek: Removing *all* contacts isn't really that useful. I think we could expect a simple JSON list of JIDs in request body instead.

kanes115:
@fenek, waiiting for changes suggestions ;)

test.disabled/ejabberd_tests/tests/rest_client_SUITE.erl
```diff
@@ -802,6 +791,106 @@ add_and_remove(Config) ->
...
+    escalus:fresh_story(
+        Config, [{alice, 1}, {bob, 1}, {kate, 1}, {carol, 1}],
+        fun(Alice, Bob, Kate, Carol) ->
+            BCred = credentials({bob, Bob}),
+            % adds all the other users
+            lists:foreach(fun(AddContact) -> 
+                                  add_contact_check_roster_push(AddContact, {bob, Bob}) end,
+                         [Alice, Kate, Carol]),
+            % Check Bob's roster now
+            {?OK, R2} = gett("/contacts", BCred),
```
fenek: The correctness of contacts adding is verified in other tests. Of course it won't hurt to double-check it but in such case I think the verification part should be extracted into helper function, so it won't cloud the purpose of this test.

test.disabled/ejabberd_tests/tests/rest_client_SUITE.erl
```diff
@@ -802,6 +791,106 @@ add_and_remove(Config) ->
...
+            true = lists:member(AliceContact, Result),
+            true = lists:member(KateContact, Result),
+            true = lists:member(CarolContact, Result),
+            % delete Alice and Kate
+            Body = binary_to_list(AliceJID) ++ "," ++ binary_to_list(KateJID),
+            {?OK, {[{<<"not_deleted">>,[]}]}} = delete("/contacts", BCred, list_to_binary(Body)),
+            % Bob's roster consists now of only Carol
+            {?OK, R4} = gett("/contacts", BCred),
+            [CarolContact] = decode_maplist(R4),
+            IsSubscriptionRemove = fun(El) ->
```
fenek: Also worth being extracted into a dedicated function `is_subscription_remove/1`.

rel/files/ejabberd.cfg
```diff
@@ -197,6 +197,7 @@
           {"_", "/api/sse", lasse_handler, [mongoose_client_api_sse]},
           {"_", "/api/messages/[:with]", mongoose_client_api_messages, []},
           {"_", "/api/contacts/[:jid]", mongoose_client_api_contacts, []},
+          {"_", "/api/contacts", mongoose_client_api_contacts, []},
```
fenek: Is it really necessary? `[]` in the line above should make the argument optional already.

doc/http-api/client_swagger.yml
```diff
@@ -254,6 +254,19 @@ paths:
...
+        description: "Deletes provided contacts from the user's roster"
+        tags:
+            - "Contacts"
+        parameters:
+            - in: body
+              name: contacts
+              required: true
+        responses:
+            200:
+                description: "In body it returns a comma-seperated list od jids of users that could not be deleted"
```
fenek: Missing response schema.

doc/http-api/client_swagger.yml
```diff
@@ -254,6 +254,19 @@ paths:
       responses:
         204:
           description: "The user was added to a contact list."
+
+    delete:
+        description: "Deletes provided contacts from the user's roster"
+        tags:
+            - "Contacts"
+        parameters:
+            - in: body
```
fenek: Missing input schema.

apps/ejabberd/src/mod_commands.erl
```diff
@@ -278,10 +291,30 @@ add_contact(Caller, JabberID, Name, Groups) ->
...
+        {true, ok} ->
+            maybe_delete_contacts(Caller, T, NotDeleted)
+    end.
+
 delete_contact(Caller, JabberID) ->
     CJid = jid:from_binary(Caller),
     mod_roster:remove_from_roster(CJid, JabberID).
 
+-spec jid_exists(binary(), binary()) -> boolean().
+jid_exists(CJid, Jid) ->
```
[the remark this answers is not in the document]
kanes115: Are you sure delete_contact will return error when there is no such contact in callers roster? From what I've observed it will not.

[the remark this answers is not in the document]
kanes115: And that's why I made this check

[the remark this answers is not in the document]
kanes115: We can also want not to return users that were not deleted because of their 'non-existence' but it's not up to me. If it was I would say it is better to return these ones to client so this check is necessary.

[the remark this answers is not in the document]
kanes115: What I could do to make it more elegant is to move this check from case clause in 'maybe_delete_contacts' to 'delete_contact', how about that?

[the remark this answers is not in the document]
fenek: Yes, you are right. Sorry. :) No need to move this check elsewhere in my opinion.

PR #1471, opened 2017-09-12
Configurable archiveid element

kanes115:
This PR makes archive id elements introduced in XEP 0313 (<stanza-id/> v0.6 and <archived/> v0.2) configurable which means they can be enabled or disabled with the ejabberd.cfg file.

* archived - disabled by default (and warning if someone enables it with suggestion to use stanza-id)
* stanza-id - enabled by default

### DIFF
--- a/apps/ejabberd/src/mod_mam.erl
+++ b/apps/ejabberd/src/mod_mam.erl
@@ -74,8 +74,7 @@
 
 %% XML
 -import(mod_mam_utils,
-        [add_archived_elem/3,
-         add_stanzaid_elem/3,
+        [maybe_add_arcid_elems/5,
          wrap_message/6,
          result_set/4,
          result_query/2,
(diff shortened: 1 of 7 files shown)


kanes115:
What should be done:
* Investigate why if muc_all is ran before mam_all init_per_testcase does not set archiveid elements options but rather uses the default ones on the VERY FIRST RUN of the tests (on every other run on the same MongooseIM instance everything's fine)
* Check why some rooms are not empty when destroying them in end_per_* in some tests (not only tests from this PR) causing the crash in end_per_* (which does not influence the result of the tests but can be dangerous.

kanes115:
@fenek, I am especially waiting for your comments on commit baea29b (Prevent end_per_testcase from crashing). PR ready to review.

fenek:
Dialyzer is still sad. :(

doc/modules/mod_mam.md
```diff
@@ -147,7 +148,7 @@ You can change the default settings using extra parameters:
 {mod_mam_meta, [
         {backend, odbc},
 
-        {add_archived_element, true},
+        {no_stanzaid_element, true},
```
kanes115: I changed it as `add_archived_element` set to true is now considered 'not recommended'. Therefore it is not a good idea to put it in an example.

apps/ejabberd/src/mod_mam.erl
```diff
@@ -189,6 +190,13 @@ start(Host, Opts) ->
         _ ->
             ok
     end,
+    case gen_mod:get_opt(add_archived_element, Opts, undefined) of
+        undefined -> ok;
+        _ ->
+            ?WARNING_MSG("Archived element is going to be deprecated in one of future releases."
+                         " It is not recommended to use it."
+                         " Consider using <stanza-id/> element instead", [])
```
goddammit: Consider using a <stanza-id/> element instead",

apps/ejabberd/src/mod_mam_muc.erl
```diff
@@ -160,6 +160,13 @@ archive_id(SubHost, RoomName) when is_binary(SubHost), is_binary(RoomName) ->
 -spec start(Host :: ejabberd:server(), Opts :: list()) -> any().
 start(Host, Opts) ->
     ?DEBUG("mod_mam_muc starting", []),
+    case gen_mod:get_opt(add_archived_element, Opts, undefined) of
+        undefined -> ok;
+        _ ->
+            ?WARNING_MSG("Archived element is going to be deprecated in one of future releases."
+                        " It is not recommended to use it."
+                        " Consider using <stanza-id/> element instead", [])
```
goddammit: Consider using a <stanza-id/> element instead",

doc/modules/mod_mam.md
```diff
@@ -25,14 +25,15 @@ For now `odbc` backend has very limited support for this feature, while `cassand
 ### Options
 
 * **backend** (atom, default: `odbc`) - Database backend to use. `odbc`, `riak` and `cassandra` are supported.
-* **add_archived_element** (boolean, default: `false`) - Add `<archived/>` element from MAM v0.2.
+* **add_archived_element** (boolean, default: `false`) - Add `<archived/>` element from MAM v0.2. **Be aware:** The element is going to be deprecated in one of future releases so it's not recommended to enable this option.
```
goddammit: Add an `<archived/>` element from MAM v0.2. **Please note:** The element is going to be deprecated in one of future releases so it's not recommended to enable this option.

?

doc/modules/mod_mam.md
```diff
@@ -25,14 +25,15 @@ For now `odbc` backend has very limited support for this feature, while `cassand
...
+* **no_stanzaid_element** (boolean, default: `false`) - Do not add `<stanza-id/>` element from MAM v0.6.
```
goddammit: Do not add a `<stanza-id/>` element from MAM v0.6.

rel/files/ejabberd.cfg
```diff
@@ -840,7 +840,12 @@
 %   {muc, [
 %     {host, "muc.@HOST@"}
       %% As with pm, top-level options can be overriden for MUC archive
-%   ]}
+%   ]},
+%
+    %% Do not use <stanza-id/> element (by default stanzaid is used)
```
goddammit: Do not use a <stanza-id/> element (by default stanzaid is used)

kanes115:
@fenek, @goddammit, corrections made!

PR #1480, opened 2017-09-20
[skip ci] mod cowboy

goddammit:
please see if makes sense

This PR addresses #

Proposed changes include:

### DIFF
--- a/doc/advanced-configuration/Listener-modules.md
+++ b/doc/advanced-configuration/Listener-modules.md
@@ -99,6 +99,10 @@ Unlike `ejabberd_c2s`, it doesn't use `ejabberd_receiver` or `ejabberd_listener`
 ### HTTP module: `mod_cowboy`
 
 This module provides an additional routing layer on top of HTTP(s) or WS(S) protocols.
+Regular cowboy handlers cannot cleanly handle both protocols on the same port. 
+The routing layer allows working with both - WebSocket connections and services connecting over HTTP.
+
+
 Here's an example of its configuration (added to ejabberd_cowboy modules list described above):
 
 ```Erlang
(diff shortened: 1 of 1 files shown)


doc/advanced-configuration/Listener-modules.md
```diff
@@ -99,6 +99,10 @@ Unlike `ejabberd_c2s`, it doesn't use `ejabberd_receiver` or `ejabberd_listener`
 ### HTTP module: `mod_cowboy`
 
 This module provides an additional routing layer on top of HTTP(s) or WS(S) protocols.
+Regular cowboy handlers cannot cleanly handle both protocols on the same port.
```
studzien: I think it's perfectly fine to have a WS and HTTP handlers attached to the same listener (port) but different resources.
@michalwski,  wasn't it about the **same resource**?

michalwski: It is all possible only not in a clean way, which is sth subjective though. IMO clean cowboy handlers operates on single transport WS or HTTP (not both) and possible single or at least (similar resources). From what I understand `mod_cowboy` itself may not be clean but makes it easier and cleaner for the layers on top of it.

Nyco: It is "Cowboy" with an uppercase "C".

fenek: I think we should rephrase it into something like this:

It allows other HTTP/WS modules to coexist under the same URL on the single port.
Packets are forwarded to them based on protocol.
This mechanism is transparent for actual handlers so the path sharing does not require additional code.
Example: If you wish, you can use BOSH and WS XMPP handlers (`mod_bosh`, `mod_websockets`) on single port and URL without any code modifications.

What do you think?

michalwski: Yes, @fenek described it in the most accurate way in my opinion.

michalwski: IMO, it doesn't have to start with the uppercase "C". It's worth surrounding it with `` though like `cowboy`.

goddammit: It allows other HTTP/WS modules to coexist under the same URL on the single port.
Packets are forwarded to them based on the protocol.
This mechanism is transparent to actual handlers so the path sharing does not require any additional code.
Example: If you wish, you can use BOSH and WS XMPP handlers (mod_bosh, mod_websockets) on a single port and a URL without any code modifications.

PR #1485, opened 2017-09-21
[skip ci] Review and modernize Getting Started doc.

kzemek:

### DIFF
--- a/doc/user-guide/Getting-started.md
+++ b/doc/user-guide/Getting-started.md
@@ -1,29 +1,42 @@
 ## Installation
 
-### Getting started with MongooseIM on Mac OS X
+### Getting started with MongooseIM on Mac
 
-This short guide will show you how to compile MongooseIM from source code or install binaries from a DMG image on Mac OS X, and get users chatting right away.
+In this short guide we will set MongooseIM up and get your users chatting right away.
+You can either compile everything from the source code or install binaries from a DMG image on Mac.
 
 #### Install from source code
 
 ##### Prerequisites for source code installation
 
-MongooseIM is supported on Mac OS X 10.6.8 and later. Before you can compile and run MongooseIM, you also need the following to be installed on your system:
+MongooseIM is supported on Mac OS X / macOS 10.6.8 and later.
+Before you can compile and run MongooseIM, you will need to install the following packages:
 
-* GNU Make and GCC (the GNU Compiler Collection). To ensure that these are available, you can install the Command Line Tools for Xcode, available via Xcode or from the Apple Developer website.
-* Git – `brew install git`
-* Erlang/OTP 18.3 or higher – `brew install erlang`
+* C and C++ compiler
+* Git
+* Erlang/OTP 18.3 or higher
+* OpenSSL
 
-An easy way to install some of the packages is by using a package manager, such as [Homebrew](http://brew.sh) – the Homebrew commands are provided here.
+We recommend using [Homebrew](http://brew.sh) to manage packages on your Mac.
+With Homebrew installed, getting all dependencies is a matter of running the following commands:
+
+```bash
+$ xcode-select --install # install compilation tools
+$ brew install git erlang openssl
+```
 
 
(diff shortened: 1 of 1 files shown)


doc/user-guide/Getting-started.md
```diff
@@ -2,59 +2,74 @@
...
-MongooseIM is supported on Mac OS X 10.6.8 and later. Before you can compile and run MongooseIM, you also need the following to be installed on your system:
+MongooseIM is supported on Mac OS X / macOS 10.6.8 and later.
+Before you can compile and run MongooseIM, you will need to install the following packages:
 
-* GNU Make and GCC (the GNU Compiler Collection). To ensure that these are available, you can install the Command Line Tools for Xcode, available via Xcode or from the Apple Developer website.
-* Git – `brew install git`
-* Erlang/OTP 17.5 or higher – `brew install erlang`
+* C and C++ compiler
+* Git
+* Erlang/OTP 17.5 or higher
```
michalwski: How about we suggest 18.3 or higher? MongooseIM will soon work with Erlang 18, 19 and 20. Support for 17 will be ditched after merging #1430

doc/user-guide/Getting-started.md
```diff
@@ -2,59 +2,74 @@
 
 ### Getting started with MongooseIM on Mac OS X
 
-This short guide will show you how to compile MongooseIM from source code or install binaries from a DMG image on Mac OS X, and get users chatting right away.
+This short guide will show you how to compile MongooseIM from source code or install binaries from a DMG image on macOS, and get users chatting right away.
```
fenek: It's called "macOS" since Sierra but what about earlier versions? :) Maybe a good compromise would be to simply say "(...)install binaries from a DMG image on Mac, and get(...)"?

doc/user-guide/Getting-started.md
```diff
@@ -2,59 +2,74 @@
...
+MongooseIM is supported on Mac OS X / macOS 10.6.8 and later.
+Before you can compile and run MongooseIM, you will need to install the following packages:
 
-* GNU Make and GCC (the GNU Compiler Collection). To ensure that these are available, you can install the Command Line Tools for Xcode, available via Xcode or from the Apple Developer website.
-* Git – `brew install git`
-* Erlang/OTP 17.5 or higher – `brew install erlang`
+* C and C++ compiler
+* Git
+* Erlang/OTP 17.5 or higher
+* OpenSSL
```
fenek: And `expat`. I'm not sure if it is built into MacOS?

doc/user-guide/Getting-started.md
```diff
@@ -74,15 +89,29 @@ Registered users wishing to add an existing account to Adium should enter the Mo
...
+
+using your own hostname, for example:
+
+```erlang
+{hosts ["example.org"] }.
+```
+
 Save the configuration file and restart the MongooseIM server.
-A user’s Jabber ID will then use the domain instead of localhost, for example: `user1@example.org`
+A user’s Jabber ID will then contain the new domain instead of localhost, for example: `user1@example.org`
```
fenek: I would add a warning that changing the list of supported domains won't automatically migrate existing accounts.

doc/user-guide/Getting-started.md
kzemek: It's built in

doc/user-guide/Getting-started.md
```diff
@@ -1,60 +1,75 @@
 ## Installation
 
-### Getting started with MongooseIM on Mac OS X
+### Getting started with MongooseIM on Mac
 
-This short guide will show you how to compile MongooseIM from source code or install binaries from a DMG image on Mac OS X, and get users chatting right away.
+This short guide will show you how to compile MongooseIM from source code or install binaries from a DMG image on Mac, and get users chatting right away.
```
goddammit: In this short guide we will set MongooseIM up and get your users chatting right away.
You can either compile everything from the source code or install binaries from a DMG image on Mac.

PR #1492, opened 2017-09-29
Update docs for mod_aws_sns

kanes115:
One of the config options needs to be changed:
token_muc_messages_topic --> muc_messages_topic
because that's how it's fetched module itself


Nyco:
+* **muc_messages_topic** (optional, string, default: unset)

fenek:
@Nyco 

What is wrong with this line?

PR #1493, opened 2017-09-29
Update docs for mod_push_service_mongoosepush

kanes115:


doc/modules/mod_push_service_mongoosepush.md
```diff
@@ -26,6 +26,7 @@ It must be defined in `http_connections` setting.
 
 * **pool_name** (atom, required) - name of the pool to use (as defined in http_connections)
 * **api_version** (string, default: `v1`) - REST API version to be used. Currently only `v1` is supported
```
fenek: This is no longer valid. Current MongoosePush expects `v2` (http://mongooseim.readthedocs.io/en/latest/user-guide/Push-notifications/#mod_push_service_mongoosepush). Also, `v2` is actually the default value now.

PR #1498, opened 2017-10-02
Update mod_carbon docs

kanes115:
There was a missing config option

### DIFF
--- a/doc/modules/mod_carboncopy.md
+++ b/doc/modules/mod_carboncopy.md
@@ -56,7 +56,8 @@ For an Erlang-based test suite, please see [/esl/ejabberd_tests/blob/master/test
 
 ### Options
 
-None.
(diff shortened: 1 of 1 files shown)


doc/modules/mod_carboncopy.md
```diff
@@ -56,7 +56,8 @@ For an Erlang-based test suite, please see [/esl/ejabberd_tests/blob/master/test
 
 ### Options
 
-None.
+* **iqdisc** (default: no_queue) - determines how many queues there will be created for serving carbon copy discovery requests.
```
fenek: Parent page explains the meaning of `iqdisc` and its values (http://mongooseim.readthedocs.io/en/latest/advanced-configuration/Modules/#module-list). In the module page, please only describe what is the default value of `iqdisc` for this module.

PR #1499, opened 2017-10-02
Update docs for mod_keystore

kanes115:


kanes115:
I think that this piece of documentation is not true anymore because I couldn't find such a function in the module. Hook API is used outside the module (mod_auth_token).

fenek:
I think it is a good idea to replace this fragment with a link to actual function in `mod_auth_token` that calls this hook as an example.

doc/modules/mod_keystore.md
```diff
@@ -40,17 +40,7 @@ The module public API is hook-based:
...
-
-```erlang
--spec get_key(Domain, KeyName) -> Result when
-      Domain :: ejabberd:server(),
-      KeyName :: mod_keystore:key_name(),
-      Result :: mod_keystore:key_list().
-get_key(Domain, KeyName) ->
-    ejabberd_hooks:run_fold(get_key, Domain, [], [{KeyName, Domain}]).
-```
+Example usage can be found in [mod_auth_token:get_key_for_user/2](https://github.com/esl/MongooseIM/blob/26a23a260b14176c103339d745037cf4e3c1c188/apps/ejabberd/src/mod_auth_token.erl#L367)
```
goddammit: An example of usage usage can be found in

PR #1512, opened 2017-10-06
Fix routing messages to bare jids

michalwski:
The `Packet` has to be passed to mongoose_acc:strip in this case.
Otherwise the recipient will get the original stanza from Acc which
may not be correct.
Simple case to reproduce:
1. `User A` sends and IQ to MongooseIM
2. This IQ results in a `message` stanza to `User B` (bare jid)
3. Withtout this change `User B` recieves the IQ sent by `User A`


fenek:
Needs a test case to reproduce it.

michalwski:
I know you do :) I'm on holiday ATM. I hope the description I left leads to
a test case easy to write. Plus I have this tested in a private fork.

09.10.2017 1:48 PM "Piotr Nosek" <notifications@github.com> napisał(a):

PR #1515, opened 2017-10-10
Make s2s files pass elivs checks

mkacper:
This PR improves s2s modules code style in order to make the modules pass Elvis checks.

Changes that was made:
* decrease nesting level in some functions
* remove if expressions
* improve spec
* remove macros from function calls
* cosmetic changes (too long lines, indentation)

### DIFF
--- a/apps/ejabberd/src/ejabberd_s2s_in.erl
+++ b/apps/ejabberd/src/ejabberd_s2s_in.erl
@@ -210,17 +210,20 @@ wait_for_stream({xmlstreamstart, _Name, Attrs}, StateData) ->
                     _Else ->
                         []
                 end,
-            StartTLS = if
-                           StateData#state.tls_enabled ->
-                               [];
-                           (not StateData#state.tls_enabled) and (not StateData#state.tls_required) ->
-                              [#xmlel{name = <<"starttls">>,
-                                      attrs = [{<<"xmlns">>, ?NS_TLS}]}];
-                           (not StateData#state.tls_enabled) and StateData#state.tls_required ->
-                              [#xmlel{name = <<"starttls">>,
-                                      attrs = [{<<"xmlns">>, ?NS_TLS}],
-                                      children = [#xmlel{name = <<"required">>}]}]
-                       end,
+            StartTLS =
+                case StateData of
+                    StateData when StateData#state.tls_enabled ->
+                        [];
+                    StateData when (not StateData#state.tls_enabled) and
+                                   (not StateData#state.tls_required) ->
+                        [#xmlel{name = <<"starttls">>,
+                                attrs = [{<<"xmlns">>, ?NS_TLS}]}];
+                    StateData when (not StateData#state.tls_enabled) and
+                                   StateData#state.tls_required ->
+                        [#xmlel{name = <<"starttls">>,
+                                attrs = [{<<"xmlns">>, ?NS_TLS}],
+                                children = [#xmlel{name = <<"required">>}]}]
+                end,
             case SASL of
                 {error_cert_verif, CertVerifyResult, Certificate} ->
                     CertError = fast_tls:get_cert_verify_string(CertVerifyResult, Certificate),
@@ -675,11 +678,7 @@ get_cert_domains(Cert) ->
         lists:flatmap(
           fun(#'Extension'{extnID = ?'id-ce-subjectAltName',
                            extnValue = Val}) ->
-                  BVal = if
-                             is_list(Val) -> list_to_binary(Val);
-                             is_binary(Val) -> Val;
-                             true -> Val
-                         end,
+                  BVal = convert_extnval_to_bin(Val),
                   DSAN = ?PKIXIMPLICIT:decode('SubjectAltName', BVal),
                   get_ld_or_pcld_from_decoded_dsan(DSAN);
              (_) ->
@@ -844,3 +843,8 @@ get_ld_from_jid(#jid{luser = <<"">>,
     [LD];
 get_ld_from_jid(_) ->
     [].
+
+convert_extnval_to_bin(Val) when is_list(Val) ->
+    list_to_binary(Val);
+convert_extnval_to_bin(Val) ->
+    Val.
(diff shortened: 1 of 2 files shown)


apps/ejabberd/src/ejabberd_s2s_in.erl
```diff
@@ -150,16 +149,17 @@ init([{SockMod, Socket}, Opts]) ->
...
+        case ejabberd_config:get_local_option(s2s_use_starttls) of
+            UseTls when (UseTls==undefined) or (UseTls==false) ->
+                {false, false, false};
+            UseTls when (UseTls==true) or (UseTls==optional) ->
+                {true, false, false};
+            required ->
+                {true, true, false};
+            required_trusted ->
+                {true, true, true}
+        end,
```
bartekgorny: I'd suggest extracting it into a pattern-matched function, like:
```
tls_params(undefined) -> {false, false, false};
tls_params(false) -> {false,false, false};
tls_params(true) -> {true, false, false};
...
```
IMO it would be much more readable.

apps/ejabberd/src/ejabberd_s2s_in.erl
```diff
@@ -202,40 +202,28 @@ wait_for_stream({xmlstreamstart, _Name, Attrs}, StateData) ->
...
+                    StateData when (not StateData#state.tls_enabled) and
+                                   (not StateData#state.tls_required) ->
+                        [#xmlel{name = <<"starttls">>,
+                                attrs = [{<<"xmlns">>, ?NS_TLS}]}];
+                    StateData when (not StateData#state.tls_enabled) and
+                                   StateData#state.tls_required ->
+                        [#xmlel{name = <<"starttls">>,
+                                attrs = [{<<"xmlns">>, ?NS_TLS}],
+                                children = [#xmlel{name = <<"required">>}]}]
+                end,
```
bartekgorny: Same as above. It is legacy code and it is horrible, we try to improve it whenever possible, this is a good opportunity do to so.

apps/ejabberd/src/ejabberd_s2s_in.erl
```diff
@@ -328,53 +317,11 @@ wait_for_feature_request({xmlstreamelement, El}, StateData) ->
...
-                                         #xmlel{name = <<"failure">>,
-                                                attrs = [{<<"xmlns">>, ?NS_SASL}]}),
-                            send_text(StateData, ?STREAM_TRAILER),
-                            {stop, normal, StateData}
-                    end;
+                    CertData = (StateData#state.sockmod):get_peer_certificate(
+                                 StateData#state.socket),
+                    AuthRes = get_auth_res(CertData, AuthDomain,
+                                                    StateData),
+                    handle_auth_res(AuthRes, AuthDomain, StateData);
```
bartekgorny: 👍

apps/ejabberd/src/ejabberd_s2s_in.erl
```diff
@@ -832,3 +720,132 @@ match_labels([DL | DLabels], [PL | PLabels]) ->
...
+    case jid:from_binary(D) of
+        #jid{luser = <<"">>,
+             lserver = LD,
+             lresource = <<"">>} ->
+            [LD];
+        _ ->
+            []
+    end.
+
+get_ld_or_pcld_from_decoded_dsan({ok, SANs}) ->
```
bartekgorny: It is a cool name for a function, but I'd suggest making it a little bit more telling :)

apps/ejabberd/src/ejabberd_s2s_out.erl
```diff
@@ -467,42 +450,18 @@ wait_for_features({xmlstreamelement, El}, StateData) ->
...
+                          get_acc_with_new_req(Attr, El1, Acc);
                      (_, Acc) ->
                           Acc
                   end, {false, false, false}, Els),
-            if
-                (not SASLEXT) and (not StartTLS) and
-                StateData#state.authenticated ->
+            case {SASLEXT, StartTLS, StartTLSRequired, StateData} of
+                {SASLEXT, StartTLS, _, StateData} when (not SASLEXT) and (not StartTLS) and
+                                                       StateData#state.authenticated ->
```
bartekgorny: ibidem

mkacper: @bartekgorny What do you mean here?

bartekgorny: I mean the same as above - turn it into a set of pattern-matching functions, would improve readability.

apps/ejabberd/src/ejabberd_s2s_out.erl
```diff
@@ -467,84 +450,16 @@ wait_for_features({xmlstreamelement, El}, StateData) ->
...
-                                  {SEXT, true, Req};
-                              _ ->
-                                  Acc
-                          end;
+                             children = Els1} = _El1, Acc) ->
+                          Attr =  xml:get_attr_s(<<"xmlns">>, Attrs1),
+                          get_acc_with_new_sext(Attr, Els1, Acc);
+                     (#xmlel{name = <<"starttls">>, attrs = Attrs1} = El1, Acc) ->
+                          Attr = xml:get_attr_s(<<"xmlns">>, Attrs1),
+                          get_acc_with_new_req(Attr, El1, Acc);
```
fenek: `get_acc_with_new_tls(_params)`

apps/ejabberd/src/ejabberd_s2s_out.erl
```diff
@@ -467,84 +450,16 @@ wait_for_features({xmlstreamelement, El}, StateData) ->
...
-                     send_db_request(StateData);
-                 true ->
-                     ?DEBUG("restarted: ~p", [{StateData#state.myname,
-                                               StateData#state.server}]),
-                     % TODO: clear message queue
-                     ejabberd_socket:close(StateData#state.socket),
-                     {next_state, reopen_socket, StateData#state{socket = undefined,
-                                                                 use_v10 = false}, ?FSMTIMEOUT}
-            end;
+            try_establish_conn({SASLEXT, StartTLS, StartTLSRequired, StateData});
```
fenek: Connection is already established and the results of this function may be very different. What do you think about `handle_parsed_features`?

PR #1538, opened 2017-10-24
introduce Tide

Nyco:
we have to be consistent here


fenek:
I think we should rather mention Amoc here, not Tide, right? Tide is read only for public.

PR #1541, opened 2017-10-25
Fix complete attribute calculation for backward pagination in MAM

arcusfelis:
This PR addresses fixes "complete" attribute when paginating through results sorted backward.

- What does "complete" attribute do?
- It says the client when to stop sending requests asking for more data.

Proposed changes include:
* attribute "complete" is set correctly.

Behaviour after: Complete attribute can be used to detect if the extracted page is the last one IN THE RESULT SET.
Behaviour before: Complete attribute can be used to detect if the extracted page is the last page (contains most recent data). So, the flag was broken when paginating in the backward direction.

From MAM XEP:

>When the results returned by the server are complete (that is: when they are the last page of the result set), the server MUST include a 'complete' attribute on the <fin> element, with a value of 'true'. If it is not the last page of the result set, the server MUST either omit the 'complete' attribute, or give it a value of 'false'.

### DIFF
--- a/apps/ejabberd/src/mam_iq.erl
+++ b/apps/ejabberd/src/mam_iq.erl
@@ -126,15 +126,19 @@ form_to_lookup_params(QueryEl) ->
     is_simple => mod_mam_utils:form_decode_optimizations(QueryEl)}.
 
 common_lookup_params(QueryEl) ->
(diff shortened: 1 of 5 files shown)


arcusfelis:
not sure why it was not merged

arcusfelis:
It's closed, because the fix is in #1541.

PR #1543, opened 2017-10-26
Use mongoose_acc in mod_amp

chrzaszcz:
- Save AMP state in the accumulator instead of storing modified rules in the message itself.
- Handle the corner case when the message is stored in MAM but sending it to the recipient fails.
- Document AMP logic in more detail

### DIFF
--- a/apps/ejabberd/include/amp.hrl
+++ b/apps/ejabberd/include/amp.hrl
@@ -4,6 +4,8 @@
 %% @copyright 2014 Erlang Solutions, Ltd.
 %% This work was sponsored by Grindr LLC
 
+-include_lib("exml/include/exml.hrl").
+
 -define(NS_AMP, <<"http://jabber.org/protocol/amp">>).
 -define(NS_AMP_FEATURE, <<"http://jabber.org/feature/amp">>).
 
@@ -11,7 +13,9 @@
 -record(amp_rule, {
           condition :: amp_condition(),
           value     :: amp_value(),
-          action    :: amp_action()
+          action    :: amp_action(),
+          element   :: #xmlel{} | undefined,
+          result    :: amp_match_result() | undefined
          }).
 
 %% @doc This includes all well-formed but nonsensical rules, like:
--- a/apps/ejabberd/src/mod_amp.erl
+++ b/apps/ejabberd/src/mod_amp.erl
@@ -151,8 +151,11 @@ process_amp_rules(Packet, From, Event, Rules) ->
             send_error_and_drop(Packet, From, ValidationError, InvalidRule);
         [] ->
             Strategy = determine_strategy(Packet, From, Event),
-            ApplyResult = fold_apply_rules(Packet, From, Strategy, ValidRules),
-            take_action(Packet, From, Event, ApplyResult)
+            RulesWithResults = apply_rules(fun(Rule) ->
+                                                   resolve_condition(From, Strategy, Event, Rule)
+                                           end, ValidRules),
+            PacketResult = take_action(Packet, From, RulesWithResults),
+            return_result(PacketResult, Packet, RulesWithResults)
     end.
 
 %% @doc ejabberd_hooks helpers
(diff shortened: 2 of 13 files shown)


apps/ejabberd/src/mod_amp.erl
```diff
@@ -166,39 +138,39 @@ determine_strategy(Packet, From, Event) ->
...
-fold_apply_rules(Packet, From, Strategy, [Rule|Rest]) ->
-    case resolve_condition(From, Strategy, Rule) of
-        no_match -> fold_apply_rules(Packet, From, Strategy, Rest);
-        Status -> {Status, Rule}
+apply_rules(F, Rules) ->
+    [Rule#amp_rule{result = F(Rule)} || Rule <- Rules].
+
+-spec resolve_condition(jid(), amp_strategy(), amp_event(), amp_rule()) -> amp_match_result().
+resolve_condition(From, Strategy, Event, Rule) ->
+    Result = ejabberd_hooks:run_fold
```
michalwski: My eye parser raised a warning reading this line. It'd read better if `(` is in this line, and not below.
This may be only my parser's oddity, though.

doc/modules/mod_amp.md
```diff
@@ -74,3 +74,56 @@ mod_amp supports.
...
+### Modifications
+
+The following behaviour differs from or extends the guidelines provided in the XEP.
+
+* The action for the `deliver` condition with value `stored` is deferred until the message is stored by `mod_mam` or `mod_offline`.
+* The action for the `deliver` condition with value `direct` is deferred until the message is sent to the recipient's socket.
+
+### Server Processing Details
+
+When a message with AMP rules is being processed by the server, several system events may occur. For a given event, the rules are processed and each of them can get the *matched* or *undecided* status or, if the conditions are not met, it gets no status. If any rules get the *matched* status, the action for the first of them is performed. After that, the rule list is filtered so that only the *undecided* ones are left in the message, as they may be matched later.
```
michalwski: We try to follow `one sentence per line` rule, which makes future editing and comparing easier. It's not sth written on stone though.

test.disabled/ejabberd_tests/tests/amp_big_SUITE.erl
```diff
@@ -279,6 +326,30 @@ notify_deliver_to_online_user_recipient_privacy_test(Config) ->
               client_receives_nothing(Bob)
       end).
 
+notify_deliver_to_online_user_broken_connection_test(Config) ->
+    escalus:fresh_story(
+      Config, [{alice, 1}, {bob, 1}],
+      fun(Alice, Bob) ->
+              %% given
+              Rule = {deliver, case ?config(offline_storage, Config) of
```
michalwski: Could you move the `case` to a separate function? It will improve readability.

doc/modules/mod_amp.md
chrzaszcz: I will respect that rule. Seems a bit counter intuitive (lines become one chunk of text anyway) but it can minimize the scope of future changes indeed. The problem with `one sentence per line` is that it often encourages rewording the text just to have fitting line lengths. The content sometimes suffers during the process :)

test.disabled/ejabberd_tests/tests/amp_big_SUITE.erl
chrzaszcz: It's always a tradeoff between readability and conciseness. I could use the helper here, for me it makes no big difference though.

apps/ejabberd/include/amp.hrl
```diff
@@ -4,14 +4,18 @@
...
 -define(NS_AMP, <<"http://jabber.org/protocol/amp">>).
 -define(NS_AMP_FEATURE, <<"http://jabber.org/feature/amp">>).
 
 %% @doc This represents well-formed and legal rules
 -record(amp_rule, {
           condition :: amp_condition(),
           value     :: amp_value(),
-          action    :: amp_action()
+          action    :: amp_action(),
+          element   :: #xmlel{} | undefined,
```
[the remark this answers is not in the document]
chrzaszcz: There was definitely a reason to use `#xmlel{}` but I can't recall it now :) so I will make the change.

doc/modules/mod_amp.md
michalwski: A single sentence may be longer than the line limit IMO. More important is to split the line after `.`

chrzaszcz: Of course it may be longer, but then it is tempting to shorten sentences just to make them easily editable :)

michalwski:
@chrzaszcz I wander if we can still have these changes in master. I know there is many conflicts now, but what's in the PR is still worth merging.

PR #1547, opened 2017-10-30
Return 404 when vcard not found

kanes115:
I found that we return code 503 when a vcard has not been found after clients' iq get request. According to xep it should return either empty vcard or a 404 (item-not-found) error code.

### DIFF
--- a/test.disabled/ejabberd_tests/tests/vcard_simple_SUITE.erl
+++ b/test.disabled/ejabberd_tests/tests/vcard_simple_SUITE.erl
@@ -136,9 +136,10 @@ retrieve_own_card(Config) ->
 
 
 
-%% If no vCard exists or the user does not exist, the server MUST
-%% return a stanza error, which SHOULD be either
-%% <service-unavailable/> or <item-not-found/>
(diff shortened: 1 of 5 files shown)


michalwski:
Would it be possible for you to find some time to change or write tests for the changed code?

kanes115:
I changed the tests. Tomorrow will check if they pass.

kanes115:
One test not passing sm_SUITE:resend_unacked_on_reconection.

test.disabled/ejabberd_tests/tests/vcard_simple_SUITE.erl
```diff
@@ -136,9 +136,10 @@ retrieve_own_card(Config) ->
 
 
 
-%% If no vCard exists or the user does not exist, the server MUST
-%% return a stanza error, which SHOULD be either
-%% <service-unavailable/> or <item-not-found/>
+%% If no vCard exists, the server MUST return a stanza error
+%% %% (which SHOULD be <item-not-found/>) or an IQ-result
```
fenek: Double `%%` :)

PR #1548, opened 2017-10-30
Change passwordlike to passterm

kanes115:
The name `passwordlike` smells. As it can't be `password` because it is either binary (word) OR TUPLE of type scram:scram_tuple() we will call it `passTERM`. 

It plays a role of passWORD but instead of being a **WORD** it is a **WORD or TUPLE**. The smallest denominator then is **TERM**. This way we also let somebody extend it for any type of pass phrase he/she needs.


kanes115:
OTP 20 build is considered not passed but there is not a single test that failed. Why? (ping: @fenek )

fenek:
@kanes115 This one failed: https://travis-ci.org/esl/MongooseIM/jobs/295211289 :)

kanes115:
@fenek , I know which job failed - this is inside the job that I don't find any failed test :(

fenek:
Ouch. :D Well, it's restarted already. Sometimes the job may fail if Cover exits with a segfault (very rare but does happen). In such case you won't see any failed test cases.

PR #1553, opened 2017-11-02
Remove unnecessary TODOs

kanes115:
This PR removes unnecessary TODOs from code

### DIFF
--- a/src/auth/ejabberd_auth_external.erl
+++ b/src/auth/ejabberd_auth_external.erl
@@ -365,7 +365,6 @@ is_fresh_enough(TimeStampLast, CacheTime) ->
 
 %% @doc Code copied from mod_configure.erl
 %% Code copied from web/ejabberd_web_admin.erl
-%% TODO: Update time format to XEP-0202: Entity Time
 -spec get_last_access(User :: jid:user(),
                       Server :: jid:server()
                       ) -> online | never | mod_last_required | integer().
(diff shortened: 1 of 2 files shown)


apps/ejabberd/src/gen_mod.erl
```diff
@@ -229,7 +229,6 @@ wait_for_stop1(MonitorReference) ->
 get_opt(Opt, Opts) ->
     case lists:keysearch(Opt, 1, Opts) of
         false ->
-            % TODO: replace with more appropriate function
```
kanes115: @ludwikbukowski, any comment on that?

ludwikbukowski: throwing exception here if fine for me.
Hence "TODO" can be removed

apps/ejabberd/src/ejabberd_auth_external.erl
```diff
@@ -366,7 +366,6 @@ is_fresh_enough(TimeStampLast, CacheTime) ->
 
 %% @doc Code copied from mod_configure.erl
 %% Code copied from web/ejabberd_web_admin.erl
-%% TODO: Update time format to XEP-0202: Entity Time
```
arcusfelis: so, why not to update? :D

michalwski: No need to, we don't return the timestamp to the users. It's used only internally so it's more efficient to operate on the raw value.

PR #1560, opened 2017-11-08
don't require auth before compress

sstrigler:
This PR addresses #

Proposed changes include:
* auth is not required before enabling stream compression, please check https://xmpp.org/extensions/xep-0138.html - clients like Psi won't work

### DIFF
--- a/apps/ejabberd/src/ejabberd_c2s.erl
+++ b/apps/ejabberd/src/ejabberd_c2s.erl
@@ -604,7 +604,7 @@ wait_for_feature_before_auth({xmlstreamelement, El}, StateData) ->
         {?NS_COMPRESS, <<"compress">>} when Zlib == true,
                                             ((SockMod == gen_tcp) or
                                              (SockMod == fast_tls)) ->
-          check_compression_auth(El, wait_for_feature_before_auth, StateData);
+          check_compression_method(El, wait_for_feature_before_auth, StateData);
         _ ->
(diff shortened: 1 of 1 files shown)


michalwski:
I remember we disabled compression before auth based on recommendation in XEP-0172: https://xmpp.org/extensions/xep-0170.html#c2s-compress

sstrigler:
*sigh* sounds legit, on the other hand clients don't seem to be aware of that. 

Also how does it make sense to restart an already authenticated stream? Don't I have to re-authenticate then? *confused*

sstrigler:
But thanks already for your super fast response! 🎉

sstrigler:
Guess it would be good to mention this in XEP-0138 then, because who reads all the other XEPs when implementing XEP-0138. No wonder when people start hating XMPP's mess.

michalwski:
> Also how does it make sense to restart an already authenticated stream? Don't I have to re-authenticate then? confused

sstrigler:
?

michalwski:
Check now, I accidentally sent the comment before finishing. Now it's updated.

sstrigler:
Just a last remark for the sake of completeness. Psi (http://psi-im.org) hangs if made to try with compression and zlib compression enabled at MIM. Neither does it authenticate before trying compression nor does it continue when it fails to do enabled compression. I'll report a bug there.

sstrigler:
https://github.com/psi-im/psi/issues/330

PR #1567, opened 2017-11-14
[skip ci] playing around with readmes

goddammit:
This PR addresses #

duplicate readmes


README.md
```diff
@@ -38,110 +52,46 @@ Check out our test results:
...
-    * [escalus](https://github.com/esl/escalus): Erlang XMPP client
-    * [amoc](https://github.com/esl/amoc): a load testing tools
-* More components? There are some ideas we're working on. Tune in for updates on
-    * MongooseICE: ICE, STUN/TURN server
-    * MongoosePush: a push notification server
+ 
+**MongooseIM documentation highligts:**
 
-### Client-side components
+When developing new features/modules, please make sure you add basic documentation to the 'doc/' directory, and add a link to your document in 'doc/README.md.'
```
fenek: There's no `doc/README.md` anymore. :)

fenek:
Unfortunately some changes have to be reverted (or not; see the last part of the comment), because:

1. `mkdocs` doesn't accept a top-level item that is both a link and has children. (e.g. crashes on `- 'User guide': 'user-guide/Get-to-know-MongooseIM.md'`)
2. If a page is not listed in `pages` section, it won't be included in the final documentation, so removing this huge module list results in modules docs not being built.

I'll try to find a way to work around these in `mkdocs` configuration.

PR #1570, opened 2017-11-16
Split header files

erszcz:
This PR splits some header files into more topic-specific ones: logging macros go to `mongoose_logger.hrl`, XMPP namespaces go to `mongoose_ns.hrl` (could we reuse the header from escalus/exml in the future?), some MAM specific records go to `mod_mam.hrl`.

It also makes a small step towards using only namespaced types throughout the codebase, by moving the definition of `rsm_in()` and `rsm_out()` from a header into `jlib.erl` and adding the module prefix in places where a global type was used before.

The motivation is that files which don't use parts defined in the big, generic header files don't include them anymore, which is nice when we want to reuse them outside the project.


michalwski:
One more thing. All jobs on travis failed, this is probably not a random failure :)

erszcz:
Not at all :) I'm working on it.

erszcz:
Seems to be ok now :)

alindfor:
Have moved all errors defined as macros to separate module 'mongoose_xmpp_errors'. Not sure about the some function names because some are very similar but unrelated, this is however a quick fix if it is not acceptable.

alindfor:
There are some macros that transform errors to binaries. These were often defined in multiple places but imo did not fit in the new `mongoose_xmpp_errors` module. They were moved to a new header file to reduce duplication instead (`ejabberd_s2s.hrl`).

apps/ejabberd/src/ejabberd_s2s.hrl
```diff
@@ -0,0 +1,23 @@
+%% These error macros will change the xmpp errors(xml) to binaries
+%% All errors are defined in mongoose_xmpp_errors.erl
+%% Previously they were defined in up to three different modules
+
+-define(INVALID_NS_ERR_TO_BIN,
```
fenek: Used only twice. We may get rid of the macro and replace it with `exml:to_binary(mongoose_xmpp_errors:invalid_namespace()))` in the code.

apps/ejabberd/src/ejabberd_s2s.hrl
```diff
@@ -0,0 +1,23 @@
...
+    exml:to_binary(mongoose_xmpp_errors:invalid_namespace())).
+
+-define(HOST_UNKNOWN_ERR_TO_BIN,
```
fenek: Used only once, no need for macro at all.

apps/ejabberd/src/ejabberd_s2s.hrl
```diff
@@ -0,0 +1,23 @@
...
+-define(INVALID_XML_ERR_TO_BIN,
+    exml:to_binary(mongoose_xmpp_errors:xml_not_well_formed())).
+
+-define(INVALID_FROM_ERR_TO_BIN,
+    exml:to_binary(mongoose_xmpp_errors:invalid_from())).
+
+-define(CONFLICT_ERR_TO_BIN,
+    exml:to_binary(mongoose_xmpp_errors:stream_conflict())).
+
+-define(SOCKET_DEFAULT_RESULT, {error, badarg}).
```
fenek: Used only in `s2s_out`, no need to keep it here.

apps/ejabberd/src/ejabberd_s2s.hrl
```diff
@@ -0,0 +1,23 @@
...
+-define(HOST_UNKNOWN_ERR_TO_BIN,
+    exml:to_binary(mongoose_xmpp_errors:host_unknown())).
+
+-define(INVALID_XML_ERR_TO_BIN,
+    exml:to_binary(mongoose_xmpp_errors:xml_not_well_formed())).
+
+-define(INVALID_FROM_ERR_TO_BIN,
+    exml:to_binary(mongoose_xmpp_errors:invalid_from())).
+
+-define(CONFLICT_ERR_TO_BIN,
```
fenek: Used only in `ejabberd_service`, no need for a macro.

apps/ejabberd/src/ejabberd_s2s.hrl
```diff
@@ -0,0 +1,23 @@
...
+-define(INVALID_NS_ERR_TO_BIN,
+    exml:to_binary(mongoose_xmpp_errors:invalid_namespace())).
+
+-define(HOST_UNKNOWN_ERR_TO_BIN,
+    exml:to_binary(mongoose_xmpp_errors:host_unknown())).
+
+-define(INVALID_XML_ERR_TO_BIN,
+    exml:to_binary(mongoose_xmpp_errors:xml_not_well_formed())).
+
+-define(INVALID_FROM_ERR_TO_BIN,
```
fenek: Used only once in `s2s_in`, no need for a macro.

apps/ejabberd/src/ejabberd_s2s.hrl
```diff
@@ -0,0 +1,23 @@
...
+%% All errors are defined in mongoose_xmpp_errors.erl
+%% Previously they were defined in up to three different modules
+
+-define(INVALID_NS_ERR_TO_BIN,
+    exml:to_binary(mongoose_xmpp_errors:invalid_namespace())).
+
+-define(HOST_UNKNOWN_ERR_TO_BIN,
+    exml:to_binary(mongoose_xmpp_errors:host_unknown())).
+
+-define(INVALID_XML_ERR_TO_BIN,
```
fenek: Could be a special, separate function in `mongoose_xmpp_errors`, e.g. `xml_not_well_formed_bin()`.

alindfor:
Rebase was rough but I think it is fine now, local tests are passing.

src/jlib.erl
```diff
@@ -63,24 +63,56 @@
...
 
--type xmlch() :: xmlel() | xmlcdata(). % (XML ch)ild
+-type xmlch() :: exml:element() | xmlcdata(). % (XML ch)ild
 
 -type binary_pair() :: {binary(), binary()}.
 
--export_type([xmlel/0, xmlstreamstart/0, xmlstreamend/0, xmlstreamel/0,
+-type iq() :: #iq{}.
+
+-type user()      :: binary().
```
fenek: It makes more sense to have all JID-related types defined in `jid.erl`.

include/mongoose.hrl
```diff
@@ -27,57 +27,13 @@
...
-
--record(route, {
-          domain :: binary(),
-          handler :: mongoose_packet_handler:t()
-         }).
-
--record(external_component, {domain, handler, node}).
+-include("mongoose_logger.hrl").
 
 -define(DEPRECATED,
```
fenek: This is a `mongoose_acc` leftover as well.

include/pubsub.hrl
```diff
@@ -136,13 +136,13 @@
     id                  ,% :: mod_pubsub:nodeIdx(),
     parents = []        ,% :: [mod_pubsub:nodeId(),...],
     type    = <<"flat">>,% :: binary(),
-    owners  = []        ,% :: [jlib:ljid(),...],
+    owners  = []        ,% :: [jlib::ljid(),...],
```
fenek: Why double `:`?

PR #1572, opened 2017-11-17
Refactor jlib, jid

SauronSauron:
Refactor jlib, jid to solve elvis warnings.

This PR addresses #

Proposed changes include:

### DIFF
--- a/src/jid.erl
+++ b/src/jid.erl
@@ -42,34 +42,25 @@
 -spec make(User :: ejabberd:user(), Server :: ejabberd:server(),
            Resource :: ejabberd:resource()) ->
     ejabberd:jid()  | error.
-
 make(User, Server, Resource) ->
-    case nodeprep(User) of
-        error -> error;
-        LUser ->
-            case nameprep(Server) of
-                error -> error;
-                LServer ->
-                    case resourceprep(Resource) of
-                        error -> error;
-                        LResource ->
-                            #jid{user = User,
-                                server = Server,
-                                resource = Resource,
-                                luser = LUser,
-                                lserver = LServer,
-                                lresource = LResource}
-                    end
-            end
-    end.
+    make_nodeprep(nodeprep(User), Server, Resource, {User, Server, Resource}).
 
-build_jid(User, Server, Resource, LUser, LServer, LResource) ->
-        #jid{user = User,
-            server = Server,
-            resource = Resource,
-            luser = LUser,
-            lserver = LServer,
-            lresource = LResource}.
+make_nodeprep(error, _, _, _) -> error;
+make_nodeprep(LUser,  Server, Resource, T) ->
+    make_nameprep(LUser, nameprep(Server), Resource, T).
+
+make_nameprep(_, error, _, _) -> error;
+make_nameprep(LUser,  LServer, Resource, T) ->
+    make_resourceprep(LUser, LServer, resourceprep(Resource), T).
+
+make_resourceprep(_, _, error, _) -> error;
+make_resourceprep(LUser,  LServer, LResource, {User, Server, Resource}) ->
+    #jid{user = User,
+        server = Server,
+        resource = Resource,
+        luser = LUser,
+        lserver = LServer,
+        lresource = LResource}.
 
 -spec make(ejabberd:simple_jid()) ->  ejabberd:jid()  | error.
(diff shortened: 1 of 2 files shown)


apps/ejabberd/src/jid.erl
```diff
@@ -42,25 +42,22 @@
...
-                            #jid{user = User,
-                                 server = Server,
-                                 resource = Resource,
-                                 luser = LUser,
-                                 lserver = LServer,
-                                 lresource = LResource}
-                    end
-            end
-    end.
+  case {nodeprep(User), nameprep(Server), resourceprep(Resource)} of
```
michalwski: The previous version is a bit more efficient when it comes to parsing wrong jids. Previously nameprep was run only if nodeprep passed. And similarly resourceprep. Error cases doesn't happen that offten so maybe may note can be ignored.

apps/ejabberd/src/jlib.erl
```diff
@@ -268,67 +268,69 @@ iq_query_info(El) ->
...
-                        sub_el = SubEl}
-            end;
-        Class == reply, Filter /= any ->
-            reply
-    end;
+  children = Els}, Filter) when Name == <<"iq">> ->
+  %% Filter is either request or any.  If it is request, any replies
+  %% are converted to the atom reply.
+  ID = xml:get_attr_s(<<"id">>, Attrs),
+  Type = xml:get_attr_s(<<"type">>, Attrs),
```
michalwski: In new and changed code we try to use 'exml_query' module to read data from XML elements instead of 'xml' module.

apps/ejabberd/src/jlib.erl
```diff
@@ -268,67 +268,69 @@ iq_query_info(El) ->
...
+  Lang = xml:get_attr_s(<<"xml:lang">>, Attrs),
+  {Type1, Class} = case Type of
```
michalwski: I think extracting this case to a function may improve readability of this function a bit.

apps/ejabberd/src/jlib.erl
```diff
@@ -268,67 +268,69 @@ iq_query_info(El) ->
...
+                     <<"set">> -> {set, request};
+                     <<"get">> -> {get, request};
+                     <<"result">> -> {result, reply};
+                     <<"error">> -> {error, reply};
+                     _ -> {invalid, invalid}
+                   end,
+  case {Type1, Class, Filter} of
+    {Type1, Class, Filter} when Type1 == invalid ->
```
michalwski: No need for the guard here. More readable would be to patter match 'invalid' in the case clause.

apps/ejabberd/src/jlib.erl
```diff
@@ -268,67 +268,69 @@ iq_query_info(El) ->
...
+      invalid;
+    {Type1, Class, Filter} when Type1 /= invalid, Class == request; Filter == any ->
```
michalwski: Again no need for guards especially if my comment above is addressed.

apps/ejabberd/src/jid.erl
fenek: I think that unless someone sends malformed JIDs on purpose, it is ver unlikely in normal operation to get errors here.

apps/ejabberd/src/jlib.erl
```diff
@@ -681,11 +683,11 @@ check_list(List) ->
...
               V = list_to_integer(L),
-              if
-                  (V >= 0) and (V =< N) ->
-                      {V, B};
-                  true ->
-                      {false, false}
+              case (V) of
+                (V) when V >= 0, V =< N ->
+                    {V, B};
+                (V) when V < 0, V > N ->
```
fenek: No need for a guard here. Also, no need to surround `V` with parentheses.

apps/ejabberd/src/jlib.erl
```diff
@@ -340,15 +345,15 @@ iq_type_to_binary(_) -> invalid.
 
 -spec iq_to_xml(ejabberd:iq()) -> xmlel().
 iq_to_xml(#iq{id = ID, type = Type, sub_el = SubEl}) ->
-    if
-        ID /= "" ->
+    case ID of
```
fenek: May be matched in function head.

PR #1573, opened 2017-11-17
mod_admin_extra: add get_presence command

fblackburn1:
Add get_presence to _mongooseimctl_ in module _mod_admin_extra_
This command has been extracted/based from _ejabberd_ source code

I don't know if you fell comfortable with this kind of PR (no tests and extracted from other project) but since we use it in our project, I submit it.


fenek:
Hi @fblackburn1 

Thank you, we appreciate every PR. It doesn't matter if it is extracted from another project, as long as it is not illegal. :)

A lack of tests on the other hand, is a kind of showstopper. We expect at least basic tests for every new code we merge (as long as it is testable, and this one is).

PR #1578, opened 2017-11-23
Configure Travis to connect to Cassandra over TLS

DenysGonchar:
Configure Travis to connect to Cassandra over TLS


tools/travis-setup-db.sh
```diff
@@ -80,19 +76,39 @@ elif [ $DB = 'riak' ]; then
...
+
+    opts="$(docker inspect -f '{{range .Config.Entrypoint}}{{println}}{{.}}{{end}}' cassandra:${CASSANDRA_VERSION})"
+    opts+="$(docker inspect -f '{{range .Config.Cmd}}{{println}}{{.}}{{end}}' cassandra:${CASSANDRA_VERSION})"
+    readarray -t -s 1 init_opts <<< "$opts"
+    echo -e "cassandra startup cmd:\n\t${init_opts[@]}"
+
+    docker_entry="${DB_CONF_DIR}/docker_entry.sh"
+
+    docker run -d -p 9042:9042                   \
+               -e MAX_HEAP_SIZE=128M             \
```
arkgil: Is there a need to set those parameters? If yes, why these particular values? I mean `MAX_HEAP_SIZE` and `HEAP_NEWSIZE`.

DenysGonchar: TBH i have no idea, but these 2 were set before so i left them as is

DenysGonchar:
few comments for the history:
### [rel/files/ejabberd.cfg](https://github.com/esl/MongooseIM/pull/1578/files#diff-1340537c9d80373207c7cf10db06b690)
use the triple [mustache](http://mustache.github.io/mustache.5.html) to prevent HTML escaping of the quotes for the filename (damn, it took me almost one day to figure out why travis test works with **verify_none** ssl option and doesn't work with **verify_peer**)

###[tools/db_configs/cassandra/docker_entry.sh](https://github.com/esl/MongooseIM/pull/1578/files#diff-441ce347718d7dd6ec472687e3b550f3)
cassandra docker image is quite user-unfriendly, here is small description of the way how it starts up:
1. [docker-entry.sh](https://github.com/docker-library/cassandra/blob/master/docker-entrypoint.sh) scripts modifies _/etc/cassandra/cassandra.yaml_ configuration file
2. then it executes default cassandra's cmd(**cassandra -f**)

the most important modification in **cassandra.yaml** file is change of **listen_address** value to docker container's IP (which is dynamic). initial value of **listen_address** is **localhost** and according to the comments in  **cassandra.yaml** it's incorrect to set it to **0.0.0.0**:
`# Setting listen_address to 0.0.0.0 is always wrong.`

the main problem here is that **docker-entry.sh** uses **sed -i** to update **cassandra.yaml**. passing **in-place** parameter to sed leads to the next actions:
1. sed generates the new temporary file with changes
2. and replaces old file with the new one

such behaviour makes it impossible to use dockers **-v** option to mount custom **cassandra.yaml** 

also it is senseless to use **-e JVM_EXTRA_OPTS=-Dcassandra.config=/custom_cassandra.yaml** environment variable, because path to **cassandra.yaml** is hardcoded in **doker_entry.sh** and we still need to change **listen_address** runtime. 

so the correct solution is the next:
1. provide custom **cassandra.yaml** with **-v** option
2. provide custom **--entrypoint** docker script and copy custom **cassandra.yaml**  to default config location
3. run default **--entrypoint** script with default parameters (default cmd) to let it change **listen_address** value runtime.

but there is one trick we can use to simplify the solution a bit, we can just append some new key-value definitions to the available **/etc/cassandra/cassandra.yaml** file. having key duplicates in **cassandra.yaml** makes it incorrect [yaml dockument](http://www.yaml.org/spec/1.2/spec.html#id2759572) but it is still ok for cassandra's yaml loader, it just picks the latest provided value.

also, cassandra requires SSL certificate in Java Key Store (JKS) format. generation can be done using **keytool** utility (part of Java packet) and to avoid new build dependancy for MongooseIM, generation of JKS certificate is done inside the cassandra's container

PR #1582, opened 2017-11-27
Fix failing mod_mam_muc when only MUC mam enabled

kanes115:
We now check wether the message is of type "groupchat" as we don't
support chat markers for these. Before this commit it wasn't checked
and the module mod_mam_params wasn't even compiled (it is compiled in
mod_mam and we enabled only mod_mam_muc) which led to crash.

### DIFF
--- a/test.disabled/ejabberd_tests/tests/mam_SUITE.erl
+++ b/test.disabled/ejabberd_tests/tests/mam_SUITE.erl
@@ -636,18 +636,21 @@ init_modules(BackendType, muc_light, Config) ->
 init_modules(cassandra, muc_all, Config) ->
     init_module(host(), mod_mam_muc_cassandra_arch, []),
     init_module(host(), mod_mam_muc, [{host, muc_domain(Config)}, add_archived_element]),
+    true = rpc_apply(code, delete, [mod_mam_params]);
     Config;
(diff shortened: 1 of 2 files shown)


test.disabled/ejabberd_tests/tests/mam_SUITE.erl
```diff
@@ -636,38 +636,44 @@ init_modules(BackendType, muc_light, Config) ->
 init_modules(cassandra, muc_all, Config) ->
     init_module(host(), mod_mam_muc_cassandra_arch, []),
     init_module(host(), mod_mam_muc, [{host, muc_domain(Config)}, add_archived_element]),
+    true = rpc_apply(code, delete, [mod_mam_params]);
```
fenek: If you're happy and you know it, syntax error.
If you're happy and you know it, syntax error.
If you're happy and you know it and you really want to show it, if you're happy and you know it, syntax error.

kanes115: WIP label, @fenek!

fenek: Oh come on, why so serious? :)

kanes115:
Moved to #1591

PR #1591, opened 2017-11-30
Decouple params modules

kanes115:
This PR fixes a bux introduced by #1381

It coupled mam_iq and mod_mam_utils to tightly with mod_mam_params and mod_mam_muc_params and sometimes it called functions from the wrong module (it could even not be compiled). 

Consider this example call sequence:
mod_mam_muc:handle_set_message_form/4 --uses--> mam_iq:form_to_lookup_params/1 --uses--> mam_iq:common_lookup_params/1 --uses--> mod_mam_params:max_result_limit/0

This lead to a situation in which mod_mam_muc calls mod_mam_params instead of mod_mam_muc_params, so when mod_mam_muc is not enabled in a configuration this will lead to crash, if it is -> to an unexpected behaviour.

The solution is simple and temporary: we inject the values fetched from mod_mam*params into mam_iq and mod_mam_utils.
In the longer term: we will need to refactor these compiled in run-time modules.

To be done:
* regression tests


arkgil:
@fenek for now I have opted for purging params modules when stopping `mod_mam` and `mod_mam_muc`. Since all MAM related modules are stopped after each group in `mod_mam_SUITE`, we should be safe to assume that `mod_mam_params` is not loaded when MUC-related cases are run, and vice versa, `mod_mam_muc_params` is not loaded when PM-related cases are run.

But I have noticed one problem with params modules in general - there is currently no way to have different params modules for different domains (and I believe it's possible to start MAM for each configured domain). It means that functions from `mod_mam_params` return values configured for whichever domain's MAM was started last. And with the changes introduced in 4da66bd, stopping MAM for one domain will purge params module for all of them.

PR #1593, opened 2017-12-04
Ejabberd auth

alindfor:
This PR addresses #

Proposed changes include:

### DIFF
--- a/src/ejabberd_auth.erl
+++ b/src/ejabberd_auth.erl
@@ -30,8 +30,16 @@
 -export([start/0,
          start/1,
          stop/1,
-         set_generic_opt/3,
+         set_generic_opts/3,
+         get_generic_opt/2,
          get_generic_opt/3,
+         get_jwt_username_key/1,
+         get_jwt_algorithm/1,
+         get_jwt_secret/1,
+         get_host/1,
+         get_extauth_program/1,
+         get_path_prefix/1,
+         get_basic_auth_header/1,
          authorize/1,
          set_password/3,
          check_password/3,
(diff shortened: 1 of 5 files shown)


src/ejabberd_auth.erl
```diff
@@ -30,6 +30,16 @@
 -export([start/0,
          start/1,
          stop/1,
+         set_generic_opts/3,
+         get_generic_opt/2,
+         get_generic_opt/3,
+         get_jwt_username_key/1,
```
fenek: Lines 36-42 are too specific, they shouldn't be provided as API here. 33-35 should be sufficient for `ejabberd_auth_*` modules. `ejabberd_auth` shouldn't be aware about specific possible config keys inside auth_opts.

src/ejabberd_auth.erl
```diff
@@ -101,6 +111,78 @@ plain_password_required(Server) ->
               M:plain_password_required()
       end, auth_modules(Server)).
 
+-spec set_generic_opts(Host :: ejabberd:server(),
+                      Opt :: atom(),
+                      KVs :: [tuple()]) ->  {atomic|aborted, _}.
+set_generic_opts(Host, Opt, KVs) ->
+    OldOpts = ejabberd_config:get_local_option(Opt, Host),
```
fenek: `set_generic_opts` should access `auth_opts` by default, so the API here should be `set_opt(Host, Opt, Val)` (including the function name change).

src/ejabberd_auth.erl
```diff
@@ -101,6 +111,78 @@ plain_password_required(Server) ->
...
+    AccFunc = fun({K, V}, Acc) ->
+                  lists:keystore(K, 1, Acc, {K, V})
+              end,
+    NewOpts = lists:foldl(AccFunc, OldOpts, KVs),
+    ejabberd_config:add_local_option({Opt, Host}, NewOpts).
+
+-spec get_generic_opt(Host :: ejabberd:server(),
+                      Opt :: atom(),
+                      Key :: ejabberd:key()) -> undefined | ejabberd:value().
+get_generic_opt(Host, Opt) ->
```
fenek: Is it that useful to fetch all options? I would rather expect an API like `get_opt(Host, Opt)` and `get_opt(Host, Opt, Default)`.

src/ejabberd_auth_http.erl
```diff
@@ -42,8 +42,8 @@
 
 -spec start(binary()) -> ok.
 start(Host) ->
-    AuthOpts = ejabberd_config:get_local_option(auth_opts, Host),
-    {_, AuthHost} = lists:keyfind(host, 1, AuthOpts),
+    AuthOpts = ejabberd_auth:get_generic_opt(Host, auth_opts),
+    AuthHost = ejabberd_auth:get_host(Host),
     PoolSize = proplists:get_value(connection_pool_size, AuthOpts, 10),
```
fenek: These should be fetched with accessors from `ejabberd_auth` as well.

PR #1597, opened 2017-12-08
C2S crash fix

alindfor:
This PR addresses crash of ejabberd_c2s process when receiving iq with blocking or privacy namespace

Proposed changes include:
Check if IQ is of record iq, if not return old state

### DIFF
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -2352,19 +2352,24 @@ process_privacy_iq(Acc0, To, StateData) ->
     IQ = mongoose_acc:get(iq_query_info, Acc),
     Acc1 = mongoose_acc:put(iq, IQ, Acc),
     From = mongoose_acc:get(from_jid, Acc1),
-    #iq{type = Type, sub_el = SubEl} = IQ,
-    {Acc2, NewStateData} = process_privacy_iq(Acc1, Type, To, StateData),
-    Res = mongoose_acc:get(iq_result, Acc2, {error, ?ERR_FEATURE_NOT_IMPLEMENTED}),
-    IQRes = case Res of
-                {result, Result} ->
-                    IQ#iq{type = result, sub_el = Result};
-                {result, Result, _} ->
-                    IQ#iq{type = result, sub_el = Result};
-                {error, Error} ->
-                    IQ#iq{type = error, sub_el = [SubEl, Error]}
-            end,
-    Acc3 = ejabberd_router:route(To, From, Acc2, jlib:iq_to_xml(IQRes)),
-    {Acc3, NewStateData}.
+    case is_record(IQ, iq) of
+        true ->
+            #iq{type = Type, sub_el = SubEl} = IQ,
+            {Acc2, NewStateData} = process_privacy_iq(Acc1, Type, To, StateData),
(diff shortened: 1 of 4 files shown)


src/ejabberd_c2s.erl
```diff
@@ -2352,19 +2352,24 @@ process_privacy_iq(Acc0, To, StateData) ->
...
-                {result, Result} ->
-                    IQ#iq{type = result, sub_el = Result};
-                {result, Result, _} ->
-                    IQ#iq{type = result, sub_el = Result};
-                {error, Error} ->
-                    IQ#iq{type = error, sub_el = [SubEl, Error]}
-            end,
-    Acc3 = ejabberd_router:route(To, From, Acc2, jlib:iq_to_xml(IQRes)),
-    {Acc3, NewStateData}.
+    case is_record(IQ, iq) of
```
fenek: It would be more common to write:
```
case IQ of
    #iq{type = Type, sub_el = SubEl} = IQ ->
        {Acc2...
```
What is more, this `case` can even begin before line 2352.

src/ejabberd_c2s.erl
```diff
@@ -2349,22 +2349,25 @@ get_priority_from_presence(PresencePacket) ->
...
-                {result, Result} ->
-                    IQ#iq{type = result, sub_el = Result};
-                {result, Result, _} ->
-                    IQ#iq{type = result, sub_el = Result};
-                {error, Error} ->
-                    IQ#iq{type = error, sub_el = [SubEl, Error]}
-            end,
-    Acc3 = ejabberd_router:route(To, From, Acc2, jlib:iq_to_xml(IQRes)),
-    {Acc3, NewStateData}.
+    case mongoose_acc:get(iq_query_info, Acc) of
```
arcusfelis: can we use a separate function instead of case?

fenek: We could, it's a tradeoff between large case and adding yet another function to `ejabberd_c2s` (and coming up with a meaningful name).

alindfor: Either way works for me, the problem I have would be to come up with a meaningful function name.

test.disabled/ejabberd_tests/tests/mod_blocking_SUITE.erl
```diff
@@ -286,6 +287,38 @@ blocking_propagates_to_resources(Config) ->
             message_is_not_delivered(User2, [User1b], <<"miau!">>)
         end).
 
+send_iq_with_blocking_ns(Config) ->
+    %% Will send iq with blocking namespace with type result and error
+    escalus:story(Config, [{alice, 1}, {bob, 1}], fun(Alice, Bob) ->
+        QueryWithPrivacyNS = escalus_stanza:query_el(?NS_BLOCKING, []),
+        BobJid = escalus_utils:get_short_jid(Bob),
+        IQError = escalus_stanza:iq(BobJid,
```
fenek: Lines 295-306 may be extracted into parametrised helper in order to reduce code duplication. The same applies to `privacy_SUITE`.

alindfor:
I am unsure about the placement of the new helper function in privacy_helper.erl

fenek:
Merging because only one test failed - resend on unacked, our known random. Thanks!

PR #1606, opened 2017-12-14
Upgrade fast_tls to a version using two-level namespace on macOS.

kzemek:
This PR upgrades fast_tls to a version that eschews -flat_namespace linking flag that - on macOS High Sierra - caused wrong TLS library to be found during runtime, which resulted in segmentation faults.

There are quite a few changes between the version we used and the version this PR upgrades to, probably the most important being [a switch from port to NIF implementation](https://github.com/processone/fast_tls/commit/17bfcb769cb6573c53497e01957fffccd384cf6a), so we should look closely at load test results.


kzemek:
latest_to_pr-1606

The darker colors represent the load test results on the PR, the lighter were done on Latest docker. The start was rougher for the PR version but later there's no significant difference either way.

fenek:
I think these difference could be caused by Docker randomness. These are a tiny fraction of the traffic anyway and are sub-8ms, so looks good to me.

PR #1608, opened 2017-12-20
Register room race

alindfor:
This PR addresses race condition that occurs when registering room processes
The register_room function will now safely check if the room already exists.
A helper function is introduced to deal with the Pid related to the newly created room and will return the correct Pid needed by other parts of the module.
Tests meck the mod_muc_room:start and route functions in order to reproduce a race condition and to pass messages to the test cases.


test.disabled/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -4286,6 +4412,80 @@ parse_result_query(#xmlel{name = <<"query">>, children = Children}) ->
...
+        %% Check if the registered pid matches the fake pid
+        ?FAKEPID = Record#muc_online_room.pid.
+
+check_presence_route_to_offline_room(Config) ->
+    escalus:story(Config, [{alice, 1}], fun(Alice) ->
+        Room = <<"testroom3">>,
+        %% Send a presence to a nonexistent room
+        escalus:send(Alice, stanza_groupchat_enter_room_no_nick(Room)),
+        timer:sleep(?WAIT_TIME),
+        ?assertReceivedMatch(fakepid, 3000)
```
rslota: There is a macro for this value :)

test.disabled/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -4286,6 +4412,80 @@ parse_result_query(#xmlel{name = <<"query">>, children = Children}) ->
...
+        timer:sleep(?WAIT_TIME),
+        %% Check if the registered pid matches the fake pid
+        ?FAKEPID = Record#muc_online_room.pid.
+
+check_presence_route_to_offline_room(Config) ->
+    escalus:story(Config, [{alice, 1}], fun(Alice) ->
+        Room = <<"testroom3">>,
+        %% Send a presence to a nonexistent room
+        escalus:send(Alice, stanza_groupchat_enter_room_no_nick(Room)),
+        timer:sleep(?WAIT_TIME),
```
rslota: Why do we wait here, when we are waiting with `assertReceivedMatch` anyway?

test.disabled/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -4286,6 +4412,80 @@ parse_result_query(#xmlel{name = <<"query">>, children = Children}) ->
...
+    escalus:story(Config, [{alice, 1}], fun(Alice) ->
+        Room = <<"testroom4">>,
+        Host = muc_host(),
+        F = fun() ->
+                mnesia:write(#muc_room{name_host = {Room, Host}, opts = []})
+            end,
+        {atomic, ok} = escalus_ejabberd:rpc(mnesia, transaction, [F]),
+        %% Send a message to an offline permanent room
+        escalus:send(Alice, stanza_room_subject(Room, <<"Subject line">>)),
+        timer:sleep(?WAIT_TIME),
```
rslota: Same as above.

test.disabled/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -4286,6 +4412,80 @@ parse_result_query(#xmlel{name = <<"query">>, children = Children}) ->
...
+
+        %% Read the room
+        F = fun() ->
+                mnesia:read(muc_online_room, {Room, Host}, write)
+            end,
+        {_, [Record]} = escalus_ejabberd:rpc(mnesia, transaction, [F]),
+
+        timer:sleep(?WAIT_TIME),
+        %% Check if the registered pid matches the fake pid
+        ?FAKEPID = Record#muc_online_room.pid.
```
rslota: Since this is a test, let's use assert macros instead.

test.disabled/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -4286,6 +4412,80 @@ parse_result_query(#xmlel{name = <<"query">>, children = Children}) ->
...
+        [Alice | _] = ?config(escalus_users, Config),
+        start_room(Config, Alice, Room, <<"aliceroom">>, default),
+
+        %% Read the room
+        F = fun() ->
+                mnesia:read(muc_online_room, {Room, Host}, write)
+            end,
+        {_, [Record]} = escalus_ejabberd:rpc(mnesia, transaction, [F]),
+
+        timer:sleep(?WAIT_TIME),
```
rslota: This sleep seems redundant.

src/mod_muc.erl
```diff
@@ -510,20 +505,28 @@ process_packet(Acc, From, To, El, #state{
...
-            mod_muc_room:route(Pid, From, Nick, Acc, Packet),
-            ok
+            route_to_online_room(Pid, Routed)
     end.
 
+route_to_online_room(Pid, Routed) ->
+    ?DEBUG("MUC: send to process ~p~n", [Pid]),
+    {From, To, Acc, Packet} = Routed,
+    {_, _, Nick} = jid:to_lower(To),
+    mod_muc_room:route(Pid, From, Nick, Acc, Packet),
```
rslota: I know that this is how it was, but it doesn't seem right to ignore the return value of this function. Maybe we should at least match it, even if the spec says it's always 'ok'. What do you think about it?

test.disabled/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -67,23 +79,48 @@
         items :: [#xmlel{}]
         }).
 
+-define(assertReceivedMatch(Expect), ?assertReceivedMatch(Expect, timer:seconds(5))).
+
+-define(assertReceivedMatch(Expect, Timeout),
```
rslota: I'm pretty sure we already use this macro somewhere. Maybe it's time to move it to some header file in order not to repeat ourselves.

test.disabled/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -4286,6 +4412,80 @@ parse_result_query(#xmlel{name = <<"query">>, children = Children}) ->
...
+
+    %% Write a permanent room
+    F1 = fun() ->
+             mnesia:write(#muc_room{name_host = {Room, Host}, opts = []})
+         end,
+    {atomic, ok} = escalus_ejabberd:rpc(mnesia, transaction, [F1]),
+
+    % Load permanent rooms and let it sink in
+    escalus_ejabberd:rpc(mod_muc, load_permanent_rooms , [Host, ServerHost, Access, HistorySize, RoomShaper, HttpAuthPool]),
+    timer:sleep(?WAIT_TIME),
```
rslota: Do we really need this sleep?

test.disabled/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -4286,6 +4412,80 @@ parse_result_query(#xmlel{name = <<"query">>, children = Children}) ->
...
+    %% Read online room
+    F2 = fun()->
+             mnesia:read(muc_online_room, {Room, Host}, write)
+         end,
+    {_, [R]} = escalus_ejabberd:rpc(mnesia, transaction, [F2]),
+    timer:sleep(?WAIT_TIME),
```
rslota: This sleep seems redundant.

test.disabled/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -432,6 +504,47 @@ init_per_testcase(CaseName, ConfigIn) ->
     Config = maybe_create_unique_room(ConfigIn),
     escalus:init_per_testcase(CaseName, Config).
 
+%% Meck will register a fake room right before a 'real' room is started
+meck_room_start() ->
```
rslota: There is one issue with this approach. It's very implementation specific. For example: If some would move `mod_muc_room:start` call before the room check (which btw. would be totally fine), those tests stop testing anything. This is only a note, since I don't really think other ideas I have are objectively better (e.g. we could slow down `mnesia:write` significantly, to like 5 seconds, which would make two requests always create race condition). Just something to think about :)

test.disabled/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -4286,6 +4412,80 @@ parse_result_query(#xmlel{name = <<"query">>, children = Children}) ->
...
+    Host = <<"localhost">>,
+    ServerHost = <<"localhost">>,
+    Access = none,
+    HistorySize = 10,
+    RoomShaper = none,
+    HttpAuthPool = none,
+
+    %% Write a permanent room
+    F1 = fun() ->
+             mnesia:write(#muc_room{name_host = {Room, Host}, opts = []})
```
rslota: Is there a way of doing that without redeclaring the record? This way we end up with two definitions that have to be maintained in sync.

src/mod_muc.erl
```diff
@@ -510,20 +486,27 @@ process_packet(Acc, From, To, El, #state{
...
 
+route_to_online_room(Pid, Routed) ->
+    ?DEBUG("MUC: send to process ~p~n", [Pid]),
+    {From, To, Acc, Packet} = Routed,
+    {_, _, Nick} = jid:to_lower(To),
+    ok = mod_muc_room:route(Pid, From, Nick, Acc, Packet).
 
--spec route_to_nonexistent_room(room(), from_to_packet(), state()) -> 'ok'.
+-spec route_to_nonexistent_room(room(), from_to_packet(), state()) -> 'ok' | {_, pid()}.
 route_to_nonexistent_room(Room, {From, To, Acc, Packet}, State) ->
```
fenek: All `route_*_to_nonexistent_room` functions should be renamed, as now they are not routing anything to any room. Example: `get_registered_room_or_route_error`.

src/mod_muc.erl
```diff
@@ -510,20 +486,27 @@ process_packet(Acc, From, To, El, #state{
...
     route_by_nick(Nick, Routed, State);
-route_to_room(Room, {From, To, Acc, Packet} = Routed, #state{host=Host} = State) ->
+route_to_room(Room, Routed, #state{host=Host} = State) ->
     case mnesia:dirty_read(muc_online_room, {Room, Host}) of
         [] ->
-            route_to_nonexistent_room(Room, Routed, State);
+            case route_to_nonexistent_room(Room, Routed, State) of
+                {ok, NewPid} ->
+                    route_to_online_room(NewPid, Routed);
+                {exists, OldPid} ->
```
fenek: No need to distinguish between "new" or "old" PID here.

src/mod_muc.erl
```diff
@@ -561,7 +542,7 @@ route_presence_to_nonexistent_room(Room, From, To, Acc, Packet,
             ejabberd_router:route(To, From, Acc, Err)
```
fenek: Lack of proper return value after this line, that could be used by `route_to_room`.

src/mod_muc.erl
```diff
@@ -561,7 +542,7 @@ route_presence_to_nonexistent_room(Room, From, To, Acc, Packet,
...
     end.
 
-route_packet_to_nonexistent_room(Room, From, To, Acc, Packet,
+route_packet_to_nonexistent_room(Room, From, To, _Acc, Packet,
```
fenek: Lack of proper return value after line 556 (error routing), that could be used by `route_to_room`.

src/mod_muc.erl
```diff
@@ -752,13 +726,25 @@ start_new_room(Host, ServerHost, Access, Room,
                                RoomShaper, HttpAuthPool, Opts)
     end.
 
+handle_room_registration(Host, Room, Pid) ->
+    case register_room(Host, Room, Pid) of
+        {_, ok} ->
+            {ok, Pid};
+        {_, {exists, OldPid}} ->
+            exit(Pid, normal),
+            {exists, OldPid}
```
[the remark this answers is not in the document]
alindfor: The function will stop the newly started process associated with Pid, but the calling function will still need OldPid in order to route the message to the room, will it not? Would the message not be lost otherwise?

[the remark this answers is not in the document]
fenek: Yes, but `{ok, OldPid}` may be returned instead.

src/mod_muc.erl
```diff
@@ -489,35 +508,34 @@ route_to_room(<<>>, {_, To, _Acc, _} = Routed, State) ->
...
-            case route_to_nonexistent_room(Room, Routed, State) of
-                {ok, NewPid} ->
-                    route_to_online_room(NewPid, Routed);
-                {exists, OldPid} ->
-                    route_to_online_room(OldPid, Routed)
+            case get_registered_room_or_route_error(Room, Routed, State) of
+                {ok, Pid} ->
+                    route_to_online_room(Pid, Routed);
+                {route_error, _ErrText} ->
+                    ok
```
alindfor: Not sure if an error needs to be returned from here or 'ok' is enough

src/mod_muc_room.erl
```diff
@@ -188,6 +189,9 @@ start_link(Host, ServerHost, Access, Room, HistorySize, RoomShaper, HttpAuthPool
                         RoomShaper, HttpAuthPool, Opts],
                        ?FSMOPTS).
 
+stop(Pid) ->
+    gen_fsm:stop(Pid).
```
alindfor: Will this change be enough to terminate the process?

arcusfelis:
Fails on CI with timeout

http://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1706/4085/odbc_pgsql_mnesia.19.3/big/ct_run.test@travis-job-03eed274-fbe1-43dd-8513-ba794840da9a.2018-02-08_14.59.40/ejabberd_tests.tests.muc_SUITE.logs/run.2018-02-08_15.11.31/muc_suite.create_already_registered_room.html

test.disabled/ejabberd_tests/tests/muc_SUITE.erl
```diff
@@ -4286,6 +4377,70 @@ parse_result_query(#xmlel{name = <<"query">>, children = Children}) ->
...
+    Room = <<"testroom1">>,
+    Host = <<"localhost">>,
+    ServerHost = <<"localhost">>,
+    Access = none,
+    HistorySize = 10,
+    RoomShaper = none,
+    HttpAuthPool = none,
+
+    %% Write a permanent room
+    {atomic, ok} = escalus_ejabberd:rpc(mod_muc,store_room,[Host, Room, []]),
```
arcusfelis: code formatting

`rpc(jid, make, [Room....`
`load_permanent_rooms,`

PR #1612, opened 2017-12-22
Fix global distribution tests for Travis

kzemek:
This PR fixes timing issues of global distrib tests on very slow machines (i.e. Travis).

### DIFF
--- a/src/global_distrib/mod_global_distrib_receiver.erl
+++ b/src/global_distrib/mod_global_distrib_receiver.erl
@@ -186,10 +186,15 @@ start_listeners() ->
 -spec start_listener(mod_global_distrib_utils:endpoint()) -> any().
 start_listener({Addr, Port} = Ref) ->
     ?INFO_MSG("Starting listener on ~s:~b", [inet:ntoa(Addr), Port]),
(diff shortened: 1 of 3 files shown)


src/global_distrib/mod_global_distrib_receiver.erl
```diff
@@ -186,10 +186,15 @@ start_listeners() ->
...
     ?INFO_MSG("Starting listener on ~s:~b", [inet:ntoa(Addr), Port]),
-    {ok, _} = ranch:start_listener(Ref, 10, ranch_tcp, [{ip, Addr}, {port, Port}],
-                                   ?MODULE, []).
+    case ranch:start_listener(Ref, 10, ranch_tcp, [{ip, Addr}, {port, Port}], ?MODULE, []) of
+        {ok, _} -> ok;
+        {error, eaddrinuse} ->
+            ?ERROR_MSG("Failed to start listener on ~s:~b: address in use. Will retry in 1 second.",
+                       [inet:ntoa(Addr), Port]),
+            timer:sleep(1000),
+            start_listener(Ref)
```
fenek: I think we should have a limited number of attempts here.

Issue #1614, opened 2017-12-26
received stanza size metric is wildly wrong for largish xmpp stanzas


igors:
MongooseIM version: current master
Installed from: source
Erlang/OTP version: 20

The current code for received stanza size metric doesn't really measure stanza size at all; instead, it measures the size of TCP stream data chunks read in a single recv() op from a socket. These data chunks may contain from 0 to many XMPP stanzas, depending on stanza size. Experimentally, this seems to work fairly well for small (i.e., under MTU size) stanzas but fails miserably for stanzas that exceed MTU. At least on Linux, no values larger than MTU size are ever recorded in this metric.

Fixing this properly would probably require changing exml parser to return the byte sizes of the parsed messages and may or may not be worth it. However, it's fairly easy to modify the code to at least approximate large stanzas sizes by accumulating received data chunks sizes that don't result in any parsed XML. This doesn't seem to have much effect on how smaller stanza sizes are recorded so it sounds like an improvement over the current behavior.

fenek:
Fixed by PR merge.

PR #1615, opened 2017-12-26
Fix up received stanza size metric

igors:
This PR addresses #1614

Proposed changes include:

Use cumulative chunk size when reporting the size of received stanzas. Chunk size "accumulates" until exml parses out a full stanza.

This is still imperfect since a single chunk can potentially contain more than one stanza but it's a much more useful approximation than never reporting anything above MTU size.

### DIFF
--- a/src/ejabberd_receiver.erl
+++ b/src/ejabberd_receiver.erl
@@ -51,6 +51,7 @@
                 shaper_state,
                 c2s_pid,
                 max_stanza_size,
+                stanza_chunk_size,
                 parser,
                 timeout}).
 -type state() :: #state{}.
@@ -123,6 +124,7 @@ init([Socket, SockMod, Shaper, MaxStanzaSize]) ->
                 sock_mod = SockMod,
                 shaper_state = ShaperState,
                 max_stanza_size = MaxStanzaSize,
+                stanza_chunk_size = 0,
                 timeout = Timeout}}.
 
 %%--------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/ejabberd_receiver.erl
```diff
@@ -123,6 +124,7 @@ init([Socket, SockMod, Shaper, MaxStanzaSize]) ->
                 sock_mod = SockMod,
                 shaper_state = ShaperState,
                 max_stanza_size = MaxStanzaSize,
+                stanza_chunk_size = 0,
```
arcusfelis: it should be set to 0, each time we reset stream.

igors: Thanks! Let me know if c107deb is not what you had in mind.

igors:
Would you rather I return a new State from reset_parser? I've no problem changing that if it's preferred from your POV - just let me know if you'd like a new commit or a rebase.

fenek:
Actually it is a pretty good idea to have `reset_parser` accept only `State` as parameter and return new state. Please change it as a new commit to this branch. :)

igors:
Take a look at c19f52d ;)

fenek:
Thanks. :) Could you please rename `ParserState` to something like `StateAfterReset`? Currently the name suggests that this could be some new type of state.

igors:
Done

fenek:
Looks good now! Thank you. :)

PR #1616, opened 2017-12-27
[don't merge] Test if escalus with get_config/2 throwing errors doesn't mess up the tests

erszcz:
This PR tests if a change to Escalus public API, i.e. `escalus_config:get_config/2`, doesn't break the tests. The change is introduced in https://github.com/esl/escalus/pull/161.

Don't merge! This PR only intends to test an Escalus side branch. Once it's merged we can switch MIM to use the new Escalus.


erszcz:
Travis is green, we got what we wanted.

PR #1618, opened 2017-12-28
Http sns doc update

alindfor:
Proposed changes include:
Add information about keys in url-encoded form for HTTP notifications
Add information about JSON schema for Amazon SNS notifications

### DIFF
--- a/doc/modules/mod_event_pusher_http.md
+++ b/doc/modules/mod_event_pusher_http.md
@@ -44,6 +44,20 @@ This module uses a connection pool created by mongoose_http_client. It must be d
 
 Notifications will be POSTed to `http://localhost:8000/webservice/notifications`.
 
+## Payload format
+The HTTP event pusher sends a POST request with Content-Type `application/x-www-form-urlencoded`. The form has the following fields:
+* `author`: lower case username of the user who authored the message
+* `server`: . lower case name of the server from where the message originates
+* `receiver`: lower case username of the user who the message is for
+* `message`: message text
+
+### Example
+Below is an example of what the body of an HTTP POST request can look like:
+```bash
+"author=alice&server=localhost&receiver=bob&message=Hi, Bob!"
+```
+
+
 ## Metrics
 
 If you'd like to learn more about metrics in MongooseIM, please visit [MongooseIM metrics](../operation-and-maintenance/Mongoose-metrics.md** page.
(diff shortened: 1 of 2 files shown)


doc/modules/mod_event_pusher_http.md
```diff
@@ -44,6 +44,20 @@ This module uses a connection pool created by mongoose_http_client. It must be d
 
 Notifications will be POSTed to `http://localhost:8000/webservice/notifications`.
 
+## Payload format
+The HTTP event pusher sends a POST request with Content-Type `application/x-www-form-urlencoded`. The form has the following fields:
+* `author`: lower case username of the user who authored the message
```
fenek: These fields are not only converted to lower case but actually are processed by specific profiles of "stringprep", as per https://tools.ietf.org/html/rfc6122. You may mention lower case conversion as one of the consequences of this process of course.

doc/modules/mod_event_pusher_http.md
```diff
@@ -44,6 +44,20 @@ This module uses a connection pool created by mongoose_http_client. It must be d
...
+* `server`: . lower case name of the server from where the message originates
+* `receiver`: lower case username of the user who the message is for
+* `message`: message text
```
fenek: "Payload from `<body/>` element of the message." would be more precise.

doc/modules/mod_event_pusher_sns.md
```diff
@@ -56,4 +56,42 @@ Full topics for notifications (ARN as defined in [Amazon Resource Names][aws-arn
...
+#### Sent messages
+The JSON format for a private message notification is:
+```JSON
+{
+    "to_user_id": "bob0.61585@localhost",
+    "message": "Hello, Bob",
+    "from_user_id": "alice0.61585@localhost"
+}
+```
+The notification is similar for group messages except that the `to_user_id` is the recipient room jid. For example:
```
fenek: Room **JID** (not jid :)).

doc/modules/mod_event_pusher_sns.md
```diff
@@ -56,4 +56,42 @@ Full topics for notifications (ARN as defined in [Amazon Resource Names][aws-arn
...
+
+The JSON format for an online presence update notification is:
+```JSON
+{
+    "user_id": "alice0.61585@localhost",
+    "present": true
+}
+```
+
+For offline presence updates the `present` boolean value is set to false:
```
goddammit: For offline presence updates, the `present` boolean value is set to false ?

Issue #1620, opened 2018-01-03
mongooseim cannot be run from a directory with spaces in its name


igors:
MongooseIM version: current master
Installed from: source
Erlang/OTP version: 19

When trying to run from a directory with spaces in the name:
```
$ pwd
/tmp/dir with spaces

$ ./mongooseim/bin/mongooseim live
ls: /tmp/dir: No such file or directory
ls: spaces/mongooseim/lib/mongooseim-*/priv/lib: No such file or directory
ls: with: No such file or directory
./mongooseim/bin/mongooseim: line 25: cd: /tmp/dir: No such file or directory
egrep: /tmp/dir: No such file or directory
egrep: with: No such file or directory
egrep: spaces/mongooseim/etc/vm.args: No such file or directory
vm.args needs to have either -name or -sname parameter.
```

fenek:
Resolved by PR merge.

PR #1621, opened 2018-01-03
Allow spaces in install dir name

igors:
This PR addresses #1620

Proposed changes include:
* the bulk of the changes is just quoting bash variables correctly
* the only "interesting" change is to the `nodetool`, which is changed to call `epmd` through `open_port` directly since `os:cmd` tries to parse its parameter and, naturally, fails on spaces too.

This seems to be fully operational with OTP 20. Unfortunately, `nodetool` invocations are still broken on prior releases; this is due to escript itself breaking when invoked from a directory with spaces - I'm not sure how much can be done about that...

### DIFF
--- a/rel/files/erl
+++ b/rel/files/erl
@@ -10,19 +10,19 @@
 ## file available in $ROOTDIR/release/VSN.
 
 # Determine the abspath of where this script is executing from.
-ERTS_BIN_DIR=$(cd ${0%/*} && pwd)
+ERTS_BIN_DIR=$(cd "${0%/*}" && pwd)
 
 # Now determine the root directory -- this script runs from erts-VSN/bin,
 # so we simply need to strip off two dirs from the end of the ERTS_BIN_DIR
 # path.
-ROOTDIR=${ERTS_BIN_DIR%/*/*}
+ROOTDIR="${ERTS_BIN_DIR%/*/*}"
 
 # Parse out release and erts info
-START_ERL=`cat $ROOTDIR/releases/start_erl.data`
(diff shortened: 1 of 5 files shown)


fenek:
Thank you for this improvement. However, it cannot be merged until it works 100% properly with all OTP versions supported by MongooseIM (18.3-20). Would you be able to work on it in foreseeable future? :)

igors:
A couple of notes here:

* This patch doesn't make things any worse than they currently are on any of the OTP versions you mentioned.
* The only things that are still broken on OTP 18.3-19.x are `mongooseim` script  invocations that make use of `nodetool`; all of them are broken in the current master as well when run from a directory with spaces in the name. Just for the record, the following commands are affected:
** `mongooseim start`
** `mongooseim stop`
** `mongooseim restart`
** `mongooseim reboot`
** `mongooseim ping`
** `mongooseim attach`
* On the other hand, a few commands are fixed even on OTP 18.3-19, e.g., `mongooseim console|live|debug`
* The reason commands making use of `nodetool` are broken is not due to anything in the (patched) `nodetool` but to `escript`; `escript` _itself_ does not work when invoked from directories with spaces in the name in OTP 18.3-19; this has been fixed in OTP 20.

I can't think of any obvious workarounds for the `escript` invocation issue so let me know if you have any ideas for a potential fix.

This hopefully gives you enough information to decide if this makes this pull request worth merging - let me know if you need more details.

fenek:
OK, since it improves the behaviour for 20.x and doesn't make it worse for older versions then it's good enough for me. :)

PR #1622, opened 2018-01-03
Allow JIDs from all MUC light domains in HTTP room API

arkgil:
This PR allows passing room JID from all configured MUC light domains in HTTP room API routes. Previously the JID needed to be in a "primary" MUC light domain on host (started via `mod_muc_light:start/2`). If only username part of JID is given, still "primary" domain is assumed.

### DIFF
--- a/src/mongoose_client_api_rooms.erl
+++ b/src/mongoose_client_api_rooms.erl
@@ -41,7 +41,6 @@ allowed_methods(Req, State) ->
 
 resource_exists(Req, #{jid := #jid{lserver = Server}} = State) ->
     {RoomIDOrJID, Req2} = cowboy_req:binding(id, Req),
-    MUCLightDomain = muc_light_domain(Server),
     case RoomIDOrJID of
         undefined ->
             {Method, Req3} = cowboy_req:method(Req2),
@@ -53,8 +52,8 @@ resource_exists(Req, #{jid := #jid{lserver = Server}} = State) ->
             end;
         _ ->
             case validate_room_id(RoomIDOrJID, Server) of
-                {ok, RoomID} ->
-                    does_room_exist(RoomID, MUCLightDomain, Req2, State);
+                {ok, RoomID, RoomHost} ->
+                    does_room_exist(RoomID, RoomHost, Req2, State);
                 _ ->
                     bad_request(Req2, State)
             end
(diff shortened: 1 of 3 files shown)


src/mongoose_client_api_rooms.erl
```diff
@@ -150,15 +149,30 @@ determine_role(US, Users) ->
...
+              error ->
+                error
+            end;
         _ ->
             error
     end.
+
+-spec validate_room_host(binary()) -> ok | error.
+validate_room_host(RoomHost) ->
+    MucLightDomains = ejabberd_router:dirty_get_routes_to_module(mod_muc_light),
```
fenek: Feels a bit hackish. Instead of adding this API, can we reuse `mongoose_subhosts:get_host/1`?

src/mongoose_client_api_rooms.erl
```diff
@@ -150,15 +149,30 @@ determine_role(US, Users) ->
...
+    case lists:member(RoomHost, MucLightDomains) of
```
arcusfelis: this would not work well in case of a lot of domains.

src/mongoose_client_api_rooms.erl
arkgil: Still, we're not sure if that subhost is a MUC light domain. IMO router is a more appropriate place to look for that information, because it is exactly the place where that information is registered (I mean the host <-> module mapping).

src/mongoose_client_api_rooms.erl
arkgil: I agree, I need to come up with better solution.

arkgil:
@arcusfelis @fenek in hope to address concerns of both of you, I propose to replace `dirty_get_routes_to_module` with `dirty_check_route_exists/2` which would take a domain and a module as a parameter, and return `true` if a route to that domain handled by such module exists.

arcusfelis:
@arkgil dirty_check_route_exists/2 makes sense.

fenek:
@arkgil 

OK, sounds good. :) Also I think that `dirty_is_route_handler_module/2` would be more accurate. `dirty_get_route_handler_module/1` would be fine as well. :)

arkgil:
@fenek @arcusfelis I've added negative test cases and one test to ensure that rooms can be actually created at different domain. However I noticed that if we'd like to fully support running multiple MUC light services on a single host, couple of changes would need to be made, e.g.:
* drop the uniqueness constraint on room IDs and force uniqueness only in the scope of a single domain
* allow to start multiple instances of the same module attached to a single XMPP host served by Mongoose

Neustradamus:
What is the situation of this PR?

PR #1626, opened 2018-01-04
Eradicate relative paths from tests

arcusfelis:
Proposed changes include:
* Add REPO_ROOT_PATH variable into test.disabled/ejabberd_tests/Makefile

TODO
----

mod_global_distrib_SUITE has canonicalize_path in it, which does not fit there. - done
These type of functions should go to a separate helper.

Also, get rid of .. in includes - vcard_SUITE.
We can also declare env varialble for mongoose root, and use it instead of relative paths everywhere.


erszcz:
@arcusfelis Nice proposition! Travis fails, though. Can you check if it's a spurious failure or a problem with the changes? Only one job is red, so it seems to be the former.

fenek:
I think that exporting a repo root (and tests root?) variables is the best option here. With a dedicated helper module to access them, relative paths will get contained as much as possible (actually limited to the Makefile).

BTW: Please don't forget that small tests use relative paths in more than one place too. :) It may be hidden in unusual way, e.g. https://github.com/esl/MongooseIM/blob/master/test/ejabberd_helper.erl#L53

PR #1627, opened 2018-01-08
Use gen_mod opts in MAM

kzemek:
This PR removes dynamically compiled `mod_mam_prefs` and `mod_mam_muc_prefs` modules in favor of fetching module options with ETS via `gen_mod:get_module_opts()` and then searching for the right option with `lists:keyfind`.

### DIFF
--- a/src/gen_mod.erl
+++ b/src/gen_mod.erl
@@ -289,33 +289,27 @@ get_module_opt_by_subhost(SubHost, Module, Opt, Default) ->
     {ok, Host} = mongoose_subhosts:get_host(SubHost),
     get_module_opt(Host, Module, Opt, Default).
 
+
 %% @doc Non-atomic! You have been warned.
 -spec set_module_opt(ejabberd:server(), module(), _Opt, _Value) -> boolean().
 set_module_opt(Host, Module, Opt, Value) ->
-    Key = {Module, Host},
-    OptsList = ets:lookup(ejabberd_modules, Key),
-    case OptsList of
+    case ets:lookup(ejabberd_modules, {Module, Host}) of
         [] ->
             false;
         [#ejabberd_module{opts = Opts}] ->
             Updated = set_opt(Opt, Opts, Value),
-            ets:update_element(ejabberd_modules, Key,
-                               {#ejabberd_module.opts, Updated})
+            set_module_opts(Host, Module, Updated)
     end.
 
-%% @doc Non-atomic! You have been warned.
--spec set_module_opts(ejabberd:server(), module(), _Opts) -> boolean().
+
+%% @doc Replaces all module options
+-spec set_module_opts(ejabberd:server(), module(), _Opts) -> true.
 set_module_opts(Host, Module, Opts0) ->
     Opts = proplists:unfold(Opts0),
-    Key = {Module, Host},
-    OptsList = ets:lookup(ejabberd_modules, Key),
-    case OptsList of
-        [] ->
-            false;
-        [#ejabberd_module{opts = _Opts}] ->
-            ets:update_element(ejabberd_modules, Key,
-                               {#ejabberd_module.opts, Opts})
-    end.
+    ets:insert(ejabberd_modules,
+               #ejabberd_module{module_host = {Module, Host},
+                                opts = Opts}).
+
 
 -spec set_module_opt_by_subhost(
         SubHost :: ejabberd:server(),
--- a/src/gen_mod.erl
+++ b/src/gen_mod.erl
@@ -52,6 +52,7 @@
          set_opt/3,
          get_module_opt/4,
          set_module_opt/4,
+         set_module_opts/3,
          get_module_opts/2,
          get_opt_subhost/3,
          get_opt_subhost/4,
@@ -302,6 +303,20 @@ set_module_opt(Host, Module, Opt, Value) ->
                                {#ejabberd_module.opts, Updated})
     end.
 
+%% @doc Non-atomic! You have been warned.
+-spec set_module_opts(ejabberd:server(), module(), _Opts) -> boolean().
+set_module_opts(Host, Module, Opts0) ->
+    Opts = proplists:unfold(Opts0),
+    Key = {Module, Host},
+    OptsList = ets:lookup(ejabberd_modules, Key),
+    case OptsList of
+        [] ->
+            false;
+        [#ejabberd_module{opts = _Opts}] ->
+            ets:update_element(ejabberd_modules, Key,
+                               {#ejabberd_module.opts, Opts})
+    end.
+
 -spec set_module_opt_by_subhost(
         SubHost :: ejabberd:server(),
         Module :: module(),
@@ -345,7 +360,8 @@ loaded_modules_with_opts(Host) ->
 
 -spec set_module_opts_mnesia(ejabberd:server(), module(), [any()]
                             ) -> {'aborted', _} | {'atomic', _}.
-set_module_opts_mnesia(Host, Module, Opts) ->
+set_module_opts_mnesia(Host, Module, Opts0) ->
+    Opts = proplists:unfold(Opts0),
     Modules = case ejabberd_config:get_local_option({modules, Host}) of
         undefined ->
             [];
--- a/src/mod_mam_muc.erl
+++ b/src/mod_mam_muc.erl
@@ -54,7 +54,6 @@
 -export([archive_message/8]).
 -export([lookup_messages/2]).
 -export([archive_id_int/2]).
--export([set_params/1]).
 -export([handle_set_message_form/3]).
 -export([handle_lookup_result/4]).
 -export([send_messages_and_iq_result/4]).
@@ -164,7 +163,6 @@ start(Host, Opts) ->
                         " It is not recommended to use it."
                                              " Consider using a <stanza-id/> element instead")
     end,
-    compile_params_module(Opts),
     %% MUC host.
     MUCHost = gen_mod:get_opt_subhost(Host, Opts, mod_muc:default_host()),
     IQDisc = gen_mod:get_opt(iqdisc, Opts, parallel), %% Type
@@ -208,8 +206,9 @@ stop(Host) ->
 %% @doc Handle public MUC-message.
 -spec filter_room_packet(Packet :: packet(), EventData :: list()) -> packet().
 filter_room_packet(Packet, EventData) ->
+    {room_jid, #jid{lserver = LServer}} = lists:keyfind(room_jid, 1, EventData),
     ?DEBUG("Incoming room packet.", []),
-    IsArchivable = mod_mam_muc_params:is_archivable_message(?MODULE, incoming, Packet),
+    IsArchivable = is_archivable_message(LServer, incoming, Packet),
     case IsArchivable of
         true ->
             {_, FromNick} = lists:keyfind(from_nick, 1, EventData),
@@ -249,8 +248,8 @@ archive_room_packet(Packet, FromNick, FromJID=#jid{}, RoomJID=#jid{}, Role, Affi
                     maybe_add_arcid_elems(RoomJID,
                                           mess_id_to_external_binary(MessID),
                                              Packet,
-                                             add_archived_element(),
-                                             add_stanzaid_element());
+                                             add_archived_element(Host),
+                                             add_stanzaid_element(Host));
                 {error, _} -> Packet
             end;
         false -> Packet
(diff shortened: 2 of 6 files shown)


kzemek:
#### Before
screen shot 2018-01-15 at 14 56 15

#### After
screen shot 2018-01-15 at 14 56 32

kzemek:
Closeup on the TTD graph after the main (~2s) peak.

#### Before
screen shot 2018-01-15 at 14 57 57

#### After
screen shot 2018-01-15 at 14 58 03

kzemek:
Closeup on the TTD graph after the second (~80ms) peak.

#### Before
screen shot 2018-01-15 at 15 04 52

#### After
screen shot 2018-01-15 at 15 05 31

src/gen_mod.erl
```diff
@@ -302,6 +303,20 @@ set_module_opt(Host, Module, Opt, Value) ->
                                {#ejabberd_module.opts, Updated})
     end.
 
+%% @doc Non-atomic! You have been warned.
+-spec set_module_opts(ejabberd:server(), module(), _Opts) -> boolean().
```
arcusfelis: It's a new function.
Can we have a test that checks that the function works correctly?
Description is nice to have too, i.e. how it works, what is Opts and what is the result.

arcusfelis: Does it replace all module opts or appends opts? - it's what I want to know from description

test.disabled/ejabberd_tests/tests/mam_SUITE.erl
```diff
@@ -813,8 +815,7 @@ init_state(C, muc_light, Config) ->
 init_state(_C, prefs_cases, Config) ->
     Config;
 init_state(_C, policy_violation, Config) ->
-    rpc_apply(mod_mam, set_params,
-              [ [{max_result_limit, 5}] ]),
+    rpc_apply(gen_mod, set_module_opt, [host(), mod_mam, max_result_limit, 5]),
```
arcusfelis: can we put this call into a helper module?

src/mod_mam_utils.erl
```diff
@@ -1062,3 +1066,30 @@ error_on_sql_error(_HostOrConn, _Query, Result) ->
 -spec wrapper_id() -> binary().
 wrapper_id() ->
     uuid:uuid_to_string(uuid:get_v4(), binary_standard).
+
+param(Module, Host, Opt, Default) ->
```
arcusfelis: specs for each exported function.

src/mod_mam_utils.erl
```diff
@@ -1062,3 +1066,30 @@ error_on_sql_error(_HostOrConn, _Query, Result) ->
...
+    ArchiveChatMarkers = param(Module, Host, archive_chat_markers, false),
+    erlang:apply(IsArchivableModule, IsArchivableFunction,
+                 [Module, Dir, Packet, ArchiveChatMarkers]).
+
+%% @doc Enable support for `<archived/>' element from MAM v0.2
+%% -spec add_archived_element(Host :: ejabberd:lserver()) -> boolean().
+add_archived_element(Module, Host) ->
+    param(Module, Host, add_archived_element, false).
+
+%% -spec add_stanzaid_element(Host :: ejabberd:lserver()) -> boolean().
```
arcusfelis: spec commented? why?

src/mod_mam_muc.erl
```diff
@@ -770,53 +769,14 @@ report_issue(Reason, Stacktrace, Issue, #jid{lserver = LServer, luser = LUser},
...
+
+is_archivable_message(Host, Dir, Packet) ->
+    mod_mam_utils:call_is_archivable_message(?MODULE, Host, Dir, Packet).
+
+add_archived_element(Host) ->
+    mod_mam_utils:add_archived_element(?MODULE, Host).
 
-    Format =
-        io_lib:format(
-          "-module(mod_mam_muc_params).~n"
```
arcusfelis: I like the param module, because I can get all MAM related settings without thinking for default value of each option.
Can we have a function, that returns all possible MAM parameters with current values?

Something like

```erlang
mod_mam:get_parameters(Host) ->
[{add_archived_element, false}, {default_result_limit, 50}, ....]
```

Same for mod_mam_muc.

src/mod_mam_muc.erl
```diff
@@ -208,8 +206,9 @@ stop(Host) ->
 %% @doc Handle public MUC-message.
 -spec filter_room_packet(Packet :: packet(), EventData :: list()) -> packet().
 filter_room_packet(Packet, EventData) ->
+    {room_jid, #jid{lserver = LServer}} = lists:keyfind(room_jid, 1, EventData),
     ?DEBUG("Incoming room packet.", []),
-    IsArchivable = mod_mam_muc_params:is_archivable_message(?MODULE, incoming, Packet),
```
arcusfelis: I can't see now that we are using function that is configuration depended.
I would rather keep mod_mam_muc_params, by making a real mod_mam_muc_params module.
But of course, you can call gen_mod from inside this new module.

mod_mam_utils is supposed to have functions, that would always work the same predictable way, not depending on configuration parameters.

kzemek: > I can't see now that we are using function that is configuration depended.

If not religiously adhering to dependency injection by parameters, *when* can you see whether called function is or is not configuration dependent?

> mod_mam_utils is supposed to have functions, that would always work the same predictable way, not depending on configuration parameters.

Why? I see `mod_mam_utils` as the place for common utility functions of `mod_mam` and `mod_mam_muc`. These new functions are not even the first ones there to call `gen_mod:get_module_opt`.

src/mod_mam_muc.erl
kzemek: > I like the param module, because I can get all MAM related settings without thinking for default value of each option.

True, that's something to improve.

> Can we have a function, that returns all possible MAM parameters with current values?

What would be the point of such function?

test.disabled/ejabberd_tests/tests/mam_SUITE.erl
kzemek: I prefer this to be explicit since there's some magic going on with external option setting.

src/mod_mam_muc.erl
arcusfelis: utils - functions that do something.
params - functions that are allowed to override.
I don't like god-modules, we would end-up with another c2s or muc_room.

src/mod_mam_muc.erl
```diff
@@ -765,12 +763,17 @@ report_issue(Reason, Stacktrace, Issue, #jid{lserver = LServer, luser = LUser},
...
 
-
 %% ----------------------------------------------------------------------
-%% Dynamic params module
+%% Dynamic params
+
+extra_params_module(Host) ->
+    mod_mam_utils:param(?MODULE, Host, extra_lookup_params, undefined).
+
+max_result_limit(Host) ->
```
arcusfelis: can we add specs for calls like this?

src/mod_mam_muc.erl
kzemek: I'm hesitant to implement such functionality for MAM modules. If we want more explicit support for defaults, we should add it to gen_mod.

kzemek: Obviously sans defaults, the runtime config should just be fetched with `gen_mod:get_module_opts`

src/mod_mam_muc.erl
kzemek: I disagree that this is a way for mod_mam_utils to become a god module, but I agree that a separation into mod_mam_params would be beneficial for code clarity.

kzemek: Done

test.disabled/ejabberd_tests/tests/mam_SUITE.erl
```diff
@@ -603,11 +603,14 @@ end_per_group(G, Config) when G == rsm_all; G == mam_purge; G == nostore;
       Config;
 end_per_group(muc_configurable_archiveid, Config) ->
     ParamsB = proplists:get_value(params_backup, Config),
-    rpc_apply(mod_mam_muc, set_params, [ParamsB]),
+    lists:foreach(fun({Key, Value}) ->
```
arcusfelis: It screams to add gen_mod:set_module_opts/3 function.

test.disabled/ejabberd_tests/tests/mam_SUITE.erl
```diff
@@ -554,9 +554,9 @@ init_per_group(muc06, Config) ->
     [{props, mam06_props()}, {with_rsm, true}|Config];
 
 init_per_group(muc_configurable_archiveid, Config) ->
-    Config ++ [{params_backup, rpc_apply(mod_mam_muc_params, params, [])}];
+    Config ++ [{params_backup, rpc_apply(gen_mod, get_module_opts, [host(), mod_mam_muc])}];
 init_per_group(configurable_archiveid, Config) ->
-    [{params_backup, rpc_apply(mod_mam_params, params, [])} | Config];
+    [{params_backup, rpc_apply(gen_mod, get_module_opts, [host(), mod_mam])} | Config];
```
arcusfelis: the whole backup mech should probably replaces with save/restore params.

src/mam/mod_mam_params.erl
```diff
@@ -0,0 +1,78 @@
...
+max_result_limit(Module, Host) ->
+    param(Module, Host, max_result_limit, 50).
+
+-spec default_result_limit(mam_module(), Host :: ejabberd:lserver()) -> pos_integer().
+default_result_limit(Module, Host) ->
+    param(Module, Host, default_result_limit, 50).
+
+-spec has_full_text_search(Module :: mod_mam | mod_mam_muc, Host :: ejabberd:server()) -> boolean().
+has_full_text_search(Module, Host) ->
+    gen_mod:get_module_opt(Host, Module, full_text_search, true).
```
fenek: Why `param/4` is not reused here?

src/mam/mod_mam_muc.erl
```diff
@@ -766,57 +763,9 @@ report_issue(Reason, Stacktrace, Issue, #jid{lserver = LServer, luser = LUser},
...
--spec add_archived_element() -> boolean().
-add_archived_element() ->
-    mod_mam_muc_params:add_archived_element().
-
--spec add_stanzaid_element() -> boolean().
-add_stanzaid_element() ->
-    mod_mam_muc_params:add_stanzaid_element().
+-spec is_archivable_message(Host :: ejabberd:lserver(), Dir :: incoming | outgoing,
+                            Packet :: exml:element()) -> boolean().
+is_archivable_message(Host, Dir, Packet) ->
```
fenek: Maybe it should be extracted to `mod_mam_utils`? I believe it is a duplicate from `mod_mam`.

src/mam/mod_mam_muc.erl
```diff
@@ -766,57 +763,9 @@ report_issue(Reason, Stacktrace, Issue, #jid{lserver = LServer, luser = LUser},
...
+    {M, F} = mod_mam_params:is_archivable_message_fun(?MODULE, Host),
+    ArchiveChatMarkers = mod_mam_params:archive_chat_markers(?MODULE, Host),
+    erlang:apply(M, F, [?MODULE, Dir, Packet, ArchiveChatMarkers]).
```
fenek: This change implies updating docs:
https://github.com/esl/MongooseIM/blob/master/doc/modules/mod_mam.md#options
https://github.com/esl/MongooseIM/blob/master/doc/modules/mod_mam.md#is_archivable_message
https://github.com/esl/MongooseIM/blob/master/doc/modules/mod_mam.md#archiving-chat-markers

src/mam/mod_mam_muc.erl
kzemek: @arcusfelis wanted to have param-driven functions out of `mod_mam_utils` and I sort-of agree. Choosing parameters went to `mod_mam_params`, actual logic went to `mod_mam_utils` and a convenience shim connecting them went to mam modules. Duplication is a pity here, but it stems more from duplication of other parts of `mod_mam`/`mod_mam_muc`.

src/mam/mod_mam_muc.erl
kzemek: How so? No logic has changed

src/mam/mod_mam_params.erl
kzemek: Omission

kzemek:
> Not all MAM modules have been moved to new folder. Please either move them or cancel the relocation in this PR.

@fenek done

src/mam/mod_mam_muc.erl
fenek: OK, I understand. May be a subject of refactoring in the future then, maybe there will be a dedicated module for param-driven MAM utils.

fenek:
@kzemek

Thanks! Only docs update remains. :)

fenek:
Apologies, docs became outdated before this PR!

PR #1633, opened 2018-01-11
Update recommended config values

alindfor:
This PR updates recommended configuration options for MySQL backend.

## Tested with v 5.7.20
Default for both variables is `on`.
### innodb_large_prefix
Default is **on**
Accepts `on/true` and `off/false`.
### innodb_file_per_table
Default is **on**
Accepts `on/off` and `true/false`

### innodb_large_prefix
Default is **off**,
Accepts `on/true` and `off/false`

### innodb_file_per_table
for v. <= 5.6.5, default is **off**
and v. >= 5.6.6, it is **on**
Accepts `on/true` and `off/false`

### DIFF
--- a/doc/advanced-configuration/database-backends-configuration.md
+++ b/doc/advanced-configuration/database-backends-configuration.md
@@ -84,7 +84,7 @@ innodb_file_format=BARRACUDA
 innodb_file_format_max=BARRACUDA
 innodb_file_per_table=true
 ```
-For versions `5.7.9` and later, all of the above options are set correctly by default.
+For versions `5.7.9` and newer, all of the above options are set correctly by default.
 ## PostgreSQL
 
 **Can be used for:**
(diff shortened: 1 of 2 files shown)


alindfor:
For the record, I noticed in the MySQL docs that the default value for `innodb_file_format_max` in versions <= 5.7.8 is Antelope. Should this be included in our docs maybe?

fenek:
@alindfor 

By all means, please update the upper version boundary to 5.7.8 then! (Don't forget to update SQL file as well)

doc/advanced-configuration/database-backends-configuration.md
```diff
@@ -76,14 +76,15 @@ Please refer to the [Advanced configuration/Database setup](../Advanced-configur
...
 
 ```bash
-innodb_large_prefix=ON
+innodb_large_prefix=true
 innodb_file_format=BARRACUDA
 innodb_file_format_max=BARRACUDA
+innodb_file_per_table=true
 ```
-
+For versions `5.7.9` and later, all of the above options are set correctly by default.
```
goddammit: and newer?

Issue #1640, opened 2018-01-14
max_stanza_size cannot be set when using websocket


igors:
MongooseIM version: master
Installed from: source
Erlang/OTP version: 20

c2s listeners allow limiting incoming XML stanza size with `max_stanza_size` config option. It'd be rather nice to have the same option when using cowboy with mod_websockets.

fenek:
Fixed by #1641 merge.

PR #1641, opened 2018-01-14
Websocket max stanza size

igors:
This PR addresses #1640 

Proposed changes include:
* Accept `max_stanza_size` as part of `mod_websockets` parameters.

### DIFF
--- a/src/mod_websockets.erl
+++ b/src/mod_websockets.erl
@@ -50,7 +50,8 @@
           open_tag :: stream | open | undefined,
           parser :: exml_stream:parser() | undefined,
           opts :: proplists:proplist() | undefined,
-          ping_rate :: integer() | none
+          ping_rate :: integer() | none,
+          max_stanza_size :: integer() | infinity
          }).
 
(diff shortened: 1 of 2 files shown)


src/mod_websockets.erl
```diff
@@ -307,6 +319,11 @@ replace_stream_ns(Element, #ws_state{ open_tag = open }) ->
 replace_stream_ns(Element, _State) ->
     Element.
 
+get_parser_opts(Text, #ws_state{ max_stanza_size = infinity } = State) ->
+    [{max_child_size, 0} | get_parser_opts(Text)];
```
arcusfelis: do we need to set it to 0?

igors: No, not really - I just wanted it to be a) explicit b) identical to ejabberd_receiver's code. Do you want me to add a commit that removes this?

fenek: I don't think it really matters, as it matches the default value assumed by `exml`.

PR #1653, opened 2018-01-15
Docker compose in docker

arcusfelis:
This PR includes:
- docker-compose inside docker
- printing docker-composing execution time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ addons:
         - slapd
         - ldap-utils
 before_install:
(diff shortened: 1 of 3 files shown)


michalwski:
@arcusfelis  What was the main motivation to use docker-compose in docker? Current travis images have docker-compose 1.17 installed. The docker compose in container is in version 14.
If the main motivation was to reduce time spent to install docker-compose, how about we use docker-compose available on travis instances?

PR #1660, opened 2018-01-16
Create pool on receiving

kanes115:
This PR adds hosts refresher which ensures in a given frequency if all hosts have a corresponding supervision tree built (mod_global_distrib_server_sup). It is run as a child of mod_global_distrib_outgoing_conns_sup.

It is introduced to not wait with establishing connection for the moment of sending an XMPP stanza to remote host.

### DIFF
--- a/src/global_distrib/mod_global_distrib_hosts_refresher.erl
+++ b/src/global_distrib/mod_global_distrib_hosts_refresher.erl
@@ -1,5 +1,5 @@
 %%%==============================================================================
-%% Copyright 2017 Erlang Solutions Ltd.
+%% Copyright 2018 Erlang Solutions Ltd.
 %%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
@@ -23,10 +23,10 @@
 -include("ejabberd.hrl").
 
 %% API
--export([start_link/0]).
+-export([start_link/1]).
 
 %% gen_server API
--export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/3]).
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2]).
 
 %% gen_mod API
 -export([deps/2, start/2, stop/1]).
@@ -35,10 +35,6 @@
 %%--------------------------------------------------------------------
 %% API
 %%--------------------------------------------------------------------
-
--spec start_link() -> pid() | term().
-start_link() ->
-    start_link(default_interval()).
 
 -spec start_link(Milliseconds :: non_neg_integer()) -> pid() | term().
 start_link(RefreshInterval) ->
@@ -67,8 +63,9 @@ deps(Host, Opts) ->
 %%--------------------------------------------------------------------
 
 init([RefreshInterval]) ->
-    self() ! refresh,
-    {ok, [{refresh_interval, RefreshInterval}]}.
+    ?DEBUG("refresher starting with interval ~p~n", [RefreshInterval]),
+    schedule_refresh(RefreshInterval),
+    {ok, RefreshInterval}.
 
 handle_call(_Request, _From, State) ->
     {noreply, State}.
@@ -76,16 +73,13 @@ handle_call(_Request, _From, State) ->
 handle_cast(_Request, State) ->
     {noreply, State}.
 
-handle_info(refresh, [{refresh_interval, Int}] = State) ->
-    ?DEBUG("Refreshing hosts, checking if there exists a supervisor for all of them~n", []),
-    Hosts = mod_global_distrib_mapping:hosts(),
-    ?DEBUG("Discovered hosts: ~p~n", [Hosts]),
-    lists:map(fun maybe_add/1, Hosts),
-    timer:send_after(Int, self(), refresh),
-    {noreply, State}.
+handle_info(refresh, Interval) ->
+    refresh(),
+    schedule_refresh(Interval),
+    {noreply, Interval}.
 
-terminate(Reason, _State, _Data) ->
-    ?DEBUG("REFRESHER has shutdown with reason ~p", [Reason]),
+terminate(Reason, _State) ->
+    ?ERROR_MSG("mod_global_distrib_refresher terminated with reason: ~p", [Reason]),
     ok.
 
 
@@ -93,19 +87,41 @@ terminate(Reason, _State, _Data) ->
 %% Helper functions
 %%--------------------------------------------------------------------
 
+
 -spec deps(Opts :: proplists:proplist()) -> gen_mod:deps_list().
 deps(Opts) ->
     [{mod_global_distrib_mapping, Opts, hard}].
 
 -spec start() -> any().
 start() ->
+    start_outgoing_conns_sup(),
+    Interval = mod_global_distrib_utils:opt(?MODULE, hosts_refresh_interval, default_interval()),
+    Child = #{
+      id => mod_global_distrib_hosts_refresher,
+      start => {?MODULE, start_link, [Interval]},
+      restart => transient,
+      shutdown => 5000,
+      modules => [mod_global_distrib_outgoing_conns_sup]
+    },
+    {ok, _} = supervisor:start_child(mod_global_distrib_outgoing_conns_sup, Child),
     ok.
 
 -spec stop() -> any().
 stop() ->
+    stop_outgoing_conns_sup(),
     ok.
 
-maybe_add(Host) ->
+refresh() ->
+    ?DEBUG("Refreshing hosts, checking if there exists a supervisor for all of them~n", []),
+    Hosts = mod_global_distrib_mapping:hosts(),
+    ?DEBUG("Discovered hosts: ~p~n", [Hosts]),
+    lists:map(fun maybe_add_host/1, Hosts),
+    ok.
+
+schedule_refresh(Interval) ->
+  erlang:send_after(Interval, self(), refresh).
+
+maybe_add_host(Host) ->
     ?DEBUG("Checking host ~p (on host ~p)~n", [Host, local_host()]),
     case local_host() of
         Host ->
@@ -118,5 +134,23 @@ default_interval() ->
     3000.
 
 local_host() ->
-    [{local_host, Host}] = ets:lookup(mod_global_distrib, local_host),
-    Host.
+    mod_global_distrib_utils:opt(?MODULE, local_host).
+
+start_outgoing_conns_sup() ->
+  ConnsSup = mod_global_distrib_outgoing_conns_sup,
+    ChildSpec = #{
+      id => ConnsSup,
+      start => {ConnsSup, start_link, []},
+      restart => permanent,
+      shutdown => 5000,
+      type => supervisor,
+      modules => [ConnsSup]
+     },
+  supervisor:start_child(ejabberd_sup, ChildSpec),
+  ok.
+
+
+stop_outgoing_conns_sup() ->
+  ConnsSup = mod_global_distrib_outgoing_conns_sup,
+  supervisor:terminate_child(ejabberd_sup, ConnsSup),
+  supervisor:delete_child(ejabberd_sup, ConnsSup).
--- a/src/global_distrib/mod_global_distrib_hosts_refresher.erl
+++ b/src/global_distrib/mod_global_distrib_hosts_refresher.erl
@@ -50,8 +50,8 @@ start_link(RefreshInterval) ->
 %%--------------------------------------------------------------------
 
 -spec start(Host :: ejabberd:lserver(), Opts :: proplists:proplist()) -> any().
-start(Host, Opts0) ->
-    Opts = [{message_ttl, 4} | Opts0],
+start(Host, Opts) ->
+    ?DEBUG("Opts in refresher: ~p~n", [Opts]),
     mod_global_distrib_utils:start(?MODULE, Host, Opts, fun start/0).
 
 -spec stop(Host :: ejabberd:lserver()) -> any().
@@ -106,8 +106,17 @@ stop() ->
     ok.
 
 maybe_add(Host) ->
-    ?DEBUG("Checking host ~p~n", [Host]),
-    mod_global_distrib_outgoing_conns_sup:ensure_server_started(Host).
+    ?DEBUG("Checking host ~p (on host ~p)~n", [Host, local_host()]),
+    case local_host() of
+        Host ->
+            ok;
+        _ ->
+            mod_global_distrib_outgoing_conns_sup:ensure_server_started(Host)
+    end.
 
 default_interval() ->
     3000.
+
+local_host() ->
+    [{local_host, Host}] = ets:lookup(mod_global_distrib, local_host),
+    Host.
--- a/src/global_distrib/mod_global_distrib_hosts_refresher.erl
+++ b/src/global_distrib/mod_global_distrib_hosts_refresher.erl
@@ -47,7 +47,7 @@ start_link(RefreshInterval) ->
 
 -spec start(Host :: ejabberd:lserver(), Opts :: proplists:proplist()) -> any().
 start(Host, Opts) ->
-    ?ERROR_MSG("Opts in refresher: ~p~n", [Opts]),
+    ?DEBUG("Opts in refresher: ~p~n", [Opts]),
     mod_global_distrib_utils:start(?MODULE, Host, Opts, fun start/0).
 
 -spec stop(Host :: ejabberd:lserver()) -> any().
@@ -63,7 +63,7 @@ deps(Host, Opts) ->
 %%--------------------------------------------------------------------
 
 init([RefreshInterval]) ->
-    ?ERROR_MSG("refresher starting with interval ~p~n", [RefreshInterval]),
+    ?DEBUG("refresher starting with interval ~p~n", [RefreshInterval]),
     schedule_refresh(RefreshInterval),
     {ok, RefreshInterval}.
 
@@ -112,9 +112,9 @@ stop() ->
     ok.
 
 refresh() ->
-    ?ERROR_MSG("Refreshing hosts, checking if there exists a supervisor for all of them~n", []),
+    ?DEBUG("Refreshing hosts, checking if there exists a supervisor for all of them~n", []),
     Hosts = mod_global_distrib_mapping:hosts(),
-    ?ERROR_MSG("Discovered hosts: ~p~n", [Hosts]),
+    ?DEBUG("Discovered hosts: ~p~n", [Hosts]),
     lists:map(fun maybe_add_host/1, Hosts),
     ok.
 
--- a/src/global_distrib/mod_global_distrib_hosts_refresher.erl
+++ b/src/global_distrib/mod_global_distrib_hosts_refresher.erl
@@ -20,7 +20,7 @@
 
 -behavior(gen_server).
 -behavior(gen_mod).
--include("ejabberd.hrl").
+-include("mongoose.hrl").
 
 %% API
 -export([start_link/1]).
@@ -47,7 +47,7 @@ start_link(RefreshInterval) ->
 
 -spec start(Host :: ejabberd:lserver(), Opts :: proplists:proplist()) -> any().
 start(Host, Opts) ->
-    ?DEBUG("Opts in refresher: ~p~n", [Opts]),
+    ?ERROR_MSG("Opts in refresher: ~p~n", [Opts]),
     mod_global_distrib_utils:start(?MODULE, Host, Opts, fun start/0).
 
 -spec stop(Host :: ejabberd:lserver()) -> any().
@@ -63,7 +63,7 @@ deps(Host, Opts) ->
 %%--------------------------------------------------------------------
 
 init([RefreshInterval]) ->
-    ?DEBUG("refresher starting with interval ~p~n", [RefreshInterval]),
+    ?ERROR_MSG("refresher starting with interval ~p~n", [RefreshInterval]),
     schedule_refresh(RefreshInterval),
     {ok, RefreshInterval}.
 
@@ -112,9 +112,9 @@ stop() ->
     ok.
 
 refresh() ->
-    ?DEBUG("Refreshing hosts, checking if there exists a supervisor for all of them~n", []),
+    ?ERROR_MSG("Refreshing hosts, checking if there exists a supervisor for all of them~n", []),
     Hosts = mod_global_distrib_mapping:hosts(),
-    ?DEBUG("Discovered hosts: ~p~n", [Hosts]),
+    ?ERROR_MSG("Discovered hosts: ~p~n", [Hosts]),
     lists:map(fun maybe_add_host/1, Hosts),
     ok.
 
--- a/src/global_distrib/mod_global_distrib_hosts_refresher.erl
+++ b/src/global_distrib/mod_global_distrib_hosts_refresher.erl
@@ -47,7 +47,6 @@ start_link(RefreshInterval) ->
 
 -spec start(Host :: ejabberd:lserver(), Opts :: proplists:proplist()) -> any().
 start(Host, Opts) ->
-    ?DEBUG("Opts in refresher: ~p~n", [Opts]),
     mod_global_distrib_utils:start(?MODULE, Host, Opts, fun start/0).
 
 -spec stop(Host :: ejabberd:lserver()) -> any().
@@ -67,10 +66,12 @@ init([RefreshInterval]) ->
     schedule_refresh(RefreshInterval),
     {ok, RefreshInterval}.
 
-handle_call(_Request, _From, State) ->
+handle_call(Request, From, State) ->
+    ?ERROR_MSG("issue=unknown_call request=~p from=~p", [Request, From]),
     {noreply, State}.
 
-handle_cast(_Request, State) ->
+handle_cast(Request, State) ->
+    ?ERROR_MSG("issue=unknown_cast request=~p", [Request]),
     {noreply, State}.
 
 handle_info(refresh, Interval) ->
@@ -78,9 +79,10 @@ handle_info(refresh, Interval) ->
     schedule_refresh(Interval),
     {noreply, Interval}.
 
-terminate(Reason, _State) ->
-    ?ERROR_MSG("mod_global_distrib_refresher terminated with reason: ~p", [Reason]),
-    ok.
+terminate(normal, _State) ->
+    ?INFO("mod_global_distrib_refresher has terminated with reason: normal", []);
+terminate(Reason, State) ->
+    ?ERROR_MSG("mod_global_distrib_refresher terminated with reason: ~p state=~p", [Reason, State]).
 
 
 %%--------------------------------------------------------------------
@@ -95,7 +97,7 @@ deps(Opts) ->
 -spec start() -> any().
 start() ->
     start_outgoing_conns_sup(),
-    Interval = mod_global_distrib_utils:opt(?MODULE, hosts_refresh_interval, default_interval()),
+    Interval = mod_global_distrib_utils:opt(?MODULE, hosts_refresh_interval, default_refresh_interval()),
     Child = #{
       id => mod_global_distrib_hosts_refresher,
       start => {?MODULE, start_link, [Interval]},
@@ -112,9 +114,8 @@ stop() ->
     ok.
 
 refresh() ->
-    ?DEBUG("Refreshing hosts, checking if there exists a supervisor for all of them~n", []),
+    ?DEBUG("Refreshing hosts, checking if there exists a connection pool for all of them", []),
     Hosts = mod_global_distrib_mapping:hosts(),
-    ?DEBUG("Discovered hosts: ~p~n", [Hosts]),
     lists:map(fun maybe_add_host/1, Hosts),
     ok.
 
@@ -122,15 +123,15 @@ schedule_refresh(Interval) ->
   erlang:send_after(Interval, self(), refresh).
 
 maybe_add_host(Host) ->
-    ?DEBUG("Checking host ~p (on host ~p)~n", [Host, local_host()]),
     case local_host() of
         Host ->
             ok;
         _ ->
+            ?DEBUG("reason=maybe_add_host host=~p local_host=~p", [Host, local_host()]),
             mod_global_distrib_outgoing_conns_sup:ensure_server_started(Host)
     end.
 
-default_interval() ->
+default_refresh_interval() ->
     3000.
 
 local_host() ->
--- a/src/global_distrib/mod_global_distrib_hosts_refresher.erl
+++ b/src/global_distrib/mod_global_distrib_hosts_refresher.erl
@@ -80,7 +80,7 @@ handle_info(refresh, Interval) ->
     {noreply, Interval}.
 
 terminate(normal, _State) ->
-    ?INFO("mod_global_distrib_refresher has terminated with reason: normal", []);
+    ?INFO_MSG("mod_global_distrib_refresher has terminated with reason: normal", []);
 terminate(Reason, State) ->
     ?ERROR_MSG("mod_global_distrib_refresher terminated with reason: ~p state=~p", [Reason, State]).
 
--- a/src/global_distrib/mod_global_distrib_hosts_refresher.erl
+++ b/src/global_distrib/mod_global_distrib_hosts_refresher.erl
@@ -1,0 +1,113 @@
+%%%==============================================================================
+%% Copyright 2017 Erlang Solutions Ltd.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%% http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%==============================================================================
+
+
+-module(mod_global_distrib_hosts_refresher).
+-author("dominik.stanaszek@erlang-solutions.com").
+
+-behavior(gen_server).
+-behavior(gen_mod).
+-include("ejabberd.hrl").
+
+%% API
+-export([start_link/0]).
+
+%% gen_server API
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/3]).
+
+%% gen_mod API
+-export([deps/2, start/2, stop/1]).
+
+
+%%--------------------------------------------------------------------
+%% API
+%%--------------------------------------------------------------------
+
+-spec start_link() -> pid() | term().
+start_link() ->
+    start_link(default_interval()).
+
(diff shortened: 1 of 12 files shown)


src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,113 @@
+%%%==============================================================================
+%% Copyright 2017 Erlang Solutions Ltd.
```
kzemek: It's 2018 now!

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,113 @@
...
+
+handle_cast(_Request, State) ->
+    {noreply, State}.
+
+handle_info(refresh, [{refresh_interval, Int}] = State) ->
+    ?DEBUG("Refreshing hosts, checking if there exists a supervisor for all of them~n", []),
+    Hosts = mod_global_distrib_mapping:hosts(),
+    ?DEBUG("Discovered hosts: ~p~n", [Hosts]),
+    lists:map(fun maybe_add/1, Hosts),
+    timer:send_after(Int, self(), refresh),
```
kzemek: [Prefer `erlang:send_after` to `timer`](http://erlang.org/doc/efficiency_guide/commoncaveats.html#id2262042).

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,113 @@
...
+    {noreply, State}.
+
+terminate(Reason, _State, _Data) ->
```
kzemek: `terminate` callback takes two arguments.

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
```diff
@@ -45,9 +42,23 @@ get_connection(Server) ->
...
 init(_) ->
     SupFlags = #{ strategy => one_for_one, intensity => 5, period => 5 },
-    {ok, {SupFlags, []}}.
+    RefresherSpec = #{
+      id => mod_global_distrib_hosts_refresher,
+      start => {mod_global_distrib_hosts_refresher, start_link, []},
+      restart => transient,
+      shutdown => 5000
+    },
+    {ok, {SupFlags, [RefresherSpec]}}.
```
kzemek: Since `mod_global_distrib_hosts_refresher` is a `gen_mod`-type module, its process should probably be started in its `start/2` callback instead of being hardcoded here.

kanes115:
@kzemek , tests are almost done. See TODO in comment to this PR. Thanks for the other comments.

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
```diff
@@ -33,10 +33,7 @@ add_server(Server) ->
 
 -spec get_connection(Server :: ejabberd:lserver()) -> pid().
 get_connection(Server) ->
-    case whereis(mod_global_distrib_utils:server_to_sup_name(Server)) of
-        undefined -> ok = add_server(Server);
-        _ -> ok
-    end,
+    ensure_server_started(Server),
```
arcusfelis: What if there are several processes calling this get_connection at the same time?

kanes115:
TODO: Investigate why the newly created test passes when run separately but fails when run along with other ones.

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
kanes115: Why should that be a problem?

kanes115:
I would also add an option to force refresh (synchronus) so that we could use it in a hook that is called on receiving packets. This could let us eliminate a problem with lost disco responses. Right now some of the disco responses are lost because we don't refresh global users on time and a node that got a request doesn't know the node it should send back the response to.

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
```diff
@@ -33,21 +33,27 @@ add_server(Server) ->
...
     mod_global_distrib_server_sup:get_connection(Server).
```
arcusfelis: hmm, there can be more than one connection between servers and you pick a random one?
No wonder this test fails

http://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1680/4000/pgsql_mnesia.20.0/big/ct_run.test@travis-job-c416160e-42d3-4434-ade5-ca16966418d8.2018-01-22_16.46.45/ejabberd_tests.tests.mod_global_distrib_SUITE.logs/run.2018-01-22_17.02.33/mod_global_distrib_suite.test_in_order_messages_on_multiple_connections_with_bounce.html

Why are you sure that the stanzas would be delivered in order?

kzemek: `get_connection` chain returns a random connection; the connection is then bound to the client process that initiated the call: https://github.com/esl/MongooseIM/blob/master/src/global_distrib/mod_global_distrib.erl#L137

kanes115: It should actually match on already_started as well (in add_server). This Is going to be changed.

kanes115:
The test might not be passing due to bug resolved [here](https://github.com/esl/MongooseIM/pull/1628)
CI checking it (in progress)

kanes115:
Trying to wait longer for a tree to build

kanes115:
Refresher does not log refreshing during the test time (even after making it long 2000ms with 200ms refresh interval)

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,156 @@
...
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [RefreshInterval], []).
+
+
+%%--------------------------------------------------------------------
+%% gen_mod callbacks
+%%--------------------------------------------------------------------
+
+-spec start(Host :: ejabberd:lserver(), Opts :: proplists:proplist()) -> any().
+start(Host, Opts) ->
+    ?DEBUG("Opts in refresher: ~p~n", [Opts]),
```
arcusfelis: comment this black magic.
In code.
I.e. runs start/0 only when Host is configured as global Host.

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,156 @@
...
+    {ok, _} = supervisor:start_child(mod_global_distrib_outgoing_conns_sup, Child),
+    ok.
+
+-spec stop() -> any().
+stop() ->
+    stop_outgoing_conns_sup(),
+    ok.
+
+refresh() ->
+    ?DEBUG("Refreshing hosts, checking if there exists a supervisor for all of them~n", []),
```
arcusfelis: `~n` in the end of line is not needed for these macro calls.

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,156 @@
...
+    Hosts = mod_global_distrib_mapping:hosts(),
+    ?DEBUG("Discovered hosts: ~p~n", [Hosts]),
+    lists:map(fun maybe_add_host/1, Hosts),
+    ok.
+
+schedule_refresh(Interval) ->
+  erlang:send_after(Interval, self(), refresh).
+
+maybe_add_host(Host) ->
+    ?DEBUG("Checking host ~p (on host ~p)~n", [Host, local_host()]),
```
arcusfelis: Checking host=~p local_host=~p
Or even

reason=maybe_add_host host=~p local_host=~p

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,156 @@
...
+    case local_host() of
+        Host ->
+            ok;
+        _ ->
+            mod_global_distrib_outgoing_conns_sup:ensure_server_started(Host)
+    end.
+
+default_interval() ->
```
arcusfelis: default_refresh_interval

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,156 @@
...
+
+
+-spec deps(Opts :: proplists:proplist()) -> gen_mod:deps_list().
+deps(Opts) ->
+    [{mod_global_distrib_mapping, Opts, hard}].
+
+-spec start() -> any().
+start() ->
+    start_outgoing_conns_sup(),
+    Interval = mod_global_distrib_utils:opt(?MODULE, hosts_refresh_interval, default_interval()),
```
arcusfelis: this can go into a subfun

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,156 @@
...
+handle_call(_Request, _From, State) ->
+    {noreply, State}.
+
+handle_cast(_Request, State) ->
+    {noreply, State}.
+
+handle_info(refresh, Interval) ->
+    refresh(),
+    schedule_refresh(Interval),
+    {noreply, Interval}.
```
arcusfelis: we can hibernate here

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,156 @@
...
+%%--------------------------------------------------------------------
+%% gen_server callbacks
+%%--------------------------------------------------------------------
+
+init([RefreshInterval]) ->
+    ?DEBUG("refresher starting with interval ~p~n", [RefreshInterval]),
+    schedule_refresh(RefreshInterval),
+    {ok, RefreshInterval}.
+
+handle_call(_Request, _From, State) ->
```
arcusfelis: `ERROR_MSG("issue=unknown_call request=~p from=~p", _)`

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,156 @@
...
+    {noreply, State}.
+
+handle_cast(_Request, State) ->
```
arcusfelis: log unknown casts

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,156 @@
...
+    {noreply, State}.
+
+handle_info(refresh, Interval) ->
+    refresh(),
+    schedule_refresh(Interval),
+    {noreply, Interval}.
+
+terminate(Reason, _State) ->
+    ?ERROR_MSG("mod_global_distrib_refresher terminated with reason: ~p", [Reason]),
```
arcusfelis: ERROR_MSG is too noisy. Can be useful, if reason is not `normal`.
For normal - it's debug or info level.

src/global_distrib/mod_global_distrib_mapping.erl
```diff
@@ -114,6 +115,9 @@ all_domains() ->
 endpoints(Host) ->
     mod_global_distrib_mapping_backend:get_endpoints(Host).
 
+hosts() ->
```
arcusfelis: what is hosts?

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
```diff
@@ -33,21 +33,27 @@ add_server(Server) ->
 
 -spec get_connection(Server :: jid:lserver()) -> pid().
 get_connection(Server) ->
```
arcusfelis: doc for this function.
Describe errors, blocking behaviour (does it wait for server to be connected too? does it return without waiting?).

arcusfelis: What if several processes try to connect to the same server which is not connected?
Will it trigger TCP connect for each call or will the calls be de-duplicated?
Should be in the function description.

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
arcusfelis: Maybe because of this?
https://github.com/esl/MongooseIM/pull/1660/files#r165373503
I know, it's very improbable to get this bug in tests, but in the real world (c) we often can get 10k routing calls at the same time. or 100k+ over several minutes.

arcusfelis: ah, so easy to reproduce the bug.
Two routes from two processes.
One creates a supervisor of connections. gen_tcp:connect can block init/1 function for some time.
So, second process would fail inside ensure_started.
at least it's not possible to start several supervisors, because the main supervisor would not allow it (but again, exactly that's why there is a badmatch in easure started).

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
arcusfelis: ok, it will block in

gen_server:call(PoolName, wait_for_connection)
and crash with lovely error no_connections. very informative, assuming that cpool does not include pool id into the error term.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -242,6 +248,32 @@ generic_end_per_testcase(CaseName, Config) ->
 %% Service discovery test
 %%--------------------------------------------------------------------
 
+test_host_refreshing(_Config) ->
+    timer:sleep(?HOSTS_REFRESH_INTERVAL + 20), % Let's wait to be sure refresher added children
```
arcusfelis: If we wait for something, we should check that some condition is true. 

```erlang
eventually_true("wait for refresher", fun() -> there_are_children(...))

eventually_true(Description, F) ->                                                                
    eventually_true(Description, F, 5).                                                           
                                                                                     
eventually_true(Description, _F, 0) ->                                                            
    ct:fail({eventually_true_failed, Description});                                              
eventually_true(Description, F, Retries) ->                                                       
    case F() of                                                                      
        true ->                                                                      
            ok;                                                                      
        false ->                                                                     
            timer:sleep(1000),                                                       
            eventually_true(Description, F, Retries-1)                                            
    end.
```

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -242,6 +248,32 @@ generic_end_per_testcase(CaseName, Config) ->
 %% Service discovery test
 %%--------------------------------------------------------------------
 
+test_host_refreshing(_Config) ->
```
arcusfelis: Test description using human words :)

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -242,6 +248,32 @@ generic_end_per_testcase(CaseName, Config) ->
...
+    ct:log("children in europe1: ~p~n", [rpc(europe_node1, supervisor, which_children, [mod_global_distrib_outgoing_conns_sup])]),
+    [EuropeHost] = lists:filtermap(fun({europe_node1, Host, _}) -> {true, list_to_binary(Host)};
+                                      (_) -> false end, get_hosts()),
+    EuropeSup = binary_to_atom(<<"mod_global_distrib_server_sup_", EuropeHost/binary>>, utf8),
+    [EuropePid] = lists:filtermap(fun({Sup, Pid, supervisor, _}) -> {Sup =:= EuropeSup, Pid};
+                                   (_) -> false end,
+                                      Children),
+
+    erlang:exit(EuropePid, kill),
+
```
arcusfelis: Add timer:sleep(5000) here and run this test.
It's how travis would execute this test. Sometimes.

Correct order

```
ChildrenNoEurope = rpc(asia_node, supervisor, which_children, [mod_global_distrib_outgoing_conns_sup]),
erlang:exit(EuropePid, kill),
```

src/global_distrib/mod_global_distrib_mapping.erl
kanes115: These are all the hosts registered in Redis. I'm gonna add a spec here. I called it according to `endpoints/1` function. What more should be changed?

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
```diff
@@ -33,21 +33,27 @@ add_server(Server) ->
...
 
 %%--------------------------------------------------------------------
 %% Helpers
 %%--------------------------------------------------------------------
 
+ensure_server_started(Server) ->
+  case whereis(mod_global_distrib_utils:server_to_sup_name(Server)) of
+    undefined ->
+      ?DEBUG("Host ~p didn't have a corresponding supervisor", [Server]),
+      ok = add_server(Server);
```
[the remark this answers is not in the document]
kanes115: I have already taken care of that [here](https://github.com/esl/MongooseIM/pull/1695/files#diff-efb82bbe04fe9e7a6904ef4e4bd732fbR31)

kanes115:
This test is outdated after this PR:
[test_in_order_messages_on_multiple_connections_with_bounce](https://github.com/esl/MongooseIM/blob/master/test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl#L553)

arcusfelis:
updates here?

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -853,6 +874,33 @@ redis_query(Node, Query) ->
     RedisWorker = rpc(Node, wpool_pool, best_worker, [mod_global_distrib_mapping_redis]),
     rpc(Node, eredis, q, [RedisWorker, Query]).
 
+eventually_true(Description, _Predicate, Interval, 0) ->
```
fenek: It would be better to reuse `wait_until`, only perhaps extend it with `Description`.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -242,6 +248,21 @@ generic_end_per_testcase(CaseName, Config) ->
 %% Service discovery test
 %%--------------------------------------------------------------------
 
+test_host_refreshing(_Config) ->
+    eventually_true("wait for refresher to converge trees states",
+                    fun() -> trees_for_connections_present() end, ?HOSTS_REFRESH_INTERVAL, 2),
+    ConnectionSups = out_connection_sups(asia_node),
+    [EuropeHost] = lists:filtermap(fun({europe_node1, Host, _}) -> {true, list_to_binary(Host)};
```
fenek: Whoa, isn't it shooting mosquito with a cannon? Why can't we use `lists:keyfind/3`?

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -242,6 +248,21 @@ generic_end_per_testcase(CaseName, Config) ->
...
+                                      (_) -> false end, get_hosts()),
+    EuropeSup = rpc(asia_node, mod_global_distrib_utils, server_to_sup_name, [EuropeHost]),
+    [EuropePid] = lists:filtermap(fun({Sup, Pid, supervisor, _}) -> {Sup =:= EuropeSup, Pid};
```
fenek: See comment above.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -853,6 +874,33 @@ redis_query(Node, Query) ->
...
+
+trees_for_connections_present() ->
+    AsiaChildren = out_connection_sups(asia_node),
+    Europe1Children = out_connection_sups(europe_node1),
+    Europe2Children = out_connection_sups(europe_node2),
+    lists:all(fun(Host) -> length(Host) > 0 end, [AsiaChildren, Europe1Children, Europe2Children]).
+
+tree_for_sup_present(Node, ExpectedSup) ->
+    Children = out_connection_sups(Node),
+    lists:any(fun({Sup, _, _, _}) -> Sup =:= ExpectedSup;
```
fenek: `lists:keyfind`?

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -242,6 +248,21 @@ generic_end_per_testcase(CaseName, Config) ->
...
+    eventually_true("wait for refresher to converge trees states",
+                    fun() -> trees_for_connections_present() end, ?HOSTS_REFRESH_INTERVAL, 2),
+    ConnectionSups = out_connection_sups(asia_node),
+    [EuropeHost] = lists:filtermap(fun({europe_node1, Host, _}) -> {true, list_to_binary(Host)};
+                                      (_) -> false end, get_hosts()),
+    EuropeSup = rpc(asia_node, mod_global_distrib_utils, server_to_sup_name, [EuropeHost]),
+    [EuropePid] = lists:filtermap(fun({Sup, Pid, supervisor, _}) -> {Sup =:= EuropeSup, Pid};
+                                     (_) -> false end,
+                                      ConnectionSups),
+    erlang:exit(EuropePid, kill),
```
fenek: I think a comment would be useful here, that it's OK to `kill` the process as it has `temporary` restart type.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -55,6 +57,8 @@ groups() ->
        test_update_senders_host_by_ejd_service
        %% TODO: Add test case fo global_distrib_addr option
       ]},
+     {host_refresher, [],
```
kanes115: This is inconsistency. Let's call it hosts_refresher.

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
arcusfelis: Can be fixed by https://github.com/esl/MongooseIM/pull/1750
But this place should be covered with tests.
This comment branch is out of scope of this PR.

src/global_distrib/mod_global_distrib_hosts_refresher.erl
kzemek: Still todo.

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,154 @@
...
+handle_cast(Request, State) ->
+    ?ERROR_MSG("issue=unknown_cast request=~p", [Request]),
+    {noreply, State}.
+
+handle_info(refresh, Interval) ->
+    refresh(),
+    schedule_refresh(Interval),
+    {noreply, Interval}.
+
+terminate(normal, _State) ->
```
kzemek: Is there a reason to restrict termination to `normal`? From experience, crash on `terminate/2` obfuscates original crash that triggered the `terminate/2` in the first place.

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,154 @@
...
+-spec deps(Opts :: proplists:proplist()) -> gen_mod:deps_list().
+deps(Opts) ->
+    [{mod_global_distrib_mapping, Opts, hard}].
+
+-spec start() -> any().
+start() ->
+    start_outgoing_conns_sup(),
+    Interval = mod_global_distrib_utils:opt(?MODULE, hosts_refresh_interval, default_refresh_interval()),
+    Child = #{
+      id => mod_global_distrib_hosts_refresher,
```
kzemek: `?MODULE`

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,154 @@
...
+      start => {?MODULE, start_link, [Interval]},
+      restart => transient,
+      shutdown => 5000,
+      modules => [mod_global_distrib_hosts_refresher]
```
kzemek: `?MODULE`

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,154 @@
...
+    [{mod_global_distrib_mapping, Opts, hard}].
+
+-spec start() -> any().
+start() ->
+    start_outgoing_conns_sup(),
+    Interval = mod_global_distrib_utils:opt(?MODULE, hosts_refresh_interval, default_refresh_interval()),
+    Child = #{
+      id => mod_global_distrib_hosts_refresher,
+      start => {?MODULE, start_link, [Interval]},
+      restart => transient,
```
kzemek: Why `transient`?

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,154 @@
...
+
+-spec stop() -> any().
+stop() ->
+    stop_outgoing_conns_sup(),
+    ok.
+
+refresh() ->
+    ?DEBUG("Refreshing hosts, checking if there exists a connection pool for all of them", []),
+    Hosts = mod_global_distrib_mapping:hosts(),
+    lists:map(fun maybe_add_host/1, Hosts),
```
kzemek: `lists:foreach`, unless you forgot to handle results.

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,154 @@
...
+
+handle_call(Request, From, State) ->
+    ?ERROR_MSG("issue=unknown_call request=~p from=~p", [Request, From]),
+    {reply, {error, unknown_request}, State}.
+
+handle_cast(Request, State) ->
+    ?ERROR_MSG("issue=unknown_cast request=~p", [Request]),
+    {noreply, State}.
+
+handle_info(refresh, Interval) ->
```
kzemek: I think we'd do well to handle other, unexpected messages here. While call and casts really shouldn't happen without us coding the wrong logic, normal messages could be generated by random things in runtime over which we have no control.

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,154 @@
...
+
+schedule_refresh(Interval) ->
+    erlang:send_after(Interval, self(), refresh).
+
+maybe_add_host(Host) ->
+    case local_host() of
+        Host ->
+            ok;
+        _ ->
+            ?DEBUG("reason=maybe_add_host host=~p local_host=~p", [Host, local_host()]),
```
kzemek: I think adding a new connection pool is `info` worthy (unless it's logged somewhere else).

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -0,0 +1,154 @@
...
+%%--------------------------------------------------------------------
+
+
+-spec deps(Opts :: proplists:proplist()) -> gen_mod:deps_list().
+deps(Opts) ->
+    [{mod_global_distrib_mapping, Opts, hard}].
+
+-spec start() -> any().
+start() ->
+    start_outgoing_conns_sup(),
```
kzemek: I'm actually confused by `outgoing_conns_sup` being started by this module (moved from `mod_global_distrib_sender`). On one hand, this does make a degree of sense, on the other... it seems weird? @fenek any thoughts?

src/global_distrib/mod_global_distrib_utils.erl
```diff
@@ -114,6 +114,15 @@ opt(Module, Key) ->
             error(atom_to_list(Module) ++ " required option unset: " ++ atom_to_list(Key))
     end.
 
+-spec opt(module(), Key :: atom(), Default :: term()) -> Value :: term().
+opt(Module, Key, Default) ->
```
kzemek: The reason there was no `opt/3` with default value is that each module specifies its default values on start.

Example: `mod_global_distrib_bounce` adds default values for `resend_after` and `max_retries`: https://github.com/esl/MongooseIM/blob/master/src/global_distrib/mod_global_distrib_bounce.erl#L43

The point was to avoid `try ... catch` because options are consulted very often, and to enable fast inspection and change of all options used by the module (`ets:tab2list(mod_global_distrib_bounce)`).

src/global_distrib/mod_global_distrib_hosts_refresher.erl
kanes115: Well, we want it to be restarted in case of error, don't we?

src/global_distrib/mod_global_distrib_hosts_refresher.erl
kanes115: Of course, I agree, but what should I do then? Not log it? What implementation of handle_call, handle_cast should I provide? Or maybe this whole module shouldn't be a gen_server as it does not expect any calls or casts? Please elaborate :)

src/global_distrib/mod_global_distrib_hosts_refresher.erl
kzemek: The full question would be "why `transient` as opposed to `permanent`?".

src/global_distrib/mod_global_distrib_hosts_refresher.erl
kzemek: The comment is not about `handle_cast` & `handle_call`, it's about providing a catch-all for `handle_info`.

src/global_distrib/mod_global_distrib_hosts_refresher.erl
kzemek: @fenek ping

fenek: It does seem weird for me as well but I didn't complain because I couldn't find more rational argument than "looks weird".

src/global_distrib/mod_global_distrib_hosts_refresher.erl
kanes115: Ahh, okey

kanes115:
It might be necessary to add stoping hosts_refresher [here](https://github.com/esl/MongooseIM/blob/48c57b1edc37557cf15bb13f3e2bf294f8a928fe/test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl#L233). I think lack of it causes end_per_testcase crash sometimes.

arcusfelis:
Travis  fails
```erlang
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
mod_global_distrib_SUITE:wait_until failed on line 789
Reason: {badmatch,[false,false]}
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Testing ejabberd_tests.tests.mod_global_distrib_SUITE: *** FAILED test case 24 of 24 ***
----------------------------------------------------
2018-03-06 19:43:14.908
{test_host_refreshing,hosts_refresher}
{{badmatch,[false,false]},
 [{mod_global_distrib_SUITE,wait_until,5,
                            [{file,"mod_global_distrib_SUITE.erl"},
                             {line,789}]},
  {mod_global_distrib_SUITE,test_host_refreshing,1,
                            [{file,"mod_global_distrib_SUITE.erl"},
                             {line,256}]},
  {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1529}]},
  {test_server,run_test_case_eval1,6,[{file,"test_server.erl"},{line,1045}]},
  {test_server,run_test_case_eval,9,[{file,"test_server.erl"},{line,977}]}]}
----------------------------------------------------
```

arcusfelis:
feel free to put "ready" label again, once CI passes.

fenek:
@arcusfelis 

It seems it is a consequence of rebase. I'll fix it.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -24,7 +24,7 @@
 -include_lib("escalus/include/escalus_xmlns.hrl").
 -include_lib("exml/include/exml.hrl").
 
--define(HOSTS_REFRESH_INTERVAL, 200).
+-define(HOSTS_REFRESH_INTERVAL, 3600*1000). %% 1h - long enough not to break any test by accident
```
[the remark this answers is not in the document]
kanes115: look [here](https://github.com/esl/MongooseIM/blob/870c27a193ca45d330e78a0a5734f02f260099e5/test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl#L142) and [here](https://github.com/esl/MongooseIM/blob/870c27a193ca45d330e78a0a5734f02f260099e5/test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl#L256).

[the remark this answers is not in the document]
fenek: They have worked on my local machine as usual but I'll double check.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -24,6 +24,8 @@
 -include_lib("escalus/include/escalus_xmlns.hrl").
 -include_lib("exml/include/exml.hrl").
 
+-define(HOSTS_REFRESH_INTERVAL, 3600*1000). %% 1h - long enough not to break any test by accident
```
arcusfelis: or :)
```erlang
-define(HOSTS_REFRESH_INTERVAL, timer:hours(1)). %% long enough not to break any test by accident
```

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -227,23 +228,47 @@ end_per_testcase(CaseName, Config) ->
 generic_end_per_testcase(CaseName, Config) ->
     lists:foreach(
       fun({NodeName, _, _}) ->
+              %% TODO: Enable refresher only for specific test cases,
```
kanes115: I'm not sure, but I feel like if we have to make such exclusion, it means we are not testing the real program. In reality the refresher will always be started and never paused.

src/global_distrib/mod_global_distrib_hosts_refresher.erl
```diff
@@ -31,6 +31,13 @@
 %% gen_mod API
 -export([deps/2, start/2, stop/1]).
 
+%% Test & debug API
+-export([pause/0, unpause/0]).
+
+-record(state, {
+          refresh_interval :: pos_integer(),
+          tref :: reference() | undefined
```
kanes115: Since it's a timer reference, can't we call it `timer_ref`? Not much longer and much more descriptive in my opinion.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
fenek: Theoretically yes but some test cases are supposed to verify correctness of on-demand connection establishment. That's why I haven't introduced it here but IMO should be a subject of separate discussion.

src/global_distrib/mod_global_distrib_hosts_refresher.erl
fenek: It's pretty common in MIM source code to abbreviate it as `tref`.

PR #1662, opened 2018-01-17
Packaging MongooseIM

erszcz:
This PR enables to build MongooseIM CentOS 7 RPMs in a containerised environment. It's possible to extend the tools to other CentOS versions as well as other package types (read: `.deb`s :wink:).

Caveats:

- The scripts reuse utilities from private ESL https://github.com/esl/package-buildscripts - if we intend to allow the community to build packages, we have to decide if/how to share the build scripts from this repo.
- The tools under `tools/pkg` will need tweaking if support for Debian or other platforms is added, since the only testing grounds so far were CentOS 6 and 7 (with CentOS 6 being dropped, as its default toolchain seems too old for current MongooseIM master, i.e. no `-std=c++11` in `gcc-c++`).


erszcz:
The README is slightly outdated currently, I'll update it shortly.

erszcz:
I think two more items need to be added:

- [ ] example git hook to keep `VERSION` and git tag in sync
- [ ] Travis job to test the build pipeline

erszcz:
Travis is green \o/

tools/pkg/README.md
```diff
@@ -0,0 +1,108 @@
...
+version went into which package.
+
+`VERSION` is also used as version of the Erlang release
+which will be installed from the package.
+The operations people will be grateful for staying sane if you make sure
+that the package name/version, release version and git tag all match.
+
+Git hooks can be a great reminder here!
+
+TODO: add git hook example
```
michalwski: Do you have the git hook example somewhere or will it require some time to prepare one?

erszcz:
> While the Dockerfiles use hardcoded version 18.3.

Good catch - I'm fixing it.

> I see you disabled to job for centos6 rpms, do you think it's worth checking why this doesn't work on travis?

Actually, it works fine. I simply came to the conclusion that if somebody uses a at least a moderately recent system, then a CentOS 7 RPM will be just fine, and we'll save a bit on total build time by having just two packaging jobs. In case we really have to build for an older system we might use the CentOS 6 platform definition anyway - for now it's still in the repo.

tools/pkg/README.md
erszcz: Indeed, I have an example hook I'll add, although in the longer term I had issues with the ergonomics of the solution. Maybe using a `post-commit` hook would work better, while I experimented with a `pre-commit` hook.

PR #1678, opened 2018-01-19
Add hex escaping of binaries to MySQL backend.

kzemek:
With a bit of refactor.

### DIFF
--- a/src/bin_to_hex.erl
+++ b/src/bin_to_hex.erl
@@ -4,7 +4,7 @@
 
 -compile([native, {hipe, [o3]}, {inline, [hex/1]}]).
 
--export([bin_to_hex/1]).
+-export([bin_to_hex/1, hex_to_bin/1]).
 
 bin_to_hex(B) when is_binary(B) ->
   bin_to_hex(B, <<>>).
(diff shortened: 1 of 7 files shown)


src/bin_to_hex.erl
```diff
@@ -24,6 +24,14 @@ bin_to_hex_(<<A:8, B:8, C:8, D:8, E:8, F:8, G:8, H:8, Rest/binary>>, Acc) ->
     <<Acc/binary,
       ?H(A), ?H(B), ?H(C), ?H(D), ?H(E), ?H(F), ?H(G), ?H(H)>>).
 
+-spec hex_to_bin(binary()) -> <<_:_*1>>.
+hex_to_bin(Bin) when is_binary(Bin) ->
```
arcusfelis: would compiling THIS function to native improve performance?

kzemek: No idea, but if it belongs in any single module it's this one.

arcusfelis: so, I suggest microbenchmark it.
bin_to_hex is optimized for native, but not all code runs good using hipe.

fenek: Microbenchmark or profiling should be a separate task.

src/rdbms/mongoose_rdbms_odbc.erl
```diff
@@ -123,3 +122,15 @@ unsplit_query([QueryHead | QueryRest], ParamMappers, [Param | Params], QueryAcc,
...
+server_type(Pool) ->
+    mongoose_rdbms_sup:get_option(Pool, odbc_server_type).
+
+-spec escape_binary(mongoose_rdbms:pool(), ServerType :: atom(), binary()) -> binary().
+escape_binary(Pool, pgsql, Bin) ->
+    mongoose_rdbms_pgsql:escape_binary(Pool, Bin);
+escape_binary(Pool, mysql, Bin) ->
+    mongoose_rdbms_mysql:escape_binary(Pool, Bin);
+escape_binary(_Pool, _ServerType, Bin) ->
+    <<"'", (bin_to_hex:bin_to_hex(Bin))/binary, "'">>.
```
arcusfelis: why not to use iolists?

src/rdbms/mongoose_rdbms_pgsql.erl
```diff
@@ -22,13 +22,20 @@
...
+-export([escape_binary/2, unescape_binary/2, connect/2, disconnect/1,
+         query/3, prepare/6, execute/4]).
 
 %% API
 
--spec escape_format(mongoose_rdbms:pool()) -> atom().
-escape_format(_Pool) ->
-    hex.
+-spec escape_binary(mongoose_rdbms:pool(), binary()) -> binary().
+escape_binary(_Pool, Bin) when is_binary(Bin) ->
```
arcusfelis: iolists?

PR #1682, opened 2018-01-22
Test

SauronSauron:
This PR addresses #

Proposed changes include:


arcusfelis:
@SauronSauron not sure what was this, BUT I propose to configure travis on your account for such tests.
Like I did https://github.com/arcusfelis/MongooseIM/pull/21
It should take you 10 minutes.
The only drawback is that you would not get ct_reports uploaded to AWS. But I guess you can use local test run or run here, but ONE task only, the one that fails.

Travis CI is limited resource and should be used only for ready pull requests (or for debugging test/build scripts).

Issue #1683, opened 2018-01-22
Disable broken test_in_order_messages_on_multiple_connections_with_bounce test case in GlobDist


arcusfelis:
MongooseIM version: 4a19c96177d50737380ea552f873aadf1c90136b

Broken test:
http://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1680/4000/pgsql_mnesia.20.0/big/ct_run.test@travis-job-c416160e-42d3-4434-ade5-ca16966418d8.2018-01-22_16.46.45/ejabberd_tests.tests.mod_global_distrib_SUITE.logs/run.2018-01-22_17.02.33/mod_global_distrib_suite.test_in_order_messages_on_multiple_connections_with_bounce.html

from mod_global_distrib_server_mgr

```erlang
-spec pick_connection(Enabled :: [endpoint_info()]) -> pid().                        
pick_connection(Enabled) ->                                                          
    #endpoint_info{ conn_pool_ref = PoolRef } = lists:nth(rand:uniform(length(Enabled)), Enabled),
    cpool:get_connection(PoolRef). 
```

This is not correct. Picking random connection would not enforce order.
There are several ways to do it right:
- use hash from `From` jid;
- use hash from `{From, To}`;
- use hash of the caller's pid. i.e. the pid of the process, sending stanza, or calling routing function.

arcusfelis:
it's should be easy to reproduce, if you pause some of tcp connections or processes, handling them on sender's side.

fenek:
PR with a fix for a race condition in GD is waiting for review, no need to disable this test case.

PR #1684, opened 2018-01-22
Disable test_in_order_messages_on_multiple_connections_with_bounce #1683

arcusfelis:
It fails to often, see #1683 with some debug info


fenek:
Obsoleted by #1689

PR #1700, opened 2018-02-07
Test component visibility

kanes115:
This PR is to reproduce and collect conclusions on visibility of different components on different nodes/datacenters.

TODO: PR is no longer needed but we might reuse new test cases.

### DIFF
--- a/rel/fed1.vars.config
+++ b/rel/fed1.vars.config
@@ -10,7 +10,12 @@
 {ejabberd_s2s_in_port, 5299}.
 {cowboy_port, 5282}.
 {cowboy_port_secure, 5287}.
-{ejabberd_service, ""}.
+{ejabberd_service, ",{9999, ejabberd_service, [\n"
+                 "                {access, all},\n"
+                 "                {shaper_rule, fast},\n"
+                 "                {ip, {127, 0, 0, 1}},\n"
+                 "                {password, \"secret\"}\n"
+                 "           ]}"}.
 {mod_roster, "{mod_roster, []},"}.
 {http_api_endpoint, "{5389, \"127.0.0.1\"}"}.
 {s2s_use_starttls, "{s2s_use_starttls, optional}."}.
--- a/test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
+++ b/test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
@@ -53,7 +53,8 @@ groups() ->
        test_component_unregister,
        test_update_senders_host,
        test_update_senders_host_by_ejd_service,
-       test_components_on_different_hosts
+       test_components_on_different_hosts,
+       test_component_referenced_before_connecting
        %% TODO: Add test case fo global_distrib_addr option
       ]},
      {cluster_restart, [],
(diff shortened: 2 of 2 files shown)


kanes115:
I have tried different configurations and came up with a table:

|who | to whom | can connect? |
|-----|----------|--------------|
alice | europe | yes
eve | europe | yes
alice | asia | no
eve | asia | yes

alice logs in in europe and eve in asia always.

Conclusion: In prepared testcase component connected to asia (host: fed1) is not visible globally - this is proved both by this table and the redis mappings (which lacks component connected to asia (fed1).

Inside ets table `external_component`:
```
asia: []
europe1: [{external_component,{<<"test_service2.fed1">>,mongooseim@localhost},
                              {packet_handler,ejabberd_service,<7650.12769.0>},
                              mongooseim@localhost},
          {external_component,{<<"test_service.localhost">>,
                               mongooseim@localhost},
                              {packet_handler,ejabberd_service,<7650.12767.0>},
                              mongooseim@localhost}]
europe2: []
```

Inside ets table `mod_global_distrib_mapping_redis_domains`:

```
asia: []
europe1: [{<<"test_service.localhost">>}]
europe2: []
```

All visible globally domains: `{ok,[<<"test_service.localhost">>]}` (result of calling `mod_global_distrib_mapping:all_domains/0`)

kanes115:
In asia (fed1):
`(<0.6608.0>) call mod_global_distrib_mapping:for_domain(<<"test_service2.localhost">>)`
The mapping is introduced by fed1 but gone later somehow (not seen from both asia and europe)

kanes115:
In europe:
`(<0.13575.0>) call mod_global_distrib_mapping:insert_for_domain(<<"test_service.localhost">>)`
So it is added in europe but there is no such debug entry on asia node. That might be the reason of the bug.

kanes115:
```
2017-11-27 20:53:21.318 [error] <0.1006.0> CRASH REPORT Process <0.1006.0> with 0 neighbours exited with reason: exists in exometer_admin:new_entry/3 line 175 in gen_server:init_it/6 line 352
2017-11-27 20:53:21.318 [error] <0.1004.0> Supervisor {<0.1004.0>,cpool_pool_connection_sup} had child connection started with {mod_global_distrib_connection,start_link,undefined} at <0.1006.0> exit with reason exists in exometer_admin:new_entry/3 line 175 in context child_terminated
```
To `mod_global_distrib_connection:start_link/2` function an argument `undefined` was passed (as a second argument) but it should be of type `binary()`

kanes115:
Getting back to error with unseen components:

we get an `error` instead of `{ok, LocalHost}` in cache tab.

kanes115:
The problem might appear in the following scenario:
component `rf11` connected to Mongoose (everything is fine with it), someone tried to connect to `rf12`, wrote `error` to cache (cause it is not yet connected) and only after that `rf12` actually connected to Mongoose. Client who asks about `rf12` looks into cache and has error there - so `rf12` is not seen.

michalwski:
@kanes115 @fenek do we still need this PR and want it to be merged with master?

kanes115:
You can delete the branch

PR #1707, opened 2018-02-08
Add conditional logging macros.

kzemek:


include/mongoose_logger.hrl
```diff
@@ -1,19 +1,37 @@
 -ifndef(MONGOOSEIM_LOGGER_HRL).
 -define(MONGOOSEIM_LOGGER_HRL, true).
 
+-define(LOG_IF(Level, Condition, Format, Args),
```
arcusfelis: I like the idea.
Would be cool to add some docs into developers-guide about logging, because currently there are none.

Issue #1639, opened 2018-01-13
Push-enabling causes repeated MongooseIM crashes


pbhowmic:
MongooseIM version: 2.1.0
Installed from: erlang solutions
Erlang/OTP version: esl-erlang 1:20.2.2

I am running a fairly standard MongooseIM server on Ubuntu Xenial in GCE with these standouts:-
Push messaging enabled and my client is Android-based with and uses the smack library and SSL/TLS disabled (for now). Bear with me as I try & describe this issue the best I can and please ask for more info as needed.
1. The following 3 are served by the same domain _X_ and MongooseIM instance.

- mod_push
- mod_pubsub
- mod_push_service_mongoosepush

2. The MongoosePush component is running as a docker container in the same GCE instance as the above and the subdomain _pubsub.X_ resolves to the same GCE host.
3. I have http authentication turned on.

The following is my ejabberd.cfg file (comments/blank lines stripped out)

```
{loglevel, 3}.
{hosts, ["chat-dev.idocrx.com"] }.
{listen,
 [
  { 5280, ejabberd_cowboy, [
      {num_acceptors, 10},
      {transport_options, [{max_connections, 1024}]},
      {modules, [
          {"_", "/http-bind", mod_bosh},
          {"_", "/ws-xmpp", mod_websockets, [{ejabberd_service, [
                                        {access, all},
                                        {shaper_rule, fast},
                                        {ip, {127, 0, 0, 1}},
                                        {password, "secret"}]}
          ]}
      ]}
  ]},
  { 5285, ejabberd_cowboy, [
        {num_acceptors, 10},
        {transport_options, [{max_connections, 1024}]},
        {ssl, [{certfile, "priv/ssl/fake_cert.pem"}, {keyfile, "priv/ssl/fake_key.pem"}, {password, ""}]},
        {modules, [
            {"_", "/http-bind", mod_bosh},
            {"_", "/ws-xmpp", mod_websockets, [
            ]}
        ]}
    ]},
  { {8088, "10.142.0.2"} , ejabberd_cowboy, [
      {num_acceptors, 10},
      {transport_options, [{max_connections, 1024}]},
      {modules, [
          {"10.142.0.2", "/api", mongoose_api_admin, []}
      ]}
  ]},
  { 8089 , ejabberd_cowboy, [
      {num_acceptors, 10},
      {transport_options, [{max_connections, 1024}]},
      {protocol_options, [{compress, true}]},
[...]

PR #1710, opened 2018-02-08
make sure we don't ping user w/o resource

sstrigler:
In some cases it can happen that a jid without resource ends up in mod_ping's internal 'database'.

In our case it was that a custom module injected messages manually into the system for users without a resource and manually ran the user_send_packet hook. 

The result is an endless loop of iq-error replies bringing the whole system to a halt within a short time (system replying a 'bad-request' error to an iq-error and so on). 

The proposed fix just makes sure we never try to ping a user without a resource.

### DIFF
--- a/src/mod_ping.erl
+++ b/src/mod_ping.erl
@@ -72,6 +72,9 @@ start_link(Host, Opts) ->
     gen_server:start_link({local, Proc}, ?MODULE, [Host, Opts], []).
 
 start_ping(Host, JID) when JID#jid.lresource =/= <<>> ->
+    %% Guard check above ensures we are not adding a user without a
+    %% resource by accident. If this happens the server ends up in a
+    %% hot loop of iq-error replies.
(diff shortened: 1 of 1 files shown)


sstrigler:
Please note, this is a first shot at the problem. As for us we could address it on a different level as well now that we know what causes it.
But then we saw that others are experiencing similar problems that might be related (#1639).

src/mod_ping.erl
```diff
@@ -71,9 +71,10 @@ start_link(Host, Opts) ->
     Proc = gen_mod:get_module_proc(Host, ?MODULE),
     gen_server:start_link({local, Proc}, ?MODULE, [Host, Opts], []).
 
-start_ping(Host, JID) ->
+start_ping(Host, JID) when JID#jid.lresource =/= <<>> ->
```
arcusfelis: description about infinite loop problems should be in a comment here.

PR #1724, opened 2018-02-13
Use different config entry for advertised endpoints

kanes115:
This PR seperates endpoints node will listen on from the ones it will be advertising. This can be useful i.e. when a node is behind a NAT and it needs to advertise its inside global and listen on its inside local.


test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -242,6 +260,12 @@ generic_end_per_testcase(CaseName, Config) ->
 %% Service discovery test
 %%--------------------------------------------------------------------
 
+test_advertised_endpoints_override_endpoints(Config) ->
```
arcusfelis: Description on how this test works like.
What exactly we are comparing?

Example:

Requires module blabla to be started with argument advertised_endpoints for each host in get_hosts().
Reads Redis to confirm that endpoints are overwritten.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -853,6 +877,9 @@ redis_query(Node, Query) ->
     RedisWorker = rpc(Node, wpool_pool, best_worker, [mod_global_distrib_mapping_redis]),
     rpc(Node, eredis, q, [RedisWorker, Query]).
 
+advertised_endpoints() ->
+    [{ {231, 110, 1, 4}, 2222}].
```
arcusfelis: Comment: This address is a fake one, we never try to connect there.
Used in test_advertised_endpoints_override_endpoints test case.

kanes115:
To decide:
do we want {adverised_endpoints, []} to mean that we don't want to advertise any endpoints or that `endpoints` entry should be taken (as if there was no advertised_endpoints entry at all)?

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
kanes115: @arcusfelis, in what form and where should such a description be put? What do you mean? (or do you mean - in a testcase name?)

arcusfelis:
@kanes115 `{adverised_endpoints, []}` should mean no endpoints.
`{adverised_endpoints, undefined}` or missing value should mean we use endpoints value.
All this should be documented.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
arcusfelis: @kanes115 We can put some small readable description for each testcase function.
Example:

```erlang
%% This case verifies advertised_endpoints configuration parameter.
%% We read a list of endpoints from redis and compare it with configured value.
%% Makes this check for each host.
%% Requires module mod_global_distrib to be loaded with advertised_endpoints parameter.
test_advertised_endpoints_override_endpoints(Config) ->
...
```

Basically, think, that one of our clients asks:
- make me a document that describes each test you make, so we can ensure that our system works.
Than you can't just copy past them a list from `all()` function.

fenek:
@kanes115 I agree with @arcusfelis on `{adverised_endpoints, []}`

kanes115:
@fenek, @arcusfelis, so that's how it's done. Thanks ;)

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
kanes115: I like that idea, but I don't see much documentation elsewhere in this suite. That's why I haven't documented it. No problem.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -242,6 +260,12 @@ generic_end_per_testcase(CaseName, Config) ->
 %% Service discovery test
 %%--------------------------------------------------------------------
 
+test_advertised_endpoints_override_endpoints(Config) ->
+    GetEndpoints = fun({NodeName, _, _}) ->
+                            rpc(europe_node1, mod_global_distrib_mapping_redis, get_endpoints, [<<"fed1">>]) end,
```
fenek: I think we want to call `NodeName`, not `europe_node1` every time? :)

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
fenek: During last grooming we've discussed that every new test case should have a brief description as a comment.

arcusfelis: @kanes115 We are going to have a meeting to discuss how docs and test cases should look like pretty soon :)
I often find myself in a situation, when I don't remember how my own tests work. But if there is some description, it's easier to dig into. Also, it helps to make a better test, the same motivation as duck programming.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
kanes115: Sure :p

src/global_distrib/mod_global_distrib_mapping.erl
```diff
@@ -298,3 +302,42 @@ do_lookup_jid({_, _, _} = Jid) ->
...
+-spec endpoints_with_ipaddr_and_domains(AllEndpoints :: [endpoint()])
+                                       -> {IpEndpoints :: [endpoint()], DomainEndpoints :: [endpoint()]}.
+endpoints_with_ipaddr_and_domains(Endpoints) ->
+    WithDomain = lists:filter(fun({Domain, _Port}) ->
+                                        is_domain(Domain) end, Endpoints),
+    WithIp = Endpoints -- WithDomain,
+    {WithIp, WithDomain}.
+
+-spec is_domain(any()) -> boolean().
+is_domain(Domain) when is_tuple(Domain) -> false;
```
kanes115: @fenek or @arcusfelis, do you think this function name is ok? I think it might be ambiguous but can't think of a nice and short name.

kanes115: Especially that it might soon go out of this module to utils :p

src/global_distrib/mod_global_distrib_server_mgr.erl
```diff
@@ -297,7 +297,7 @@ get_endpoints(Server) ->
         undefined -> mod_global_distrib_mapping:endpoints(Server);
         Endpoints -> {ok, Endpoints}
     end,
-    {ok, mod_global_distrib_utils:resolve_endpoints(EndpointsToResolve)}.
+    {ok, EndpointsToResolve}.
```
kanes115: @fenek, @arcusfelis, All tests pass after I deleted it, but could you, please, take a look and tell whether it is ok?

kanes115: Ok, so we probably don't have a test for `global_distrib_addr` and that is a case where this would fail after deletion of this function call

kanes115:
TODO: move some functions to utils!!

kanes115:
Why does mod_global_distrib_utils:resolve_endpoints not return all ip addresses? Why it prefers ipv6?

kanes115:
Now it resolves ok:
```
(ejabberd2@localhost)1> mod_global_distrib_utils:resolve_endpoints([{"www.google.pl", 80}]).
[{{10752,5200,16411,2052,0,0,0,8195},80},
 {{172,217,16,3},80}]
```

kanes115:
@arcusfelis asked about the status - so I'm pinging you here.
@fenek, just to make sure, this PR's status is `waiting-for-review`. I've made corrections and Igor has checked if it works for him (and it does). You can take a look and unless something is wrong - merge it.

src/global_distrib/mod_global_distrib_utils.erl
```diff
@@ -244,12 +250,25 @@ translate_opt(Opt) ->
...
-        {ok, Av6s} -> {ok, Av6s};
-        {error, Reasonv6} ->
-            case inet:getaddrs(Addr, inet) of
-                {ok, Av4s} -> {ok, Av4s};
-                {error, Reasonv4} -> {error, {Reasonv6, Reasonv4}}
-            end
+    case {inet:getaddrs(Addr, inet6), inet:getaddrs(Addr, inet)} of
+        {{error, Reason6}, {error, Reason4}} ->
+            {error, {Reason6, Reason4}};
+        {Addrs, {error, _}} ->
```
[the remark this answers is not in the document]
kanes115: And that's how it was done before as well. I have not changed this way of handling errors here. I suspect that we simply want all working IPs and only return error when both fail (because we lack connection now). 

We don't print error further on so as long as there are any IPs it is fine - we can connect. But as I said it is not even my solution, it was here before, this commit is not modifying it.

If this should be changed - I think it is a seperate issue. We would need to think about a certain way of informing user that errors appeared when fetching one version of IP address and still give him a connection with IPs from the other version.

I hope that makes sense, @arcusfelis :)

src/global_distrib/mod_global_distrib_server_mgr.erl
```diff
@@ -297,7 +296,9 @@ get_endpoints(Server) ->
         undefined -> mod_global_distrib_mapping:endpoints(Server);
         Endpoints -> {ok, Endpoints}
     end,
-    {ok, mod_global_distrib_utils:resolve_endpoints(EndpointsToResolve)}.
+    Resolved = mod_global_distrib_utils:resolve_endpoints(EndpointsToResolve),
+    ?ERROR_MSG("To resolve: ~p, resolved to: ~p", [EndpointsToResolve, Resolved]),
```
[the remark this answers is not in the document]
kanes115: Outdated comment

src/global_distrib/mod_global_distrib_mapping_redis.erl
```diff
@@ -279,15 +285,23 @@ set_endpoints(Endpoints) ->
...
+endpoint_to_binary({IpAddr, Port}) when is_tuple(IpAddr) ->
+    iolist_to_binary([inet:ntoa(IpAddr), "#", integer_to_binary(Port)]);
+endpoint_to_binary({Domain, Port}) when is_list(Domain) ->
+    iolist_to_binary([list_to_binary(Domain), "#", integer_to_binary(Port)]).
 
 -spec binary_to_endpoint(binary()) -> mod_global_distrib_utils:endpoint().
 binary_to_endpoint(Bin) ->
     [Addr, BinPort] = binary:split(Bin, <<"#">>),
-    {ok, IpAddr} = inet:parse_address(binary_to_list(Addr)),
+    IpAddrOrDomain = case mod_global_distrib_utils:is_domain(Addr) of
```
fenek: `is_domain` does exactly the same thing as line 298. Can we move it here and match on the result?

doc/modules/mod_global_distrib.md
```diff
@@ -121,6 +121,7 @@ Global distribution modules expose several per-datacenter metrics that can be us
 * **endpoints** (list, default: `[{LocalHost, 5555}]`): A list of `{Host, Port}` tuples on which the server will listen for connections.
   `Host` can be given as a hostname, in which case it will be resolved to an IP address before first on module start.
   The endpoint list will be shared with other datacenters via the replicated backend.
+* **advertised_endpoints** (list | false, default: false): A list of `{Host, Port}` tuples which will be advertised in Redis and therefore used to establish connection with this node by other nodes. If not specified, `endpoints` value (after resolution) is considered advertised_endpoints. Host may be either IP or domain, just like in case of endpoints. The difference is, domain name won't be resolved but inserted directly to mappings backend instead.
```
goddammit: **advertised_endpoints** (list | false, default: false): A list of `{Host, Port}` tuples which will be advertised in Redis and therefore used to establish connection with this node by other nodes. 
If not specified, `endpoints` value (after resolution) is considered advertised_endpoints. 
The host may be either an IP address or a domain, just like in case of endpoints. (?)
The difference is, the domain name won't be resolved but inserted directly to the mappings backend instead.

kanes115: What's wrong here, @goddammit?

fenek: **The** host may be either (...)
The difference is, **the** domain name won't be resolved but inserted directly to **the** mappings (...)

doc/modules/mod_global_distrib.md
```diff
@@ -121,6 +121,7 @@ Global distribution modules expose several per-datacenter metrics that can be us
 * **endpoints** (list, default: `[{LocalHost, 5555}]`): A list of `{Host, Port}` tuples on which the server will listen for connections.
   `Host` can be given as a hostname, in which case it will be resolved to an IP address before first on module start.
   The endpoint list will be shared with other datacenters via the replicated backend.
+* **advertised_endpoints** (list | false, default: false): A list of `{Host, Port}` tuples which will be advertised in Redis and therefore used to establish connection with this node by other nodes. If not specified, `endpoints` value (after resolution) is considered `advertised_endpoints`. The host may be either IP or domain, just like in case of endpoints. The difference is, the domain name won't be resolved but inserted directly to the mappings backend instead.
```
arcusfelis: This would go to a separate config almost immediately.
I would probably have an ejabberd_local.cfg, defining a macro with adverised_endpoints value and than this local.cfg is included from the main ejabberd.cfg.
And than we use a macro from ejabberd.cfg.

(a bit offtopic, but can be a good practice to reduce number of configs)

src/global_distrib/mod_global_distrib_mapping.erl
```diff
@@ -298,3 +302,14 @@ do_lookup_jid({_, _, _} = Jid) ->
                 BareJid -> ets_cache:lookup(?JID_TAB, BinJid, LookupInDB(jid:to_binary(BareJid)))
             end
     end.
+
+-spec get_advertised_endpoints(Opts :: [{Key :: term(), Value :: term()}]) -> [endpoint()].
+get_advertised_endpoints(Opts) ->
```
arcusfelis: get_endpoints/1 returns `{ok, [endpoints()]}`.
But this get_adverised_endpoints/1 returns `[endpoint()]` or false (not in the spec, expect warning from dialyzer).

I would like to be more consistent, even if the first argument has different type.

 I am feeling like adding gen_mod:opts() type is a good idea, instead of `[{Key :: term(), Value :: term()}]`.

src/global_distrib/mod_global_distrib_mapping_redis.erl
```diff
@@ -279,15 +285,17 @@ set_endpoints(Endpoints) ->
     refresh_set(EndpointsKey, BinEndpoints).
 
 -spec endpoint_to_binary(mod_global_distrib_utils:endpoint()) -> binary().
```
arcusfelis: Why not `IP:PORT` format instead of `IP#PORT`, which is almost standard way to specify addresses?

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -854,6 +887,58 @@ redis_query(Node, Query) ->
...
+    execute_on_each_node(code, load_binary, [?MODULE, Filename, Binary]).
+
+mock_inet_on_each_node() ->
+    Nodes = lists:map(fun({NodeName, _, _}) -> ct:get_config(NodeName) end, get_hosts()),
+    lists:map(fun(Node) -> rpc:block_call(Node, ?MODULE, mock_inet, []) end, Nodes).
+
+execute_on_each_node(M, F, A) ->
+    lists:map(fun({NodeName, _, _}) -> rpc(NodeName, M, F, A) end, get_hosts()).
+
+mock_inet() ->
```
arcusfelis: Can we have gd_inet in mongoose, which we would mock? With a bit of docs there, why it's done.
That's way we would ensure, that if meck breaks, it would break just GD, not undefined amount of other features.

arcusfelis: Or you can pass inet module as a parameter for mod_global_distribution.
Something like `{resolv_module, inet}` (by default). That's way you can make a separate module `inet_localhost_only` and not use meck at all.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -854,6 +887,58 @@ redis_query(Node, Query) ->
...
+iptuples_to_string([E | Endps]) ->
+    [E | iptuples_to_string(Endps)].
+
+maybe_add_advertised_endpoints(NodeName, Opts, Config) ->
+    Endpoints = proplists:get_value(NodeName, ?config(add_advertised_endpoints, Config), []),
+    case Endpoints of
+        [] ->
+            Opts;
+        E ->
+            Connections = case lists:keyfind(connections, 1, Opts) of
```
arcusfelis: Very non-uniform syntax of clauses (sorry).

arcusfelis: fine with
```
false -> [];
C -> C
```

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -854,6 +887,58 @@ redis_query(Node, Query) ->
...
+            [NewConnections | Opts]
+    end.
+
+load_suite_module_on_nodes() ->
+    {_Module, Binary, Filename} = code:get_object_code(?MODULE),
+    execute_on_each_node(code, load_binary, [?MODULE, Filename, Binary]).
+
+mock_inet_on_each_node() ->
+    Nodes = lists:map(fun({NodeName, _, _}) -> ct:get_config(NodeName) end, get_hosts()),
+    lists:map(fun(Node) -> rpc:block_call(Node, ?MODULE, mock_inet, []) end, Nodes).
```
arcusfelis: escalus_ct:rpc_call/6 should probably used here. It would report an error.
We should check rpc call result at least.
Not sure, if we need a block call here, spawn should be enough.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -169,6 +182,10 @@ init_per_group(_, Config0) ->
     NodesKey = rpc(SomeNode, mod_global_distrib_mapping_redis, nodes_key, []),
     [{nodes_key, NodesKey}, {escalus_user_db, xmpp} | Config2].
 
+end_per_group(advertised_endpoints, Config) ->
+    Pids = ?config(meck_handlers, Config),
+    unmock_inet(Pids),
+        Config;
```
arcusfelis: indention

src/global_distrib/mod_global_distrib_mapping.erl
kanes115: So, for now, I will make `Opts` a `list()` as it's done in similar situations (ie. in [mod_mam](https://github.com/esl/MongooseIM/blob/master/src/mod_mam.erl#L178)). We can think about introducing such a type but as it requires changing this type in the whole src/ I'll leave this job for now. How about that?

kanes115: I also weren't thinking about consistency of the result of this function cause it's not even exported. I'll change it though :)

kanes115: Are you ok with just changing Opts type to `list()` for now?

src/global_distrib/mod_global_distrib_mapping_redis.erl
kanes115: Good question! But wrong place to pose it! :D We could change it but this PR does not try to do it.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
kanes115: I have actually changed from "spawn-like" approach to "block_call-like" approach in result of a comment on this PR. So which one should be eventually used? @arcusfelis, @fenek?

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
arcusfelis: Tests should not require root access.
So, we can't change `/etc/hosts`.

Actually, there is one exception: I've added `muc.localhost` on travis to be able to run s2s+muc tests. But the same test runs just fine on my local machine without changing hosts, so it's most likely broken resolve function on travis.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
arcusfelis: ok, I see reasoning behind block call.
Be aware, that if meck does not return in the call, we have a broken rpc calls.
Meck would also most likely link to the rpc process, which can cause issues. The maximum I can think of is crashing rpc process, that would be restarted. Can be fine with tests, but not OK for the server code ;)

So, you can keep it as it is for now.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
arcusfelis: 2nd is a general way of "Dependency injection". 
You can even specify just resolv_function.
I don't see a problem to have an extra argument. Just document the argument as for internal usage, can removed anytime.

src/global_distrib/mod_global_distrib_mapping_redis.erl
arcusfelis: resolved for this PR. will discuss and create a redmine task.

doc/modules/mod_global_distrib.md
arcusfelis: resolved, can go into best practices guide, if we would have one.

src/global_distrib/mod_global_distrib_utils.erl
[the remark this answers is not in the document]
arcusfelis: resolved

src/global_distrib/mod_global_distrib_mapping_redis.erl
fenek: Simple answer: With `:` it would be less straightforward to parse binary with IPv6+Port. With `#` it is as easy as it can be. :)

kzemek: Because a serialized representation of `IP#PORT` requires much less deserialization work when IPv6 is a factor.

Anyway, this is an internal representation of data that is human readable mostly because I fancied it rather than anything else. Does this really warrant a task? Why?

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
kzemek: > In the future we are, apparently, going to get rid of all mecks

That's a contested opinion

src/global_distrib/mod_global_distrib_utils.erl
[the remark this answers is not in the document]
kzemek: @kanes115 Actually, the original version did not fetch "all working IPs":

src/global_distrib/mod_global_distrib_server_mgr.erl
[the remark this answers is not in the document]
kzemek: I can still see it though

src/global_distrib/mod_global_distrib_mapping_redis.erl
fenek: @kzemek We need a tech meeting to discuss whether we want the separator to be 😂 or 😍

kanes115:
@arcusfelis, I have implemented a version not using meck. I have mixed feelings about this as it:
* makes the test less black-box: we now actually start the module with other option (so it's not the same program flow as just starting it with default values). If somebody changes the default `{inet, getaddrs}`, we will not cover this with our test and a developer who commited such a change won't be warned.
* mocks as well, it just now does this explicitly and meck did it implicitly which is the way things are mocked in general (again keeping things more black-boxed)

While I don't see any clear benefits of this implementation.

kzemek:
@kanes115 -1 for meck change for the sake of change

fenek:
Reverted meck removal. We'll think of better replacement as a separate task.

src/global_distrib/mod_global_distrib_mapping.erl
```diff
@@ -298,3 +302,14 @@ do_lookup_jid({_, _, _} = Jid) ->
                 BareJid -> ets_cache:lookup(?JID_TAB, BinJid, LookupInDB(jid:to_binary(BareJid)))
             end
     end.
+
+-spec get_advertised_endpoints(Opts :: list()) -> [endpoint()].
+get_advertised_endpoints(Opts) ->
+    Conns = proplists:get_value(connections, Opts, []),
+    AdvEndps = proplists:get_value(advertised_endpoints, Conns),
+    case AdvEndps of
+        undefined ->
```
kzemek: Probably easier to just return `proplists:get_value(advertised_endpoints, Conns, false)`

src/global_distrib/mod_global_distrib_mapping_redis.erl
```diff
@@ -279,15 +285,17 @@ set_endpoints(Endpoints) ->
...
-endpoint_to_binary({IpAddr, Port}) ->
-    iolist_to_binary([inet:ntoa(IpAddr), "#", integer_to_binary(Port)]).
+endpoint_to_binary({IpAddr, Port}) when is_tuple(IpAddr) ->
+    iolist_to_binary([inet:ntoa(IpAddr), "#", integer_to_binary(Port)]);
+endpoint_to_binary({Domain, Port}) when is_list(Domain) ->
+    iolist_to_binary([list_to_binary(Domain), "#", integer_to_binary(Port)]).
```
kzemek: Since you already use `iolist_to_binary` here, you can drop `list_to_binary`.

Or, since domains might have unicode characters, replace `list_to_binary` with `unicode:characters_to_binary`

src/global_distrib/mod_global_distrib_utils.erl
```diff
@@ -155,18 +157,22 @@ create_ets(Name, Type) ->
...
-      end,
-      Endpoints).
+    lists:foldl(fun resolve_endpoint/2, [], Endpoints).
+
+resolve_endpoint({Addr, _Port} = E, Endpoints) when is_tuple(Addr) ->
+    [E | Endpoints];
+resolve_endpoint({Addr, Port}, Endpoints) ->
+    case to_ip_tuples(Addr) of
+        {ok, IpAddrs} ->
+            Resolved = Endpoints ++ [{IpAddr, Port} || IpAddr <- IpAddrs],
```
kzemek: This is pretty much precisely why I `flatmap`ed it before. I think it would do well for this code too to replace `lists:foldl` with `lists:flatmap` in `resolve_endpoints/1`

src/global_distrib/mod_global_distrib_utils.erl
```diff
@@ -155,18 +157,22 @@ create_ets(Name, Type) ->
...
+            ?INFO_MSG_IF(is_domain(Addr), "Domain ~p resolved to: ~p", [Addr, IpAddrs]),
+            Resolved;
+        {error, {Reasonv6, Reasonv4}} ->
+            ?ERROR_MSG("Cannot convert ~p to IP address: IPv6: ~s. IPv4: ~s.",
+                       [Addr, inet:format_error(Reasonv6), inet:format_error(Reasonv4)]),
+            error({Reasonv6, Reasonv4})
```
kzemek: We have the reasons but we don't know what error it is - add an atom describing what's going on.

src/global_distrib/mod_global_distrib_utils.erl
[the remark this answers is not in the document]
kzemek: I propose to at least add a debug log for this case.

src/global_distrib/mod_global_distrib_utils.erl
```diff
@@ -155,18 +157,22 @@ create_ets(Name, Type) ->
...
-      Endpoints).
+    lists:foldl(fun resolve_endpoint/2, [], Endpoints).
+
+resolve_endpoint({Addr, _Port} = E, Endpoints) when is_tuple(Addr) ->
+    [E | Endpoints];
+resolve_endpoint({Addr, Port}, Endpoints) ->
+    case to_ip_tuples(Addr) of
+        {ok, IpAddrs} ->
+            Resolved = Endpoints ++ [{IpAddr, Port} || IpAddr <- IpAddrs],
+            ?INFO_MSG_IF(is_domain(Addr), "Domain ~p resolved to: ~p", [Addr, IpAddrs]),
```
kzemek: Will this message ever trigger? `Addr` cannot be `{domain, _}` because it's used as an argument for `to_ip_tuples/1` which takes a string or an IP tuple.

kanes115: `Addr` doesn't have to be of the form `{domain, _}` for `is_domain/1` to return true.

src/global_distrib/mod_global_distrib_mapping_redis.erl
arcusfelis: Why not `{"ip": "127.0.0.1", "port": 6666}` format? :trollface:

kzemek: That's hard to parse too

arcusfelis: Agree

```erlang
jiffy:decode(<<"{\"ip\": \"127.0.0.1\", \"port\": 6666}">>, [return_maps]).
#{<<"ip">> => <<"127.0.0.1">>,<<"port">> => 6666}
```

fenek:
Waiting for final confirmation from one of GD users.

PR #1725, opened 2018-02-14
`mod_mam` custom backends

sstrigler:
This patch allows to have custom backends for `mod_mam` that handle archive-IDs on their own when those are encoded as UUIDs canonical textual representation (timed UUIDs).

Also fixes a bug where mod_mam would just crash when given an archive-id that is not parseable as an integer.

Also allows backends to set more meaningful error reporting (eg: 'item-not-found' etc).

### DIFF
--- a/big_tests/tests/mam_SUITE.erl
+++ b/big_tests/tests/mam_SUITE.erl
@@ -2052,8 +2052,8 @@ range_archive_request_not_empty(Config) ->
         [_, _, StartMsg, StopMsg | _] = Msgs,
         {{StartMsgId, _}, _, _, _, _StartMsgPacket} = StartMsg,
         {{StopMsgId, _}, _, _, _, _StopMsgPacket} = StopMsg,
-        {StartMicro, _} = rpc_apply(mod_mam_utils, decode_compact_uuid, [StartMsgId]),
-        {StopMicro, _} = rpc_apply(mod_mam_utils, decode_compact_uuid, [StopMsgId]),
+        StartMicro = rpc_apply(mod_mam_utils, maybe_message_id_to_timestamp, [StartMsgId]),
+        StopMicro = rpc_apply(mod_mam_utils, maybe_message_id_to_timestamp, [StopMsgId]),
         StartTime = make_iso_time(StartMicro),
         StopTime = make_iso_time(StopMicro),
         %% Send
--- a/big_tests/tests/mam_helper.erl
+++ b/big_tests/tests/mam_helper.erl
@@ -980,8 +980,8 @@ generate_msg_for_date_user(Owner, {RemoteBin, _, _} = Remote, DateTime, Content)
     MicrosecDateTime = datetime_to_microseconds(DateTime),
     NowMicro = rpc_apply(mod_mam_utils, now_to_microseconds, [rpc_apply(erlang, now, [])]),
     Microsec = min(NowMicro, MicrosecDateTime),
-    MsgIdOwner = rpc_apply(mod_mam_utils, encode_compact_uuid, [Microsec, random:uniform(20)]),
-    MsgIdRemote = rpc_apply(mod_mam_utils, encode_compact_uuid, [Microsec+1, random:uniform(20)]),
+    MsgIdOwner = rpc_apply(mod_mam_utils, maybe_timestamp_to_message_id, [Microsec]),
+    MsgIdRemote = rpc_apply(mod_mam_utils, maybe_timestamp_to_message_id, [Microsec+1]),
     Packet = escalus_stanza:chat_to(RemoteBin, Content),
     {{MsgIdOwner, MsgIdRemote}, Owner, Remote, Owner, Packet}.
 
--- a/big_tests/tests/mam_helper.erl
+++ b/big_tests/tests/mam_helper.erl
@@ -980,8 +980,8 @@ generate_msg_for_date_user(Owner, {RemoteBin, _, _} = Remote, DateTime, Content)
     MicrosecDateTime = datetime_to_microseconds(DateTime),
     NowMicro = rpc_apply(mod_mam_utils, now_to_microseconds, [rpc_apply(erlang, now, [])]),
     Microsec = min(NowMicro, MicrosecDateTime),
-    MsgIdOwner = rpc_apply(mod_mam_utils, maybe_timestamp_to_message_id, [Microsec]),
-    MsgIdRemote = rpc_apply(mod_mam_utils, maybe_timestamp_to_message_id, [Microsec+1]),
+    MsgIdOwner = rpc_apply(mod_mam_utils, maybe_timestamp_to_message_id, [Microsec, min]),
+    MsgIdRemote = rpc_apply(mod_mam_utils, maybe_timestamp_to_message_id, [Microsec+1, max]),
     Packet = escalus_stanza:chat_to(RemoteBin, Content),
     {{MsgIdOwner, MsgIdRemote}, Owner, Remote, Owner, Packet}.
 
--- a/doc/developers-guide/mod_mam_custom_backend.md
+++ b/doc/developers-guide/mod_mam_custom_backend.md
@@ -1,0 +1,83 @@
+# `mod_mam` Custom Backend
+
+You can implement a custom backend for mod_mam in order to support
+ additional databases or your own proprietary service.
+
+To configure MongooseIM to use your custom module make it have its own
+ toplevel place within the modules section of your configuration file,
+ eg:
+
+``` erlang
+{modules, [
+  ...
+  {mod_mam_custom_arch, []},
+  {mod_mam, [...]},
+  ...
+]}
+```
+
+This module needs to implement the `gen_mod` behavior and typically
+needs to add handlers for a (subset of) hooks triggered by mod_mam:
+
+ * mam_archive_message
+ * mam_archive_size
+ * mam_lookup_messages
+ * mam_remove_archive
+ * mam_purge_single_message
+ * mam_purge_multiple_messages
+
+See official backend implementations (like `mod_mam_odbc_arch.erl`) to get an idea what their
+signature looks like and what they are meant to do.
+
+A minimalistic example that only implements message lookup (without
+the actual data access):
+
+``` erlang
+start(Host, _Opts) ->
+  ...
+  ejabberd_hooks:add(mam_lookup_messages, Host, ?MODULE, lookup_messages, 50),
+  ...
+  ok.
+
+...
+
+-spec lookup_messages(any(), ejabberd:server(), map()) -> {ok, mod_mam:lookup_result()} |
+                                                          {error, 'policy-violation'} |
+                                                          {error, any()}.
+lookup_messages({error, _Reason} = Result, _Host, _Params) -> Result;
+lookup_messages(_Result, Host, #{owner_jid        := OwnerJid,
+                                 rsm              := RSM,
+                                 start_ts         := undefined,
+                                 end_ts           := undefined,
+                                 with_jid         := undefined,
+                                 search_text      := undefined,
+                                 page_size        := PageSize,
+                                 limit_passed     := _LimitPassed,
+                                 max_result_limit := _MaxResultLimit,
+                                 is_simple        := _IsSimple      }) ->
+    %% Your implementation goes here
+    ...
+    ;
+lookup_messages(_Result, _Host, Params) ->
+    lager:info("unsupported params: ~p", [Params]),
+    {error, 'not-supported'}.
+
+...
+
+```
+
+If this module uses a different format for its archive-ids other than
+what's in `mam_uid_nodetime` you can configure a custom `uid_module`:
+
+``` erlang
+{modules, [
+  ...
+  {mod_mam, [{uid_module, mam_uid_custom}]},
+  ...
+]}
+```
+
+Your custom UID module needs to implement the `mam_uid` behaviour.
+
+There's `mam_uid_uuid` provided for convenience, it implements time base UUIDs.
+See https://www.famkruithof.net/guid-uuid-timebased.html for details.
--- a/src/mam/mam_uid.erl
+++ b/src/mam/mam_uid.erl
@@ -4,8 +4,8 @@
 %% This module is global, so it works for both mod_mam and mod_mam_muc
 -module(mam_uid).
 
--callback generate_message_id() -> integer().
--callback encode_compact_uuid(integer(), integer()) -> integer().
--callback decode_compact_uuid(integer()) -> {integer(),byte()}.
--callback mess_id_to_external_binary(integer()) -> binary().
--callback external_binary_to_mess_id(binary()) -> integer().
+-callback generate_message_id() -> any().
+-callback encode_compact_uuid(integer(), byte()) -> any().
+-callback decode_compact_uuid(any()) -> {integer(), byte()}.
+-callback mess_id_to_external_binary(any()) -> binary().
+-callback external_binary_to_mess_id(binary()) -> any().
--- a/src/mam/mam_uid.erl
+++ b/src/mam/mam_uid.erl
@@ -1,0 +1,11 @@
+%% @doc MAM message unique ID
+%%
+%% How to generate or represent message ids
+%% This module is global, so it works for both mod_mam and mod_mam_muc
+-module(mam_uid).
+
+-callback generate_message_id() -> integer().
+-callback encode_compact_uuid(integer(), integer()) -> integer().
+-callback decode_compact_uuid(integer()) -> {integer(),byte()}.
+-callback mess_id_to_external_binary(integer()) -> binary().
+-callback external_binary_to_mess_id(binary()) -> integer().
--- a/src/mam/mam_uid.erl
+++ b/src/mam/mam_uid.erl
@@ -11,4 +11,4 @@
 -callback mess_id_to_external_binary(any()) -> binary().
 -callback external_binary_to_mess_id(binary()) -> any().
 -callback message_id_to_timestamp(any()) -> integer().
--callback timestamp_to_message_id(integer()) -> any().
+-callback timestamp_to_message_id(integer(), atom()) -> any().
--- a/src/mam/mam_uid.erl
+++ b/src/mam/mam_uid.erl
@@ -1,11 +1,14 @@
 %% @doc MAM message unique ID
 %%
-%% How to generate or represent message ids
-%% This module is global, so it works for both mod_mam and mod_mam_muc
+%% This behaviour defines the way message IDs are generated and transformed into timestamps or vice
+%% versa. It is assumed that generated message IDs always hold the timestamp of their time of
+%% generation so that they can be compared with timestamps by mod_mam and database backends.
+%%
+%% This module is global, so it works for both mod_mam and mod_mam_muc.
 -module(mam_uid).
 
 -callback generate_message_id() -> any().
--callback encode_compact_uuid(integer(), byte()) -> any().
--callback decode_compact_uuid(any()) -> {integer(), byte()}.
 -callback mess_id_to_external_binary(any()) -> binary().
 -callback external_binary_to_mess_id(binary()) -> any().
+-callback message_id_to_timestamp(any()) -> integer().
+-callback timestamp_to_message_id(integer()) -> any().
--- a/src/mam/mam_uid_nodetime.erl
+++ b/src/mam/mam_uid_nodetime.erl
@@ -29,13 +29,13 @@ generate_message_id() ->
 %% It removes a leading 0 from 64-bit binary representation.
 %% It puts node id as a last byte.
 %% The maximum date, that can be encoded is `{{4253,5,31},{22,20,37}}'.
--spec encode_compact_uuid(integer(), integer()) -> integer().
+-spec encode_compact_uuid(integer(), byte()) -> integer().
 encode_compact_uuid(Microseconds, NodeId)
     when is_integer(Microseconds), is_integer(NodeId) ->
     (Microseconds bsl 8) + NodeId.
 
 %% @doc Extract date and node id from a message id.
--spec decode_compact_uuid(integer()) -> {integer(),byte()}.
+-spec decode_compact_uuid(integer()) -> {integer(), byte()}.
 decode_compact_uuid(Id) ->
     Microseconds = Id bsr 8,
     NodeId = Id band 255,
--- a/src/mam/mam_uid_nodetime.erl
+++ b/src/mam/mam_uid_nodetime.erl
@@ -11,7 +11,7 @@
          mess_id_to_external_binary/1,
          external_binary_to_mess_id/1,
          message_id_to_timestamp/1,
-         timestamp_to_message_id/1]).
+         timestamp_to_message_id/2]).
 
 %% -----------------------------------------------------------------------
 %% UID
@@ -43,6 +43,8 @@ message_id_to_timestamp(MessID) ->
 
 
 %% @doc Transform a timestamp to a message ID
--spec timestamp_to_message_id(mod_mam:posix_timestamp()) -> integer().
-timestamp_to_message_id(Microseconds) ->
-    Microseconds bsl 8.
+-spec timestamp_to_message_id(mod_mam:posix_timestamp(), atom()) -> integer().
+timestamp_to_message_id(Microseconds, min) ->
+    Microseconds bsl 8;
+timestamp_to_message_id(Microseconds, max) ->
+    (Microseconds bsl 8) + 255.
--- a/src/mam/mam_uid_nodetime.erl
+++ b/src/mam/mam_uid_nodetime.erl
@@ -1,0 +1,54 @@
+%% @doc Message id based on node_id and unique timestamp
+%%
+%% Message ids unique unless nodes get restarted
+%% We use HEX encoding for external ids
+%% Ids are not transarent for the clients (they can't get timestamp from id)
+-module(mam_uid_nodetime).
+-behaviour(mam_uid).
+
+%% UID
+-export([generate_message_id/0,
+         encode_compact_uuid/2,
+         decode_compact_uuid/1,
+         mess_id_to_external_binary/1,
+         external_binary_to_mess_id/1]).
+
+%% -----------------------------------------------------------------------
+%% UID
+
+-spec generate_message_id() -> integer().
+generate_message_id() ->
+    {ok, NodeId} = ejabberd_node_id:node_id(),
+    CandidateStamp = p1_time_compat:os_system_time(micro_seconds),
+    UniqueStamp = mongoose_mam_id:next_unique(CandidateStamp),
+    encode_compact_uuid(UniqueStamp, NodeId).
+
+
+%% @doc Create a message ID (UID).
+%%
+%% It removes a leading 0 from 64-bit binary representation.
+%% It puts node id as a last byte.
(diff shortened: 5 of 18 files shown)


src/mod_mam_utils.erl
```diff
@@ -207,16 +207,31 @@ encode_compact_uuid(Microseconds, NodeId)
 
 %% @doc Extract date and node id from a message id.
 -spec decode_compact_uuid(integer()) -> {integer(), byte()}.
-decode_compact_uuid(Id) ->
+decode_compact_uuid(Id) when is_integer(Id) ->
```
arcusfelis: For one of our client we patched this to use timestamps as uuids.
Or anything else. The point was to introduce mam_uid, that would encode from/to compact uids.

```diff
commit 2873535835e4c1f2385253516e389cdae9f2d4b4
Author: Michael Uvarov <arcusfelis@gmail.com>
Date:   Fri Nov 25 14:15:11 2016 +0100

    Separate message id logic into mam_uid

diff --git a/apps/ejabberd/src/mam_uid.erl b/apps/ejabberd/src/mam_uid.erl
new file mode 100644
index 000000000..968f9f157
--- /dev/null
+++ b/apps/ejabberd/src/mam_uid.erl
@@ -0,0 +1,11 @@
+%% @doc MAM message unique ID
+%%
+%% How to generate or represent message ids
+%% This module is global, so it works for both mod_mam and mod_mam_muc
+-module(mam_uid).
+
+-callback generate_message_id() -> integer().
+-callback encode_compact_uuid(integer(), integer()) -> integer().
+-callback decode_compact_uuid(integer()) -> {integer(),byte()}.
+-callback mess_id_to_external_binary(integer()) -> binary().
+-callback external_binary_to_mess_id(binary()) -> integer().

diff --git a/apps/ejabberd/src/mam_uid_nodetime.erl b/apps/ejabberd/src/mam_uid_nodetime.erl
new file mode 100644
index 000000000..f55458d7e
--- /dev/null
+++ b/apps/ejabberd/src/mam_uid_nodetime.erl
@@ -0,0 +1,54 @@
+%% @doc Message id based on node_id and unique timestamp
+%%
+%% Message ids unique unless nodes get restarted
+%% We use HEX encoding for external ids
+%% Ids are not transarent for the clients (they can't get timestamp from id)
+-module(mam_uid_nodetime).
+-behaviour(mam_uid).
+
+%% UID
+-export([generate_message_id/0,
+         encode_compact_uuid/2,
+         decode_compact_uuid/1,
+         mess_id_to_external_binary/1,
+         external_binary_to_mess_id/1]).
+
+%% -----------------------------------------------------------------------
+%% UID
+
+-spec generate_message_id() -> integer().
+generate_message_id() ->
+    {ok, NodeId} = ejabberd_node_id:node_id(),
+    %% Use monotone function here.
+    encode_compact_uuid(mod_mam_utils:now_to_microseconds(now()), NodeId).
+
+
+%% @doc Create a message ID (UID).
+%%
+%% It removes a leading 0 from 64-bit binary representation.
+%% It puts node id as a last byte.
+%% The maximum date, that can be encoded is `{{4253,5,31},{22,20,37}}'.
+-spec encode_compact_uuid(integer(), integer()) -> integer().
+encode_compact_uuid(Microseconds, NodeId)
+    when is_integer(Microseconds), is_integer(NodeId) ->
+    (Microseconds bsl 8) + NodeId.
+
+%% @doc Extract date and node id from a message id.
+-spec decode_compact_uuid(integer()) -> {integer(),byte()}.
+decode_compact_uuid(Id) ->
+    Microseconds = Id bsr 8,
+    NodeId = Id band 255,
+    {Microseconds, NodeId}.
+
+
+%% @doc Encode a message ID to pass it to the user.
+-spec mess_id_to_external_binary(integer()) -> binary().
+mess_id_to_external_binary(MessID) when is_integer(MessID) ->
+    list_to_binary(integer_to_list(MessID, 32)).
+
+
+%% @doc Decode a message ID received from the user.
+-spec external_binary_to_mess_id(binary()) -> integer().
+external_binary_to_mess_id(BExtMessID) when is_binary(BExtMessID) ->
+    list_to_integer(binary_to_list(BExtMessID), 32).
diff --git a/apps/ejabberd/src/mod_mam.erl b/apps/ejabberd/src/mod_mam.erl
index e889b86f9..bc857675e 100644
--- a/apps/ejabberd/src/mod_mam.erl
+++ b/apps/ejabberd/src/mod_mam.erl
@@ -59,6 +59,8 @@
 
 %% for feature (escalus) tests
 -export([set_params/1]).
+%% Called from mod_mam_utils
+-export([uid_module/0]).
 
 %% ----------------------------------------------------------------------
 %% Imports
@@ -990,11 +992,13 @@ params_helper(Params) ->
         "is_complete_message() -> ~p.~n"
         "default_result_limit() -> ~p.~n"
         "max_result_limit() -> ~p.~n"
+        "uid_module() -> ~p.~n"
         "params() -> ~p.~n",
         [proplists:get_bool(add_archived_element, Params),
          proplists:get_value(is_complete_message, Params, mod_mam_utils),
          proplists:get_value(default_result_limit, Params, 50),
          proplists:get_value(max_result_limit, Params, 50),
+         proplists:get_value(uid_module, Params, mam_uid_nodetime),
          Params
         ]))).
 
@@ -1009,3 +1013,9 @@ add_archived_element() ->
 is_complete_message(Module, Dir, Packet) ->
     M = mod_mam_params:is_complete_message(),
     M:is_complete_message(Module, Dir, Packet).
+
+uid_module() ->
+    %% Ensure that it's still works in tests
+    try mod_mam_params:uid_module()
+    catch _:_ -> mam_uid_nodetime
+    end.


diff --git a/apps/ejabberd/src/mod_mam_utils.erl b/apps/ejabberd/src/mod_mam_utils.erl
index d5263aab9..faa392995 100644
--- a/apps/ejabberd/src/mod_mam_utils.erl
+++ b/apps/ejabberd/src/mod_mam_utils.erl
@@ -163,10 +163,8 @@ microseconds_to_datetime(MicroSeconds) when is_integer(MicroSeconds) ->
 
 -spec generate_message_id() -> integer().
 generate_message_id() ->
-    {ok, NodeId} = ejabberd_node_id:node_id(),
-    %% Use monotone function here.
-    encode_compact_uuid(now_to_microseconds(now()), NodeId).
-
+    M = mod_mam:uid_module(),
+    M:generate_message_id().
 
 %% @doc Create a message ID (UID).
 %%
@@ -174,23 +172,22 @@ generate_message_id() ->
 %% It puts node id as a last byte.
 %% The maximum date, that can be encoded is `{{4253,5,31},{22,20,37}}'.
 -spec encode_compact_uuid(integer(), integer()) -> integer().
-encode_compact_uuid(Microseconds, NodeId)
-    when is_integer(Microseconds), is_integer(NodeId) ->
-    (Microseconds bsl 8) + NodeId.
-
+encode_compact_uuid(Microseconds, NodeId) ->
+    M = mod_mam:uid_module(),
+    M:encode_compact_uuid(Microseconds, NodeId).

 %% @doc Extract date and node id from a message id.
 -spec decode_compact_uuid(integer()) -> {integer(),byte()}.
 decode_compact_uuid(Id) ->
-    Microseconds = Id bsr 8,
-    NodeId = Id band 255,
-    {Microseconds, NodeId}.
+    M = mod_mam:uid_module(),
+    M:decode_compact_uuid(Id).
 
 
 %% @doc Encode a message ID to pass it to the user.
 -spec mess_id_to_external_binary(integer()) -> binary().
-mess_id_to_external_binary(MessID) when is_integer(MessID) ->
-    list_to_binary(integer_to_list(MessID, 32)).
+mess_id_to_external_binary(MessID) ->
+    M = mod_mam:uid_module(),
+    M:mess_id_to_external_binary(MessID).
 
 
 -spec maybe_external_binary_to_mess_id(binary()) -> undefined | integer().
@@ -202,8 +199,9 @@ maybe_external_binary_to_mess_id(BExtMessID) ->
 
 %% @doc Decode a message ID received from the user.
 -spec external_binary_to_mess_id(binary()) -> integer().
-external_binary_to_mess_id(BExtMessID) when is_binary(BExtMessID) ->
-    list_to_integer(binary_to_list(BExtMessID), 32).
+external_binary_to_mess_id(BExtMessID) ->
+    M = mod_mam:uid_module(),
+    M:external_binary_to_mess_id(BExtMessID).
 
 %% -----------------------------------------------------------------------
 %% XML
```

We also added file that allows to return timestamp in the same format as ejabberd does if mod_mam is configured with {uid_module, mam_uid_timestamp} option.

```erlang
%% @doc Message id based on timestamp                                                
-module(mam_uid_timestamp).                                                          
-behaviour(mam_uid).                                                                 
                                                                                     
%% UID                                                                               
-export([generate_message_id/0,                                                      
         encode_compact_uuid/2,                                                      
         decode_compact_uuid/1,                                                      
         mess_id_to_external_binary/1,                                               
         external_binary_to_mess_id/1]).                                             
                                                                                     
%% -----------------------------------------------------------------------           
%% UID                                                                               
                                                                                     
-spec generate_message_id() -> integer().                                            
generate_message_id() ->                                                             
    mod_mam_utils:now_to_microseconds(jlib:unique_timestamp()).                      
                                                                                     
                                                                                     
%% @doc Create a message ID (UID).                                                   
-spec encode_compact_uuid(integer(), integer()) -> integer().                        
encode_compact_uuid(Microseconds, NodeId)                                            
    when is_integer(Microseconds), is_integer(NodeId) ->                             
    Microseconds.                                                                    
                                                                                     
                                                                                     
%% @doc Extract date and node id from a message id.                                  
-spec decode_compact_uuid(integer()) -> {integer(),byte()}.                          
decode_compact_uuid(Id) ->                                                           
    Microseconds = Id,                                                               
    NodeId = 0,                                                                      
    {Microseconds, NodeId}. 

%% @doc Encode a message ID to pass it to the user.                                  
-spec mess_id_to_external_binary(integer()) -> binary().                             
mess_id_to_external_binary(MessID) when is_integer(MessID) ->                        
    list_to_binary(integer_to_list(MessID)).                                         
                                                                                     
                                                                                     
%% @doc Decode a message ID received from the user.                                  
-spec external_binary_to_mess_id(binary()) -> integer().                             
external_binary_to_mess_id(BExtMessID) when is_binary(BExtMessID) ->                 
    list_to_integer(binary_to_list(BExtMessID)). 
```

I know, that this patch is NOT compatible with the latest MongooseIM.
But I would love to see uid_module introduced and guid-uuid stuff introduced as a configurable option.
Because there is always more than one way to encode these uids.

src/mongoose_xmpp_errors.erl
```diff
@@ -12,6 +12,14 @@
 -compile([export_all]).
 -include("mongoose_ns.hrl").
 
+to_error(A) ->
```
arcusfelis: can we pass a map here?
Something like `#{xmpp_error => conflict}`.
Because I would expect backend modules to add an extra info.
I don't like how we pass stacktrace, for example.
Than it would be easier to extend later and to refactor that stacktrace code, while don't care about breaking this feature much.

src/mod_mam_utils.erl
sstrigler: Sorry if stupid question but why after all does the upper level need to know anything about UUIDs and convert them back and forth between external and internal representation?

Also in our scenario (we do not persist here at the moment) generate_message_id wouldn't work for us because IDs are created at some other external service (REST) when persisting a message and not before.

Apart from that I like the idea except for making configuration more complicated again. But maybe I misunderstood your approach. If not couldn't we go for some conventional naming here instead (eg: `mod_mam_cassandra_arch` -> `mod_mam_cassandra_uid`) or just make the interface part of the backend behaviour?

src/mongoose_xmpp_errors.erl
sstrigler: Use a map here at this place or rather at the handler in `mod_mam`? What other data would you `mongoose_xmpp_error:to_error/1` expect to handle and how? Error reporting/logging?

arcusfelis: Backend module should return a map to mod_mam.
One of the fields in a map can be used as a reason of error, which can be formatted by this module. We can introduce the map in a separate PR, than the code here is fine.

I don't see a point of passing the whole map here, ErrorType atom would be enough, like it's in the PR.

src/mod_mam_utils.erl
arcusfelis: We can pass stanza into generate_message_id as an argument.
Than we can extract UUID from the stanza itself.

We can return nothing from the function, if backend does not need it.

I just want to keep the current behaviour as the basic point and if some project needs to change something, it can create another module of mam_uid and keep the logic here.

To answer your question, we need to know UUID to allow async message writers to work.

About naming. I am just using the same mechanism as `mam_jid.erl`.

There is a file `mam_jid` that controls, how jids should be encoded in the database.
There are two modules:
- mam_jid_mini - something that was originally
- mam_jid_rfc - something that we decided to switch to.

The same with `mam_message` module:
- mam_message_xml
- mam_message_eterm
- mam_message_compressed_eterm

Our conversation is that only modules that should have `mod_` prefix should be modules that implement `gen_mod`. At least the new ones :)

fenek: Maybe we can introduce `mam_uid` in second iteration of this improvement, as a part of our next sprint? And stick to Stefan's implementation for iteration 1?

src/mongoose_xmpp_errors.erl
fenek: I agree that map is probably not needed here, as this function might be reused by other modules (MUC Light perhaps?)

src/mod_mam_utils.erl
sstrigler: +1 for how you want to handle 'mod_' prefixes :p 

As for me I'm fine with merging of course but I can also offer to contribute to any more sophisticated solution. Be it within the scope of this PR or any other.

That said, I'm not working full time currently and will be 'off' for two weeks now. This does not include responses here but potential work on this or any other PR.

GalaxyGorilla:
@arcusfelis @fenek As far as I see this PR is ok for you? Can you review/merge it? Then I can test the current master together with our umbrella app and hopefully give you some meaningful feedback :).

fenek:
@GalaxyGorilla I need to take a look one more time but I won't have time for it before Thursday. :(

src/mod_mam_utils.erl
arcusfelis: We usually don't have time for iterations 2 :)
I just don't want to have _more_ logic behind ids by _default_.

src/mod_mam_utils.erl
```diff
@@ -207,16 +207,31 @@ encode_compact_uuid(Microseconds, NodeId)
...
     Microseconds = Id bsr 8,
     NodeId = Id band 255,
-    {Microseconds, NodeId}.
+    {Microseconds, NodeId};
+decode_compact_uuid(UUID) when is_binary(UUID) ->
```
arcusfelis: At _least_ this should be covered with tests.

arcusfelis:
Also, we need docs how this feature can be used.

(because we already have message "purge" feature, that nobody knows how to use) :)

sstrigler:
What's the best place then where to document that?

src/mod_mam_utils.erl
sstrigler: Where should that go? Would eunit be ok or just ct?

GalaxyGorilla:
@fenek somewhere you asked if we are good with the current master. I just tested the master together with this PR and our QA system looks absolutely fine :). Would be cool if we can get this PR into the upcoming release.

src/mod_mam_utils.erl
arcusfelis: I guess into big tests mam_SUITE.
Because we already have tests for all other MAM features there.

src/mod_mam_utils.erl
```diff
@@ -207,16 +207,31 @@ encode_compact_uuid(Microseconds, NodeId)
...
+    UUIDTimestamp = binary:decode_unsigned(<<TimeHi:16, TimeMid:16, TimeLow:32>>),
+
+    %% uuid offset and nanosecends step
+    UnixTimestamp = (UUIDTimestamp - 122192928000000000) div 10,
+
+    {UnixTimestamp , NodeId band 255}.
 
 
 %% @doc Encode a message ID to pass it to the user.
 -spec mess_id_to_external_binary(integer()) -> binary().
```
arcusfelis: spec violation.

arcusfelis:
@sstrigler `doc/modules/mod_mam.md` can be a good place to put the docs.
I usually do something like:

```
# A cool new feature

How it works, how to use it, stanza examples, how to ensure that it works.
```

For bigger extentions we use `doc/open-extensions/` (i.e. for inbox, muc-light, MAM message deletion and editing would go into `open-extentions` too, if we would try to implement it).

By having description of a feature with examples, it's very easy to implement feature tests.

On another hand, without having docs or tests, the feature can be broken over time, during some kind of refactoring. Or will not work correctly with some backends. But most likely, the feature would not be used by other people, because they would not know how. Something similar we have with message purging implementation now.

I also don't expect a lot of docs :)
Something like

```
It's possible to design backend modules, that would store message ids as UUIDs.
In such case, this backend would return MAM ID in for UUID (here is some info of the UUID format).
This UUID should be timestamp-based.
```

# P.S., that you can safely ignore for this PR, but about direction we can go with MAM in MongooseIM

Later we can extend the docs, describing how MAM ID is generated. And maybe we will change some backends to just store timestamp, instead of timestamp+node_id. But it would need mam_uid module in place.

doc/modules/mod_mam.md
```diff
@@ -145,6 +145,61 @@ You can change the default settings using extra parameters:
...
+``` erlang
+{modules, [
+  ...
+  {mod_mam_custom_arch, []},
+  {mod_mam, [...]},
+  ...
+]}
+```
+
+Typically your module would implement add handlers for hooks triggered by mod_mam. See other backend implementations to get an idea.
```
GalaxyGorilla: s/implement//

sstrigler: thx, fixed

doc/modules/mod_mam.md
```diff
@@ -145,6 +145,61 @@ You can change the default settings using extra parameters:
...
+                                                          {error, any()}.
+lookup_messages({error, _Reason} = Result, _Host, _Params) -> Result;
+lookup_messages(_Result, Host, #{owner_jid        := OwnerJid,
+                                 rsm              := RSM,
+                                 start_ts         := undefined,
+                                 end_ts           := undefined,
+                                 with_jid         := undefined,
+                                 search_text      := undefined,
+                                 page_size        := PageSize,
+                                 limit_passed     := _LimitPassed,    % ?
```
arcusfelis: there is logic in MAM 0.3. (the version, that does not use forms for lookups).

- if an user provides a limit of messages in his request, than we are perfectly fine
- if an user does not provide the limit, and there are a lot of messages matching, the server CAN return `policy-violation` error (see in spec!).

So, yep, limit_passed flag just makes backends logic harder. I would rather make the logic in mod_mam/mod_mam_muc.

max_result_limit is `min(Limit_specified_by_user, Limit_specified_in_mongoose_config)`.

big_tests/tests/mam_SUITE.erl
```diff
@@ -2974,6 +2975,19 @@ check_user_exist(Config) ->
...
   ok = escalus_ejabberd:rpc(ejabberd_auth, remove_user, [AdminU, AdminS]).
 
+decode_compact_uuid(_Config) ->
+    %% ensure we can handle time based UUIDs
+    quickrand:seed(),
+    UState = uuid:new(self()),
+    {Uuid, _NewState} = uuid:get_v1(UState),
+
+    %% would crash if not...
+    mod_mam_utils:decode_compact_uuid(list_to_binary(uuid:uuid_to_string(Uuid))),
```
arcusfelis: Big tests don't have access to mongoose modules (or at least we don't do calls to mongoose modules like this).
Still, an RPC should be possible.

arcusfelis: To enable the test, it should be added into 

```erlang
impl_specific() ->                                                                   
  [check_user_exist,
   decode_compact_uuid].
```

A good test would also try to query archive, using compact uuids.

sstrigler: Yea, I'm aware it's a bit weird what I'm doing. That's why I asked for eunit tests, since I think that would be the correct approach to show this kind of functionality works as expected.

For a true system test I'd have to ship a custom backend module that would handle external UUIDs. What would be the right place to have this? Also I'd have to change the configuration of the node under test to use that custom backend. Any pointers how to do that?

big_tests/tests/mam_SUITE.erl
```diff
@@ -2974,6 +2976,24 @@ check_user_exist(Config) ->
   %% cleanup
   ok = escalus_ejabberd:rpc(ejabberd_auth, remove_user, [AdminU, AdminS]).
 
+decode_compact_uuid(_Config) ->
```
fenek: I'd suggest to create a "small" suite for MAM for this case (e.g. `[repo]/test/mam_SUITE.erl`), as it does RPC calls and does not generate any XMPP traffic.

doc/modules/mod_mam.md
```diff
@@ -145,6 +145,61 @@ You can change the default settings using extra parameters:
...
+
+Typically your module would add handlers for hooks triggered by mod_mam. See other backend implementations to get an idea.
+
+Eg.:
+
+``` erlang
+start(Host, _Opts) ->
+  ...
+  ejabberd_hooks:add(mam_lookup_messages, Host, ?MODULE, lookup_messages, 50),
+  ...
```
fenek: This description should be more precise and I'd actually move it to a new page in "Developer's Guide" and leave a reference here.

If you don't have time to create a better description, I'd appreciate at least if you could move it to a new page.

doc/modules/mod_mam.md
```diff
@@ -145,6 +145,61 @@ You can change the default settings using extra parameters:
...
+    ;
+lookup_messages(_Result, _Host, Params) ->
+    lager:info("unsupported params: ~p", [Params]),
+    {error, 'not-supported'}.
+
+...
+
+```
+
+Archive-IDs handled by your backend can either be integers as generated by `mod_mam_utils:generate_message_id/0` or a canonical textual representation of a time based UUID, eg `58e0a7d7-eebc-11d8-9669-0800200c9a66`.
```
fenek: `handled by` -> `returned by`?

big_tests/tests/mam_SUITE.erl
arcusfelis: `[repo]/test/mod_mam_SUITE.erl`.

I would love to not have name duplicates.

fenek: So actually it should be `mod_mam_utils_SUITE` :)

big_tests/tests/mam_SUITE.erl
arcusfelis: ok, as Piotr said, small tests are in test/ directory.
I am not sure if we need a system wide test here.

But if I would need to write one, that the module would go to src/ OR to big_tests/tests/ and dynamically loaded (code:load_binary). Than mam_SUITE can be used to test the backend. mam_SUITE already tests different configurations of MAM (for example, async writers).

sstrigler:
Haven't had time yet to get back to you but I will, just want to ensure you know I haven't forgotten about this PR.

sstrigler:
Something went really wrong now with the latest rebase on origin/master. Any idea what might have happened?

sstrigler:
sorry, mistake

sstrigler:
*ping*

Checks failing are no issues I could address, right?

src/mod_mam_utils.erl
```diff
@@ -234,7 +249,12 @@ maybe_external_binary_to_mess_id(BExtMessID) ->
 %% @doc Decode a message ID received from the user.
 -spec external_binary_to_mess_id(binary()) -> integer().
```
arcusfelis: spec violation.

arcusfelis:
Reopening for CI.

src/mod_mam_utils.erl
sstrigler: Ok, sorry, this is a mess. Sorry this is taking so long. Out of curiosity how did you find that? I've had a failing spec reported before but not this time. What did I overlook?

Anyway, I think I need a bit more time again to get this right. I'll introduce a type but want to think a bit to provide a clean solution this time.

arcusfelis: Dialyzer does not like to find errors in `try...catch`-es.

It's pretty easy, I really dislike functions that return more than one type. ;)
I.e. `integer() | binary()`. Because it's leads to code branching. This means potentially some bugs, specs are the easiest ones.
The next kind of bugs you would need to look at runtime is with some backend modules, expecting integer-id (ok, we have handled this by specifying the case in docs).

fenek: 1. UIDs may be tagged values and backends could report and error when they get an unsupported UID format.
2. Backend may check the configured UID provider on startup and refuse to start when it's not supported.
3. Backends may expose extra `supported_uid_formats/0` callback and `mod_mam` should verify it on startup.
4. UID providers may expose `uid_format/0` callback because there may be several providers that return a valid integer, a string etc., so the validation won't be based only on provider name.

For sake of simplicity we may stick to (3) for now and perhaps extend it with (4) if the UID provider list grows.

michalwski:
Hi @sstrigler 

Looks like now only dialyzer is not happy about the PR https://travis-ci.org/esl/MongooseIM/jobs/403521457#L1417 Could you find some time to address this?

@arcusfelis @fenek can you approve the PR now or are there still things worth changing / improving?

GalaxyGorilla:
@michalwski, I am currently working on that patch since @sstrigler is on holiday for some months. I think I fixed the dialyzer issues. However, looking through the code I think the new UUIDs don't work together with the current MAM :(. There are a lot of numerical comparisons going on within MAM, for example here:

https://github.com/esl/MongooseIM/blob/984dd6df7ee850f232e64077e3074cc6a2579fb2/src/mam/mod_mam_utils.erl#L949

Those comparisons are not possible with the binary UUIDs. I need to wrap my head around how to solve this ... the encode/decode_compact_uuid functions in `mod_mam_uid_uuid` IMHO need an overhaul.

fenek:
We'll consider a merge after Tuesday. I'd prefer to avoid such an invasive change right before the release. :)

GalaxyGorilla:
@fenek, I updated the PR in a way that time comparisons are now possible within mod_mam (even though the tests are not ok yet). For UIDs it is now always assumed that they contain timestamp information and those comparisons are now always done with those timestamps and not (as before) with the UIDs.
Could comeone give me a hint/script on how to setup one of the failing test cases from the CI? I really have a hard time to reproduce those errors :-/.

GalaxyGorilla:
So ... I spent quite a lot of time now on this PR and finally I have to say that the idea of the overall PR will not work out in a way which I consider acceptable for you as maintainers. The main problem is that the IDs are just way too much coupled to timestamps. I don't understand the necessity of that, but that's how it is. Integrating other UIDs here means to compare timestamps all the way around with those UIDs and this is just a clusterfuck when it comes to the handling of RSM elements in context of MAM. The cost-benefit ratio is just way off. The good thing is that your tests made me aware of that ;)).

IMHO you can close this, we will just keep our little fork with this patch.

fenek:
Hi @GalaxyGorilla 

Thanks for checking this. :) We may consider decoupling MAM IDs and timestamps but I don't think it would be a priority right now (especially that it might break backward compatibility).

Issue #1722, opened 2018-02-13
mod_event_pusher_push errors


Beisenbek:
MongooseIM version: 2.1.1
Installed from: source
Erlang/OTP version: 18

Getting dozens of errors with next format:

`2018-02-13 04:23:24.152 [error] <0.20442.0>@ejabberd_hooks:run_fold1:264 {function_clause,[{exml_query,cdata,[undefined],[{file,"/opt/xmpp/MongooseIM/_build/default/lib/exml/src/exml_query.erl"},{line,112}]},{mod_event_pusher_push,push_notification_iq,5,[{file,"/opt/xmpp/MongooseIM/_build/prod/lib/mongooseim/src/event_pusher/mod_event_pusher_push.erl"},{line,246}]},{mod_event_pusher_push,'-publish_message/3-fun-3-',6,[{file,"/opt/xmpp/MongooseIM/_build/prod/lib/mongooseim/src/event_pusher/mod_event_pusher_push.erl"},{line,168}]},{lists,foreach,2,[{file,"lists.erl"},{line,1337}]},{mod_event_pusher_push,publish_message,3,[{file,"/opt/xmpp/MongooseIM/_build/prod/lib/mongooseim/src/event_pusher/mod_event_pusher_push.erl"},{line,166}]},{mod_event_pusher,'-push_event/3-lc$^0/1-0-',3,[{file,"/opt/xmpp/MongooseIM/_build/prod/lib/mongooseim/src/event_pusher/mod_event_pusher.erl"},{line,40}]},{mod_event_pusher,push_event,3,[{file,"/opt/xmpp/MongooseIM/_build/prod/lib/mongooseim/src/event_pusher/mod_event_pusher.erl"},{line,40}]},{mod_event_pusher_hook_translator,filter_local_packet,1,[{file,"/opt/xmpp/MongooseIM/_build/prod/lib/mongooseim/src/event_pusher/mod_event_pusher_hook_translator.erl"},{line,53}]}]}
running hook: {filter_local_packet,[]}`

**MongooseIM has next config for push:**

` {mod_ping, [{send_pings, true},
            {ping_interval,40},
            {timeout_action,kill},
            {ping_req_timeout,20}]},
  {mod_pubsub, [
               {plugins, [<<"push">>]}
  ]},
  {mod_push_service_mongoosepush, [
      {pool_name, mongoose_push_http},
      {api_version, "v2"}
  ]},
  %{mod_push, [
  %    {backend, mnesia},
[...]

PR #1726, opened 2018-02-14
Don't try to push message without body

fenek:
This PR addresses #1722

It adds a hard check for `body` subelement in message processed by `mod_event_pusher_push`. It overrides `should_publish` result, as this element is mandatory in constructed payload.

### DIFF
--- a/src/event_pusher/mod_event_pusher_push.erl
+++ b/src/event_pusher/mod_event_pusher_push.erl
@@ -105,8 +105,9 @@ push_event(_, _) ->
     ok.
 
 do_push_event(Host, #chat_event{from = From, to = To, packet = Packet}) ->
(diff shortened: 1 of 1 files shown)


michalwski:
I'm waiting for this to be merged!

src/event_pusher/mod_event_pusher_push.erl
```diff
@@ -105,8 +105,9 @@ push_event(_, _) ->
     ok.
 
 do_push_event(Host, #chat_event{from = From, to = To, packet = Packet}) ->
-    mod_event_pusher_push_plugin:should_publish(Host, From, To, Packet) andalso
-        publish_message(From, To, Packet).
+    mod_event_pusher_push_plugin:should_publish(Host, From, To, Packet)
```
arcusfelis: can we add a comment?
- do not push messages without body element
- but push messages with empty body element

PR #1728, opened 2018-02-16
Update accumulator to currently routed element in ejabberd_router:route/4

arkgil:
This change makes `ejabberd_router:route/4` always update the accumulator to contain information about currently routed element. Before this change, element and accumulator being routed could be out of sync.

If the accumulator is already in sync with the element, `ejabberd_route:route/3` should be used.

### DIFF
--- a/src/ejabberd_router.erl
+++ b/src/ejabberd_router.erl
@@ -111,7 +111,8 @@ route(From, To, Acc) ->
     El = mongoose_acc:get(element, Acc),
     route(From, To, Acc, El, routing_modules_list()).
 
-route(From, To, Acc, El) ->
+route(From, To, Acc0, El) ->
+    Acc = mongoose_acc:update_element(Acc0, El, From, To),
     ?DEBUG("route~n\tfrom ~p~n\tto ~p~n\tpacket ~p~n",
         [From, To, Acc]),
     route(From, To, Acc, El, routing_modules_list()).


src/ejabberd_router.erl
```diff
@@ -111,7 +111,8 @@ route(From, To, Acc) ->
     El = mongoose_acc:get(element, Acc),
     route(From, To, Acc, El, routing_modules_list()).
 
-route(From, To, Acc, El) ->
+route(From, To, Acc0, El) ->
+    Acc = mongoose_acc:update_element(Acc0, From, To, El),
```
fenek: `update_element(Acc, Element, From, To)`

arkgil: Oops, thanks for that 😅

fenek:
@arkgil Do you have time to debug why all tests suddenly fail?

arkgil:
@fenek Currently no, but I'll handle it when I find some time.

bartekgorny:
Wait, it was by design - what is inside the acc is supposed to be what initially entered the system, it is not meant to be changed. Sometimes what is later sent out is the same (e.g. message), sometimes it can be entirely different (like when mongoose replies with an error message). Initial design was that the element to be sent out was one of the acc fields (labeled to_send IIRC), but later it was changed so that acc and element to be sent out are two separate arguments.

arkgil:
@bartekgorny Should we close the PR in this case? The change proposed here was motivated by the fact that I called `update_element/3` + `route/4` quite often on one of the projects. Calling `route/4` means that we're most likely crossing the process boundary - to me it's intuitive that the process receiving the element and accumulator expects  them to be "in sync".  However, I understand your reasoning here.

I'm fine with closing the PR if my explanation doesn't convince anyone. This has probably more to do with assumptions each developer has on behaviour of accumulators.

/cc @fenek

bartekgorny:
When an acc is sent to another process it is usually stripped, which basically means recreating from the element while preserving some attrs. If in your case it is not stripped, while you feel it should be, then it needs some examination.

arkgil:
@bartekgorny the problem with `strip/1` is that it preserves only "some" attributes, the list of which is baked in the code of that function. 

Now that I think about it, probably `persistent_properties` were designed for use cases like that. What I don't like about them, though, is that they are a list, and new elements are always appended to that list, which means we can't override things.

Anyway, this PR definitely shouldn't be merged, because it's not obvious what problem it solves. I'm closing it for now. Thanks for the discussion @bartekgorny! 👍 🙂

PR #1311, opened 2017-05-08
Feature odbc backend for mod muc

arcusfelis:
This PR adds SQL backend for mod_muc.

Proposed changes include:
* Separate query logic into separate Mnesia backend
* Add ODBC backend
* New option load_permanent_rooms_at_startup option that keep rooms hibernated after whole cluster restart.

PR #1736, opened 2018-02-19
Dedicate fed1 node for s2s tests

arcusfelis:
Separate s2s and global distribution test nodes.
Global distribution tests should use new node reg1 (region 1)
localhost should not be loaded on fed1, because it would make
any additional s2s tests hard to implement.
An example of such test is muc and s2s combination.
Remove debug printing to console from mod_global_distrib_SUITE.
Use generic rpc function provided by escalus from mod_global_distrib_SUITE.
Update docs.

This PR addresses "I can't merge #1311" without this patch.

### DIFF
--- a/rel/reg1.vars.config
+++ b/rel/reg1.vars.config
@@ -1,17 +1,20 @@
-%% This node is for s2s testing.
-%% "localhost" host should NOT be defined.
-{hosts, "[ \"fed1\" ]"}.
+%% This node is for global distribution testing.
+%% reg is short for region.
+%% Both local and global hosts should be defined.
+%% "localhost" is a global host.
+%% "reg1" is a local host.
+{hosts, "[ \"reg1\", \"localhost\" ]"}.
 {sm_backend, "{mnesia, []}"}.
 {auth_method, "internal"}.
 {outgoing_s2s_port, 5269}.
 {odbc_server, ""}.
 {s2s_addr, "{ {s2s_addr, \"localhost\"}, {127,0,0,1} }. { {s2s_addr, \"localhost.bis\"}, {127,0,0,1} }."}.
 {s2s_default_policy, allow}.
-{node_name, "fed1@localhost"}.
-{ejabberd_c2s_port, 5242}.
-{ejabberd_s2s_in_port, 5299}.
-{cowboy_port, 5282}.
-{cowboy_port_secure, 5287}.
+{node_name, "reg1@localhost"}.
+{ejabberd_c2s_port, 5252}.
+{ejabberd_s2s_in_port, 5298}.
+{cowboy_port, 5272}.
+{cowboy_port_secure, 5277}.
 {ejabberd_service, ""}.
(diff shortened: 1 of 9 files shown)


kzemek:
> This PR addresses "I can't merge #1311" without this patch.
Can you elaborate?

FWIW, I reused fed1 for global distribution because the pain of testing increases nonlinearly with the number of nodes (esp. for local tests).

arcusfelis:
@kzemek sure, that PR has additional tests for MUC over s2s.
So, the tests would not work, because stanzas, routed to "muc.localhost" would e routed to the local node (i.e. fed1). I need to remove "localhost" from fed1 to fix that tests.

arcusfelis:
But overall, any XMPP interaction should work over s2s.
The fact that we don't test every feature over s2s can hide some bugs.
So, one of the way to start testing, is to override user specs in init_per_group and test several user-configurations.

arcusfelis:
yeah, a bit too many nodes :)
But separate nodes provide better isolation - that means easier debugging/refactoring.

rel/reg1.vars.config
```diff
@@ -0,0 +1,37 @@
+%% This node is for global distribution testing.
+%% reg is short for region.
+%% "localhost" host should be defined.
+{hosts, "[ \"reg1\", \"localhost\" ]"}.
```
ludwikbukowski: `"reg1"` domain not needed

arcusfelis: done.

fenek: It's not funny. Every host in GD setup needs both global domain and local domain.

test.disabled/ejabberd_tests/tests/s2s_SUITE.erl
```diff
@@ -96,13 +96,13 @@ init_per_suite(Config0) ->
 
     Config1 = [{s2s_opts, S2S} | escalus:init_per_suite(Config0)],
     Config2 = [{escalus_user_db, xmpp} | Config1],
-    escalus:create_users(Config2, escalus:get_users([alice2, alice_bis, bob])).
+    escalus:create_users(Config2, escalus:get_users([alice2, alice, bob])).
```
fenek: What is the reason behind these changes?

arcusfelis: Reverts s2s_SUITE back before 85cc37284fae81f25d14f71cfb3418f4a2402b83.
In that commit, alice was replaced with alice_bis, because it's was not possible to use alice in s2s.
I reverted it back, because we use alice everywhere else.

arcusfelis:
@fenek ok, I've updated "header" of reg1.vars.config, so we would not remove some of the hosts, trying to "optimize" stuff :)
Still not sure, how mim1 host works without a "local" host.

fenek:
@arcusfelis 
`localhost.bis` is the local host of mim1.

PR #1737, opened 2018-02-20
Put node starting code into a separate script

arcusfelis:
Split start and started steps when in travis-test

### DIFF
--- a/tools/start-nodes.sh
+++ b/tools/start-nodes.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+# This script has no arguments
+
+source tools/travis-common-vars.sh
+
+MIM1=${BASE}/_build/mim1/rel/mongooseim
+MIM2=${BASE}/_build/mim2/rel/mongooseim
+MIM3=${BASE}/_build/mim3/rel/mongooseim
+FED1=${BASE}/_build/fed1/rel/mongooseim
+MIM1CTL=${MIM1}/bin/mongooseimctl
+MIM2CTL=${MIM2}/bin/mongooseimctl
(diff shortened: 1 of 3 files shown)


tools/start-nodes.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
```
kzemek: At least `set -e` would be useful here.

tools/start-nodes.sh
```diff
@@ -0,0 +1,38 @@
...
+# This script has no arguments
+
+source tools/travis-common-vars.sh
+
+MIM1=${BASE}/_build/mim1/rel/mongooseim
+MIM2=${BASE}/_build/mim2/rel/mongooseim
+MIM3=${BASE}/_build/mim3/rel/mongooseim
+FED1=${BASE}/_build/fed1/rel/mongooseim
```
kzemek: Can we offload `MIM{1,2,3}` & `FED1` variables to `tools/travis-common-vars.sh`?

tools/travis-test.sh
```diff
@@ -142,9 +123,7 @@ run_tests() {
   echo "Running big tests (tests/ejabberd_tests)"
   echo "############################"
 
-  for node in ${NODES[@]}; do
-    start_node $node;
-  done
+  time ${TOOLS}/start-nodes.sh
 
   # Start all additional services
   start_services
```
kzemek: Would it make sense to also reorder instructions here to `start_services` before `start-nodes.sh`?

kzemek: Ok, I think that's out of scope here.

PR #1739, opened 2018-02-26
Let reload_cluster not check if configs are the same

kanes115:
Give user a possibility to specify whether to check that config files are exactly the same on each node in cluster when using mongooseimctl's `reload_cluster` command. It is done by adding options to this command. These are:
* hard - exactly how it's now done by default
* soft - it reloads files on a node if the file is modified more recently than current config in memory 
* none - no checks are performed, it reloads files no matter how different they are

### DIFF
--- a/doc/operation-and-maintenance/Reloading-configuration-on-a-running-system.md
+++ b/doc/operation-and-maintenance/Reloading-configuration-on-a-running-system.md
@@ -10,10 +10,14 @@
 This might introduce inconsistencies between different nodes of the cluster.
 It's available as a safety mechanism for the rare case of a cluster-global reload failing.
 
-`reload_cluster` is generally safe. It will try to apply the configuration
+`reload_cluster [safety_check]` is generally safe. It will try to apply the configuration
 on all nodes of the cluster.
 The prerequisite is that the modified config file must be available on
 all nodes at the same location (the location where MongooseIM expects its config file).
+Option `safety_check`can be one of the following:
+* hard - ctl will check whether all configs are the same on all the nodes vefore reloading changes
+* soft - ctl will only check if config files on nodes are modified after config was loaded to memory and reload changes on nodes that satisfy this contraint
+* none - will not check consistency between nodes' configs
 
 ### Non-reloadable options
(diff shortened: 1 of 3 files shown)


kanes115:
TODO:
* Implement `soft` mode
* Adjust conf_reload_SUITE and exactly (probably) funciton reload_helper:reload_through_ctl/2 to use `hard` mode

doc/operation-and-maintenance/Reloading-configuration-on-a-running-system.md
```diff
@@ -10,10 +10,14 @@
 This might introduce inconsistencies between different nodes of the cluster.
 It's available as a safety mechanism for the rare case of a cluster-global reload failing.
 
-`reload_cluster` is generally safe. It will try to apply the configuration
+`reload_cluster [safety_check]` is generally safe. It will try to apply the configuration
 on all nodes of the cluster.
 The prerequisite is that the modified config file must be available on
 all nodes at the same location (the location where MongooseIM expects its config file).
+Option `safety_check`can be one of the following:
+* hard - ctl will check whether all configs are the same on all the nodes vefore reloading changes
```
goddammit: vefore = before?;)

src/ejabberd_config.erl
```diff
@@ -876,30 +898,104 @@ reload_cluster() ->
...
+reload_softly_locally({CC, LC, LHC}, State1, ConfigVersion) ->
+    reload_softly(fun apply_changes/5, [CC, LC, LHC, State1, ConfigVersion]).
+
+reload_softly_remote(ConfigFile, ConfigDiff) ->
+    reload_softly(fun apply_changes_remote_unsafe/2, [ConfigFile, ConfigDiff]).
+
+reload_softly(ApplyChangesFun, Args) ->
+    case is_config_file_fresh() of
+        false ->
+            {error, node(), "Config file was modified before current config was laoded and `soft` safety check was chosen."};
```
goddammit: "The config file was modified before the current config was loaded and  the`soft` safety check was chosen."};

fenek:
Superseded by #1752

Issue #1741, opened 2018-02-27
REST API + MUC Light odbc backend not tested


ludwikbukowski:
MongooseIM version: 80b56ec32160c90
Installed from: source
Erlang/OTP version: 19

When we check the tests for REST API for admin for Muc light management[1]
we can see that `mnesia` backend is forced to use (as a default one). The `odbc` is not tested at all

[1] https://github.com/esl/MongooseIM/blob/master/test.disabled/ejabberd_tests/tests/muc_light_http_api_SUITE.erl#L61

fenek:
Fixed by #1742

PR #1742, opened 2018-02-27
Test http admin + muc light odbc backend

ludwikbukowski:
This PR addresses https://github.com/esl/MongooseIM/issues/1741

It wasn't tested before so lets see what might fail.
If so, all the fixes will be introduced in this PR

### DIFF
--- a/doc/rest-api/Administration-backend_swagger.yml
+++ b/doc/rest-api/Administration-backend_swagger.yml
@@ -444,9 +444,9 @@ paths:
           schema:
             title: roomJID
             type: string
-  /muc-lights/{XMPPHost}/{roomName}/participants:
+  /muc-lights/{XMPPMUCHost}/{roomName}/participants:
     parameters:
-      - $ref: '#/parameters/hostName'
+      - $ref: '#/parameters/MUCServer'
       - $ref: '#/parameters/roomName'
     post:
       tags:
@@ -476,9 +476,9 @@ paths:
       responses:
         204:
           description: An invite was sent out
-  /muc-lights/{XMPPHost}/{roomName}/messages:
+  /muc-lights/{XMPPMUCHost}/{roomName}/messages:
     parameters:
-      - $ref: '#/parameters/hostName'
+      - $ref: '#/parameters/MUCServer'
       - $ref: '#/parameters/roomName'
     post:
       tags:
@@ -508,9 +508,9 @@ paths:
       responses:
         204:
           description: Message was sent to the MUC Light room
-  /muc-lights/{XMPPHost}/{roomName}/{user}/management:
+  /muc-lights/{XMPPMUCHost}/{roomName}/{user}/management:
     parameters:
-      - $ref: '#/parameters/hostName'
+      - $ref: '#/parameters/MUCServer'
       - $ref: '#/parameters/roomName'
       - user: owner
         in: path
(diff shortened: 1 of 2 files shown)


kanes115:
The problem is with call `mongoose_subhosts:get_hosts/1` [here](https://github.com/esl/MongooseIM/blob/ad0cd061bee1bed403fbd33763921020a96c21d5/src/mod_muc_light_db_odbc.erl#L505) which returns `undefined` for some input (that it shouldn't)

kanes115:
```
(<0.965.0>) call mod_muc_light_commands:invite_to_room(<<"localhost">>,<<"wonderland">>,<<"alicE74.651030@localhost">>,<<"bOb74.651030@localhost">>)
(<0.965.0>) call mod_muc_light_commands:muc_light_room_name_to_jid_and_aff({jid,<<"alicE74.651030">>,<<"localhost">>,<<>>,<<"alice74.651030">>,
     <<"localhost">>,<<>>},<<"wonderland">>,<<"localhost">>)
(<0.965.0>) call mod_muc_light_commands:get_user_rooms({<<"alice74.651030">>,<<"localhost">>},<<"localhost">>)
(<0.965.0>) call mod_muc_light_db_odbc:get_user_rooms({<<"alice74.651030">>,<<"localhost">>},<<"localhost">>)
(<0.965.0>) call mod_muc_light_db_odbc:main_host(<<"localhost">>)
(<0.965.0>) exception_from {mod_muc_light_db_odbc,main_host,1} {error,{badmatch,undefined}}
```

get_user_rooms should refer to `muclight.localhost` and instead it refers to `localhost`.

kanes115:
[In mod_muc_light_commands](https://github.com/esl/MongooseIM/blob/ad0cd061bee1bed403fbd33763921020a96c21d5/src/mod_muc_light_commands.erl#L239) there is `Domain` being passed (<<"localhost">>). Whereas in [backend module](https://github.com/esl/MongooseIM/blob/ad0cd061bee1bed403fbd33763921020a96c21d5/src/mod_muc_light_db_odbc.erl#L119) we expect a `MUCServer` (<<"muclight.localhost">>).

kanes115:
I have implemented a host fix to this problem. Locally tests are passing. I uploaded it to test in the CI tool. For now this commit should be considered as WIP.

ludwikbukowski:
I just realised that the Client REST API is not tested with odbc as well.
I commited a change that adds possibility to test it

src/mod_muc_light_db_odbc.erl
```diff
@@ -502,5 +502,9 @@ update_room_version_transaction({RoomU, RoomS}, Version) ->
 main_host({_, RoomS}) ->
```
arcusfelis: this function should be more simple:
- accept only MucServer or US;
- don't make assumptions about host type (i.e. MucServer should be always MucServer, not Server sometimes).

kzemek: I agree, the main cause not that `main_host/1` doesn't handle the case when it's given a `Server` instead of `MucServer`; the main cause is that a function calling `main_host/1` gets a  `Server` when it doesn't expect it.

kanes115: Yes, of course it is the cause.

kzemek: And this is the thing that should be fixed

PR #1744, opened 2018-02-28
Make HTTP one-to-one messages handler use the accumulator

arkgil:
This patch makes the handler's behaviour consistent with what ejabberd_c2s module does when routing any stanzas:
* prepare initial accumulator from the packet sent by the user
* run `user_send_packet` hook (`rest_user_send_packet` in case of HTTP) and use the returned accumulator
* use the `element` field of the new accumulator as the new packet to route instead of the packet originally sent by the user


arcusfelis:
@arkgil should it be waiting-for-review or WIP?

PR #1747, opened 2018-02-28
Run riak in a docker container for CI

arcusfelis:
This PR allows to run Riak in a container.

Also adds:
- Adds healthcheck for Docker MySQL container.

See previous work:
- https://github.com/esl/MongooseIM/pull/1564 - SSL for Riak driver


tools/travis-setup-db.sh
```diff
@@ -1,8 +1,11 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
kzemek: Is `env` is more likely to be in `/usr/bin` than `bash`?

tools/travis-setup-db.sh
```diff
@@ -68,18 +71,50 @@ Debug                = 1
...
+    TEMP_RIAK_CONF=$(mktemp)
+    # Export config from a container
+    docker cp "mongooseim-riak:/etc/riak/riak.conf" "$TEMP_RIAK_CONF"
+    # Enable search
+    sed -i "s/^search = \(.*\)/search = on/" "$TEMP_RIAK_CONF"
+    # Enable ssl by appending settings from riak.conf.ssl
+    cat "${DB_CONF_DIR}/riak.conf.ssl" >> "$TEMP_RIAK_CONF"
+    # Import config back into container
+    docker cp "$TEMP_RIAK_CONF" "mongooseim-riak:/etc/riak/riak.conf"
+    # Erase temporary config file
```
kzemek: You're really going overboard with the comments here.

tools/travis-setup-db.sh
```diff
@@ -68,18 +71,50 @@ Debug                = 1
...
-    sudo cp ${SSLDIR}/fake_key.pem ${RIAK_DIR}/key.pem
-    sudo cp ${SSLDIR}/ca/cacert.pem ${RIAK_DIR}/cacertfile.pem
-    sudo cp ${DB_CONF_DIR}/advanced.config ${RIAK_DIR}/advanced.config
-    sudo cp ${DB_CONF_DIR}/riak.conf ${RIAK_DIR}/riak.conf
-    sudo service riak restart
-    echo "Setup Riak"
-    sudo tools/setup_riak mongooseim_secret
-
-elif [ $DB = 'cassandra' ]; then
+    # Instead of docker run, use "docker create" + "docker start".
```
kzemek: What if instead of doing `create` + `start` trick we did the following?:

```bash
RIAK_CONF=$(mktemp)
docker run --rm --entrypoint bash michalwski/docker-riak:1.0.6 -c 'cat /etc/riak/riak.conf' > ${RIAK_CONF}

```

tools/wait_for_healthcheck.sh
```diff
@@ -0,0 +1,47 @@
...
+    exit "Illegal number of parameters"
+fi
+CONTAINER="$1"
+
+# Default timeout is 1 minute
+TIMEOUT="${TIMEOUT:-60}"
+
+# Stop waiting after timeout using a background task
+MAIN_PID=$BASHPID
+(sleep "$TIMEOUT"; echo ""; echo "Killed by timeout"; kill $MAIN_PID) &
```
kzemek: Why go through the complications of setting up a separate timeout process? We can achieve less fidelity but much more clarity with simply manipulating the loop:

```bash
for i in $(seq ${TIMEOUT}); do
    if [ $(health_status "$CONTAINER")"" = "\"healthy\"" ]; then
        echo "\nWaiting is done"
        exit 0
    fi
    echo -n "."
    sleep 1
done
echo "\nKilled by timeout"
exit 1
```

kzemek: Or, with the same amount of fidelity:

kzemek: Or, with the same amount of fidelity:

```

Comment (kzemek) [reply tools/wait_for_healthcheck.sh]: Or, with around the same amount of fidelity with a bit less clarity:

```bash
DIE_AFTER=$(( $(date '+%s') + ${TIMEOUT} ))
until [ $(health_status "$CONTAINER")"" = "\"healthy\"" ]; do
    if [ "$(date '+%s')" -gt "${DIE_AFTER}" ]; then
        echo "\nKilled by timeout"
        exit 1
    fi
    echo -n "."
    sleep 1
done
echo "\nWaiting is done"
```

tools/travis-setup-db.sh
arcusfelis: @kzemek we use `/usr/bin/env` in many scripts, it's just another way to do the same thing.
Never seen a system without` /bin/bash` or without `/usr/bin/env`.

kzemek: `/usr/bin/env <sth>` is a trick to find `<sth>` in caller's `$PATH`.
I'm not saying that the change is bad. I'm not saying it's good either.

tools/travis-setup-db.sh
arcusfelis: Two times container creation in your case vs one time container creation in mine.
I want to measure both, but yours version would crash.
Because of the reason, described in a comment here in the code:

```
+    # We can't use volumes for riak.conf, because:
+    # - we want to change it
+    # - container starting code runs sed on it and gets IO error,
+    #   if it's a volume (THIS)
```

tools/wait_for_healthcheck.sh
arcusfelis: ok, I like your code more.
We can use also `timeout` utility.

kzemek: There are two main problems with `timeout`:
* still requires trapping exists, otherwise SIGINT from user will be lost
* not present on macOS unless you install coreutils (and then present as `gtimeout`

tools/travis-setup-db.sh
kzemek: > Two times container creation in your case vs one time container creation in mine.

I don't think this is a problem

> I want to measure both, but yours version would crash.

Yep, that's a problem

tools/wait_for_healthcheck.sh
arcusfelis: `echo -e "\n"`
heh, it's tricky

kzemek: Well just `echo` then :)

kzemek: I definitely can.
It's also possible that `kill` wouldn't kill the process if it hanged in docker call, anyway.

tools/travis-setup-db.sh
arcusfelis: Feeling the same.
But there was some conversation on hipchat about using env everywhere. It was more about "one way to do this", and env was chosen.
Maybe we should document it.

tools/wait_for_healthcheck.sh
arcusfelis: OK, rewrote using cycle.
blocking docker is super rare, I would care, if it would cause problems for _me_.
Until then it should be enough.

PR #1748, opened 2018-03-02
Fix saving unicode MAM messages.

kzemek:
This PR allows storing and retrieving unicode messages in MAM. It changes a few places to properly handle unicode, as well as sidesteps escaping needed for some databases by using prepared queries also in synchronous writes (previously only in asynchronous)

### DIFF
--- a/src/mod_mam_muc_odbc_arch.erl
+++ b/src/mod_mam_muc_odbc_arch.erl
@@ -71,6 +71,8 @@ start(Host, Opts) ->
             error(hand_made_partitions_not_supported)
     end,
 
+    prepare_insert(insert_mam_muc_message, 1),
+
     start_muc(Host, Opts).
 
 -spec stop(jid:server()) -> 'ok'.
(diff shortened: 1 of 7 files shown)


src/mod_mam_muc_odbc_arch.erl
```diff
@@ -138,6 +138,8 @@ archive_message(_Result, Host, MessID, RoomID,
     try
         archive_message_unsafe(Host, MessID, RoomID, FromNick, Packet)
     catch _Type:Reason ->
+            ?ERROR_MSG("event=archive_message_failed reason='~p' stacktrace=~p",
```
arcusfelis: We want also see MessID, RoomID and FromNick in logs for debugging.

src/mod_mam_odbc_arch.erl
```diff
@@ -212,6 +212,8 @@ archive_message(Result, Host, MessID, UserID,
         do_archive_message(Result, Host, MessID, UserID,
                            LocJID, RemJID, SrcJID, Dir, Packet)
     catch _Type:Reason ->
+            ?ERROR_MSG("event=archive_message_failed reason='~p' stacktrace=~p",
```
arcusfelis: same here, extra fields are good to have.

test.disabled/ejabberd_tests/tests/mam_SUITE.erl
```diff
@@ -1347,6 +1351,38 @@ long_text_search_request(Config) ->
         end,
     escalus_fresh:story(Config, [{alice, 1}, {bob, 1}], F).
 
+save_unicode_messages(Config) ->
+    P = ?config(props, Config),
+    F = fun(Alice, Bob) ->
+                escalus:send(Alice, escalus_stanza:chat_to(Bob, <<"Hi! this is an unicode character ȥ"/utf8>>)),
```
arcusfelis: should be information with a link to codepoints.

LATIN SMALL LETTER Z WITH HOOK
http://www.fileformat.info/info/unicode/char/0225/index.htm

test.disabled/ejabberd_tests/tests/mam_SUITE.erl
```diff
@@ -1347,6 +1351,38 @@ long_text_search_request(Config) ->
...
+                escalus:send(Alice, escalus_stanza:chat_to(Bob, <<"this is another one ȸ"/utf8>>)),
```
arcusfelis: LATIN SMALL LETTER DB DIGRAPH

test.disabled/ejabberd_tests/tests/mam_SUITE.erl
```diff
@@ -1347,6 +1351,38 @@ long_text_search_request(Config) ->
...
+
+                %% 'ȸ' query
+                escalus:send(Alice, stanza_text_search_archive_request(P, <<"q2">>, <<"ȸ"/utf8>>)),
+                Res2 = wait_archive_respond(P, Alice),
+                assert_respond_size(1, Res2),
+                assert_respond_query_id(P, <<"q2">>, parse_result_iq(P, Res2)),
+                [Msg3] = respond_messages(Res2),
+                #forwarded_message{message_body = Body3} = parse_forwarded_message(Msg3),
+                ?assert_equal(<<"this is another one ȸ"/utf8>>, Body3),
+
```
arcusfelis: Don't forget to test stuff that requires surrogates when encoded in UTF16.
For example:
𐀀 (Unicode Linear B Syllable B008 A)

```erlang
unicode:characters_to_binary([65536]).
<<240,144,128,128>>
```

And smiles 
https://en.wikipedia.org/wiki/Emoticons_(Unicode_block)

kzemek: The tested characters are already multibyte, and the fix/test are about unicode characters storage not about (currently not-our-implementation-defined) search capabilities.

arcusfelis: At least in mysql emoticons were broken without mb4 encoding (while lower multibyte code points were working just fine), so it's a real case.
It we have a test with them, than we would be sure that it works with all of our backends.

Search is completely different topic...

src/mod_mam_utils.erl
```diff
@@ -757,11 +757,11 @@ form_to_text(El) ->
 %% "my%cat%was%eaten%by%my%dog%why"
 %% @end
 %% -----------------------------------------------------------------------
--spec normalize_search_text(binary() | string() | undefined) -> string() | undefined.
+-spec normalize_search_text(binary() | undefined) -> binary() | undefined.
 normalize_search_text(Text) ->
```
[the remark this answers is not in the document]
arcusfelis: Incorrect:

```erlang
re:replace(<<"юникод">>, "\s+", <<"%">>, [{return, list}, global, unicode, ucp]).                       
"N=8:>4"
```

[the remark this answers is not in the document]
fenek: I believe it is because you wrote `<<"юникод">>` instead of `<<"юникод"/utf8>>`?

PR #1750, opened 2018-03-02
Fix a race condition in a call to GD conn manager

fenek:
This PR addresses an issue observed in load tests, where sometimes a call to `get_connection` crashed with `noproc` exception (that applies to connection manager process).

It may happen if a process starts outgoing connections supervisor and second one tries to access the pool in a state, when supervisor is already running but manager (child of the supervisor) is not (yet).

Also please check the comments in the code.

Load tests confirm that this branch fixes the issue.

### DIFF
--- a/src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
+++ b/src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
@@ -26,19 +26,42 @@ add_server(Server) ->
       type => supervisor,
       modules => dynamic
      },
+    ?INFO_MSG("event=outgoing_conn_start_progress,server='~s',state='starting'", [Server]),
     case supervisor:start_child(?MODULE, ServerSupSpec) of
-        {ok, _} -> ok;
-        {error, {already_started, _}} -> ok;
-        Error -> Error
+        {ok, Pid} ->
+            ?INFO_MSG("event=outgoing_conn_start_progress,server='~s',"
+                      "state='started',pid='~p'", [Server, Pid]),
+            ok;
+        {error, {already_started, Pid}} ->
+            ?INFO_MSG("event=outgoing_conn_start_progress,server='~s',"
+                      "state='started_by_other',pid='~p'", [Server, Pid]),
+            ok;
+        Error ->
+            ?INFO_MSG("event=outgoing_conn_start_progress,server='~s',"
+                      "state='failed',error='~p'", [Server, Error]),
+            Error
     end.
 
 -spec get_connection(Server :: jid:lserver()) -> pid().
 get_connection(Server) ->
-    case whereis(mod_global_distrib_utils:server_to_sup_name(Server)) of
-        undefined -> ok = add_server(Server);
-        _ -> ok
-    end,
-    mod_global_distrib_server_sup:get_connection(Server).
+    get_connection(Server, 5).
+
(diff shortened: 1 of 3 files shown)


src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
```diff
@@ -26,19 +26,42 @@ add_server(Server) ->
...
+                      "state='started_by_other',pid='~p'", [Server, Pid]),
+            ok;
+        Error ->
+            ?INFO_MSG("event=outgoing_conn_start_progress,server='~s',"
+                      "state='failed',error='~p'", [Server, Error]),
+            Error
     end.
 
 -spec get_connection(Server :: jid:lserver()) -> pid().
 get_connection(Server) ->
```
arcusfelis: Can be still relevant
https://github.com/esl/MongooseIM/pull/1660/files#r165371814

I am feeling that we would find more bugs in this part of GD.

src/global_distrib/mod_global_distrib_server_sup.erl
```diff
@@ -33,9 +33,23 @@ start_link(Server) ->
     SupName = mod_global_distrib_utils:server_to_sup_name(Server),
     supervisor:start_link({local, SupName}, ?MODULE, [Server]).
 
--spec get_connection(Server :: jid:lserver()) -> pid().
+-spec get_connection(Server :: jid:lserver()) -> {ok, pid()} | {error, not_available}.
 get_connection(Server) ->
-    mod_global_distrib_server_mgr:get_connection(Server).
+    case catch mod_global_distrib_server_mgr:get_connection(Server) of
```
arcusfelis: I would love to see a test for this fix.

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
```diff
@@ -26,19 +26,42 @@ add_server(Server) ->
...
-    mod_global_distrib_server_sup:get_connection(Server).
+    get_connection(Server, 5).
+
+-spec get_connection(Server :: jid:lserver(), RetriesLeft :: non_neg_integer()) ->
+    pid() | no_return().
+get_connection(Server, 0) ->
+    ?ERROR_MSG("event=cannot_acquire_outgoing_connection,server='~s'", [Server]),
+    throw({error, {cannot_acquire_outgoing_connection, Server}});
+get_connection(Server, RetriesLeft) ->
+    case mod_global_distrib_server_sup:get_connection(Server) of
```
arcusfelis: Still not covered with tests.

src/global_distrib/mod_global_distrib_server_sup.erl
fenek: I'll add a TODO comment here because we would need to have a uniform method of simulating race conditions. Without it, I would be forced to use meck+RPC which is evil and failure-prone.

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
fenek: See above.

src/global_distrib/mod_global_distrib_outgoing_conns_sup.erl
fenek: Since it was mentioned in another PR, maybe it should rather be resolved there? But +1 for adding some comment about the behaviour.

PR #1752, opened 2018-03-05
Let reload_cluster not check some config values

kanes115:
Right now we just filter this option out of config when calculating sha with `ejabberd_config:filter_out_node_specific_endpoints/1` called in `compute_config_file_version/1` and `compute_config_version/1`. In the future it should probably be more general (if the number of such different-among-nodes options increase).

### DIFF
--- a/src/ejabberd_config.erl
+++ b/src/ejabberd_config.erl
@@ -1123,15 +1123,17 @@ methods_to_auth_modules(A) when is_atom(A) ->
 
 compute_config_version(LC, LCH) ->
     L0 = lists:filter(mk_node_start_filter(), LC ++ LCH),
-    L1 = filter_out_node_specific_options(L0),
+    L1 = filter_out_node_specific_options(L0, config),
     L2 = sort_config(L1),
     crypto:hash(sha, term_to_binary(L2)).
 
 compute_config_file_version(#state{opts = Opts, hosts = Hosts}) ->
-    Opts2 = filter_out_node_specific_options(Opts),
+    Opts2 = filter_out_node_specific_options(Opts, config_file),
     L = sort_config(Opts2 ++ Hosts),
     crypto:hash(sha, term_to_binary(L)).
 
+filter_out_node_specific_options(Opts, Label) ->
+    filter_out_node_specific_options(Opts).
 
 filter_out_node_specific_options([]) ->
     [];
--- a/src/ejabberd_config.erl
+++ b/src/ejabberd_config.erl
@@ -1127,7 +1127,10 @@ compute_config_version(LC, LCH) ->
     crypto:hash(sha, term_to_binary(L1)).
 
 compute_config_file_version(#state{opts = Opts, hosts = Hosts}) ->
+    ?ERROR_MSG("[reload_cluster] Opts in:~n~p", [Opts]),
     Opts2 = filter_out_node_specific_options(Opts),
+    ?ERROR_MSG("[reload_cluster] Opts out:~n~p", [Opts2]),
+    ?ERROR_MSG("hosts:~n~p", []),
     L = sort_config(Opts2 ++ Hosts),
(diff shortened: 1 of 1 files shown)


src/ejabberd_config.erl
```diff
@@ -1127,9 +1127,29 @@ compute_config_version(LC, LCH) ->
...
 
+filter_out_gd_endpoints([]) ->
+    [];
+filter_out_gd_endpoints([{local_config, {modules, Host}, Mods} | Opts]) ->
+    NewMods = delete_path_in_proplist(Mods, [mod_global_distrib, connections, endpoints]),
+    [{local_config, {modules, Host}, NewMods} | Opts];
+filter_out_gd_endpoints([Opt | Opts]) ->
+    [Opt | filter_out_gd_endpoints(Opts)].
+
+delete_path_in_proplist(Plist, [Step]) ->
```
kanes115: @fenek, if you know of any function in stdlib that could replace this definition (that is a solution for a very generic problem of nested proplists), let me know - I haven't found one.

kanes115:
Make it even more general: extend it for the whole config -  so that `node_specific_module_options/0`  would contain a tuple `{config_type(), option_path()}`. ie: `{local_config, [mod_global_distrib, connections, endpoints]}`.

Make this entry a type

fenek:
Let's confirm that it works on dev machine, as there are no `reload_cluster` tests at all. They should be added as a separate task.

src/ejabberd_config.erl
```diff
@@ -1127,9 +1127,39 @@ compute_config_version(LC, LCH) ->
...
+                          Mods, node_specific_module_options()),
+    [{local_config, {modules, Host}, NewMods} | Opts];
+filter_out_node_specific_options([Opt | Opts]) ->
+    [Opt | filter_out_node_specific_options(Opts)].
+
+% @doc The list of options that should not be checked if equal
+% between nodes in cluster. Each option is expressed as
+% a list of keys in proplist that should be deleted (if such
+% path exists). The `root path` for them is a list of modules.
+node_specific_module_options() ->
```
arcusfelis: Hmm, can this path go into ejabberd.cfg itself?
I don't like it to be hardcoded.
It feels extremely installation-specific. For different clients we can have different per-node parameters.

I even think there would be installations with the same endpoints for all nodes.

kanes115: I'm not sure if it's not too meta. These hardcoded are the options that MIGHT be different and we don't require them to be the same (but they might be). It is not user's role to define them, I think.

fenek: IMO can be done as a separate task. This PR solves immediate issue.

fenek: @kanes115 I agree with @arcusfelis. It wouldn't be too meta and we already have config options only for advanced users, so why not affect the `reload_cluster` logic?

kanes115: Yea, I agree

src/ejabberd_config.erl
```diff
@@ -1123,13 +1123,49 @@ methods_to_auth_modules(A) when is_atom(A) ->
...
+    Opts2 = filter_out_node_specific_options(Opts, compute_config_file_version),
+    L = sort_config(Opts2 ++ Hosts),
     crypto:hash(sha, term_to_binary(L)).
 
+filter_out_node_specific_options(Opts, _Label) ->
+    filter_out_node_specific_options(Opts).
+
+filter_out_node_specific_options([]) ->
+    [];
+filter_out_node_specific_options([{local_config, {modules, Host}, Mods} | Opts]) -> % use record
```
kanes115: Use record here in pattern matching

michalwski:
@kanes115 @fenek can we close this PR? I understand the issue was solved by #1948. Is that right?

PR #1758, opened 2018-03-06
Separated database logic from mod_muc into backend modules

arcusfelis:
It's a **subset** of https://github.com/esl/MongooseIM/pull/1311
Includes mnesia backend
Includes common backend code
Includes registration tests

We want to add odbc backend as a separate PR.
Because we first need to discuss schema.
Maybe we would have two RDBMS backends, one normalized, one denormalized and compare performance.

Add load_permanent_rooms_at_startup option

New result format for some mod_muc functions

Document mod_muc.backend option

Remove mnesia migrations from mod_muc

Replace catch with try..catch

Simplify mod_muc Mnesia code

Subtruct unset_nick from set_nick function

Do not allow to use a nick if DB returns an error

Add MUC registration tests

Test registration over S2S

Move generic s2s code into s2s_helper

Use s2s_helper in muc_SUITE

Disable fed1 cover for MUC-over-S2S tests

Increase timeout for hibernation tests

Add muc_SUITE:can_found_in_db_when_stopped testcase/1

### DIFF
--- a/src/jlib.erl
+++ b/src/jlib.erl
@@ -58,7 +58,9 @@
          stanza_errort/5,
          stream_error/1,
          stream_errort/3,
-         remove_delay_tags/1]).
+         remove_delay_tags/1,
+         expr_to_term/1,
+         term_to_expr/1]).
 
 -include_lib("exml/include/exml.hrl").
 -include_lib("exml/include/exml_stream.hrl"). % only used to define stream types
@@ -868,3 +870,12 @@ remove_delay_tags(#xmlel{children = Els} = Packet) ->
                               El ++ [R]
                 end, [], Els),
     Packet#xmlel{children=NEl}.
+
+expr_to_term(Expr) ->
+    Str = binary_to_list(<<Expr/binary, ".">>),
+    {ok, Tokens, _} = erl_scan:string(Str),
+    {ok, Term} = erl_parse:parse_term(Tokens),
+    Term.
+
+term_to_expr(Term) ->
+    list_to_binary(io_lib:print(Term)).
--- a/src/jlib.erl
+++ b/src/jlib.erl
@@ -58,9 +58,7 @@
          stanza_errort/5,
          stream_error/1,
          stream_errort/3,
-         remove_delay_tags/1,
-         expr_to_term/1,
-         term_to_expr/1]).
+         remove_delay_tags/1]).
 
 -include_lib("exml/include/exml.hrl").
 -include_lib("exml/include/exml_stream.hrl"). % only used to define stream types
@@ -870,12 +868,3 @@ remove_delay_tags(#xmlel{children = Els} = Packet) ->
                               El ++ [R]
                 end, [], Els),
     Packet#xmlel{children=NEl}.
-
-expr_to_term(Expr) ->
-    Str = binary_to_list(<<Expr/binary, ".">>),
-    {ok, Tokens, _} = erl_scan:string(Str),
-    {ok, Term} = erl_parse:parse_term(Tokens),
-    Term.
-
-term_to_expr(Term) ->
-    list_to_binary(io_lib:print(Term)).
--- a/src/mod_muc.erl
+++ b/src/mod_muc.erl
@@ -535,10 +535,27 @@ get_registered_room_or_route_error_from_presence(Room, From, To, Acc, Packet,
                    http_auth_pool = HttpAuthPool,
                    default_room_opts = DefRoomOpts} = State,
             {_, _, Nick} = jid:to_lower(To),
-            {ok, Pid} = start_new_room(Host, ServerHost, Access, Room,
+            Result = start_new_room(Host, ServerHost, Access, Room,
                                        HistorySize, RoomShaper, HttpAuthPool,
                                        From, Nick, DefRoomOpts),
-            register_room_or_stop_if_duplicate(Host, Room, Pid);
+            case Result of
+                {ok, Pid} ->
+                    register_room_or_stop_if_duplicate(Host, Room, Pid);
+                {error, {failed_to_restore, Reason}} ->
+                    %% Notify user about our backend module error
+                    ?WARNING_MSG("event=send_service_unavailable room=~ts reason=~p",
+                                 [Room, Reason]),
+                    Lang = exml_query:attr(Packet, <<"xml:lang">>, <<>>),
+                    ErrText = <<"Service is temporary unavailable">>,
+                    Err = jlib:make_error_reply(
+                            Packet, mongoose_xmpp_errors:service_unavailable(Lang, ErrText)),
+                    ejabberd_router:route(To, From, Acc, Err),
+                    {route_error, ErrText};
+                _ ->
+                    %% Unknown error, most likely a room process failed to start.
+                    %% Do not notify user (we can send "internal server error").
+                    erlang:error({start_new_room_failed, Room, Result})
+            end;
         false ->
             Lang = exml_query:attr(Packet, <<"xml:lang">>, <<>>),
             ErrText = <<"Room creation is denied by service policy">>,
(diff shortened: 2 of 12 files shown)


src/mod_muc.erl
```diff
@@ -733,14 +705,14 @@ load_permanent_rooms(Host, ServerHost, Access, HistorySize, RoomShaper, HttpAuth
 start_new_room(Host, ServerHost, Access, Room,
                HistorySize, RoomShaper, HttpAuthPool, From,
                Nick, DefRoomOpts) ->
-    case mnesia:dirty_read(muc_room, {Room, Host}) of
-        [] ->
+    case mod_muc_db_backend:restore_room(ServerHost, Host, Room) of
+        {error, _} ->
```
kzemek: This is different behaviour than original. Originally, we'd create a room if it was not found; here we create it on all possible errors.

src/mod_muc.erl
```diff
@@ -330,9 +292,17 @@ init([Host, Opts]) ->
     ejabberd_router:register_route(MyHost, mongoose_packet_handler:new(?MODULE, State)),
     mongoose_subhosts:register(Host, MyHost),
 
-    load_permanent_rooms(MyHost, Host,
-                         {Access, AccessCreate, AccessAdmin, AccessPersistent},
-                         HistorySize, RoomShaper, HttpAuthPool),
+    case gen_mod:get_module_opt(Host, mod_muc, load_permanent_rooms_at_startup, false) of
+        false ->
+            ?INFO_MSG("issue=load_permanent_rooms_at_startup, skip=true, "
```
kzemek: What's the difference between `issue` and `event` in logs?

src/mod_muc_db_mnesia.erl
```diff
@@ -0,0 +1,182 @@
...
+
+-spec forget_room(ejabberd:server(), ejabberd:server(), mod_muc:room()) -> 'ok'.
+forget_room(_ServerHost, MucHost, RoomName) ->
+    F = fun() ->
+                mnesia:delete({muc_room, {RoomName, MucHost}})
+        end,
+    Result = mnesia:transaction(F),
+    case Result of
+        {atomic, _} ->
+            ejabberd_hooks:run(forget_room, MucHost, [MucHost, RoomName]),
```
kzemek: A backend does not seem to be a good place to run hooks from; every backend will have to duplicate that line on success in `forget_room/3`.

src/mod_muc_db.erl
```diff
@@ -0,0 +1,46 @@
...
+
+-type room_opts() :: [{OptionName :: atom(), OptionValue :: term()}].
+
+
+
+%% Called when MUC service starts or restarts for each domain
+-callback init(server_host(), ModuleOpts :: list()) -> ok.
+
+-callback store_room(server_host(), muc_host(), mod_muc:room(), room_opts()) ->
+    {aborted, _} | {atomic, _}.
```
kzemek: These return values stand out, as you made other returns non-mnesia-specific (`{ok, _}`, `{error, _}`). I'd suggest doing the same with `store_room/4`.

src/mod_muc_db.erl
```diff
@@ -0,0 +1,46 @@
...
+
+-callback restore_room(server_host(), muc_host(), mod_muc:room()) ->
+    {ok, room_opts()} | {error, room_not_found} | {error, term()}.
+
+-callback forget_room(server_host(), muc_host(), mod_muc:room()) -> ok.
+
+-callback can_use_nick(server_host(), muc_host(),
+                       client_jid(), mod_muc:nick()) -> boolean().
+
+-callback get_rooms(server_host(), muc_host()) -> [#muc_room{}].
```
kzemek: Again, I think we should standardize callback behaviour. This one returns a value or exception while other ones return `{ok, _}`, `{error, _}`.

src/mod_muc_db_mnesia.erl
```diff
@@ -0,0 +1,182 @@
...
+
+restore_room(_ServerHost, MucHost, RoomName) ->
+    try mnesia:dirty_read(muc_room, {RoomName, MucHost}) of
+        [#muc_room{opts = Opts}] ->
+            {ok, Opts};
+        [] ->
+            {error, room_not_found};
+        Other ->
+            {error, Other}
+        catch Class:Reason ->
```
kzemek: You seem to consistently ident `catch` in this PR as if it were a clause inside `try`. This is inconsistent with how we (and [most of the Erlang world](http://erlang.org/doc/reference_manual/expressions.html#try), including indentation tools I use) do it everywhere else.

src/mod_muc_db_mnesia.erl
```diff
@@ -0,0 +1,182 @@
...
+            {error, not_registered};
+        [#muc_registered{nick = Nick}] ->
+            {ok, Nick}
+        catch Class:Reason ->
+            ?ERROR_MSG("issue=get_nick_failed jid=~ts reason=~p:~p",
+                       [jid:to_binary(From), Class, Reason]),
+            {error, {Class, Reason}}
+    end.
+
+set_nick(_ServerHost, _MucHost, _From, <<>>) ->
```
kzemek: I think this should be handled on the frontend-module side, else every backend will have to include this exact clause returning `{error, should_not_be_empty}`.

src/mod_muc_db_mnesia.erl
```diff
@@ -0,0 +1,182 @@
...
+    case mnesia:transaction(F) of
+        {atomic, _} ->
+            ok;
+        ErrorResult ->
+            ?ERROR_MSG("issue=unset_nick_failed jid=~ts reason=~1000p",
+                       [jid:to_binary(From), ErrorResult]),
+            {error, ErrorResult}
+    end.
+
+jid_to_lower_us(JID) ->
```
kzemek: There's also `jid:to_lus/1`.

src/jlib.erl
```diff
@@ -58,7 +58,9 @@
          stanza_errort/5,
          stream_error/1,
          stream_errort/3,
-         remove_delay_tags/1]).
+         remove_delay_tags/1,
+         expr_to_term/1,
```
kzemek: Changes in this file seem to be unrelated to this PR.

src/mod_muc.erl
arcusfelis: Ideological issues. Once we choose one, we can replace it everywhere.

```erlang
~/erlang/esl/2017/MongooseIM|feature-muc-backends-revived $ git grep "event=" | wc -l                                             
34                              
~/erlang/esl/2017/MongooseIM|feature-muc-backends-revived $ git grep "issue=" | wc -l                                             
35 
```

src/mod_muc_db.erl
arcusfelis: yeah, but mod_muc_room just ignores output value for now

kzemek: The other point is that other backends, for example a future Riak backend, will have to replicate alien Mnesia interface.

src/mod_muc_db_mnesia.erl
arcusfelis: oh, no, I don't like the world.

kzemek:
LGTM

src/mod_muc.erl
```diff
@@ -562,20 +565,29 @@ get_registered_room_or_route_error_from_presence(Room, From, To, Acc, Packet,
...
             Err = jlib:make_error_reply(
                     Packet, mongoose_xmpp_errors:item_not_found(Lang, ErrText)),
             ejabberd_router:route(To, From, Err),
             {route_error, ErrText};
-        Opts ->
+        {error, Reason} ->
+            ?WARNING_MSG("event=send_service_unavailable room=~ts reason=~p",
+                         [Room, Reason]),
+            Lang = exml_query:attr(Packet, <<"xml:lang">>, <<>>),
+            ErrText = <<"Service is temporary unavailable">>,
```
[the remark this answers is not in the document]
arcusfelis: @michalslaski sure, but in a separate small PR.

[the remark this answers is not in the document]
fenek: @arcusfelis You fell into michalslaski/michalwski trap :)

PR #1760, opened 2018-03-07
Move all path related functions into path_helper in big tests

arcusfelis:
This PR is continuation of #1626 

Proposed changes include:
* new helper module path_helper
* new env variables REPO_DIR and TEST_DIR passed to VM, running the big tests


test.disabled/ejabberd_tests/Makefile
```diff
@@ -20,6 +21,8 @@ endif
 
 COMMON_OPTS := -sname test -setcookie ejabberd -hidden \
                $(TLS_DIST_OPTS) \
+               -env REPO_DIR "$(shell readlink -f $(REPO_DIR))" \
```
erszcz: [We have a `readlink.sh` file stating that it's needed for MacOS](https://github.com/esl/MongooseIM/blob/master/readlink.sh) - how does this change work on Mac?

arcusfelis: Lovely, was thinking about this

```erlang
function readlink() {
  DIR=$(echo "${1%/*}")
  (cd "$DIR" && echo "$(pwd -P)")
}
```

erszcz: I don't know which is better: `readlink.sh` that we have in the repo or your proposed function, but I bet you can work it out :)

arcusfelis: This script looks broken. I will add another one into tools.

```erlang
./readlink.sh ../                                                          
/home/user/erlang/esl/2017/MongooseIM/.. 
```

arcusfelis: done

PR #1761, opened 2018-03-07
Change Global Distribution loglevels + some new logs

fenek:
This pull request:
* Changes level of some less important logs to `debug`.
* Fixes the check for existing metric.
* Adds 2 new logs: with resolved endpoints diff (info) and full resolved endpoints list (debug).
* Renames Redis refresher logs, so their content can't be confused with server manager refresh logs.

### DIFF
--- a/src/global_distrib/mod_global_distrib_utils.erl
+++ b/src/global_distrib/mod_global_distrib_utils.erl
@@ -46,7 +46,7 @@ binary_to_metric_atom(Binary) ->
     list_to_atom(List).
 
 ensure_metric(Metric, Type) ->
-    case catch mongoose_metrics:ensure_metric(global, Metric, Type) of
+    case mongoose_metrics:ensure_metric(global, Metric, Type) of
         ok ->
             Reporters = exometer_report:list_reporters(),
             Interval = mongoose_metrics:get_report_interval(),
--- a/src/global_distrib/mod_global_distrib_utils.erl
+++ b/src/global_distrib/mod_global_distrib_utils.erl
@@ -57,7 +57,7 @@ ensure_metric(Metric, Type) ->
               end,
               Reporters);
         {ok, already_present} ->
-            ?INFO_MSG("issue=metric_already_exists,metric=\"~p\",type=\"~p\"", [Metric, Type]),
+            ?DEBUG("issue=metric_already_exists,metric=\"~p\",type=\"~p\"", [Metric, Type]),
             ok;
         Other ->
             ?WARNING_MSG("issue=cannot_create_metric,metric=\"~p\",type=\"~p\",reason=\"~p\"",
--- a/src/mongoose_metrics.erl
+++ b/src/mongoose_metrics.erl
@@ -81,7 +81,8 @@ init_subscriptions() ->
                 subscribe_to_all(Name, Interval)
         end, Reporters).
 
--spec create_generic_hook_metric(jid:lserver(), atom()) -> ok | {ok, already_present}.
+-spec create_generic_hook_metric(jid:lserver(), atom()) ->
+    ok | {ok, already_present} | {error, any()}.
 create_generic_hook_metric(Host, Hook) ->
     FilteredHookName = filter_hook(Hook),
     do_create_generic_hook_metric(Host, FilteredHookName).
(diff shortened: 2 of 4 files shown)


src/global_distrib/mod_global_distrib_utils.erl
```diff
@@ -46,7 +46,7 @@ binary_to_metric_atom(Binary) ->
     list_to_atom(List).
 
 ensure_metric(Metric, Type) ->
```
arcusfelis: should not be really here.
One of our clients (c) has this code in mongoose_metrics:

```erlang
-spec ensure_update(Host :: ejabberd:lserver() | global,                             
                    Name :: term() | list(),                                         
                    Change :: term(),                                                
                    Type :: atom()) -> any().                                        
ensure_update(Host, Name, Value, Type) when is_atom(Type) ->                         
    Res = mongoose_metrics:update(Host, Name, Value),                                
    case Res of                                                                      
        {error,not_found} ->                                                         
            mongoose_metrics:ensure_metric(Host, Name, Type),                        
            mongoose_metrics:update(Host, Name, Value);                              
        _ ->                                                                         
            Res                                                                      
    end.  

-spec ensure_metric(ejabberd:server() | global, metric(), atom() | tuple(), atom()) -> ok.
ensure_metric(Host, Metric, Type, ShortType) when is_atom(Metric) ->                 
    ensure_metric(Host, [Metric], Type, ShortType);                                  
ensure_metric(Host, Metric, Type, ShortType) when is_list(Metric) ->                 
    %% the split into ShortType and Type is needed because function metrics are      
    %% defined as tuples (that is Type), while exometer:info returns only 'function' 
    PrefixedMetric = name_by_all_metrics_are_global(Host, Metric),                   
    case exometer:info(PrefixedMetric, type) of                                      
        ShortType -> {ok, already_present};                                          
        undefined ->                                                                 
            Result =                                                                 
                try                                                                  
                    exometer:new(PrefixedMetric, Type)                               
                catch error:exists ->                                                
                          ok                                                         
                end,                                                                 
            %% exometer_report has a dead lock, so make it in a separate process     
            proc_lib:spawn(fun() ->                                                  
               init_metric_subscriptions(PrefixedMetric, Type)                       
               end),                                                                 
            Result                                                                   
    end. 
```

We can prepare a patch.

This TODO is not relevant to this PR.

src/mongoose_metrics.erl
```diff
@@ -328,7 +328,12 @@ ensure_metric(Host, Metric, Type, ShortType) when is_list(Metric) ->
     PrefixedMetric = name_by_all_metrics_are_global(Host, Metric),
     case exometer:info(PrefixedMetric, type) of
         ShortType -> {ok, already_present};
-        undefined -> exometer:new(PrefixedMetric, Type)
+        undefined ->
```
arcusfelis: Yep, race condition between ensure and create.
Occurs under heavy load, when metric is not created.
One process creates the metric, all other processes crash.

src/mongoose_metrics.erl
```diff
@@ -328,7 +328,12 @@ ensure_metric(Host, Metric, Type, ShortType) when is_list(Metric) ->
...
+            case catch exometer:new(PrefixedMetric, Type) of
```
arcusfelis: `try...catch`, please. There is an example above in my comment.

arcusfelis: which can be added when we would refactor the whole mongoose_metrics module.

src/global_distrib/mod_global_distrib_utils.erl
```diff
@@ -57,7 +57,7 @@ ensure_metric(Metric, Type) ->
               end,
               Reporters);
         {ok, already_present} ->
```
arcusfelis: Dialyzer
```erlang
_build/default/lib/mongooseim/src/global_distrib/mod_global_distrib_utils.erl
  59: The pattern {'ok', 'already_present'} can never match the type {'error',_}
===> Warnings written to /home/travis/build/esl/MongooseIM/_build/default/20.0.dialyzer_warnings
```

src/global_distrib/mod_global_distrib_utils.erl
fenek: I'll create "Metrics refactoring" project.

src/mongoose_metrics.erl
fenek: OK, will be a part of Metrics Refactoring

src/global_distrib/mod_global_distrib_utils.erl
fenek: Fixed

PR #1763, opened 2018-03-08
Make packet_to_component GD handler more robust

fenek:
* `mod_global_distrib:get_metadata/2,3` now may return `{error, missing_gd_structure}` error.
* Fixed ETS creation in `mod_global_distrib_utils` - `{heir, none, testing}` parameter is illegal.
* `maybe_update_mapping/2` won't crash now if `global_distrib` structure is missing in acc.
* New integration test is a byproduct of attempts to reproduce the issue; since it covers a new scenario, I haven't removed it.
* Fixed `component_SUITE:connect_component/2` as it was returning invalid component address.

### DIFF
--- a/src/global_distrib/mod_global_distrib.erl
+++ b/src/global_distrib/mod_global_distrib.erl
@@ -43,15 +43,21 @@ start(Host, Opts0) ->
 stop(Host) ->
     mod_global_distrib_utils:stop(?MODULE, Host, fun stop/0).
 
--spec get_metadata(mongoose_acc:t(), Key :: term(), Default :: term()) -> Value :: term().
+-spec get_metadata(mongoose_acc:t(), Key :: term(), Default :: term()) ->
+    Value :: term() | {error, missing_gd_structure}.
 get_metadata(Acc, Key, Default) ->
-    GD = mongoose_acc:get(global_distrib, Acc),
-    maps:get(Key, GD, Default).
+    case mongoose_acc:get(global_distrib, Acc, undefined) of
+        undefined -> {error, missing_gd_structure};
+        GD -> maps:get(Key, GD, Default)
+    end.
 
--spec get_metadata(mongoose_acc:t(), Key :: term()) -> Value :: term().
+-spec get_metadata(mongoose_acc:t(), Key :: term()) ->
+    Value :: term() | {error, missing_gd_structure}.
 get_metadata(Acc, Key) ->
-    GD = mongoose_acc:get(global_distrib, Acc),
-    maps:get(Key, GD).
+    case mongoose_acc:get(global_distrib, Acc, undefined) of
+        undefined -> {error, missing_gd_structure};
+        GD -> maps:get(Key, GD)
+    end.
 
 -spec put_metadata(mongoose_acc:t(), Key :: term(), Value :: term()) -> mongoose_acc:t().
 put_metadata(Acc, Key, Value) ->
--- a/src/global_distrib/mod_global_distrib.erl
+++ b/src/global_distrib/mod_global_distrib.erl
@@ -23,8 +23,9 @@
 -include("jlib.hrl").
 -include("global_distrib_metrics.hrl").
 
--export([deps/2, start/2, stop/1, get_metadata/2, get_metadata/3, remove_metadata/2, put_metadata/3,
-         maybe_reroute/1]).
+-export([deps/2, start/2, stop/1]).
+-export([find_metadata/2, get_metadata/3, remove_metadata/2, put_metadata/3]).
+-export([maybe_reroute/1]).
 
 %%--------------------------------------------------------------------
 %% gen_mod API
@@ -43,17 +44,16 @@ start(Host, Opts0) ->
 stop(Host) ->
     mod_global_distrib_utils:stop(?MODULE, Host, fun stop/0).
 
--spec get_metadata(mongoose_acc:t(), Key :: term(), Default :: term()) ->
-    {ok, Value :: term()}.
+-spec get_metadata(mongoose_acc:t(), Key :: term(), Default :: term()) -> Value :: term().
 get_metadata(Acc, Key, Default) ->
     case mongoose_acc:find(global_distrib, Acc) of
-        error -> {ok, Default};
-        {ok, GD} -> {ok, maps:get(Key, GD, Default)}
+        error -> Default;
+        {ok, GD} -> maps:get(Key, GD, Default)
     end.
 
--spec get_metadata(mongoose_acc:t(), Key :: term()) ->
+-spec find_metadata(mongoose_acc:t(), Key :: term()) ->
     {ok, Value :: term()} | {error, missing_gd_structure | undefined}.
-get_metadata(Acc, Key) ->
+find_metadata(Acc, Key) ->
     case mongoose_acc:find(global_distrib, Acc) of
         error ->
             {error, missing_gd_structure};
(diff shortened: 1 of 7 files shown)


src/global_distrib/mod_global_distrib.erl
```diff
@@ -43,15 +43,21 @@ start(Host, Opts0) ->
 stop(Host) ->
     mod_global_distrib_utils:stop(?MODULE, Host, fun stop/0).
 
--spec get_metadata(mongoose_acc:t(), Key :: term(), Default :: term()) -> Value :: term().
+-spec get_metadata(mongoose_acc:t(), Key :: term(), Default :: term()) ->
```
arcusfelis: Feels like it makes sense to start using `{ok, Value} | {error, Reason}` return format.

arcusfelis: Also, it makes sense to just return `Default` instead.

src/global_distrib/mod_global_distrib.erl
```diff
@@ -43,15 +43,21 @@ start(Host, Opts0) ->
...
 get_metadata(Acc, Key, Default) ->
-    GD = mongoose_acc:get(global_distrib, Acc),
-    maps:get(Key, GD, Default).
+    case mongoose_acc:get(global_distrib, Acc, undefined) of
+        undefined -> {error, missing_gd_structure};
+        GD -> maps:get(Key, GD, Default)
+    end.
 
--spec get_metadata(mongoose_acc:t(), Key :: term()) -> Value :: term().
+-spec get_metadata(mongoose_acc:t(), Key :: term()) ->
```
arcusfelis: tricky error handling.
badkey as an exception, `{error, Reason}`, high chance that we would not match on `{error,_}` in the caller code and the tuple would "walk" around our code.

src/global_distrib/mod_global_distrib_utils.erl
```diff
@@ -145,12 +145,12 @@ create_ets(Names, Type) when is_list(Names) ->
 create_ets(Name, Type) ->
     Self = self(),
     Heir = case whereis(ejabberd_sup) of
```
arcusfelis: not the first time I see this pattern, usually in module `start` function.
Maybe we should have a dedicated process for keeping our ets tables.
Also, explicit `ets:give_away` is preferable, because the ets always would be owned by the same process.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -371,6 +372,26 @@ test_component_on_one_host(Config) ->
...
     [escalus:fresh_story(Config, [{User, 1}], Story) || User <- [alice, eve]].
 
+%% Ensures that 2 components in distinct data centers can communicate.
+test_components_in_different_regions(_Config) ->
+    ComponentCommonConfig = [{host, <<"localhost">>}, {password, <<"secret">>},
+                             {server, <<"localhost">>}, {component, <<"test_service">>}],
+    Component1Config = [{port, 8888}, {component, <<"service1">>} | ComponentCommonConfig],
+    Component2Config = [{port, 9990}, {component, <<"service2">>} | ComponentCommonConfig],
+
+    {Comp1, Addr1, _Name1} = component_SUITE:connect_component(Component1Config),
```
arcusfelis: `connect_component ` should into `component_helper`. Not in this PR.

test.disabled/ejabberd_tests/tests/component_SUITE.erl
```diff
@@ -468,8 +468,8 @@ connect_component(ComponentOpts, StartStep) ->
     case Res of
     {ok, Component, _} ->
         {component, ComponentName} = lists:keyfind(component, 1, ComponentOpts),
-        {host, ComponentHost} = lists:keyfind(host, 1, ComponentOpts),
-        ComponentAddr = <<ComponentName/binary, ".", ComponentHost/binary>>,
+        {server, ComponentServer} = lists:keyfind(server, 1, ComponentOpts),
+        ComponentAddr = <<ComponentName/binary, ".", ComponentServer/binary>>,
         {Component, ComponentAddr, ComponentName};
```
arcusfelis: map is preferable as a return value (maybe not in this pr).

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -833,6 +854,8 @@ get_mapping(Node, Client) ->
     {ok, What} = redis_query(Node, [<<"GET">>, FullJid]),
     What.
 
+%% Warning! May not work properly with alice or any other user whose
+%% stringprepped JID is different than original one
 delete_mapping(Node, Client) ->
```
arcusfelis: `escalus_utils:jid_to_lower(BinJid)` can be used.

fenek: When I tried to experiment with this function, some tests suddenly broke. :P So I'm putting a warning for now so the PR won't be blocked... but this suite should be refactored definitely.

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
fenek: I've created a dedicated task for this.

src/global_distrib/mod_global_distrib_utils.erl
fenek: `mongoose_ets`? :) I'll create a task for this.

test.disabled/ejabberd_tests/tests/component_SUITE.erl
fenek: Can be refactored as a part of "move to helper" PR.

src/global_distrib/mod_global_distrib.erl
kzemek: Since there's no possibility for error I'd stick with returning a single value. AFAIK that's normal for functions taking a default (`proplists:get_value`, `map:get`, `application:get_env`).

src/global_distrib/mod_global_distrib_utils.erl
```diff
@@ -200,8 +200,14 @@ maybe_update_mapping(#jid{luser = <<>>, lserver = LServer} = From, Acc) ->
 maybe_update_mapping(From, Acc) ->
     case mod_global_distrib_mapping:for_jid(From) of
         error ->
-            Origin = mod_global_distrib:get_metadata(Acc, origin),
-            mod_global_distrib_mapping:insert_for_jid(From, Origin);
+            case mod_global_distrib:get_metadata(Acc, origin) of
+                %% Lack of 'global_distrib' indicates 100% local routing...
```
kzemek: How did we get to `maybe_update_mapping` with non-GD message? 😕

fenek: When `mongoose_router_external*` are moved to the beginning of router list and the message is routed locally in cluster, a packet to component will be processed by `packet_to_component` hook, which is handled by `mod_global_distrib_mapping`.

src/global_distrib/mod_global_distrib.erl
fenek: I wanted to keep consistent API for `get_` functions... so maybe we can rename them to `get_metadata/3` and `find_metadata/2`, so the API will be similar to `map` equivalents? Single value returned by `/3` and tuple or error for `/2`.

src/global_distrib/mod_global_distrib_utils.erl
kzemek: Guess I missed the change to the routing list.

fenek: The change is done only in one of the deployments, it won't be a default for open source.

src/global_distrib/mod_global_distrib.erl
kzemek: Sure, go ahead :)

test.disabled/ejabberd_tests/tests/mod_global_distrib_SUITE.erl
arcusfelis: yep, there is some regex inside, not sure it does normalization too

PR #1765, opened 2018-03-09
Bump cpool version to 0.1.1

kzemek:
Minor bugfix in cpool.


fenek:
Fixes `cpool` and breaks GD actually. Will require different approach probably.

PR #1770, opened 2018-03-13
Add missing GD small suite

fenek:
I forgot to add this suite to PR #1763 :(

### DIFF
--- a/test/mod_global_distrib_SUITE.erl
+++ b/test/mod_global_distrib_SUITE.erl
@@ -1,0 +1,175 @@
+%%==============================================================================
+%% Copyright 2014 Erlang Solutions Ltd.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%% http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%==============================================================================
+
+-module(mod_global_distrib_SUITE).
+-compile(export_all).
+-author('piotr.nosek@erlang-solutions.com').
+
+-include_lib("common_test/include/ct.hrl").
+-include_lib("exml/include/exml.hrl").
+
+%%--------------------------------------------------------------------
+%% Suite configuration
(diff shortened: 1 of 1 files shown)


test/mod_global_distrib_SUITE.erl
```diff
@@ -0,0 +1,175 @@
+%%==============================================================================
+%% Copyright 2014 Erlang Solutions Ltd.
```
arcusfelis: year :D

fenek: Hahaha :D Copy & paste issues. :D Thanks.

test/mod_global_distrib_SUITE.erl
```diff
@@ -0,0 +1,175 @@
...
+    fake_stop(),
+    unset_meck(),
+    Config.
+
+%%--------------------------------------------------------------------
+%% Hook handlers tests
+%%--------------------------------------------------------------------
+
+
+%% missing_struct_ tests verify the behaviour of packet_to_component handler,
```
arcusfelis: good :)

test/mod_global_distrib_SUITE.erl
```diff
@@ -0,0 +1,175 @@
...
+-author('piotr.nosek@erlang-solutions.com').
+
+-include_lib("common_test/include/ct.hrl").
+-include_lib("exml/include/exml.hrl").
+
+%%--------------------------------------------------------------------
+%% Suite configuration
+%%--------------------------------------------------------------------
+
+all() ->
```
arcusfelis: we can list cases in all() directly.
I don't see any point of having groups for small suites.
(the only point is that we use groups everywhere).

test/mod_global_distrib_SUITE.erl
```diff
@@ -0,0 +1,175 @@
...
+    % The handler must not crash and return unchanged Acc
+    Acc = mod_global_distrib_mapping:packet_to_component(Acc, From, To).
+        
+%% Update logic has two separate paths: when a packet is sent by a user or by another
+%% component. This test covers the latter.
+missing_struct_in_message_from_component(_Config) ->
+    From = jid:make(<<"">>, <<"from_service.", (global_host())/binary>>, <<"">>),
+    {Acc, To} = fake_acc_to_component(From),
+    % The handler must not crash and return unchanged Acc
+    Acc = mod_global_distrib_mapping:packet_to_component(Acc, From, To).
```
arcusfelis: I am usually using `?assert_equal`.
Recently started to use 
`?assert_equal_extra(A, B, #{comment => "Do you have something to say?", some_var => X, some_other_var => Y})`
which contains enough info to figure out what is wrong, without reading logs or suite code in 50% of cases.

PR #1773, opened 2018-03-15
Push notifications refactoring

michalwski:
The main change is that the final push notification delivery was moved to plugin module. This allows to create a plugin which can push an event to a queue so some backend service may decide what to do next with the push notification.

Additionally following changes were mad:
* Accumulator is passed to mod_event backends
* Push notification tests were refactored - helper module was created


michalwski:
@fenek The doc is updated. I also rebased on top of current master.

PR #1775, opened 2018-03-19
Introduce 'foreground' mode

GalaxyGorilla:
In rebar the 'foreground' mode is usually used when the erlang app is
running as a daemon within a supervising service such as systemd or
upstart. To work correctly within those services some flags for 'erl'
are necessary so that for example unix pipelines can be used. This
patch introduces the 'foreground' mode as known from rebar which is
very similar to live/console but sets the following additional flags:

    -noshell -noinput +Bd

Those are documented at:

    http://erlang.org/doc/man/erl.html

The very same flags are also used in relx and distillery:

https://github.com/bitwalker/distillery/blob/master/priv/libexec/commands/foreground.sh#L18
https://github.com/erlware/relx/blob/86d415f33c500d432d6c9fa2fcbff3bb2fff3e8f/priv/templates/extended_bin#L623

It should definitely be thought about adjusting the systemd unit file so
that 'foreground' is used instead of 'start'. Otherwise the beam process
is not considered as main PID and when the beam is killed you are not
notified properly via the standard systemd tooling (e.g. 'systemctl
status' returns 0).

EDIT: Aligned the PR desciption according to discussion below.

### DIFF
--- a/rel/files/mongooseim
+++ b/rel/files/mongooseim
@@ -166,10 +166,10 @@ case "$1" in
         # Start the VM
         exec_echo "$BINDIR/erlexec" $FOREGROUNDOPTIONS \
             -boot "$RUNNER_BASE_DIR/releases/$APP_VSN/$BOOTFILE" \
+            -embedded \
             -config "$RUNNER_ETC_DIR/app.config" \
             -args_file "$RUNNER_ETC_DIR/vm.args"  \
             -args_file "$RUNNER_ETC_DIR/vm.dist.args" \
-            -mode embedded \
             -- ${1+"$@"}
         ;;
 
--- a/rel/files/mongooseim
+++ b/rel/files/mongooseim
@@ -216,7 +216,7 @@ case "$1" in
         fi
         ;;
     *)
-        echo "Usage: $SCRIPT {start|stop|restart|reboot|ping|live|console|console_clean|attach|debug|version}"
+        echo "Usage: $SCRIPT {start|stop|restart|reboot|ping|live|console|console_clean|foreground|attach|debug|version}"
         exit 1
         ;;
 esac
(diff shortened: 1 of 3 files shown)


arcusfelis:
Live mode is the same as console mode. It's for "live" interactions. I.e. you can't just disable shell and console input.
Live name has come from ejabberd world, console - from rebar world. The change would be confusing for people, who use live mode for dev.

I would suggest to add a new 'foreground' mode, that would run erl with these flags.
And also add information about it into our documentation in doc/ directory.

GalaxyGorilla:
@arcusfelis OK, thanks for clearing that up. Actually the 'live' mode I have in my umbrella app is working differently (because the wrapper script are different from upstream MIM), that's why I thought in another direction. I will adjust this MR to introduce a 'foreground' mode :).

GalaxyGorilla:
I updated the PR accordingly. Is the provided docu enough? I also changed some description for the other modes slightly so that the difference between them is hopefully clear.

GalaxyGorilla:
Puh, can s1 decypher whats going on with the CI? ;)

rel/files/mongooseim
```diff
@@ -157,7 +164,13 @@ case "$1" in
         logger -t "$SCRIPT[$$]" "Starting up"
 
         # Start the VM
-        exec_echo "$BINDIR/erlexec" -boot "$RUNNER_BASE_DIR/releases/$APP_VSN/$BOOTFILE" -embedded -config "$RUNNER_ETC_DIR/app.config" -args_file "$RUNNER_ETC_DIR/vm.args" -args_file "$RUNNER_ETC_DIR/vm.dist.args" -- ${1+"$@"}
+        exec_echo "$BINDIR/erlexec" $FOREGROUNDOPTIONS \
```
arcusfelis: Something wrong with embedded argument. 
Wondering, if erl wants the argument before any config argument. Will look deeper on Monday.

arcusfelis: This exact commit is needed to fix CI
https://github.com/arcusfelis/MongooseIM/pull/33/commits/58ee78fafba2474a421ba5478533b93b3144faa4

arcusfelis: So, cherry-pick the commit or apply the changes, push, let's wait for CI to pass on the updated build.
Once it's ready - we can merge it.

arcusfelis:
Thanks!

PR #1776, opened 2018-03-20
Improve room http api

michalwski:
With this PR MongooseIM will respond with the following error for room's message sending via client's HTTP API
* `400 Bad Request` if the request body is not valid:
    * the body is not JSON
    * or doesn't have `body` element
    * or the `body` element is not string
* `403 Forbidden` if the user sending the message is not member of the room

### DIFF
--- a/src/mongoose_client_api_rooms_messages.erl
+++ b/src/mongoose_client_api_rooms_messages.erl
@@ -73,6 +73,8 @@ to_json(Req, #{jid := UserJID, room := Room} = State) ->
     JSONData = [make_json_item(Msg) || Msg <- Msgs],
     {jiffy:encode(JSONData), Req3, State}.
 
+from_json(Req, #{role_in_room := none} = State) ->
+    mongoose_client_api:forbidden_request(Req, State);
 from_json(Req, #{user := User, jid := JID, room := Room} = State) ->
     {ok, Body, Req2} = cowboy_req:body(Req),
     try
--- a/src/mongoose_client_api_rooms_messages.erl
+++ b/src/mongoose_client_api_rooms_messages.erl
@@ -43,7 +43,7 @@ allow_missing_post(Req, State) ->
     {false, Req, State}.
 
 to_json(Req, #{role_in_room := none} = State) ->
-    mongoose_client_api_rooms:forbidden_request(Req, State);
+    mongoose_client_api:forbidden_request(Req, State);
 to_json(Req, #{jid := UserJID, room := Room} = State) ->
     RoomJID = maps:get(jid, Room),
     Server = UserJID#jid.server,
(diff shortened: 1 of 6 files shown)


src/mongoose_client_api_rooms_messages.erl
```diff
@@ -73,16 +73,50 @@ to_json(Req, #{jid := UserJID, room := Room} = State) ->
     JSONData = [make_json_item(Msg) || Msg <- Msgs],
     {jiffy:encode(JSONData), Req3, State}.
 
+from_json(Req, #{role_in_room := none} = State) ->
+    mongoose_client_api:forbidden_request(Req, State);
```
fenek: Same as above.

src/mongoose_client_api_rooms_users.erl
```diff
@@ -51,10 +51,10 @@ from_json(Req, #{user := User,
                                               UserToInvite, <<"member">>),
     {true, Req2, State};
 from_json(Req, State) ->
-    mongoose_client_api_rooms:forbidden_request(Req, State).
+    mongoose_client_api:forbidden_request(Req, State).
 
 delete_resource(Req, #{role_in_room := none} = State) ->
-    mongoose_client_api_rooms:forbidden_request(Req, State);
+    mongoose_client_api:forbidden_request(Req, State);
```
fenek: 404?

src/mongoose_client_api_rooms_users.erl
```diff
@@ -51,10 +51,10 @@ from_json(Req, #{user := User,
                                               UserToInvite, <<"member">>),
     {true, Req2, State};
 from_json(Req, State) ->
-    mongoose_client_api_rooms:forbidden_request(Req, State).
+    mongoose_client_api:forbidden_request(Req, State).
```
fenek: Separate `from_json` clause for `role_in_room := none` with 404?

src/mongoose_client_api_rooms_messages.erl
```diff
@@ -43,7 +43,7 @@ allow_missing_post(Req, State) ->
     {false, Req, State}.
 
 to_json(Req, #{role_in_room := none} = State) ->
-    mongoose_client_api_rooms:forbidden_request(Req, State);
+    mongoose_client_api:forbidden_request(Req, State);
```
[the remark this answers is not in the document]
michalwski: Yes, from security and consistency with XMPP interface perspective 404 is better. From development perspective when the client needs to know if the request was sent to existing room, but the caller is not member of, returning 404 in both cases doesn't make things easier...

src/mongoose_client_api_rooms_users.erl
michalwski: It was 403 here before already :) If you want I can change to 404.

fenek: PR is about "improve", not "refactor", so I think it's worth including in the same branch. :)

src/mongoose_client_api_rooms_messages.erl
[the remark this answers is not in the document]
fenek: Hmm.... okaaaay, I understand how it could make REST client's life easier. A bit unfortunate but I agree - 403 should stay, at least for now.

src/mongoose_client_api_rooms_users.erl
michalwski: Good question, if we return 403 if user is not member and tries to post I think we should also return 403 in this case. If we choose to return 404 here, I think 404 should be returned in the other case as well.

michalwski:
@fenek I've just rebased, updated swagger spec and addressed on comment to the code. What is the final decision about the 404 vs 403 for non-members trying to do sth with the room?

arkgil:
AFAIR MUC light over XMPP returns `item-not-found` for non-members, so we could return 404 here for consistency (although I don't know if that does make sense).

michalwski:
Re-opening to trigger travis.

PR #1782, opened 2018-03-26
Tidy test helpers

kanes115:
This PR extracts common functions to helpers so that no SUITE calls another's SUITE's function.


kanes115:
TODO:
- [ ] export only public functions (in modules that have `-compile(export_all)`

kanes115:
When deleting user, the user cannot be authorised now. It manifests in end_per_groups.

kanes115:
IMPORTANT: Before merging we need to merge changes in escalus and chage rebar.config in mongoose's tests accordingly !!!

big_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -586,8 +587,8 @@ test_pm_with_ungraceful_reconnection_to_different_server(Config0) ->
               escalus:assert(is_chat_message, [<<"Hi from Europe1!">>], FromAlice),
               escalus:assert(is_chat_message, [<<"Hi again from Europe1!">>], AgainFromAlice),
               escalus:assert(is_chat_message, [<<"Hi from Asia!">>], AgainFromEve)
-      end),
-    escalus_users:delete_users(Config, [{eve, [{port, 5222} | EveSpec]}]).
```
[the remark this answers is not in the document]
kanes115: Escalus fresh deletes users on its own as well in end_per_*. This line caused some trouble with double-deleting user.

big_tests/tests/mongoose_helper.erl
```diff
@@ -21,8 +21,10 @@
 -export([ensure_muc_clean/0]).
 -export([successful_rpc/3]).
 -export([logout_user/2]).
+-export([wait_until/3, wait_until/4]).
 
 -define(RPC(M, F, A), escalus_ejabberd:rpc(M, F, A)).
```
fenek: This macro is not used now, right?

kanes115: sure

kanes115:
@fenek ping

fenek:
You've mentioned "export only public functions (in modules that have `-compile(export_all)`" as one of TODOs. Some helper modules still use it. Do you plan to improve it as well or would you prefer to leave it for next sprint?

kanes115:
@fenek, I would prefer not to loose focus on my current tasks and eradicate those export_alls in the next sprint. How about that?

PR #1787, opened 2018-03-30
Bump MySQL database driver version

varnerac:
The latest release includes useful changes, like support for recent MariaDB versions.


fenek:
Hi @varnerac 

Your PR has to update `rebar.lock` file as well. :)

kzemek:
The easiest way to do that is to execute `./rebar3 upgrade mysql` in the mongooseim directory.

PR #1790, opened 2018-04-04
[do not merge] Check codecov integration

fenek:
don't merge


fenek:
Codecov integration works.

PR #1797, opened 2018-04-11
Jingle SIP

michalwski:
This PR adds a jingle sip translator in both ways:
1. When Mongoose receives a Jingle stanzas from a user it translates it to a SIP message and sends to configured SIP Proxy
2. When Mongoose receives a SIP message form a SIP Proxy, it translates it to a jingle stanza and delivers to the user

Additionally the macro `VERSION` is renamed to `MONGOOSE_VERSION` because nksip.hrl needed for this feature also defines macro of the same name.


big_tests/tests/push_helper.erl
```diff
@@ -65,15 +65,17 @@ become_unavailable(Client) ->
...
+                   escalus_utils:jid_to_lower(escalus_client:server(Client)),
+                   escalus_utils:jid_to_lower(escalus_client:resource(Client)))
     end). %% There is no ACK for unavailable status
 
-is_offline(LUser, LServer) ->
-    case catch lists:max(escalus_ejabberd:rpc(ejabberd_sm, get_user_present_pids, [LUser, LServer])) of
-        {Priority, _} when is_integer(Priority), Priority >= 0 ->
-            false;
+is_offline(LUser, LServer, LRes) ->
+    Resources =  escalus_ejabberd:rpc(ejabberd_sm, get_user_present_resources, [LUser, LServer]),
```
arcusfelis: Variable should be `PResources` or `PriorityResources`.
Than I know, why we need to call keyfind for element 2.

rebar.config
```diff
@@ -1,6 +1,7 @@
 {erl_opts, [debug_info,
             {i, ["include"]},
             {d, xml_nif},
+            {platform_define, "^18", no_typed_maps},
```
arcusfelis: what is no_typed_maps here?

tools/configure
```diff
@@ -44,6 +44,7 @@ app_opts() ->
      {"pgsql",     ["epgsql"],          include_driver("pgsql")},
      {"redis",     ["redo"],            include_driver("redis")},
      {"riak",      ["riakc"],           include_driver("riak")},
+     {"jingle-sip",["nksip"],           include_driver("nksip")},
```
arcusfelis: make sense to change:

```erlang
include_driver(DB) ->                                                                
    ["include ", DB, " driver"].
```

to

```erlang
include_driver(Driver) ->                                                                
    ["include ", Driver, " driver"].
```

big_tests/tests/push_helper.erl
```diff
@@ -65,15 +65,17 @@ become_unavailable(Client) ->
...
+    case lists:keyfind(LRes, 2, Resources) of
```
arcusfelis: lists:keymember?

src/jingle_sip/mod_jingle_sip_backend.erl
```diff
@@ -0,0 +1,199 @@
...
+                             state,
+                             direction,
+                             request,
+                             node,
+                             meta}).
+
+init(_Host, _Opts) ->
+    mnesia:create_table(jingle_sip_session,
+                        [{ram_copies, [node()]},
+                         {attributes, record_info(fields, jingle_sip_session)}]).
```
arcusfelis: add_local_copy?

src/jingle_sip/sip_to_jingle.erl
```diff
@@ -0,0 +1,300 @@
...
+                            {<<"value">>, Value}]};
+        [Value] ->
+            #xmlel{name = <<"parameter">>,
+                   attrs = [{<<"value">>, Value}]}
+    end.
+
+parse_nksip_media_attrs(Attrs) ->
+    parse_nksip_media_attrs(Attrs, #{}).
+
+parse_nksip_media_attrs([], Acc) ->
```
arcusfelis: Can we try

```erlang
+parse_nksip_media_attrs([], Acc) ->
+    Acc;
+parse_nksip_media_attrs([{AttrName, AttrValue} | Rest], Acc) ->
+    NewAcc = parse_nksip_media_attr(AttrName, AttrValue, Acc),
+    parse_nksip_media_attrs(Rest, NewAcc).
```

instead?

Than parse_nksip_media_attr/3 would be more readable

```erlang
+parse_nksip_media_attr(<<"sendrecv">>, _, Acc) ->
+    Acc#{sender => <<"both">>};
+parse_nksip_media_attr(<<"recvonly">>, _, Acc) ->
+    Acc#{sender => <<"responder">>};
+parse_nksip_media_attr(<<"sendonly">>, _, Acc) ->
+    Acc#{sender => <<"initiator">>};
```

big_tests/tests/jingle_SUITE.erl
```diff
@@ -0,0 +1,495 @@
...
+
+end_per_group(_GroupName, Config) ->
+    Config.
+
+init_per_testcase(CaseName, Config) ->
+    escalus:init_per_testcase(CaseName, Config).
+
+end_per_testcase(CaseName, Config) ->
+    escalus:end_per_testcase(CaseName, Config).
+
```
arcusfelis:
```erlang
+%% Cases
```

big_tests/tests/push_helper.erl
michalwski: Not really, the first element of the tuple is priority which we don't know (easily) when calling here. It's enough to check if the resource we ask for is removed from the list.

src/jingle_sip/mod_jingle_sip_backend.erl
michalwski: Good question, I'm quite sure it's not needed. I'll double check.

big_tests/tests/jingle_SUITE.erl
```diff
@@ -0,0 +1,495 @@
...
+    end).
+
+
+
+jingle_session_is_established_and_terminated_by_initiator(Config) ->
+    escalus:fresh_story(Config, [{alice, 1}, {bob, 1}], fun(Alice, Bob) ->
+        {InviteStanza, InviteRequest} = initiate_jingle_session(Alice, Bob),
+
+        accept_jingle_session(Alice, Bob, InviteStanza, InviteRequest),
+        timer:sleep(1000),
```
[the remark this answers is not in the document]
michalwski: Just to simulate that the following action doesn't happen immediately.

src/jingle_sip/mod_jingle_sip_backend.erl
michalwski: During the test, when node2 joined the `mnesia:info/0` showed:
```running db nodes   = [mongooseim@localhost,ejabberd2@localhost]
stopped db nodes   = []
master node tables = []
remote             = []
ram_copies         = [acl,anonymous,bosh_session,config,external_component,
                      external_component_global,iq_response,
                      jingle_sip_session,local_config,mod_register_ip,node,
                      reg_users_counter,route,s2s,session,sm_session]```
Looks like the `jingle_sip_session` is in local ram copies.

File: src/jingle_sip/mod_jingle_sip_backend.erl [hunk 9]
```diff
@@ -0,0 +1,199 @@
...
+                                          dialog = undefined,
+                                          state = undefined,
+                                          direction = in,
+                                          node = node(),
+                                          meta = Meta},
+            mnesia:write(Session)
+    end.
+
+set_outgoing_request(CallID, ReqID, From, To) ->
+    TFun = pa:bind(fun set_outgoing_request_tr/4, CallID, ReqID, From, To),
```
arcusfelis: The version with `fun() -> ... end` would still be dialyzer type transparent. This means you can find more potential errors with dialyzer.

```erlang
TFun = fun() -> set_outgoing_request_tr(CallID, ReqID, From, To) end,
run_transaction(TFun).
```

Using `pa` can be useful, if we want to use partial application, which is not the case here.

michalwski: I simply try to avoid unnamed funs, that's the reason I used `pa` here.

src/jingle_sip/mod_jingle_sip_backend.erl
```diff
@@ -0,0 +1,214 @@
...
+init(_Host, _Opts) ->
+    mnesia:create_table(jingle_sip_session,
+                        [{ram_copies, [node()]},
+                         {attributes, record_info(fields, jingle_sip_session)}]).
+
+-spec set_incoming_request(CallID :: call_id(), ReqID :: binary(),
+                           From :: jid:jid(), To :: jid:jid(), exml:element()) ->
+    ok | {error, any()}.
+set_incoming_request(CallID, ReqID, From, To, JingleEl) ->
+    TFun = pa:bind(fun set_incoming_request_tr/5, CallID, ReqID, From, To, JingleEl),
```
arcusfelis: pa:bind way of passing arguments would disable some dializer checks.

Compare:

```erlang
    -module(pa_check).  
    -export([call_my_fun/0]).  
    -export([call_pa_fun/0]).                                                              

    call_my_fun() ->                                                                       
        my_fun(1).                                                                         

    call_pa_fun() ->                                                                       
        pa_fun(1).                                                                         

    my_fun(A) ->                                                                         
        mnesia:transaction(fun() -> my_fun_tr(A) end).                                   

    pa_fun(A) ->                                                                         
        mnesia:transaction(pa:bind(fun my_fun_tr/1, A)).                                 

    my_fun_tr(x) ->                                                                      
        ok. 
```

and code

```erlang
    _build/default/lib/mongooseim/src/pa_check.erl                                                                                                                                
      12: The call pa_check:my_fun_tr(A::1) will never return since it differs in the 1st argument from the success typing arguments: ('x')  
```

I don't see why we need it here.

src/jingle_sip/jingle_sip_callbacks.erl
```diff
@@ -0,0 +1,276 @@
...
+        false ->
+            translate_and_deliver_invite(Req, FromJID, FromBinary, ToJID, ToBinary);
+        _ ->
+            {reply, temporarily_unavailable}
+    end.
+
+translate_and_deliver_invite(Req, FromJID, FromBinary, ToJID, ToBinary) ->
+    CallID = nksip_sipmsg:header(<<"call-id">>, Req),
+    Body = nksip_sipmsg:meta(body, Req),
+
```
arcusfelis: I am not a fun of empty lines inside a function.

src/jingle_sip/jingle_sip_callbacks.erl
```diff
@@ -0,0 +1,276 @@
...
+
+sip_info(Req, _Call) ->
+    {FromJID, FromBinary} = get_user_from_sip_msg(from, Req),
+    {ToJID, ToBinary} = get_user_from_sip_msg(to, Req),
+
+    CallID = nksip_sipmsg:header(<<"call-id">>, Req),
+    Body = nksip_sipmsg:meta(body, Req),
+
+    ?INFO_MSG("Info to: ~p in: ~p,~nbody: ~s", [ToBinary, CallID, Body]),
```
arcusfelis: We usually use this format:

```erlang
"event=sip_info to=~p call_id=~p body=~s"
```

src/jingle_sip/jingle_sip_callbacks.erl
```diff
@@ -0,0 +1,276 @@
...
+    CallID = nksip_sipmsg:header(<<"call-id">>, SIPMsg),
+
+    ReasonEl = #xmlel{name = <<"reason">>,
+                      children = [#xmlel{name = <<"decline">>}]},
+    JingleEl = jingle_sip_helper:jingle_element(CallID, <<"session-terminate">>, [ReasonEl]),
+    IQEl = jingle_sip_helper:jingle_iq(ToBinary, FromBinary, JingleEl),
+    Acc = mongoose_acc:from_element(IQEl, FromJID, ToJID),
+    maybe_route_to_all_sessions(FromJID, ToJID, Acc, IQEl),
+    ok;
+invite_resp_callback({resp, ErrorCode, SIPMsg, _Call}) when ErrorCode >= 400, ErrorCode =< 600 ->
```
arcusfelis: `ErrorCode >= 400, ErrorCode =< 600` good to make a comment why this range.

arcusfelis: because the comment above says:
```erlang
all error responses between 400 and 700 result in genering session-terminate reason
```

fenek:
Dialyzer error. :(

src/jingle_sip/sip_to_jingle.erl
michalwski: Done.

big_tests/tests/jingle_SUITE.erl
michalwski: Done.

src/jingle_sip/jingle_sip_callbacks.erl
michalwski: They sometimes do organise code to be me a bit more readable :)

src/jingle_sip/jingle_sip_callbacks.erl
michalwski: good catch!

PR #1798, opened 2018-04-11
Fixes to MUC configuration cancellation and form submission handling

arkgil:
Fix handling of form submits and cancels in MUC rooms

Previously any cancel request sent at any time would trigger room destruction. Additionally, form with non-default values could be submitted without fetching the form first (which is disallowed in the XEP).

The new behaviour is as follows:
* cancel IQs sent when room is in the locked state destroy the room ([link](https://xmpp.org/extensions/xep-0045.html#example-162))
* cancel IQs sent when the room is unlocked, but before fetching the
  form return an error
* cancel IQs sent when the room is unlocked, but after fetching the
  form return an IQ result and don't modify any values in the room ([link](https://xmpp.org/extensions/xep-0045.html#example-166))
* submit IQs without modified values sent when the room is locked
  create an instant room ([link](https://xmpp.org/extensions/xep-0045.html#createroom-instant))
* submit IQs with modified values sent before fetching the form return
  an error
* submit IQs with modified values sent when the room is locked, but
  after fetching the form, create a reserved room ([link](https://xmpp.org/extensions/xep-0045.html#createroom-reserved))
* submit IQs with modififed valued sent when the room is unlocked, but
  after fetching the form, reconfigure the room ([link](https://xmpp.org/extensions/xep-0045.html#roomconfig))

While fixes to cancel IQ handling are necessary and need to be applied (because the rooms were being destroyd unexpectedly), changes made to form submission handling are subject to the discussion. Now, each time a client wants to reconfigure the room, they have to fetch the form first, and only submit it then. This might give significant performance overhead when configuration is changed frequently. On the other hand, these changes make our implementation more in line with the [MUC XEP](https://xmpp.org/extensions/xep-0045.html).

### DIFF
--- a/big_tests/tests/muc_SUITE.erl
+++ b/big_tests/tests/muc_SUITE.erl
@@ -3446,6 +3446,7 @@ cancel_iq_sent_to_locked_room_destroys_it(Config) ->
         Nick = <<"the-owner">>,
 
         PresenceResp = escalus:send_and_wait(Alice, stanza_muc_enter_room(Room, Nick)),
+        escalus:wait_for_stanza(Alice),
         ?assert(is_presence_with_affiliation(PresenceResp, <<"owner">>)),
         ?assert(is_presence_with_role(PresenceResp, <<"moderator">>)),
 
@@ -3461,6 +3462,7 @@ cancel_iq_sent_to_unlocked_room_is_unexpected(Config) ->
         Nick = <<"the-owner">>,
 
         PresenceResp = escalus:send_and_wait(Alice, stanza_muc_enter_room(Room, Nick)),
+        escalus:wait_for_stanza(Alice),
         ?assert(is_presence_with_affiliation(PresenceResp, <<"owner">>)),
         ?assert(is_presence_with_role(PresenceResp, <<"moderator">>)),
 
@@ -3481,6 +3483,7 @@ reserved_room_cant_be_created_before_form_is_fetched(Config) ->
         Nick = <<"the-owner">>,
 
         PresenceResp = escalus:send_and_wait(Alice, stanza_muc_enter_room(Room, Nick)),
+        escalus:wait_for_stanza(Alice),
         ?assert(is_presence_with_affiliation(PresenceResp, <<"owner">>)),
         ?assert(is_presence_with_role(PresenceResp, <<"moderator">>)),
 
@@ -3497,6 +3500,7 @@ room_cant_be_reconfigured_before_form_is_fetched(Config) ->
         Nick = <<"the-owner">>,
 
         PresenceResp = escalus:send_and_wait(Alice, stanza_muc_enter_room(Room, Nick)),
+        escalus:wait_for_stanza(Alice),
         ?assert(is_presence_with_affiliation(PresenceResp, <<"owner">>)),
         ?assert(is_presence_with_role(PresenceResp, <<"moderator">>)),
 
--- a/big_tests/tests/muc_SUITE.erl
+++ b/big_tests/tests/muc_SUITE.erl
@@ -20,6 +20,7 @@
 -include_lib("escalus/include/escalus.hrl").
 -include_lib("escalus/include/escalus_xmlns.hrl").
 -include_lib("common_test/include/ct.hrl").
+-include_lib("eunit/include/eunit.hrl").
 -include_lib("exml/include/exml.hrl").
 -include("assert_received_match.hrl").
 
@@ -236,7 +237,11 @@ groups() -> [
                 configure,
                 configure_logging,
                 %% fails, see testcase
-                configure_anonymous
+                configure_anonymous,
+                cancel_iq_sent_to_locked_room_destroys_it,
+                cancel_iq_sent_to_unlocked_room_is_unexpected,
+                reserved_room_cant_be_created_before_form_is_fetched,
+                room_cant_be_reconfigured_before_form_is_fetched
                 ]},
         {owner_no_parallel, [], [
                                  room_creation_not_allowed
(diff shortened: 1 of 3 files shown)


arcusfelis:
> Now, each time a client wants to reconfigure the room, they have to fetch the form first, and only submit it then.

There is no information in XEP, what should we do if we just send configuration form without fetching it.
It's pretty often use case to send a form to change only one specific configuration parameter (for example, room's title).

I feel that `configuration_initiated` flag only introduces complexity.

arkgil:
@arcusfelis I agree, that was my main concern. I'll happily change it back, but I'd like to have input from more people first. @fenek @michalwski maybe?

arkgil:
@arcusfelis Although when creating a room, the XEP states that the user *MUST* fetch the form and only then submit it, quoting:

> If the initial room owner would like to create and configure a reserved room, the room owner MUST then request a configuration form by sending an IQ stanza of type "get" to the room containing an empty <query/> element qualified by the 'http://jabber.org/protocol/muc#owner' namespace, then complete Steps 4 and 5. If the room owner would prefer to create an instant room, the room owner MUST send a query element qualified by the 'http://jabber.org/protocol/muc#owner' namespace and containing an empty <x/> element of type "submit" qualified by the 'jabber: x:data' namespace, then skip to Step 6.

fenek:
In my opinion the current implementation doesn't "offend" the XEP in awful way. It is in fact an optimisation and it doesn't force any client to skip form fetch. I think only the issue with `cancel` should be fixed.

arcusfelis:
@arkgil okey, must request than must. Still we used not to ask for it :)

big_tests/tests/muc_SUITE.erl
```diff
@@ -1507,6 +1512,7 @@ admin_member_list_allowed(Config) ->
         %% Kate can't get member list
         check_memberlist(Kate, no, Config),
         %% setup room - allow getmemberlist for moderator
+        escalus:send_and_wait(Alice, stanza_configuration_form_request(?config(room, Config))),
```
arcusfelis: We should check, that we received what we supposed to receive.
I propose to have a function send_iq_and_wait_for_result/1.
That would:
- check that IQ result id matches IQ request id
- handle error and print both request and error as a readable error.

Motivation: if a service is broken, it would return error service-unavailable and we just continue the test execution.
From time to time there are "unexpected" presences routed in our tests, that also can be consumed by send_and_wait without any assertion.

src/mod_muc_room.erl
```diff
@@ -3211,32 +3212,45 @@ send_kickban_presence1(UJID, Reason, Code, Affiliation, StateData) ->
...
+process_authorized_iq_owner(From, set, Lang, SubEl, StateData, StateName) ->
     #xmlel{children = Els} = SubEl,
     case xml:remove_cdata(Els) of
         [#xmlel{name = <<"x">>} = XEl] ->
             case {xml:get_tag_attr_s(<<"xmlns">>, XEl),
-                  xml:get_tag_attr_s(<<"type">>, XEl)} of
-                {?NS_XDATA, <<"cancel">>} ->
+                  xml:get_tag_attr_s(<<"type">>, XEl),
+                  StateName,
+                  StateData#state.configuration_initiated} of
```
arcusfelis: ok, so if we WANT to keep it.
there can be more than one room administrator.

configuration_initiated should be PER room administrator device jid (i.e. alice@localhost/res1).
we also should reset configuration_initiated, when the administrator leaves the room (it's just logical, not something described in XEP).

Than, extra cases that should be tested and documented.

Admin asks for the form several times.
Admin sends the same form several times.
Admin sends the form once. Then sends the form again with different data.
There are two admins, one initiates configuration. one sends the form (AND VARIATIONS).
There are two admins, both fetched form and both submitted (last write wins? or do we want to check that the form is not expired yet?).

All these cases should be documented.
It's actually the main reason, why there was no check for the room.
Because any extra "if" should be tested.

Error reporting should be consistent and contain description, in which there should be information, that allows to find corresponding XEP part AND corresponding code in MongooseIM.

The problem that the error has no example in the XEP, so we should say something like "Form MUST be fetched before being submitted. See xep-0045, version 1.31.1. 10.1 Creating a Room => 10.1.1 General Considerations => workflow step 3.".

We need that step, because not all front-end developers read XEP, they just use libraries, which can have the form fetching step missing.

arkgil:
@arcusfelis @fenek Alright, I'll revert back to not requiring the config form to be fetched. Indeed, it is more sane to allow users to configure the room right away. Also, it will simplify the implementation a bit.

big_tests/tests/muc_SUITE.erl
```diff
@@ -3442,6 +3440,42 @@ configure_anonymous(Config1) ->
...
+        ?assert(is_presence_with_role(PresenceResp, <<"moderator">>)),
+
+        InstantRoomIq = stanza_instant_room(Room),
+        InstantRoomIqResp = escalus:send_and_wait(Alice, InstantRoomIq),
+        escalus:assert(is_iq_result, [InstantRoomIq], InstantRoomIqResp),
+        RoomsIqResp = escalus:send_and_wait(Alice, stanza_room_list_request(<<"id">>, undefined)),
+        has_room(room_address(Room), RoomsIqResp),
+
+        CancelIqResp = escalus:send_and_wait(Alice, stanza_cancel(Room)),
+        escalus:assert(is_iq_result, [CancelIqResp], CancelIqResp)
```
arcusfelis: It's optional.

Still functional means that it's still in normal state and we can join, leave and send messages (+ more checks if needed).
The check can go into a helper function `assert_room_is_functional`.

src/mongoose_xmpp_errors.erl
```diff
@@ -91,6 +91,9 @@ unexpected_request() ->
 unexpected_request_cancel() ->
     jlib:stanza_error(<<"401">>,<<"cancel">>,<<"unexpected-request">>).
 
+unexpected_request_modify() ->
```
arcusfelis: Do we really need that function?

arkgil:
@arcusfelis Yep, this PR will grow too much. Let's do this in another one if there will be need/will to do so.

arkgil:
So this [single job](1) fails constantly on Travis and always passes locally for me. I need to try debug-build it.

[1]: https://travis-ci.org/esl/MongooseIM/jobs/370490331

arcusfelis:
@arkgil which? mysql job? 
Yep, it got broken, after we upgraded mysql on travis. For some reason.

arkgil:
@arcusfelis yep, but it's the same, single test case failing every time. I am not sure if it is related to MySQL version changes.

The failing test is `mod_blocking_SUITE:messages_arrive_after_unblock_and_relogin/1`.

fenek:
@arkgil 

Seems unlikely to be affected by your changes but if it manifests in this PR for some reason, let's try to find a fix in a branch based on yours.

PR #1802, opened 2018-04-14
Allow to enable preset on just some nodes

arcusfelis:
Allows to start just some nodes for tests:

```erlang
./tools/configure with-all
DB=mysql ./tools/travis-setup-db.sh
make mim1
%% just run with one node
%% be aware, that tests, that need other nodes would fail :(
%% but I usually test with modified test spec file, that does not have all the tests enabled.
TEST_HOSTS="mim" DEV_NODES="mim1" ./tools/travis-test.sh -c false -s false -e false -p mysql_mnesia
```

or

```erlang
TEST_HOSTS="mim fed" DEV_NODES="mim1 fed1" ./tools/travis-test.sh -c false -s false -e false -p mysql_mnesia
```

Without this PR we need to run all nodes to run a test.

### DIFF
--- a/big_tests/run_common_test.erl
+++ b/big_tests/run_common_test.erl
@@ -1,3 +1,5 @@
+%% During dev you would use something similar to:
+%% DEV_NODES="mim1" ./tools/travis-test.sh -e false -c false -s false -p odbc_mssql_mnesia
 -module(run_common_test).
 
 -export([main/1, analyze/2]).
@@ -182,7 +184,7 @@ run_config_test({Name, Variables}, Test, N, Tests) ->
 enable_preset(Name, PresetVars, Test, N, Tests) ->
     %% TODO: Do this with a multicall, otherwise it's not as fast as possible (not parallelized).
     %%       A multicall requires the function to be defined on the other side, though.
-    Rs = [ enable_preset_on_node(host_node(H), PresetVars, host_vars(H))
+    Rs = [ maybe_enable_preset_on_node(host_node(H), PresetVars, host_vars(H), rel_name(H))
            || H <- get_hosts(Test) ],
     [ok] = lists:usort(Rs),
     error_logger:info_msg("Configuration ~p of ~p: ~p started.~n",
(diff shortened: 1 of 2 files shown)


big_tests/run_common_test.erl
```diff
@@ -420,6 +443,8 @@ group_by(F, L) ->
 host_cluster(Host) -> host_param(cluster, Host).
 host_node(Host)    -> host_param(node, Host).
 host_vars(Host)    -> host_param(vars, Host).
+%% Directory name in _build
+rel_name(Host)     -> host_param(relname, Host).
```
fenek: Perhaps we could simply rename keys in `hosts` so additional param won't be necessary at all?

arcusfelis: I was thinking about it.

We can't really.
Because relname is not nodename.

Rebar allows to build a mix of presets, so the name would be something like "preset1+preset2".

I would rather keep relname as a separate entity from any other key, we currently have in hosts (i.e. node, vars, cluster, domain and some more).

arcusfelis: also, in DEV_NODES we specify relnames.
and the whole meaning of DEV_NODES is "enable preset on a node in the release directory. Use hosts key in tests.config to figure out which node is which".

BTW, cool feature in presets. We can actually build releases like "mim1+with_all", if we create with_all preset.

arcusfelis: I will think, what changes are required to be able to specify HOSTS in DEV_NAMES, instead of relnames.

fenek: OK, I understand. Indeed release names could change and it would be pretty inconvenient to change these names in all tests that use node names. :)

fenek:
Is it still WIP?

big_tests/run_common_test.erl
arcusfelis: Ah, we already use DEV_NODES in `tools/start-nodes.sh` and it wants build names.
There is no easy way to access test.config from bash.
That was the reasoning behind adding an extra key.

arcusfelis: And it's not easy to write testhostname_to_relname and relname_to_testhostname functions without explicitly defining mapping. 

So, I would rename fed test host into fed1 and reg into reg1.

big_tests/run_common_test.erl
```diff
@@ -1,3 +1,18 @@
+%% During dev you would use something similar to:
+%% TEST_HOSTS="mim1" ./tools/travis-test.sh -e false -c false -s false -p odbc_mssql_mnesia
```
arcusfelis: `-e` flag is missing YET. But will appear after #1816 is merged.

arcusfelis:
@fenek it's waiting for another round of review.

PR #1803, opened 2018-04-14
Ignore stanzas addressed to an old c2s process

arcusfelis:
This PR addresses build issues.

Proposed changes include:
* Introduce ejabberd_c2s:process_incoming_stanza_with_conflict_check.
* Fix a race-condition in privacy_SUITE tests.
* New accumulator props sender_sid and sender_jid.

```erlang
13:13:57.921 [error] event=conflict_check_failed jid=alicE@localhost/res1 
c2s_sid={{1523,704437,851989},<0.2078.0>} sender_sid={{1523,704437,409824},<0.2074.0>} 
acc=#{attrs => [{<<"from_sender_element">>,<<"1">>},{<<"type">>,<<"unavailable">>}],
element => {xmlel,<<"presence">>,[{<<"from_sender_element">>,<<"1">>},{<<"type">>,<<"unavailable">>}],[]},from => <<"alicE@localhost/res1">>,
from_jid => {jid,<<"alicE">>,<<"localhost">>,<<"res1">>,<<"alice">>,<<"localhost">>,<<"res1">>},mongoose_acc => true,name => <<"presence">>,persistent_properties => [{sender_jid,{jid,<<"alicE">>,<<"localhost">>,<<"res1">>,<<"alice">>,<<"localhost">>,<<"res1">>}},{sender_sid,{{1523,704437,409824},<0.2074.0>}}],
ref => #Ref<0.2819752192.2344615938.90969>,server => <<"localhost">>,timestamp => {1523,704437,818568},type => <<"unavailable">>,user => <<"alicE">>}
```

We often use non-fresh users in our tests.
Sometimes, we have two stories one after another.
In such cases, connections from a test earlier can interfere into work of the next test.
Usually, by sending presence unavailable or by replying on presence probe.

This test fixes sending presence unavailable to the same client.

Example, when the bug is happening:

```erlang                                                                  
activate(Config) ->                                                                  
    escalus:story(Config, [{alice, 1}, {bob, 1}], fun(Alice, Bob) ->                 
                                                                                     
        privacy_helper:set_list(Alice, {<<"deny_client">>, Bob}),                    
                                                                                     
        Request = escalus_stanza:privacy_activate(<<"deny_client">>),                
        escalus_client:send(Alice, Request),                                         
                                                                                     
        Response = escalus_client:wait_for_stanza(Alice),                            
        escalus:assert(is_iq_result, Response)                                       
                                                                                     
        end).                                                                        
                                                                                     
activate_nonexistent(Config) ->                                                      
    escalus:story(Config, [{alice, 1}], fun(Alice) ->                                
                                                                                     
        Request = escalus_stanza:privacy_activate(<<"some_list">>),                  
        escalus_client:send(Alice, Request),                                         
                                                                                     
        Response = escalus_client:wait_for_stanza(Alice),                            
        escalus:assert(is_error, [<<"cancel">>, <<"item-not-found">>], Response)     
                                                                                     
        end). 
```

There is also another type of bug, when your roster contact replies on your presence.
This can't be fixed in c2s module, we need to call 

```erlang
mongoose_helper:kick_everyone(), 
```

Another case the fix resolves is when in first case we send IQ-request, that is executed in a separate process (when iqdisc is parallel).
Than reply from first story can arrive into second story and cause unexpected stanza errors.

This PR has no test, but the test is possible :)
Test should be:

```
WHEN route stanza with the same user_jid and different user_sid.
THEN it should be dropped.
```

It's still "covered" by tests, because the check is triggered by CI several times during a build randomly.

### DIFF
--- a/doc/developers-guide/accumulators.md
+++ b/doc/developers-guide/accumulators.md
@@ -13,6 +13,9 @@ At the beginning of the main processing chain an accumulator is created containi
 * attrs - attributs of the root xml element of the stanza
 * from, to - full jids of the sender and recipient in binary form
 * from_jid, to_jid - same as #jid{} records.
+* sender_sid - session id (SID) of `ejabberd_c2s`. Set when we receive a stanza
+  from the user device. *property*, *optional*.
+* sender_jid - full sender JID. Set the same time as `sender_sid`. *property*, *optional*.
 
 It is then passed through all the stages until it reaches the end of its life.
 Throughout the process it is the very same accumulator; it is therefore possible to store a value in it on one stage of the processing and retrieve the same value later on.
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -455,7 +455,7 @@ wait_for_auth({xmlstreamelement,
                #xmlel{name = <<"enable">>} = El}, StateData) ->
     maybe_unexpected_sm_request(wait_for_auth, El, StateData);
 wait_for_auth({xmlstreamelement, El}, StateData) ->
-    Acc0 = mongoose_acc:from_element(El),
+    Acc0 = from_sender_element(El, StateData),
     User = StateData#state.user,
     Server = StateData#state.server,
     Acc = mongoose_acc:update(Acc0, #{user => User, server => Server}),
@@ -777,7 +777,7 @@ wait_for_session_or_sm({xmlstreamelement, El}, StateData0) ->
                                             StateData0),
     case jlib:iq_query_info(El) of
         #iq{type = set, xmlns = ?NS_SESSION} ->
-            Acc0 = mongoose_acc:from_element(El),
+            Acc0 = from_sender_element(El, StateData0),
             User = StateData#state.user,
             Server = StateData#state.server,
             Acc = mongoose_acc:update(Acc0, #{user => User, server => Server}),
@@ -965,7 +965,7 @@ session_established({xmlstreamelement, El}, StateData) ->
                                                    StateData),
             % initialise accumulator, fill with data
             El1 = fix_message_from_user(El, StateData#state.lang),
-            Acc0 = mongoose_acc:from_element(El1),
+            Acc0 = from_sender_element(El1, StateData),
             User = NewState#state.user,
             Server = NewState#state.server,
             To = exml_query:attr(El, <<"to">>),
@@ -1218,7 +1218,7 @@ handle_info({force_update_presence, LUser}, StateName,
                            StateData#state.pres_last,
                            [LUser, LServer]),
             StateData2 = StateData#state{pres_last = PresenceEl},
-            Acc = mongoose_acc:from_element(PresenceEl),
+            Acc = from_sender_element(PresenceEl, StateData),
             presence_update(Acc, StateData2#state.jid, StateData2),
             StateData2;
         _ ->
@@ -1246,7 +1246,7 @@ handle_info({store_session_info, User, Server, Resource, KV, _FromPid}, StateNam
 handle_info(Info, StateName, StateData) ->
     handle_incoming_message(Info, StateName, StateData).
 
-%%% incoming messages
+%%% incoming messages from other users or services to this device
 handle_incoming_message({send_text, Text}, StateName, StateData) ->
     % it seems to be sometimes, by event sent from s2s
     send_text(StateData, Text),
@@ -1271,7 +1271,7 @@ handle_incoming_message({route, From, To, Acc0}, StateName, StateData) ->
     Acc = setup_accum(Acc0, StateData),
     Acc1 = ejabberd_hooks:run_fold(c2s_loop_debug, Acc, [{route, From, To}]),
     Name = mongoose_acc:get(name, Acc1),
-    process_incoming_stanza(Name, From, To, Acc1, StateName, StateData);
+    process_incoming_stanza_with_conflict_check(Name, From, To, Acc1, StateName, StateData);
 handle_incoming_message({send_filtered, Feature, From, To, Packet}, StateName, StateData) ->
     % this is used by pubsub and should be rewritten when someone rewrites pubsub module
     Acc0 = mongoose_acc:new(),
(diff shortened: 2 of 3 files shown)


src/ejabberd_c2s.erl
```diff
@@ -3350,3 +3380,9 @@ setup_accum(Acc, StateData) ->
     Server = StateData#state.server,
     mongoose_acc:update(Acc, #{server => Server, user => User}).
 
+%% @doc This function is executed when c2s receives a stanza from TCP connection.
+-spec from_sender_element(jlib:xmlel(), StateData :: state()) -> mongoose_acc:t().
+from_sender_element(El, #state{sid = SID, jid = JID}) ->
+    Acc = mongoose_acc:from_element(El),
+    Acc1 = mongoose_acc:add_prop(sender_sid, SID, Acc),
+    mongoose_acc:add_prop(sender_jid, JID, Acc1).
```
michalwski: In my PR #1797 I put JID to `user_jid` key in the acc. Would be good to use the same name, I'm not sure which one is better suited.

src/ejabberd_c2s.erl
```diff
@@ -3350,3 +3380,9 @@ setup_accum(Acc, StateData) ->
     Server = StateData#state.server,
     mongoose_acc:update(Acc, #{server => Server, user => User}).
 
+%% @doc This function is executed when c2s receives a stanza from TCP connection.
+-spec from_sender_element(jlib:xmlel(), StateData :: state()) -> mongoose_acc:t().
+from_sender_element(El, #state{sid = SID, jid = JID}) ->
```
michalwski: Maybe, a better name for this function would be sth like: `acc_from_element_with_sender`

src/ejabberd_c2s.erl
```diff
@@ -1311,6 +1311,35 @@ handle_incoming_message(Info, StateName, StateData) ->
...
+%% The conflict can happen, when actions with an accumulator were initiated by
+%% one process but the resulting stanzas were routed to another process with
+%% the same JID but different SID.
+%% The conflict is usually happens when an user is reconnecting.
+-spec is_old_session_conflict(mongoose_acc:t(), state()) -> boolean().
+is_old_session_conflict(Acc, #state{sid = SID, jid = JID}) ->
+    SenderSID = mongoose_acc:get_prop(sender_sid, Acc),
+    SenderJID = mongoose_acc:get_prop(sender_jid, Acc),
+    (SenderJID =:= JID)
+    andalso (SenderSID =/= undefined)
```
michalwski: how about `not SenderSID == SID` instead of 2 checks?

src/ejabberd_c2s.erl
```diff
@@ -1311,6 +1311,35 @@ handle_incoming_message(Info, StateName, StateData) ->
...
+            finish_state(ok, StateName, StateData)
+    end.
+
+%% If jid is the same, but sid is not.
+%% jid example is alice@localhost/res1.
+%% sid example is `{now(), pid()}'.
+%% The conflict can happen, when actions with an accumulator were initiated by
+%% one process but the resulting stanzas were routed to another process with
+%% the same JID but different SID.
+%% The conflict is usually happens when an user is reconnecting.
```
michalwski: no need for `is` in this sentence.

michalwski:
> This PR has no test, but the test is possible :)

It be great to have reproducible  test for this instead of counting on travis-ci to trigger it randomly.

src/ejabberd_c2s.erl
arcusfelis: Offtop: It's what my problem with accs. Nobody knows how to use them. 
Ontop: sender_jid is a property, so it would be never stripped. And nobody would change it.
user_jid is an accumulator field, so it would be stripped, the moment we resend accumulator to another process.

These accumulator fields are not stripped `from, from_jid, to, to_jid, persistent_properties, global_distrib`. Properties are not stripped.

I can't use `from_jid`, because it's the same as `From :: #jid{}`. I.e. it's different for request and reply.

`sender_jid` semantics is that it would be the same for the whole live of accumulator. I.e. we set it once, and never touch it again.

We touch `from_jid`, when we reply.

michalwski: Thanks for the clarifications. Looks like I'll adjust my code in already merged #1797 to use `sender_jid` after this PR is merged. There is no need to have duplicate keys if they hold the same value.

src/ejabberd_c2s.erl
arcusfelis: `element_to_sender_acc`. Because we use to in `list_to_binary`-family of functions.

p.s.: I wanted to use `element_to_outgoing_acc`, but I have problems with outgoing and incoming atoms in our code. There is `incoming` message from a client connection socket. But `outgoing` message from c2s to be routed.

src/ejabberd_c2s.erl
arcusfelis: I've refactored the function.

doc/developers-guide/accumulators.md
```diff
@@ -13,6 +13,9 @@ At the beginning of the main processing chain an accumulator is created containi
 * attrs - attributs of the root xml element of the stanza
 * from, to - full jids of the sender and recipient in binary form
 * from_jid, to_jid - same as #jid{} records.
+* sender_sid - session id (SID) of `ejabberd_c2s`. Set when we receive a stanza
+  from the user device. *property*, *optional*.
+* sender_jid - full sender JID. Set the same time as `sender_sid`. *property*, *optional*.
```
arcusfelis: Need to decide for a better name:

- creator_jid
- c2s_jid
- intial_jid

src/ejabberd_c2s.erl
```diff
@@ -1246,7 +1246,7 @@ handle_info({store_session_info, User, Server, Resource, KV, _FromPid}, StateNam
 handle_info(Info, StateName, StateData) ->
     handle_incoming_message(Info, StateName, StateData).
 
-%%% incoming messages
+%%% incoming messages from other users or services to this device
```
fenek: Me Gusta! At last a clear definition in at least one place what do "incoming" and "outgoing" mean. :D

fenek:
As far as I understand, it is only a partial solution that applies to stanzas that are routed as replies sent by the server but won't help in case of messages sent between users, correct?

Wouldn't it make more sense to add a property named `destination_sid` and have it set by `ejabberd_sm:do_route`? Obviously we can still add `sender_sid` (or `origin_sid` :)) for debug purposes but not use it in any logic for now.

fenek:
I'd pick names: `origin_jid` and `origin_sid`

fenek:
I think we should have a dedicated module in MongooseIM like `mongoose_delay_simulator` able to subscribe to chosen hook and introduce provided delay in execution. Useful for at least some tests and would allow reproducing this race condition WAY easier. What do you think? Could you make it a part of this PR?

arcusfelis:
rebased on top of d9ddfd90fabbe6b001e5cfc58e224e00cb1cc92d

src/ejabberd_c2s.erl
ludwikbukowski: device ?

PR #1808, opened 2018-04-16
Properly handle resource reusing between stories in ejabberdctl_SUITE

arcusfelis:
This PR is similar to #1803.
It also addresses the same resource reuse between several stories.

Proposed changes include:
* ejabberdctl_helper prints command, when a timeout error occurs
* kick_everyone between test cases
* Actually wait for cleanup to finish, not just fire a stanza and forget (because the result can arrive to a reconnected device)

### DIFF
--- a/big_tests/tests/ejabberdctl_SUITE.erl
+++ b/big_tests/tests/ejabberdctl_SUITE.erl
@@ -194,6 +194,9 @@ end_per_testcase(delete_old_users, Config) ->
         end, Users),
     escalus:end_per_testcase(delete_old_users, Config);
 end_per_testcase(CaseName, Config) ->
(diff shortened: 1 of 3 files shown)


big_tests/tests/ejabberdctl_SUITE.erl
```diff
@@ -194,6 +194,9 @@ end_per_testcase(delete_old_users, Config) ->
         end, Users),
     escalus:end_per_testcase(delete_old_users, Config);
 end_per_testcase(CaseName, Config) ->
+    %% Because kick_session fails with unexprected stanza received:
```
fenek: `unexpected`

Add `TODO: Remove when escalus learns how to automatically deal with 'unavailable' stanzas on client stop.`

PR #1820, opened 2018-04-23
Add an option for disabling Riak SSL configuration in setup_riak script

mkacper:
This PR addresses adding an option for disabling Riak SSL configuration in setup_riak script.

The script is used by Tide in order to provision Riak but we do not want to use SSL for connection to the DB. Introduced changes allows to disable the configuration while it's still enabled by default for testing purposes.

Proposed changes include:
* add an option for disabling SSL configuration in the script that provisions Riak

### DIFF
--- a/tools/setup_riak
+++ b/tools/setup_riak
@@ -4,6 +4,7 @@
 # - RIAK_HOST
 # - RIAK_PORT
 # - RIAK_PASSWORD
+# - RIAK_SECURITY - if enabled Riak requires SSL connections
 
 set -e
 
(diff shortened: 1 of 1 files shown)


mkacper:
@michalwski thanks a lot for the review. Can you merge the PR please? I know that one test case fails but it seems that this also happens for other PRs ([example](https://travis-ci.org/esl/MongooseIM/builds/369178174?utm_source=github_status&utm_medium=notification)).

michalwski:
@fenek are you OK with merging any PR before the issue with mysql container is solved?

michalwski:
Ok, the issue with mysql was fixed in master branch.

PR #1821, opened 2018-04-24
Add hibernate_after listener opt with 0 default value.

kzemek:
This PR introduces a new configuration variable for listeners: `hibernate_after`.
This variable configures the timeout (in milliseconds) after which a c2s and receiver process will hibernate if they're idle. It replaces old in-code constants that hibernated these processes after 90 seconds of idle time.

If `hibernate_after` is 0 (the default), the processes will hibernate as soon as they get a chance. Load tests have shown that this **greatly** reduces MongooseIM's memory consumption while producing no noticeable slowdown or increase in CPU load under common scenarios.

### DIFF
--- a/src/ejabberd_receiver.erl
+++ b/src/ejabberd_receiver.erl
@@ -31,7 +31,6 @@
 
 %% API
 -export([start_link/4,
-         start/3,
          start/4,
          change_shaper/2,
          starttls/2,
@@ -54,10 +53,9 @@
                 max_stanza_size,
                 stanza_chunk_size,
                 parser,
-                timeout}).
+                timeout,
+                hibernate_after = 0 :: non_neg_integer()}).
 -type state() :: #state{}.
-
--define(HIBERNATE_TIMEOUT, 90000).
 
 %%====================================================================
 %% API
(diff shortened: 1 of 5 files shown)


arcusfelis:
Do we exchange a lot of presences in our tests?
I would like to see how the change behaves with more presence probes or with MUC presences workload.

kzemek:
@arcusfelis These are presence-based load tests: [without hibernate](https://tide.erlang-solutions.com/tests/12caec65-2fdf-4bc4-b6c4-a3355c3a7032) and [with hibernate](https://tide.erlang-solutions.com/tests/6ff2b56a-890e-4ffb-9ac5-1307c092d8b5).

kzemek:
And here are load tests with users sending messages rapidly: [without hibernate](https://tide.erlang-solutions.com/tests/734603fa-9252-4cf2-9229-1c0cca066aea) and [with hibernate](https://tide.erlang-solutions.com/tests/173ab54d-f139-44a0-8f70-5597b983470e).

arcusfelis:
one_to_one test is more about message sending.

Presence based should have:
- a lot of login/logouts
- a lot of presence updates


The roster size also plays a role.

I am fine with the change, but still it would be nice to have presence-oriented tests as part of our testing on tide.

kzemek:
@arcusfelis Have you looked at the scenarios tested? The tests I linked as "presence tests" only test sending presence updates. Each user has 8-user roster and updates presence every time when in a normal test he would send a message to his 8 neighbors.

In the "rapid send" test, each user sends 5 messages/second and receives 40 messages/second.

src/ejabberd_receiver.erl
```diff
@@ -421,3 +423,15 @@ gen_server_call_or_noproc(Pid, Message) ->
     end.
 
 gen_fsm() -> p1_fsm.
+
+hibernate(Result) ->
+    case process_info(self(), message_queue_len) of
```
arcusfelis: With p1_fsm some messages can be inside an internal message queue.
This means that you should check '$internal_queue_len' process dictionary entry too.

I would rather put hibernate logic inside p1_fsm_old.

src/ejabberd_receiver.erl
```diff
@@ -421,3 +423,15 @@ gen_server_call_or_noproc(Pid, Message) ->
     end.
 
 gen_fsm() -> p1_fsm.
+
+hibernate(Result) ->
```
arcusfelis: Both hibernate/1 and maybe_hibernate/1 needs specs.
I am not fan of element/2 call to get state, because it's not dialyzer verifiable.
Just case can be a better option.

arcusfelis:
So, if we send a message each minute, than in the old test we never hibernate at all => can lead to higher memory usage.

src/ejabberd_receiver.erl
fenek: :+1: for moving the logic to `p1_fsm_old`.

PR #1823, opened 2018-04-24
Fix Travis tests with new MySQL

kzemek:
From 8.0.11 release notes:

The following features related to account management have been removed:

Using GRANT to modify account properties other than privilege assignments. This includes authentication, SSL, and resource-limit properties. Instead, establish such properties at account-creation time with CREATE USER or modify them afterward with ALTER USER.

### DIFF
--- a/tools/travis-setup-db.sh
+++ b/tools/travis-setup-db.sh
@@ -28,6 +28,8 @@ if [ "$DB" = 'mysql' ]; then
     mkdir -p ${SQL_TEMP_DIR}
     cp ${SSLDIR}/fake_cert.pem ${SQL_TEMP_DIR}/.
     openssl rsa -in ${SSLDIR}/fake_key.pem -out ${SQL_TEMP_DIR}/fake_key.pem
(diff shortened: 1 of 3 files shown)


arkgil:
Looks good to me 🎉

tools/travis-setup-db.sh
```diff
@@ -39,7 +41,8 @@ if [ "$DB" = 'mysql' ]; then
         -v ${BASE}/${TOOLS}/docker-setup-mysql.sh:/docker-entrypoint-initdb.d/docker-setup-mysql.sh \
         -v ${SQL_TEMP_DIR}:${SQL_TEMP_DIR} \
         --health-cmd='mysqladmin ping --silent' \
-        -p 3306:3306 --name=mongooseim-mysql mysql
+        -p 3306:3306 --name=mongooseim-mysql \
+        mysql --default-authentication-plugin=mysql_native_password
```
michalwski: I think it'd be good to freeze the container version so that we don't have similar unexpected issue in the future when latest changes and doesn't run.

PR #1827, opened 2018-04-26
Fix caching iq info in accumulator

bartekgorny:
This PR addresses the issue of wrong iq info being used in some places. The reason is that iq_query_info, being a bit costly to generate out of an xmlel, is cached in the accumulator (using `mongoose_acc:require/2`) and then reused. Sometimes, though, we are processing a different stanza then the original one, e.g. when we are replying with an error - iq_query_info retrieved from accumulator is in such case inaccurate.

Proposed solution: when creating an acc lets copy its reference to the element being stored. Then we can easily check whether an element being processed is the initial one or something else, and proceed accordingly.

### DIFF
--- a/big_tests/rebar.config
+++ b/big_tests/rebar.config
@@ -9,7 +9,7 @@
         {erlsh, ".*", {git, "git://github.com/proger/erlsh.git", "2fce513"}},
         {jiffy, ".*", {git, "git://github.com/davisp/jiffy.git", "0.14.8"}},
         {proper, ".*", {git, "https://github.com/manopapad/proper.git", "v1.2"}},
-        {exml, ".*", {git, "git://github.com/esl/exml.git", "013fc58"}},
+        {exml, ".*", {git, "git://github.com/esl/exml.git", {branch, "add_ref_to_xmlel"}}},
         {escalus, ".*", {git, "git://github.com/esl/escalus.git", "aa77ba506fd46cf2bf9a0ac3639bc2ed6e1a8173"}},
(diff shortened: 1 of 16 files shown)


arcusfelis:
I propose as a solution to reset cache, each time a stanza is updated in accum.

fenek:
@arcusfelis @bartekgorny 

Maybe instead of references, we could compare `phash2`es? The `phash2` of a stanza in acc maybe cached in acc itself.

bartekgorny:
> I propose as a solution to reset cache, each time a stanza is updated in accum.

I explained it already many times: by design, stanza is never updated in accum.

bartekgorny:
> Maybe instead of references, we could compare phash2es? The phash2 of a stanza in acc maybe cached in acc itself.

Of course, this is the first idea that springs to mind. The reason I haven't offered this solution is that I'm not sure about performance - the reason we want to cache stuff like iq_query_info is to save time by not drawing it from the stanza every time. Is calculating phash2 significantly cheaper then reading xmlel attrs?

bartekgorny:
No, indeed, we are not certain. I think I'll do some testing today to see how it is, maybe we can stop caching it altogether (which is actually one of the options listed in redmine ticket).

fenek:
I discussed it with @bartekgorny and we'd like to explore the explicit distinction between `ejabberd_router:route/3` and `/4` use cases. `/4` would clean all stanza-related or "volatile", cached information because it would be clearly meant only for routing **new** stanzas.

fenek:
Rejected in favour of acc refactoring. What is more, we're not changing `#xmlel{}`.

PR #1830, opened 2018-04-27
Escalus 115cffb

michalwski:
This is to test MongooseIM tests with current esl/escalus@115cffb.
It was never tested and most probably recent changes introduced to escalus makes our test even more vulnerable to random failures.

I hope it's the last time when a PR to escalus was merged without testing it with MongooseIM. Tests in escalus are not enough to prove that changes introduce in a PR doesn't break MongooseIM tests.


michalwski:
Ok, looks like the this instability was introduced by esl/escalus#169 (esl/escalus@7081f90). On my MongooseIM fork  I've created michalwski/MongooseIM#111 with some debug logs in this randomly failing global_dist test case (`test_pm_with_ungraceful_reconnection_to_different_server`).

The root cause of failures is the first message delivered to second's Eve connection from stream management. @kzemek is this expected? Should the test always wait for 2 copies of the `Hi from Europe1!` message (one regular and one from stream management)?

michalwski:
Apparently the delayed `Hi from Europe1!` is not always delivered from SM. See https://github.com/michalwski/MongooseIM/pull/111/commits/2fad2109aac0ee2506c873d523af3d8e18af16c6 and results in http://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/111/993/internal_mnesia.19.3/big/ct_run.test@travis-job-fc796232-f4af-40e7-8fc5-c7c0aa19ee76.2018-04-27_19.01.50/big_tests.tests.mod_global_distrib_SUITE.logs/run.2018-04-27_19.03.18/mod_global_distrib_suite.test_pm_with_ungraceful_reconnection_to_different_server.html

My conclusion - the test fails randomly because sometimes Eve gets `Hi from Europe1` message twice - one from stream management and the other one regular. @kzemek it'd be good to know what is the exact purpose of the test case and if SM is really needed.

PR #1846, opened 2018-05-11
Partial RPC cleanup in big tests

erszcz:
This PR tests Escalus with changes from https://github.com/esl/escalus/pull/170 as well as somewhat cleans up the way RPC is done in big tests.

With regard to RPC in big tests, this is not complete. There are still things which require refactoring:
- other `escalus_ejabberd` calls which hide RPC in the library, but by default access a particular node
- `rpc_apply` in `mam_SUITE`
- modules importing `escalus_ejabberd`, but not calling `escalus_ejabberd:rpc` directly - these cases weren't caught by `sed`
- quite likely other cases, `grep rpc` or `rg rpc` in `big_tests/` should give hints.

These bullet points do not suggest this stuff should go into this PR. Rather, than even when this PR is merged, the cleanup won't be complete yet.

### DIFF
--- a/big_tests/tests/distributed_helper.erl
+++ b/big_tests/tests/distributed_helper.erl
@@ -62,7 +62,7 @@ rpc(Node, M, F, A) ->
 
 rpc(Node, M, F, A, TimeOut) ->
     Cookie = ct:get_config(ejabberd_cookie),
-    escalus_ct:rpc_call(Node, M, F, A, TimeOut, Cookie).
+    escalus_rpc:call(Node, M, F, A, TimeOut, Cookie).
 
 %% @doc Require nodes defined in `test.config' for later convenient RPCing into.
 %%
--- a/big_tests/tests/distributed_helper.erl
+++ b/big_tests/tests/distributed_helper.erl
@@ -3,7 +3,7 @@
 -include_lib("common_test/include/ct.hrl").
 -include_lib("eunit/include/eunit.hrl").
 
--import(ejabberd_node_utils, [get_cwd/2, mim/0, mim2/0, fed/0]).
+-import(ejabberd_node_utils, [get_cwd/2]).
 
 -compile(export_all).
 
@@ -65,6 +65,24 @@ rpc(Node, M, F, A, TimeOut) ->
     escalus_ct:rpc_call(Node, M, F, A, TimeOut, Cookie).
 
 %% @doc Require nodes defined in `test.config' for later convenient RPCing into.
+%%
+%% The use case would be to require and import the same names in your suite like:
+%%
+%%  -import(distributed_helper, [mim/0, fed/0,
+%%                               require_rpc_nodes/1,
+%%                               rpc/4]).
+%%
+%%  ...
+%%
+%%  suite() ->
+%%      require_rpc_nodes([mim, fed]) ++ escalus:suite().
+%%
+%%  ...
+%%
+%%  example_test(_Config) ->
+%%      RPCResult = rpc(mim(), remote_mod, remote_fun, [arg1, arg2]),
+%%      ...
+%%
 require_rpc_nodes(Nodes) ->
     [ {require, {hosts, Node, node}} || Node <- Nodes ].
 
@@ -72,9 +90,15 @@ require_rpc_nodes(Nodes) ->
 mim() ->
     get_or_fail({hosts, mim, node}).
 
+mim2() ->
+    get_or_fail({hosts, mim2, node}).
+
+mim3() ->
+    get_or_fail({hosts, mim3, node}).
+
 %% @doc Shorthand for hosts->fed->node from `test.config'.
 fed() ->
-    get_or_fail(fed_node).
+    get_or_fail({hosts, fed, node}).
 
 get_or_fail(Key) ->
     Val = ct:get_config(Key),
--- a/big_tests/tests/distributed_helper.erl
+++ b/big_tests/tests/distributed_helper.erl
@@ -64,6 +64,23 @@ rpc(Node, M, F, A, TimeOut) ->
     Cookie = ct:get_config(ejabberd_cookie),
     escalus_ct:rpc_call(Node, M, F, A, TimeOut, Cookie).
 
+%% @doc Require nodes defined in `test.config' for later convenient RPCing into.
+require_rpc_nodes(Nodes) ->
+    [ {require, {hosts, Node, node}} || Node <- Nodes ].
+
+%% @doc Shorthand for hosts->mim->node from `test.config'.
+mim() ->
+    get_or_fail({hosts, mim, node}).
+
+%% @doc Shorthand for hosts->fed->node from `test.config'.
+fed() ->
+    get_or_fail(fed_node).
+
+get_or_fail(Key) ->
+    Val = ct:get_config(Key),
+    Val == undefined andalso error({undefined, Key}),
+    Val.
+
 start_node(Node, Config) ->
     {_, 0} = ejabberdctl_helper:ejabberdctl(Node, "start", [], Config),
     {_, 0} = ejabberdctl_helper:ejabberdctl(Node, "started", [], Config),
--- a/big_tests/tests/distributed_helper.erl
+++ b/big_tests/tests/distributed_helper.erl
@@ -8,7 +8,7 @@
 -compile(export_all).
 
 is_sm_distributed() ->
-    Backend = escalus_ejabberd:rpc(ejabberd_sm_backend, backend, []),
+    Backend = rpc(mim(), ejabberd_sm_backend, backend, []),
     is_sm_backend_distributed(Backend).
 
 is_sm_backend_distributed(ejabberd_sm_mnesia) -> true;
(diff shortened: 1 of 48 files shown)


big_tests/tests/dynamic_modules.erl
```diff
@@ -31,26 +33,25 @@ to_replace([RequiredModule | Rest], CurrentModules, ReplaceAcc, ReplaceWithAcc)
...
+    rpc(mim(), gen_mod, loaded_modules_with_opts, [Domain]).
 
 stop(Domain, Mod) ->
     Node = escalus_ct:get_config(ejabberd_node),
     stop(Node, Domain, Mod).
 
 stop(Node, Domain, Mod) ->
     Cookie = escalus_ct:get_config(ejabberd_cookie),
-    IsLoaded = escalus_ct:rpc_call(Node, gen_mod, is_loaded, [Domain, Mod], 5000, Cookie),
+    IsLoaded = escalus_rpc:call(Node, gen_mod, is_loaded, [Domain, Mod], 5000, Cookie),
```
michalwski: Looks like 2 ways of using `rpc` is mixed in this module. The one imported from `distributed_helper` and direct use of `escalus_rpc:call`. Are you going to unify it in this PR or is there a reason to do so?

erszcz:
I changed `escalus_ct:rpc_call` to `escalus_rpc:call`, because the module name changed and stuff needed patching to work. Then I used `sed` to change `escalus_ejabberd:rpc` to using `distributed_helper` & co., but it took quite a while since there were a lot of changes (the last commit is still in the pipeline for Travis).

I think we should gather the remaining points into an issue on Redmine to groom.

big_tests/tests/mam_helper.erl
```diff
@@ -950,7 +953,7 @@ parse_messages(Messages) ->
 
 bootstrap_archive(Config) ->
     random:seed(os:timestamp()),
-    Users = escalus_ct:get_config(escalus_users),
+    _Users = escalus_ct:get_config(escalus_users),
```
fenek: Hmm... so we don't need this line at all, right?

big_tests/tests/mod_aws_sns_SUITE.erl
```diff
@@ -273,7 +273,7 @@ start_publish_listener(Config) ->
 rpc(M, F, A) ->
     Node = ct:get_config({hosts, mim, node}),
     Cookie = escalus_ct:get_config(ejabberd_cookie),
-    escalus_ct:rpc_call(Node, M, F, A, 10000, Cookie).
+    escalus_rpc:call(Node, M, F, A, 10000, Cookie).
```
fenek: Why not `distributed_helper:rpc/4`?

big_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -1082,3 +1087,16 @@ refresh_mappings(NodeName, Config) ->
 trigger_rebalance(NodeName, DestinationDomain) ->
     rpc(NodeName, mod_global_distrib_server_mgr, force_refresh, [DestinationDomain]),
     timer:sleep(1000).
+
+wait_for(Label, Pred) ->
+    wait_for(Label, Pred, timer:seconds(5)).
+
+wait_for(Label, _Pred, RemainingTime) when RemainingTime =< 0 ->
```
fenek: Almost duplicate of `mongoose_helper:wait_until`.

big_tests/tests/muc_helper.erl
```diff
@@ -41,10 +44,10 @@ foreach_recipient(Users, VerifyFun) ->
       end, Users).
 
 load_muc(Host) ->
-    Backend = case mongoose_helper:is_odbc_enabled(<<"localhost">>) of
-                  true -> odbc;
-                  false -> mnesia
-              end,
+    _Backend = case mongoose_helper:is_odbc_enabled(<<"localhost">>) of
```
fenek: Or simply remove `_Backend = ` :)

big_tests/tests/push_pubsub_SUITE.erl
```diff
@@ -366,7 +366,7 @@ parse_form(Fields) when is_list(Fields) ->
 rpc(M, F, A) ->
     Node = ct:get_config({hosts, mim, node}),
     Cookie = escalus_ct:get_config(ejabberd_cookie),
-    escalus_ct:rpc_call(Node, M, F, A, 10000, Cookie).
+    escalus_rpc:call(Node, M, F, A, 10000, Cookie).
```
fenek: Why not `distributed_helper:rpc`?

big_tests/tests/mam_helper.erl
erszcz: It's obviously a leftover from a debug print/log that somebody's forgotten to remove. I just figured it might be convenient to leave it in case of future debugging, but yes - let's remove it.

big_tests/tests/muc_helper.erl
erszcz: Same as `_Users` above.

big_tests/tests/distributed_helper.erl
```diff
@@ -62,7 +62,48 @@ rpc(Node, M, F, A) ->
...
+%%  suite() ->
+%%      require_rpc_nodes([mim, fed]) ++ escalus:suite().
+%%
+%%  ...
+%%
+%%  example_test(_Config) ->
+%%      RPCResult = rpc(mim(), remote_mod, remote_fun, [arg1, arg2]),
+%%      ...
+%%
+require_rpc_nodes(Nodes) ->
```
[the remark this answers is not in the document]
erszcz: I'm against hiding IO in a function which is now completely declarative. We will know if a node is down anyway, because the RPC calls now throw instead of just returning an error by value and we will know with better granularity (the specific file, test and line).

erszcz:
@fenek I've addressed some issues with commits and some with comments. This PR is a an extra to the task of decreasing reliance of Escalus on Common Test, not a result of targeted effort to completely clean up RPC in big tests - one step at a time. As I've suggested above we should have a separate issue for complete RPC cleanup, groom it and then resolve as appropriate.

The main pain point is that there's currently an unrelated problem in MUC HTTP API suite which requires like 10 Travis restarts, so I don't want to add anything that requires a CI rerun to this particular PR.

big_tests/tests/ejabberdctl_SUITE.erl
```diff
@@ -1147,10 +1149,10 @@ match_roster(ItemsValid, Items) ->
     ItemsTokens = [ string:tokens(ItemToken, "\t") || ItemToken <- string:tokens(Items, "\n") ],
 
     true = (length(ItemsValid) == length(ItemsTokens)),
-    true = lists:all(fun({Username, Domain, Nick, Group, Sub}) ->
+    true = lists:all(fun({Username, Domain, _Nick, _Group, _Sub}) ->
                     JID = escalus_utils:jid_to_lower(<<Username/binary, "@", Domain/binary >>),
                     lists:any(fun
-                                ([RosterJID, Nick, Sub, "none", Group]) ->
+                                ([RosterJID, __Nick, __Sub, "none", __Group]) ->
```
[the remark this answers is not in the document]
erszcz: Ultimately, you're right about the intention (the function was initially introduced in https://github.com/esl/ejabberd_tests/commit/09438c5a#diff-217f8e16b78b791f657c1119ea504f64R519), but I think that since https://github.com/esl/ejabberd_tests/commit/95f4794a#diff-217f8e16b78b791f657c1119ea504f64R670 it doesn't match on `Nick`, `Sub`, or `Nick`.

fenek:
- I agree that improvement with pinging inside `require_rpc_nodes` may be done later. Can you please create a ticket for this, so we won't forget to do it?
- Same for `ejabberdctl_helper:rpc_call`
- Same for `distributed_helper:rpc` suggestions
- Comment about `wait_for` is still valid because currently this PR adds a function that is already implemented in even more generic way.

erszcz:
> I agree that improvement with pinging inside require_rpc_nodes may be done later. Can you please create a ticket for this, so we won't forget to do it?

I'm keeping my stance that we should not do it.

> Same for ejabberdctl_helper:rpc_call
> Same for distributed_helper:rpc suggestions

#1853 created.

> Comment about wait_for is still valid because currently this PR adds a function that is already implemented in even more generic way.

Apart from `mongoose_helper:wait_until/3,4` we also have `push_SUITE:wait_for/2` and `push_helper:wait_for/2` and a number of ad-hoc `wait_for_SOMETHING_SPECIFIC` here and there. This is another whole topic which could use some refactoring, so I'd rather not wait a whole day for Travis to recheck this PR if we're going back to this topic in the future.

fenek:
OK, please create a ticket for this as well.

Anyway, I think we need one last Travis confirmation anyway, correct? Last commit has `skip ci` flag.

erszcz:
All the commits after 32badd4 have `[skip ci]`, while 32badd4 is green on Travis. There's a minuscule coverage drop, but to be honest I don't know where it comes from - it seems like a spurious change (e.g. different days are hit on a code path formatting dates - o_O).

I think they are simple enough to assume correctness and we'll need to do the job restart magic on master anyway after the PR is merged.

PR #1847, opened 2018-05-14
Cassandra refactor

rslota:
This PR introduces new implementation of `mongoose_cassandra`. There are no changes to the API, so the code that uses Cassandra remains unchanged.
The new implementation of Cassandra layer moves all request processing to a worker pool (previously only asynchronous requests were handled this way). This reduces code duplication and improves error handling while also fixes some race conditions that were present in previous implementation.

This PR also adds `mongoose_cassandra_SUITE` that has some simple "cassandra-only" tests for `mongoose_cassandra` API. This test suite requires a TCP proxy to be active on top of Cassandra (to enable to connection manipulations like connection drop or connection latency).

### DIFF
--- a/big_tests/default.spec
+++ b/big_tests/default.spec
@@ -69,6 +69,7 @@
 {suites, "tests", websockets_SUITE}.
 {suites, "tests", xep_0352_csi_SUITE}.
 {suites, "tests", mod_global_distrib_SUITE}.
+{suites, "tests", mongoose_cassandra_SUITE}.
 
 {config, ["test.config"]}.
 {logdir, "ct_report"}.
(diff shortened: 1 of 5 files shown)


rslota:
@kzemek Actually, no, I haven't considered using `worker_pool`. This is part of the old implementation that I haven't changed since it was simple irrelevant to the changes in the general worker logic that were the main focus of this refactor. 

I'll reformat the code with emacs, I have no idea when and how the code got reformatted. Also, I'll look into the viability of using `worker_pool`.  If it's gonna simplify the code, I'll go for it.

rslota:
@kzemek I have reformatted the code with emacs, but introducing `worker_pool` is not the best idea right now. The changes are not as simple as removing the custom supervisor since `worker_pool` doesn't allow for nested calls (i.e. worker selects another worker) - such calls may end up in deadlock. 
I still think that using `worker_pool` will be cleaner, but it requires non-trivial changes to the worker's logic, therefore I would prefer to do it as a separate PR.

PR #1848, opened 2018-05-15
Added image about accumulator's lifecycle

bartekgorny:
Only some visual documentation, hopefully makes clear what the accumulator actually carries.


fenek:
Won't updating it be a real hell? I think it's better to embed a PNG image. And attach some source file, editable with a free/open tool.

fenek:
Is it a Monodraw file? If so, then it's not free for sure. :)

fenek:
To update after acc refactor, with different (free) diagram tool.

michalwski:
I'm closing this one since it's too old and out of date.

PR #1850, opened 2018-05-16
[clt-rabbit] Handle presence events with Rabbit

mkacper:
This issue closes esl/tide#396, closes esl/tide#413.


big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
```diff
@@ -0,0 +1,321 @@
...
+              AliceJID = nick_to_jid(alice, Config),
+              Exchanges = [BobJID, AliceJID],
+              %% WHEN users login
+              escalus:wait_for_stanzas(Bob, 1),
+              escalus:wait_for_stanzas(Alice, 1),
+              %% THEN exchanges are created
+              Result = [is_exchange_present(Connection, Channel, Exchange,
+                                 ?IF_EXCHANGE_EXISTS_RETRIES)
+                        || Exchange <- Exchanges],
+              ?assert(lists:all(fun(E) -> E == true end, Result))
```
bszaf: According to `is_exchange_present/4` it raises an exception if an error occurs. Therefore, you will never reach this line if error occurs (or exchange is not present).

big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
```diff
@@ -0,0 +1,321 @@
...
+bind_queue_to_exchange(Channel, {Queue, Exchange, RoutingKey}) ->
+    amqp_channel:call(Channel, #'queue.bind'{exchange = Exchange,
+                                             routing_key = RoutingKey,
+                                             queue = Queue}).
+
+-spec is_exchange_present(Connection :: pid(), Channel :: pid(),
+                                      Exchange :: binary(),
+                                      Retries :: non_neg_integer()) ->
+    true | no_return().
+is_exchange_present(_Connection, Channel, Exchange, 0) ->
```
bszaf: If you count down to `0`, you have more runs, than specified in `Retries` :P . You should countdown to `1`

big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
```diff
@@ -0,0 +1,321 @@
...
+              BobJID = nick_to_jid(bob, Config),
+              AliceJID = nick_to_jid(alice, Config),
+              Exchanges = [BobJID, AliceJID],
+              %% WHEN users login and logout
+              escalus:wait_for_stanzas(Bob, 1),
+              escalus:wait_for_stanzas(Alice, 1),
+              escalus:send(Bob, escalus_stanza:presence(<<"unavailable">>)),
+              escalus:send(Alice, escalus_stanza:presence(<<"unavailable">>)),
+              %% THEN exchanges are deleted
+              [?assertException(_Error, _Reason,
```
bszaf: I find it unusual to write helpers, which raises an Exception and we assert on Exception it raised. I think `error_when_exchange_is_absent/4` should be changed, see my comments on it.

big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
```diff
@@ -0,0 +1,321 @@
...
+        _Error:_Reason ->
+            timer:sleep(?WAIT_FOR_EXCHANGE_INTERVAL),
+            %% The old channel `Channel` is closed.
+            {ok, NewChannel} = amqp_connection:open_channel(Connection),
+            is_exchange_present(Connection, NewChannel, Exchange, Retries - 1)
+    end.
+
+error_when_exchange_is_absent(Connection, Channel, Exchange, 0) ->
+    is_exchange_present(Connection, Channel, Exchange, 0);
+error_when_exchange_is_absent(Connection, Channel, Exchange, Retries) ->
```
bszaf: - You have double countdown - each call `error_when_exchange_is_absent(Connection, Channel, Exchange, Retries)` causes `Retries` calls to `is_exchange_present/4`
- I would rewrite this function. Body may be copied from `is_exchange_present/4`, but it will return true when exception is raised and retry when exchange is present. Code duplication is not significant, but it will look better in test case (asserting on boolean is better than expecting an exception).
- Rename to `is_exchange_absent/4`
- Be aware of countdown to `0` vs `1` as in `is_exchange_absent/4`
- Missing spec

big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
```diff
@@ -0,0 +1,321 @@
...
+        Else ->
+            Else
+    end.
+
+-spec subscribe_to_rabbit_queue(Channel :: pid(), Queue :: binary()) -> ok.
+subscribe_to_rabbit_queue(Channel, Queue) ->
+    amqp_channel:subscribe(Channel, #'basic.consume'{queue = Queue,
+                                                     no_ack = true}, self()),
+    receive
+        #'basic.consume_ok'{} -> ok
```
bszaf: Please match on returned value from `amqp_channel:subscribe/3` - `#basic.consume_ok{}`.
I would also add `After Timeout -> error("some description")` in receive - you never know if Rabbit is dead.

big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
```diff
@@ -0,0 +1,321 @@
...
+presence_bindings(Queue, JIDs) ->
+    [{Queue, JID, presence_rk(JID)} || JID <- JIDs].
+
+-spec bind_queues_to_exchanges(Channel :: pid(),
+                               Bindings :: [rabbit_binding()]) ->
+    [amqp_client:amqp_method() | ok | blocked | closing].
+bind_queues_to_exchanges(Channel, Bindings) ->
+    [bind_queue_to_exchange(Channel, Binding) || Binding <- Bindings].
+
+bind_queue_to_exchange(Channel, {Queue, Exchange, RoutingKey}) ->
```
bszaf: - no spec
- maybe match on returned value.

big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
```diff
@@ -0,0 +1,321 @@
...
+    Channel = proplists:get_value(rabbit_channel, Config),
+    Queue = declare_rabbit_queue(Channel),
+    QueueBindings = presence_bindings(Queue, JIDs),
+    wait_for_exchanges_to_be_created(Connection, JIDs),
+    bind_queues_to_exchanges(Channel, QueueBindings),
+    subscribe_to_rabbit_queue(Channel, Queue).
+
+-spec wait_for_exchanges_to_be_created(Connection :: pid(),
+                                       Exchanges :: [binary()]) -> pid().
+wait_for_exchanges_to_be_created(Connection, Exchanges) ->
```
bszaf: You may add similiar function `wait_for_exchanges_to_be_deleted`.

bszaf: It could be used in test `disconnected_users_delete_exchanges_when_unavailable(Config)`

src/event_pusher/mod_event_pusher_rabbit.erl
```diff
@@ -0,0 +1,90 @@
...
+pool_name(Host) ->
+    gen_mod:get_module_proc(Host, ?MODULE).
+
+-spec jid(User :: binary(), Host :: binary()) -> binary().
+jid(User, Host) ->
+    <<User/binary, "@", Host/binary>>.
+
+%% Getter for module options
+-spec opt(Host :: jid:lserver(), Option :: atom()) ->
+    Value :: term() | undefined.
```
bszaf: `undefined ` ∊ term().

From docs:
>A current restriction, which currently results in a warning (not an error) by the compiler, is that the domains of the argument types cannot overlap. For example, the following specification results in a warning:
```
  -spec foo(pos_integer()) -> pos_integer()
         ; (integer()) -> integer().
```

src/mongoose_rabbit_worker.erl
```diff
@@ -0,0 +1,216 @@
...
+%% Initializes the server
+%% @end
+%%--------------------------------------------------------------------
+-spec init(Args :: term()) -> {ok, State :: term()} |
+                              {ok, State :: term(), Timeout :: timeout()} |
+                              {ok, State :: term(), hibernate} |
+                              {stop, Reason :: term()} |
+                              ignore.
+init([]) ->
+    process_flag(trap_exit, true),
```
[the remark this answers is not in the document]
mkacper: Yeah, but it's necessary to force a gen_server to run `terminate` when a supervisor kills it.

mkacper:
@bszaf please see the improvements.

fenek:
@mkacper Is it a must-have for you for 3.0.0? I'd prefer not to change scope of this sprint and add such a big change. So this feature would be included in 3.1.0 (given this will be the next number). Is it OK with you?

mkacper:
@fenek No, it's not :) This PR introduces only a basic feature regarding RabbitMQ integration. We're going to extend the feature (by handling other user related events and dealing with AMQP ACKs). I'd like to put the final version of the feature on `feature/clt-rabbit-mod_event_pusher` branch and then ask someone from MIM team for review. Once the PR is ready you could include it in a version you consider the most appropriate :)

PR #1854, opened 2018-05-17
Revert "Fix broken lager_console_backend backend"

fenek:
Reverts esl/MongooseIM#1832 as it actually has broken the console logger.

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -6,7 +6,7 @@
     {log_root, "{{mongooseim_log_dir}}"},
     {crash_log, "crash.log"},
     {handlers, [
-        {lager_console_backend, [{level, info}]},
+        {lager_console_backend, [info, {lager_default_formatter,[{eol, "\r\n"}]}]},
 %% use below line to add syslog backend for Lager
 %        {lager_syslog_backend, [ "mongooseim", local0, info]},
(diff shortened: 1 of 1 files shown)


arcusfelis:
Aaaand it breaks lager for me again.
@fenek what kind of behaviour / error you get with this change not reverted?

arcusfelis:
The way we use lager would be deprecated in the new lager:

https://github.com/erlang-lager/lager/blob/master/src/lager_console_backend.erl

```erlang
init([Level]) when is_atom(Level) ->
    ?DEPRECATED([{level, Level}]),
    init([{level, Level}]);
init([Level, true]) when is_atom(Level) -> % for backwards compatibility
    ?DEPRECATED([{level, Level}, {formatter_config, [{eol, "\\r\\n\\"}]}]),
    init([{level, Level}, {formatter_config, ?FORMAT_CONFIG_OFF}]);
init([Level, false]) when is_atom(Level) -> % for backwards compatibility
    ?DEPRECATED([{level, Level}]),
init([{level, Level}]);
```


But lager 3.2.4 expects it in the strange format we use in master
https://github.com/erlang-lager/lager/blob/3.2.4/src/lager_console_backend.erl#L47

But it looks like 3.2.4 lager version is buggy. Maybe we should just use the latest lager?

fenek:
@arcusfelis 

With "fixed" version I wasn't getting any console logs and when I tried to change the loglevel, the list of backends returned by `set` function had only file backend.

PR #1859, opened 2018-05-18
Accumulator refactoring - setters, getters and specs

arkgil:
The intention of this PR is to introduce more "safe" interface to the accumulator module. While accumulators proved to be very useful when values need to be exchanged between multiple modules or for debugging purposes (e.g. registering hook calls and send results), I believe that they're underused due to lack of confidence when working with them - we never know if an accumulator field is there and what is its value.

My goal is to remove completely the "generic" part of the accumulator - no more `get`, `put` and update. All fields should have explicit, documented and typespeced getters and setters.

## Changes so far

The code in this PR won't pass any tests - I have refactored maybe 1/3 of the whole accumulator usages. Changes so far include:

- When accumulator is created, module, function and line number need to be passed to `new/3,4,5,6`. This help immensely when debugging, when we're not sure if accumulator comes from HTTP interface, c2s, external component etc.
- The `element` attribute can be only set as a whole, i.e. the whole record, its name, type and attributes are always in sync. There is a `set_element/2` function, but it should be removed and the stanza should be an immutable field, set only when creating an accumulator.
- `from` and `to` properties can be set only when creating an accumulator, they are immutable due to the lack of setters.
- `iq_query_info`, as previously, is computed lazily and stored in the accumulator and is always in sync with corresponding `exml:element()` (although that wouldn't be an issue if element was immutable)
- `server` is a regular, mutable property

## Concerns so far

I'd really appreciate anyone's input on these topic.

- getters currently raise an error when there is no property set. However, sometimes we'd like to perform some logic based on the fact that the property is not there. We can do it in two ways:
  - make all getters return `{ok, ...} | error`. The downside is that we lose nice error message and instead will end up with `badmatch`es in cases when we're willing to crash when property is not there.
  - add additional `has_` function for each property, but that would make the interface really bloated. The upside is that we retain a nice error message when the property is not there.
- `element` should be completely immutable, it should be set to the stanza entering the node and initiating the "routing chain"
- `server` and `user` attributes should be completely immutable as well. They should represent the user and server of the user who initiated the current routing chain (in other words, user and server of the current c2s process or HTTP handler or CLI process). Maybe we could name those properties differently?
- How `from` differs from `user` and `server`? 
- `to` should be completely immutable as well.

## TODOs and questions for the future

Here is a list of things which I've found that are left to do. Please point out anything that you think is missing here:

#### Properties

- [ ] `result` property - 
- [ ] `subscription_lists` property - mutable, used by `roster_get_subscription_lists` hook handlers.
- [ ] `roster` property - filled by modules building roster. It should be mutable, getter should return empty list if the value hasn't been set yet.
- [ ] `amp_check_result` property - also mutable
- [ ] `privacy_check` property - immutable per privacy check cache key, see https://github.com/esl/MongooseIM/blob/master/src/mongoose_privacy.erl#L65
- [ ] `send_result` property - append-only list of records of sending the stanza. I'm not sure if anyone is using that, maybe we should remove it? It's used in https://github.com/esl/MongooseIM/blob/master/src/ejabberd_c2s.erl#L1769, but from what I see its value doesn't matter anyway. It's probably only useful for debugging.
- [ ] `ref` property - immutable, set on creation. Again, I'm not sure if we need it, but AFAIK @fenek has used it in the past
- [ ] `timestamp` property - immutable, set on creation. I don't know anyone who's used it, I guess we could remove it.
- [ ] `global_distrib` property - regular, mutable property
- [ ] `return_type` - used by `jlib` in https://github.com/esl/MongooseIM/blob/master/src/jlib.erl#L150. To me it could be removed - based on that `jlib` decides whether IQ passed to the linked function is already an error - but one could pass unrelated accumulator and error IQ there, and the function won't return an error. I saw that these lines are covered by tests, but I believe it comes only from unit tests which call it directly.
- [ ] `iq_result` property. I'm not sure if it has any meaningful usage. I see that it's set in a couple of places in `mod_muc_light` but it's only retrieved in `c2s`. Maybe someone know what it's created for.
- [ ] `offline_messages` - append only list, should be removable because c2s clears them
- [ ] `hook_runs` and `handler_runs` - record fired hooks and handlers, probably useful for debugging. They're not retrieved anywhere.

#### Functions

- [ ] - `strip` - we should figure out which elements are worth to keep persistent. Currently persistent properties are: `ref`, `timestamp`, `from`, `to`, `global_distrib` and `persistent_properties` (of which at this point only `origin_jid` is used). Do we want to make this list smaller?

# But is it worth it?

As you can see, these modifications would bring huge changes to the code base. IMO it's worth it, I'd definitely fill more confident when using accumulators, but I'd like us all to decide.

### DIFF
--- a/include/mongoose_acc.hrl
+++ b/include/mongoose_acc.hrl
@@ -1,0 +1,8 @@
+%%%-------------------------------------------------------------------
+%%% @doc Creates a new accumulator instance setting location arguments to the caller of this macro.
+%%% @end
+%%%-------------------------------------------------------------------
+-define(new_acc(), mongoose_acc:new(?MODULE, ?FUNCTION_NAME, ?LINE)).
+-define(new_acc(Element), mongoose_acc:new(Element, ?MODULE, ?FUNCTION_NAME, ?LINE)).
+-define(new_acc(Element, From), mongoose_acc:new(Element, From, ?MODULE, ?FUNCTION_NAME, ?LINE)).
+-define(new_acc(Element, From, To), mongoose_acc:new(Element, From, To, ?MODULE, ?FUNCTION_NAME, ?LINE)).
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -1029,7 +1029,7 @@ process_outgoing_stanza(Acc, StateData) ->
     fsm_next_state(session_established, NState).
 
 process_outgoing_stanza(Acc, error, _Name, StateData) ->
-    case mongoose_acc:get(type, Acc) of
+    case mongoose_acc:get_element_type(Acc) of
         <<"error">> -> StateData;
         <<"result">> -> StateData;
         _ ->
@@ -1397,7 +1397,7 @@ preprocess_and_ship(Acc, From, To, El, StateData) ->
 response_negative(<<"iq">>, forbidden, From, To, Acc) ->
     send_back_error(mongoose_xmpp_errors:forbidden(), From, To, Acc);
 response_negative(<<"iq">>, deny, From, To, Acc) ->
-    IqType = mongoose_acc:get(type, Acc),
+    IqType = mongoose_acc:get_element_type(Acc),
     response_iq_deny(IqType, From, To, Acc);
 response_negative(<<"message">>, deny, From, To, Acc) ->
     mod_amp:check_packet(Acc, From, delivery_failed),
@@ -1544,7 +1544,7 @@ handle_routed_presence(From, To, Acc, StateData) ->
     Packet = mongoose_acc:get_element(Acc),
     State = ejabberd_hooks:run_fold(c2s_presence_in, StateData#state.server,
                                     StateData, [{From, To, Packet}]),
-    case mongoose_acc:get(type, Acc) of
+    case mongoose_acc:get_element_type(Acc) of
         <<"probe">> ->
             {LFrom, LBFrom} = lowcase_and_bare(From),
             NewState = case am_i_available_to(LFrom, LBFrom, State) of
@@ -2018,7 +2018,7 @@ specifically_visible_to(LFrom, #state{pres_invis = Invisible} = S) ->
                       State :: state()) -> {mongoose_acc:t(), state()}.
 presence_update(Acc, From, StateData) ->
     Packet = mongoose_acc:get_element(Acc),
-    case mongoose_acc:get(type, Acc) of
+    case mongoose_acc:get_element_type(Acc) of
         <<"unavailable">> ->
             Status = case xml:get_subtag(Packet, <<"status">>) of
                          false ->
@@ -2156,7 +2156,7 @@ presence_track(Acc, StateData) ->
     LTo = jid:to_lower(To),
     User = StateData#state.user,
     Server = StateData#state.server,
-    case mongoose_acc:get(type, Acc) of
+    case mongoose_acc:get_element_type(Acc) of
         <<"unavailable">> ->
             Acc1 = check_privacy_and_route(Acc, StateData),
             I = gb_sets:del_element(LTo, StateData#state.pres_i),
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -1022,7 +1022,7 @@ session_established(closed, StateData) ->
 %% comes back whence it originated
 -spec process_outgoing_stanza(mongoose_acc:t(), state()) -> fsm_return().
 process_outgoing_stanza(Acc, StateData) ->
-    ToJID = mongoose_acc:get(to_jid, Acc),
+    ToJID = mongoose_acc:get_to_jid(Acc),
     Name = mongoose_acc:get(name, Acc),
     NState = process_outgoing_stanza(Acc, ToJID, Name, StateData),
     ejabberd_hooks:run(c2s_loop_debug, [{xmlstreamelement, mongoose_acc:get_element(Acc)}]),
@@ -2151,7 +2151,7 @@ presence_update_to_available(false, Acc, OldPriority, NewPriority, From, Packet,
 -spec presence_track(Acc :: mongoose_acc:t(),
                      State :: state()) -> {mongoose_acc:t(), state()}.
 presence_track(Acc, StateData) ->
-    To = mongoose_acc:get(to_jid, Acc),
+    To = mongoose_acc:get_to_jid(Acc),
     From = mongoose_acc:get_from_jid(Acc),
     LTo = jid:to_lower(To),
     User = StateData#state.user,
@@ -2221,7 +2221,7 @@ check_privacy_and_route(Acc, StateData) ->
                               StateData :: state()) -> mongoose_acc:t().
 check_privacy_and_route(Acc, FromRoute, StateData) ->
     From = mongoose_acc:get_from_jid(Acc),
-    To = mongoose_acc:get(to_jid, Acc),
+    To = mongoose_acc:get_to_jid(Acc),
     {Acc1, Res} = privacy_check_packet(Acc, To, out, StateData),
     Packet = mongoose_acc:get_element(Acc1),
     case Res of
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -1438,9 +1438,8 @@ handle_routed(_, _From, _To, Acc, StateData) ->
                        Acc :: mongoose_acc:t(),
                        StateData :: state()) -> routing_result().
 handle_routed_iq(From, To, Acc, StateData) ->
-    Acc1 = mongoose_acc:require(iq_query_info, Acc),
-    Qi = mongoose_acc:get(iq_query_info, Acc1),
-    handle_routed_iq(From, To, Acc1, Qi, StateData).
+    Qi = mongoose_acc:get_element_iq_query_info(Acc),
+    handle_routed_iq(From, To, Acc, Qi, StateData).
 
 -spec handle_routed_iq(From :: jid:jid(),
                        To :: jid:jid(),
@@ -2438,10 +2437,9 @@ get_priority_from_presence(PresencePacket) ->
                          To :: jid:jid(),
                          StateData :: state()) -> {mongoose_acc:t(), state()}.
 process_privacy_iq(Acc0, To, StateData) ->
-    Acc = mongoose_acc:require(iq_query_info, Acc0),
-    case mongoose_acc:get(iq_query_info, Acc) of
+    case mongoose_acc:get_element_iq_query_info(Acc0) of
         #iq{type = Type, sub_el = SubEl} = IQ ->
-            Acc1 = mongoose_acc:put(iq, IQ, Acc),
+            Acc1 = mongoose_acc:put(iq, IQ, Acc0),
             From = mongoose_acc:get_from_jid(Acc1),
             {Acc2, NewStateData} = process_privacy_iq(Acc1, Type, To, StateData),
             Res = mongoose_acc:get(iq_result, Acc2, {error, mongoose_xmpp_errors:feature_not_implemented()}),
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -62,6 +62,7 @@
          print_state/1]).
 
 -include("mongoose.hrl").
+-include("mongoose_acc.hrl").
 -include("ejabberd_c2s.hrl").
 -include("jlib.hrl").
 -include_lib("exml/include/exml.hrl").
@@ -1192,7 +1193,7 @@ handle_info(replaced, _StateName, StateData) ->
 handle_info(new_offline_messages, session_established,
             #state{pres_last = Presence, pres_invis = Invisible} = StateData)
   when Presence =/= undefined orelse Invisible ->
-    resend_offline_messages(mongoose_acc:new(), StateData),
+    resend_offline_messages(?new_acc(), StateData),
     {next_state, session_established, StateData};
 handle_info({'DOWN', Monitor, _Type, _Object, _Info}, _StateName, StateData)
   when Monitor == StateData#state.socket_monitor ->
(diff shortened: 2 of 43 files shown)


bartekgorny:
> getters currently raise an error when there is no property set.
how about getters with a default  value?

> How from differs from user and server
I think 'from' is a jid while 'user' and 'server' are parts of it. Such distinction, with the same naming, is used in many places in the old code.

1. Part of the bloat stems from the fact that hook handlers return values by adding them to the accumulator, and there is no naming requirement, so it can be 'result', but can also be 'iq_result' or whatever else, provided the caller knows what it is going to be. Some of them may be reused later, some not. 

2. The problem indeed is that acc api does not impose any logic, so you can store anything you want there. Which, quite understandably, prompted developers of global distribution to escape the mess by storing all their stuff under a separate key. In a while, mod_amp and mod_mam will be doing the same.

3. To summarise what the acc really stores:
a) some hard-wired stuff which never changes or disappears, used mostly for debugging and profiling (ref, timestamp)
b) the stanza which started the accumulator
c) some values derived from the initial stanza, for easier access
d) other attributes denoting the accumulator's origin (from and its derivatives)
e) results of hook calls, usually as 'result', volatile
f) cached results of some operations, e.g. privacy check
g) cached info derived from initial stanza (e.g. iq_query_info)
h) additional debugging info, collected along the way (hook_runs, handler_runs, send_result)
i) arbitrary values which are stored in one place to be used somewhere else (global_distrib, and mod_amp will certainly do it)

Plus, some of those need to be passed to the other process (not stripped), those are 'persistent'.
This is indeed quite a lot of use cases for a single entity, and designing an api to make it clean and consistent is a non-trivial task.

4. IMO if global_distrib is to be persistent it should be set as a persistent property, not added separately. This is the kind of bloat that current architecture seems to be causing.

5. return_type is a somewhat hackish way to stop infinite iq error loop. It is tested in a sense that there is a test which triggers the loop, and without the return_type safeguard mongoose would run out of memory and crash. Fixing acc caching logic will solve the deeper cause of the problem, then the return_type will not be needed anymore.

bartekgorny:
This one may be handy:
[https://raw.githubusercontent.com/esl/MongooseIM/accumulator_doc/doc/developers-guide/accumulator.png](url)

arkgil:
> getters currently raise an error when there is no property set.

What I don't like about default values is that they don't have semantics. What does the default mean? Also I believe it would allow more relaxed typespecs, because dialyzer would see that any value can be returned by the getter.

> I think 'from' is a jid while 'user' and 'server' are parts of it. Such distinction, with the same naming, is used in many places in the old code.

I don't think that's the case, `user` and `server` are sometimes set in places where `from` is not. There is a difference here, but I'm not sure what it is. 

> Personally I'd prefer pattern matching. Having typespecs is good, but pattern matching is way more readable.

Current implementation of accumulators can be used to pattern match, but I haven't found a single piece of code which uses that in Mongoose. Additionally, allowing to pattern-match defeats the point of opaqueness of the type, in which case, I suppose, we will be using it as a map.

**Re: 1.** This sentence
> it can be 'result', but can also be 'iq_result' or whatever else, provided the caller knows what it is going to be

and this one

> Some of them may be reused later, some not.

are exactly the reasons why I'd like to refactor the accumulator. `result`, or `iq_result` have no meaning at all or meaning depending hugely on the context they're used in. In addition, I may call some hook, handler puts its result in the `result` field,  and then other handler may override it. Or other hook which is ran after the first one. I, as a caller, have no idea  what might be inside.

I think that if hook handlers return a meaningful result, let's make it an explicit accumulator field. Or, let's scope results to hook names, so that different hooks don't override their results.

And as a final note, there are only 4 places in the whole codebase where the result is retrieved from the accumulator.

**Re: 2.** And that's exactly the point of this PR, to escape the mess 😄 


**Re: 3**

a) But has it ever been used for those purposes?

That point is a really nice summarization 🙂 

**Re: 4**

I'd remove the special `persistent_properties` at all. In fact, only `origin_jid` is put there, so why not add it to the list of not-stripped attributes.

**Re: 5**

Could you point me to the place where this error occures? Has it ever manifested itself? I guess simpler solution is to check if the element's name passed to `jlib:make_error_reply` is `"iq"` and that it has `"error"` type. If I wouldn't know that function, I could pass fresh accumulator there every time and the said infinite loop of errors would occur anyway.

arkgil:
@bartekgorny that's a dead link.

bartekgorny:
Link fixed.
About pattern matching, I tried to say that I prefer `get(attr, Container)` over `get_attr(Container)`.

> Has it ever manifested itself? 

Oh yes, it killed some production instances of MongooseIM. But this is not the scope of this PR, let's discuss it somewhere else.

bartekgorny:
Sounds reasonable. One minor addition:

> 1. what is a purpose of that field
> 2. what is a type of that field
> 3. is that field immutable
> 4. is that field always set, or maybe it appears only in some special cases

5. Should this field be carried on to the other c2s

arkgil:
@bartekgorny we can't assume that accumulator is in the "c2s" context in the first place. It can be HTTP handler process, or a CLI process.

> I prefer get(attr, Container) over get_attr(Container).

Why is that the case? It's the same number of characters 😄

bartekgorny:
I think we should allow for some non-predefined attrs, for the following reason: one of the nice things about MIM is it's hook/handler based architecture - you can add a custom module with its own hook handlers and register it in config, without modifying core code of MIM. Your handler may need to store something in the acc to be read by another handler. If we impose a strict requirement that every acc variable be predefined in mongoose_acc, we'd throw away this flexibility. Still, I'm all for setting up a dedicated place and maybe api for such custom storage.

erszcz:
What do you think of accessor naming used by the QT API? The idea is that the getters and setters are called the same and only differ by arity (in case of QT which is in C++ by signature, but close enough). I mean that instead of:

```erlang
get_some_field(Struct) -> ...

set_some_field(Struct, NewFieldValue) -> ...
```

we could have

```erlang
some_field(Struct) -> ...

some_field(Struct, NewFieldValue) -> ...
```

My impression is that the `get_`/`set_` prefixes carry little to no value, while having to type them is tedious of itself and also makes using autocompletion less convenient (having to type the prefix which is identical for a lot of functions before hitting the magic __give me completions__ button). What do you think?

arkgil:
@erszcz Yes,  I was thinking about it but I went with get/set route so that it's clearly visible what different functions do. I like the idee, though, and I'd go for it.

michalwski:
Few cents from me:
1. There is indeed a need for immutable fields, not all of them are known when the acc is created, though. A special setter is needed for that.
2. Accessing fields through accessors seems like a good idea to me, it for sure simplifies reasoning about the code.
3. It's good idea to have a hook name tagged handler results, and not use `result` or `iq_result` from different hooks.
4. Some flexibility in acc fields is needed, mainly to make it easier to add a custom hook in not open source code without changing core implementation. This could be done as a `misc` field which is also a map. In open source code such a field should be empty in my opinion.
5.  The idea with `same_field/1` for getting the value and `some_field/2` for setting the value seems good to me.

arkgil:
@michalwski the problem I see with `misc` field is that everyone will use it instead of adding a dedicated field. This may sound brutal, but in my opinion the discipline here needs to be imposed from the very beginning.

michalwski:
@arkgil that's why I said that in open source Mongoose this should be empty and not probably not used. I just want to make it as easy as possible for anyone using MongooseIM to add their own hook or handler when needed.
If they put sth custom in the `misc` field then it will be easier for them to update their code to MongooseIM's master. If they have to modify core Mongoose code around accumulators then staying up-to-date with open source MongooseIM may be harder for them. That's my main concern.

arkgil:
@michalwski that makes sense, you're right. Although I still think that usage of this field will be abused :P

kzemek:
I have one bar that I thing a good accumulator implementation **must pass**: it works with global distrib *and* there is no mention of global distrib anywhere in `mongoose_acc.erl`.

Related, module-specific fields like `amp_check_result`, `privacy_check` - should IMO also **not** be present in good accumulator implementation.

E.g. add a macro so that each mod can put/read things to/from accumulator inside its own namespace (`?MODULE`?), and other code can *explicitly* access other namespaces if needed. But I'm of a really strong opinion that the accumulator design cannot both be good and tightly coupled to mods.

bartekgorny:
> I still think that usage of this field will be abused 

That's what code reviewers are for - to prevent such abuse.

arcusfelis:
`set_user_jid` is easier to grep (less noise).

include/mongoose_acc.hrl
```diff
@@ -0,0 +1,8 @@
+%%%-------------------------------------------------------------------
+%%% @doc Creates a new accumulator instance setting location arguments to the caller of this macro.
+%%% @end
+%%%-------------------------------------------------------------------
+-define(new_acc(), mongoose_acc:new(?MODULE, ?FUNCTION_NAME, ?LINE)).
```
arcusfelis:
```erlang
#{module => ?MODULE, function => ?FUNCTION_NAME, line => ?LINE})
```

or

```erlang
-define(LOCATION, {?MODULE, ?FUNCTION_NAME, ?LINE}).
mongoose_acc:new(?LOCATION).
```

erszcz:
> `set_user_jid` is easier to grep (less noise)

Unless you import the function, you'd grep for `mongooseim_acc:user_jid`, wouldn't you? If you import, you can still grep for `import.mongooseim_acc` and building a regex alternative for the two is also an option.

fenek:
My humble suggestion. Personally I'm for flexible API but a one that makes it harder to clutter acc. It means lack of flat namespace. :)

```
#{
    %% No "global" ref for acc!
    location => #{
        timestamp => Timestamp,
        module => Module,
        function => Function,
        line => Line
    },

    origin => #{
        ref => Ref,
        el => El, %% XML element that triggered the processing
        iq_info => Info,
        jid => Jid
        % ... and some other cached values
    },

    stanza => #{
        ref = Ref2 %% Updated every time when 'stanza' key is updated
        %% Exactly (or almost) the same schema as origin
    },

    %% Has to be load tested
    hooks => #{
        history => [ {last_hook, result}, {previous_hook, result}, {two_hooks_ago, result} ]
    },

    %% Now time for per-module groups
    %% Maintained by convention, not enforcing API
    mod_global_distrib => #{
        key => val,
        strippable => false
    },
    mod_offline => #{
        stored => true,
        strippable => true
    }
}

%% MACROS:

-define(new_acc(), mongoose_acc:new(#{module => ?MODULE, function => ?FUNCTION_NAME, line => ?LINE})).
-define(new_acc(Element), mongoose_acc:new(#{module => ?MODULE, function => ?FUNCTION_NAME, line => ?LINE, el => Element})).

%% API:

-spec update(Acc :: acc(),
             Group :: term(), % atom?
             KV :: map() | [{K :: term(), V :: term()}]) -> acc().
update(_Acc, location, KV) ->
    error(immutable);
update(Acc, origin, KV) ->
    error_if_any_key_exists_in_this_group;
update(Acc, Group, KV) ->
    update.

update(Group, Key, Val) -> update(Group, [{Key, Val}]).

'get'(Acc, Group, Key)

'get'(Acc, Group, Key, Default)

%% Retain API for counters?

strip(Acc)

%% Further functions...
```

arcusfelis:
Often in code we have `proplists:get_value(key, List, default)`.
It's useful to have something similar for accs.

src/ejabberd_c2s.erl
```diff
@@ -1057,9 +1058,9 @@ process_outgoing_stanza(Acc, ToJID, <<"presence">>, StateData) ->
     NState;
 process_outgoing_stanza(Acc0, ToJID, <<"iq">>, StateData) ->
     Acc = mongoose_acc:require(xmlns, Acc0),
-    FromJID = mongoose_acc:get(from_jid, Acc),
-    Server = mongoose_acc:get(server, Acc),
-    El = mongoose_acc:get(element, Acc),
+    FromJID = mongoose_acc:get_from_jid(Acc),
+    Server = mongoose_acc:get_server(Acc),
+    El = mongoose_acc:get_element(Acc),
```
arcusfelis: sometimes I am dreaming about

```erlang
{FromJID, Server, El} = mongoose_acc:get3(from_jid, server, element, Acc)
```

arcusfelis: ok, I can almost tolerate style like this

```erlang
do_something_cool(Acc) ->
   %% so stupid, nothing is happening here, woohoo!
    FromJID = mongoose_acc:get_from_jid(Acc),
    Server = mongoose_acc:get_server(Acc),
    El = mongoose_acc:get_element(Acc),
    do_something_cool(FromJID, Server, El, Acc).

do_something_cool(FromJID, Server, El, Acc) ->
    %% actually some logic here
    ...
```


Not unpacking Acc in a middle of a function, than packing again and having Acc1, Acc2, .... Acc66 in 500 loc code :D 


TLDR: not related to this PR at all, but acc introduces a lot of boilerplate code.

src/ejabberd_c2s.erl
```diff
@@ -1258,26 +1259,26 @@ handle_incoming_message({send_text, Text}, StateName, StateData) ->
...
 handle_incoming_message({route, From, To, Acc0}, StateName, StateData) ->
     Acc = setup_accum(Acc0, StateData),
     Acc1 = ejabberd_hooks:run_fold(c2s_loop_debug, Acc, [{route, From, To}]),
-    Name = mongoose_acc:get(name, Acc1),
+    Name = mongoose_acc:get_element_name(Acc1),
     process_incoming_stanza_with_conflict_check(Name, From, To, Acc1, StateName, StateData);
 handle_incoming_message({send_filtered, Feature, From, To, Packet}, StateName, StateData) ->
     % this is used by pubsub and should be rewritten when someone rewrites pubsub module
-    Acc0 = mongoose_acc:new(),
+    Acc0 = ?new_acc(),
```
arcusfelis: Alternative idea to have descriptive type as an atom `mongoose_acc:new(pubsub_hack_in_c2s)`.
Than we can always grep to get location.

src/ejabberd_hooks.erl
```diff
@@ -128,7 +129,7 @@ run(Hook, Args) ->
           Host :: jid:server() | global,
           Args :: [any()]) -> ok.
 run(Hook, Host, Args) ->
```
kanes115: Maybe we should drop this function as it's deprecated and won't fit mongoose_acc's future form?

arkgil: Yes! It's confusing to see this in a codebase, since hook handlers receive accumulators anyway.

fenek:
Superseded by acc 2.0

PR #1864, opened 2018-05-18
[clt-rabbit] Add initial version of load testing scenario for XMPP with RabbitMQ

bszaf:
This is initial version of Amoc scenario, which works with MongooseIM pushing events to Rabbit.


load_test/xmpp_amqp.erl
```diff
@@ -0,0 +1,621 @@
...
+    Port = amoc_config:get(rabbit_port),
+    Username = amoc_config:get(rabbit_username),
+    Password = amoc_config:get(rabbit_password),
+    Params = #amqp_params_network{host = Host,
+                                  port = Port,
+                                  username = Username,
+                                  password = Password},
+    [connect_amqp(Params) || _ <- lists:seq(1, ?RABBIT_CONNECTIONS)].
+
+% @doc This helper function utilizes internal amqp_client supervisor as
```
mkacper: This function really took me by surprise. I think is't a little bit hacky in a way 😃 Maybe we can get rid of it by storing AMQP connections in an application configuration (we can achieve this using `application:set_env/3`) or in a dedicated ETS table? I believe it's worth considering because it'll definitely make the code easier to understand.

load_test/xmpp_amqp.erl
```diff
@@ -0,0 +1,621 @@
...
+        Error ->
+            exometer:update(?OPEN_CHANNEL_FAILURE, 1),
+            lager:error("Could not open channel to Rabbit, reason=~p", [Error]),
+            open_channel(Retries - 1)
+    end.
+
+-spec create_queue(ChannelPid :: pid(), QueueName :: binary()) -> ok | error.
+create_queue(Channel, QueueName) ->
+    Declare = #'queue.declare'{queue = QueueName},
+    case amqp_channel:call(Channel, Declare) of
```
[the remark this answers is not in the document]
bszaf: It is not clear for me how to handle potential error (like opening new channel may be required). I decided to leave it like that. Still we are able to notice if something is not working properly.

Same comment may be applied to `bind_queue_to_exchange/4` and `subscribe_to_queue/2`.

load_test/xmpp_amqp.erl
```diff
@@ -0,0 +1,647 @@
...
+
+%================================================
+% XMPP actions
+%================================================
+
+-spec connect_xmpp(proplists:proplist()) -> escalus:client().
+connect_xmpp(Cfg) ->
+    connect_xmpp(Cfg, ?MAX_RETRIES).
+
+-spec connect_xmpp(proplists:proplist(), non_neg_integer()) -> escalus:client().
```
ludwikbukowski: do we return `escalus:client()` in case of error?

load_test/xmpp_amqp.erl
```diff
@@ -0,0 +1,647 @@
...
+                               consumer_tag = <<"amoc">>,
+                               no_ack = true},
+    try amqp_channel:subscribe(ChannelPid, Consume, self()) of
+        #'basic.consume_ok'{} ->
+            lager:info("Successfully subscribed to queue ~p ", [QueueName]),
+            ok
+    catch
+        Err ->
+            lagger:error("Couldnt subscribe to queue ~p, reason: ~p", [QueueName, Err]),
+            error
```
ludwikbukowski: Can we return `{error, Err}` ? It's more meaningful

load_test/xmpp_amqp.erl
```diff
@@ -0,0 +1,647 @@
...
+-define(IS_XMPP(ID), ID rem 10 =/= 0).
+
+%% @doc This function returns role for specific user.
+-spec get_my_role(amoc_scenario:user_id()) -> role().
+get_my_role(ID) when ?IS_AMQP(ID) -> amqp;
+get_my_role(ID) when ?IS_XMPP(ID) -> xmpp.
+
+%% @doc This function returns a list of neighbours which connects via xmpp
+-spec xmpp_neighbors(amoc_scenario:user_id()) -> [amoc_scenario:user_id()].
+xmpp_neighbors(MyID) ->
```
mentels: Why such complicated _algorithm_ when you can go as simple as this (made appropriately readable):

```erlang
lists:filter(Pred,
   lists:seq(erlang:max(0,Id-Prev), Id-1) ++ lists:seq(Id+1,erlang:min(Id+Prev,Limit))
).
```

mentels: What does actually lazy mean here?

bszaf: I like `get_sequence` function, but I came with simpler solution:
```
-spec xmpp_neighbors(amoc_scenario:user_id()) -> [amoc_scenario:user_id()].
xmpp_neighbors(MyID) ->
    AmocUsers = amoc_config:get(amoc_users),

    PrevNeighbors = [ ID || ID <- lists:seq(1, MyID-1), ?IS_XMPP(ID) ],
    LimitedPrevNeighbors = lists:sublist(lists:reverse(PrevNeighbors), ?NUMBER_OF_PREV_NEIGHBOURS),

    NextNeighbors = [ ID || ID <- lists:seq(MyID+1, AmocUsers), ?IS_XMPP(ID) ],
    LimitedNextNeighbors = lists:sublist(NextNeighbors, ?NUMBER_OF_PREV_NEIGHBOURS),

    LimitedPrevNeighbors ++ LimitedNextNeighbors.
```
It works in the same way:
 - do not exceed boundaries - `1 .. amoc_config:get(amoc_users]`
 - filters out AMQP clients
 - returns ?NUMBER_OF_*_NEIGHBOURS ids, but does not exceed boundaries

```
amoc_users = 20.
-define(IS_AMQP(ID), ID rem 10 == 0).
-define(IS_XMPP(ID), ID rem 10 =/= 0).

1> xmpp_amqp:xmpp_neighbors(1).
[2,3,4,5]
2> xmpp_amqp:xmpp_neighbors(5).
[4,3,2,1,6,7,8,9]
3> xmpp_amqp:xmpp_neighbors(9).
[8,7,6,5,11,12,13,14]
4> xmpp_amqp:xmpp_neighbors(11).
[9,8,7,6,12,13,14,15]
5> xmpp_amqp:xmpp_neighbors(20).
[19,18,17,16]
6> xmpp_amqp:xmpp_neighbors(0).
** exception error: no function clause matching lists:seq(1,-1) (lists.erl, line 243)
     in function  xmpp_amqp:xmpp_neighbors/1 (/Users/bartoszszafran/projects/amoc/_build/default/lib/amoc/scenarios/xmpp_amqp.erl, line 195)
7> xmpp_amqp:xmpp_neighbors(21).
** exception error: no function clause matching lists:seq(22,20) (lists.erl, line 243)
     in function  xmpp_amqp:xmpp_neighbors/1 (/Users/bartoszszafran/projects/amoc/_build/default/lib/amoc/scenarios/xmpp_amqp.erl, line
```

What do you think @mentels?

bszaf:
@mentels @mkacper @ludwikbukowski Thanks for the review.

Addressed your comments, and added a few improvements. Please check it again.

load_test/xmpp_amqp.erl
```diff
@@ -242,11 +242,19 @@ xmpp_do(#{neighborIDs := NeighborIDs,
             send_presence_available(Client);
         {scheduled, send_message} ->
             schedule(send_message, SendMessageAfter),
-            send_messages_to_neighbors(Client, NeighborIDs)
+            send_messages_to_neighbors(Client, NeighborIDs);
+        {system, {From, Mref}, get_state} -> % allows to do sys:get_state(Pid)
```
mkacper: Why do we need this?

load_test/amoc_scenarios/xmpp_amqp.erl
```diff
@@ -302,6 +310,18 @@ amqp_do(State) ->
...
     amqp_do(NewState).
 
+-spec process_presence_message(#'amqp_msg'{}) -> ok.
+process_presence_message(#'amqp_msg'{payload = P}) ->
+    lager:debug("Got presence message ~p", [P]),
+    ok.
+
+-spec process_chat_message(#'amqp_msg'{}) -> ok.
+process_chat_message(#'amqp_msg'{payload = P}) ->
+    lager:debug("Got chat message ~p", [P]),
```
mkacper: Maybe we should first update metrics to make measurement as accurate as we can? I also think that it could be better to log a decoded message.

load_test/xmpp_amqp.erl
```diff
@@ -242,11 +242,19 @@ xmpp_do(#{neighborIDs := NeighborIDs,
...
+            From ! {Mref, S}
     end,
     xmpp_do(S).
 
 -spec process_stanza(exml:element()) -> ok.
+process_stanza(#xmlel{name = <<"message">>, attrs = A}) ->
```
mentels: What is stanza is not a `message`?

bszaf: If stanza is not the message, it will match into second clause, which literally does nothing.

Apart from that there is `escalus_connection:set_filter_predicate(Client, fun escalus_pred:is_chat_message/1)` in `xmpp_start/1`.

load_test/xmpp_amqp.erl
bszaf: I found it pretty useful for debugging purpose.

Is possible to do:
```
[{_, P, _, _} | _] = supervisor:which_children(amoc_users_sup),
sys:get_state(P)
```

Of course we can obtain such info with tracing, however I find this solutions more suitable for some cases. You just do not have to wait for some event, you just ask a worker for its state.

bszaf:
@mkacper 

I will rebase when scenario gets accepted. It will be easier for you to follow changes, when history is not rewritten.

bszaf:
@mentels Ok, Im going to prefix all metrics, if they are related to `xmpp` or `amqp`.

load_test/amoc_scenarios/xmpp_amqp.erl
bszaf: - You are right, its better to make measurement as fast as we can.
- I think it is better to log RAW message. It may be malformed.

mkacper: The only way that a message gets malformed and decoded process fails is when you change the Mongoose source code. In such a situation tests wouldn't pass so I'm not sure if the case you described fits here. IMO logging decoded message would increase log readability.

PR #1868, opened 2018-05-21
[clt-rabbit] Handle chat messages

mkacper:
Closes esl/tide#415.

### DIFF
--- a/big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
+++ b/big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
@@ -26,6 +26,7 @@
 -import(distributed_helper, [mim/0,
                              rpc/4]).
 
+-define(QUEUE_NAME, <<"test_queue">>).
 -define(PRESENCE_EXCHANGE, <<"custom_presence_exchange">>).
 -define(MOD_EVENT_PUSHER_RABBIT_CFG, [{presence_exchange, ?PRESENCE_EXCHANGE},
                                       {amqp_host, "localhost"},
@@ -201,13 +202,19 @@ close_rabbit_connection(Config) ->
                                             Config :: proplists:proplist()) ->
     ok | term().
 listen_to_presence_events_from_rabbit(JIDs, Config) ->
+    QueueBindings = presence_bindings(?QUEUE_NAME, JIDs),
+    listen_to_events_from_rabbit(QueueBindings, Config).
+
+-spec listen_to_events_from_rabbit(QueueBindings :: [rabbit_binding()],
+                                   Config :: proplists:proplist()) ->
+    ok | term().
+listen_to_events_from_rabbit(QueueBindings, Config) ->
     Connection = proplists:get_value(rabbit_connection, Config),
     Channel = proplists:get_value(rabbit_channel, Config),
-    Queue = declare_rabbit_queue(Channel),
-    QueueBindings = presence_bindings(Queue, JIDs),
+    declare_rabbit_queue(Channel, ?QUEUE_NAME),
     wait_for_exchanges_to_be_created(Connection, [?PRESENCE_EXCHANGE]),
     bind_queues_to_exchanges(Channel, QueueBindings),
-    subscribe_to_rabbit_queue(Channel, Queue).
+    subscribe_to_rabbit_queue(Channel, ?QUEUE_NAME).
 
 -spec wait_for_exchanges_to_be_created(Connection :: pid(),
                                        Exchanges :: [binary()]) -> pid().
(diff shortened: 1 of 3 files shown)


src/event_pusher/mod_event_pusher_rabbit.erl
```diff
@@ -85,6 +96,53 @@ publish_user_presence_change(JID, Status) ->
...
+                                                          packet => Packet,
+                                                          topic => Topic}).
+
+-spec publish_user_chat_message_received_event(From :: jid:jid(),
+                                               To :: jid:jid(),
+                                               Packet :: exml:element()) -> ok.
+publish_user_chat_message_received_event(From = #jid{lserver = Host}, To,
+                                         Packet) ->
+    Topic = opt(Host, chat_msg_recv_topic, ?DEFAULT_CHAT_MSG_RECV_TOPIC),
+    publish_user_chat_message_event(user_chat_msg_recv, #{from => From,
```
bszaf: I can see not point to use maps here. It is used just to pass parameters (All keys are mandatory). I think those options should be passed as separate arguments.

src/mongoose_rabbit_worker.erl
```diff
@@ -133,12 +139,34 @@ declare_exchange(Channel, Exchange) ->
...
+
+-spec publish_chat_msg_received(Channel :: pid(), EventData :: map()) -> ok.
+publish_chat_msg_received(Channel, EventData = #{to_jid := To,
+                                                 topic := Topic}) ->
+    RoutingKey = user_topic_routing_key(To, Topic),
+    publish_chat_msg_event(Channel, RoutingKey, EventData).
+
+-spec publish_chat_msg_event(Channel :: pid(), RoutingKey :: binary(), map()) ->
+    term().
+publish_chat_msg_event(Channel, RoutingKey, #{from_jid := From,
```
bszaf: Im not sure if this helper is necessary. It just wraps 2 function calls. I think such decoupling is not helpful, because without this function, code it really readable and code duplication is minimal.
```
    RoutingKey = user_topic_routing_key(From/To, Topic),
    Message = make_chat_msg(From, To, UserMsg),
    publish(Channel, Exchange, RoutingKey, Message).
```

Also, after removal we would have consistent flow for all events:

```
+---------------------------------------+        +----------------------+
|                                       |        |                      |
| handle_cast(user_presence_changed, _) +-------->  publish_status(...) +---------------------------+
|                                       |        |                      |                           |
+---------------------------------------+        +----------------------+                           |
                                                                                                    |
                                                                                                    |
+---------------------------------------+        +-----------------------------+         +----------v-------+
|                                       |        |                             |         |                  |
| handle_cast(user_chat_msg_sent, _)    +-------->  publish_chat_msg_sent(...) +--------->   publish(...)   |
|                                       |        |                             |         |                  |
+---------------------------------------+        +-----------------------------+         +----------^-------+
                                                                                                    |
                                                                                                    |
+---------------------------------------+        +-----------------------------+                    |
|                                       |        |                             |                    |
| handle_cast(user_chat_msg_recv, _)    +-------->  publish_chat_msg_rec (...) +--------------------+
|                                       |        |                             |
+---------------------------------------+        +-----------------------------+

```

src/event_pusher/mod_event_pusher_rabbit.erl
```diff
@@ -85,6 +96,53 @@ publish_user_presence_change(JID, Status) ->
...
+    ChatMsgExchange = opt(Host, chat_msg_exchange, ?DEFAULT_CHAT_MSG_EXCHANGE),
+    wpool:cast(pool_name(Host), {Event, #{from_jid => FromJID,
+                                          to_jid => ToJID,
+                                          msg => Message,
+                                          exchange => ChatMsgExchange,
+                                          topic => Topic}},
+               available_worker).
+
+-spec jid_from_chat_event(JID :: jid:jid()) -> binary().
+jid_from_chat_event(JID) ->
```
bszaf: Function name is misleading. Actually it is `jid_to_lower_binary`. Also it may be reused in `publish_user_presence_change`.

big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
```diff
@@ -231,17 +373,27 @@ wait_for_exchanges_to_be_deleted(Connection, Exchanges) ->
      || Exchange <- Exchanges],
     ok.
 
--spec declare_rabbit_queue(Channel :: pid()) -> binary().
-declare_rabbit_queue(Channel) ->
-    #'queue.declare_ok'{queue = Queue} =
-        amqp_channel:call(Channel, #'queue.declare'{exclusive = true}),
-    Queue.
+-spec declare_rabbit_queue(Channel :: pid(), Queue :: binary()) -> binary().
+declare_rabbit_queue(Channel, Queue) ->
```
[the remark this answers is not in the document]
mkacper: Good point, I'll add an explanatory comment here :)

src/event_pusher/mod_event_pusher_rabbit.erl
mkacper: I decided to use a map here because I'd like to avoid writing a function that accepts too many parameters. Without that it would be 5 arguments passed to the function. Apart from that I think that those arguments are connected with each other and make a logical whole. @bszaf do you see my point now?

src/event_pusher/mod_event_pusher_rabbit.erl
mkacper: 👍

PR #1873, opened 2018-05-22
Use mediumblob for mam message storage

arcusfelis:
Proposed changes include:
* upgraded eodbc, that can handle case when a part of binary data is in preallocated buffer already
* schema change for mysql to use mediumblob where blob was used.


fenek:
Sorry for cancelling Travis but we need it for release for now. :(

big_tests/tests/mam_helper.hrl
```diff
@@ -18,15 +18,31 @@
...
+            ExtraArgs = maps:fold(fun(MK,MV,A) -> A ++ [MK,MV] end, [], Extra0),
+            ExtraFmt = lists:append(lists:duplicate(maps:size(Extra0), "~n~p = ~p")),
+            ?safe_fail("assert_equal_extra(~s, ~s)~n\t"
+                        "Expected ~p~n\tValue ~p" ++ ExtraFmt,
+                       [??E, ??V, E0, V0|ExtraArgs]);
+        (E0, V0, Extra0) when E0 =/= V0 ->
+            ?safe_fail("assert_equal_extra(~s, ~s)~n\t"
+                        "Expected ~p~n\tValue ~p~nExtra ~p~n",
+                       [??E, ??V, E0, V0, Extra0]);
+        (_, _, _) ->
```
ludwikbukowski: Can we name and define this fun somwhere (e.g `fun perserve_ordering/3`) and call it? It's hard to read what does it really do here

big_tests/tests/rdbms_SUITE.erl
```diff
@@ -117,9 +142,20 @@ binary_values() ->
     %% https://bugs.erlang.org/browse/ERL-421
```
fenek: Like above

big_tests/tests/rdbms_SUITE.erl
```diff
@@ -117,9 +142,20 @@ binary_values() ->
...
     %% I like otp
     binary:copy(<<5>>, 1024*5),
-    binary:copy(<<8>>, 8002)
+    binary:copy(<<8>>, 8002),
+    binary:copy(<<0>>, 100000)
+%   binary:copy(<<16>>, 16777215) %% causes timeout with mssql
```
fenek: Wouldn't it be better to skip this part only for MSSQL?

PR #1875, opened 2018-05-22
remove `pooler` dependency from rebar.config

ludwikbukowski:
It's not used in Mongooseim at all


arcusfelis:
cqerl uses pooler. 
So, the dep would be in rebar.lock all the time

PR #1879, opened 2018-05-22
Skip install steps for pkg preset

arcusfelis:
This PR addresses "Travis takes forever"

Proposed changes include:
- skipping some steps
- upload logs only on error - reverted for now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,10 +36,6 @@ after_failure:
         - if [ -s _build/mim2/rel/mongooseim/log/crash.log ]; then cat _build/mim2/rel/mongooseim/log/crash.log; fi
         - tail -100 _build/mim1/rel/mongooseim/log/ejabberd.log
         - tail -100 _build/mim2/rel/mongooseim/log/ejabberd.log
-        - tools/travis-prepare-log-dir.sh
-        - if [ -n "${AWS_SECRET_ACCESS_KEY}" ]; then tools/travis-upload-to-s3.sh; fi
-        - if [ -n "${GDRIVE_SERVICE_ACCOUNT_CREDENTIALS}" ]; then tools/travis-upload-to-gdrive.sh; fi
-        - tools/travis-secure-erase.sh /tmp/serviceAccountCredentials
 
 after_success:
         # env variables help to distinguish CI builds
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -35,22 +36,20 @@ after_failure:
...
-        - ./rebar3 coveralls send
+        - test 1 = "$SKIP_COV" || codecov --disable=gcov --env PRESET
+        - test 1 = "$SKIP_COV" || ./rebar3 coveralls send
         - if [ -n "${DOCKERHUB_PASS}" ] && [ $PRESET = 'internal_mnesia' ] && [ $TRAVIS_OTP_RELEASE = "19.3" ];
             then tools/travis-build-and-push-docker.sh;
           fi
 
-after_script:
-        - tools/travis-prepare-log-dir.sh
-        - if [ -n "${AWS_SECRET_ACCESS_KEY}" ]; then tools/travis-upload-to-s3.sh; fi
```
michalwski: We've been here some time ago. In my opinion it's needed to upload results to s3 in any case. I remember a situation when we needed to access old results from successful jobs, but they weren't there. That's why since then test results are uploaded after every job (failed or not).

.travis.yml
```diff
@@ -77,14 +76,17 @@ matrix:
     include:
         - otp_release: 20.0
           env: PRESET=dialyzer_only
+               SKIP_RELEASE=1 SKIP_BUILD_TESTS=1 SKIP_COV=1
```
michalwski: That's a good idea.

bartekgorny:
A nice thing would be to upload only tests which failed, or generate a summary with links to them, so that one doesn't have to click his way through them all to find the one that failed.

.travis.yml
arcusfelis: Can we upload it when:
- branch is master
- or build is failing?

Because I don't want to wait 3 minutes per task for PR builds.

arcusfelis: reverted for now.
Any ideas how to optimize uploading is a topic for a separate PR.

michalwski: >
>Can we upload it when:
>  * branch is master
>  * or build is failing?
> Because I don't want to wait 3 minutes per task for PR builds.

I think it's enough to upload them for master branch build, tag builds and all failed builds.

.travis.yml
```diff
@@ -77,14 +83,19 @@ matrix:
     include:
         - otp_release: 20.0
           env: PRESET=dialyzer_only
+               SKIP_RELEASE=1 SKIP_BUILD_TESTS=1 SKIP_COV=1 SKIP_REPORT_UPLOAD=1
         - otp_release: 20.0
           env: PRESET=pgsql_mnesia DB=pgsql REL_CONFIG="with-pgsql with-jingle-sip"
         - otp_release: 18.3
           env: PRESET=riak_mnesia DB=riak REL_CONFIG=with-riak
-        - otp_release: 19.3
+        - language: generic
```
michalwski: How much time does this change save?

.travis.yml
```diff
@@ -77,14 +83,19 @@ matrix:
     include:
         - otp_release: 20.0
           env: PRESET=dialyzer_only
+               SKIP_RELEASE=1 SKIP_BUILD_TESTS=1 SKIP_COV=1 SKIP_REPORT_UPLOAD=1
```
michalwski: `SKIP_CONFIGURE` could be added here as well. Moreover, configure makes sense if combined with a release. I suggest that we use only one var, for instance `SKIP_RELEASE`. What do you think?

.travis.yml
arcusfelis: 10 seconds * two builds. But I just don't want to install stuff we don't need for packaging.

.travis.yml
arcusfelis: good idea, pushed.

PR #1883, opened 2018-05-22
Parallel pull for docker-compose

arcusfelis:
This PR addresses "Travis is slow!".

Proposed changes include:
* use docker pull --parallel
* use "docker-compose tries to resolve an ip, so slow" - not sure, if we actually need it.


michalwski:
This is quite good idea, on the other hand after quick comparison with old jobs, this change doesn't save much time (around 10-20s). It still may be useful when we have more containers in tests.

arcusfelis:
true, also sometimes the docker pull can take 3-11 minutes for some reason (travis is slooow?).
In this case parallel pull can be useful.

PR #1885, opened 2018-05-23
Use s3-parallel-put instead of aws-s3-sync

arcusfelis:
aws-s3-sync: 180 seconds
aws-s3-copy: 100 seconds
s3-parallel-put: 27 seconds

### DIFF
--- a/tools/travis-upload-to-s3.sh
+++ b/tools/travis-upload-to-s3.sh
@@ -6,11 +6,21 @@ set -euo pipefail
 
 CT_REPORTS=$(ct_reports_dir)
 
+if [ ! -d "${CT_REPORTS}" ]; then
+    echo "Skip uploading, because $CT_REPORTS directory does not exist"
+    exit 0
+fi
+
 echo "Uploading test results to s3"
 echo $(s3_url ${CT_REPORTS})
 
 echo "Installing awscli"
-sudo time pip install awscli --ignore-installed six
+sudo time pip install boto python-magic
+wget -O tools/s3-parallel-put https://raw.githubusercontent.com/arcusfelis/s3-parallel-put/e2f66d215eb6b3e376f9bafc4f52891ce671eff0/s3-parallel-put
+chmod +x tools/s3-parallel-put
+
+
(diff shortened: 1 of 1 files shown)


michalwski:
@arcusfelis this is very interesting research you are doing in this PR. There is one thing worth checking/remembering when optimising test results uploads. In some cases jobs are restarted, currently with `aws s3 sync` old data is preserved and new logs are added. I'd like to have this functionality in the optimised uploader.

tools/travis-upload-to-s3.sh
```diff
@@ -22,5 +22,6 @@ chmod +x tools/s3-parallel-put
 
 AWS_S3_HOST="${AWS_S3_HOST:-s3.amazonaws.com}"
 
-echo "Uploading"
+FILE_COUNT=$(find "${CT_REPORTS}" -type f | wc -l)
```
michalwski: How long does it take to calculate number of files to upload? Maybe it's not worth counting?

arcusfelis: very fast, less than a half of a second.

arcusfelis: We need to know number of files, because it would be a next optimization most likely.
Because each build generates 30k files (5k files per task * 6 tasks)

Maybe we can compress successful builds test results with zip before uploading...

arcusfelis:
@michalwski sync tries to optimize situation, when only small number of files were modified.
cp --recursive would overwrite files on conflict.
s3-parallel-put should do the same.

michalwski:
>aws-s3-sync: 180 seconds
>aws-s3-copy: 100 seconds
>s3-parallel-put: 27 seconds

On what job? I hope you checked on the same job, didn't you?

michalwski:
I see that first job passed: http://opensource.erlang-solutions.com/mongooseim-ct-reports/s3_reports.html?prefix=PR/1885/4752/internal_mnesia.19.3/big/ I'm restarting it to make sure that new results are uploaded and old are preserved.

michalwski:
All right, restarted test added new logs without removing old ones. We are good to merge this. Good job @arcusfelis !

michalwski:
Let's save some minutes!

PR #1892, opened 2018-05-29
Faster rebar3 deps

arcusfelis:
This PR addresses "slow travis?".

Proposed changes include:
* Use rebar3 for big_tests
* Cache deps archives in `~/.cache/rebar3/`

Before: get-deps step 41 seconds
Now: 12 seconds

TODO:
- clean expired cache entries from `.cache/rebar3/gitcache`
Does hex clean its cache? 

Does removing file from a directory is enough to remove it from the cache?
We can clean cache from time-to-time, when needed.

DONE:
- README https://github.com/arcusfelis/rebar3-faster-deps-plugin


erszcz:
It wouldn't hurt if you rewrote the README of https://github.com/arcusfelis/rebar3-faster-deps-plugin to actually describe what the plugin does, instead of leaving the original one from `rebar_tidy_deps` ;) Your code might be way more useful if people actually know how to use it.

arcusfelis:
@erszcz nobody reads readmes :D

arcusfelis:
@erszcz updated readme

rebar.lock
```diff
@@ -240,5 +241,6 @@
  {<<"jsx">>, <<"749BEC6D205C694AE1786D62CEA6CC45A390437E24835FD16D12D74F07097727">>},
  {<<"lhttpc">>, <<"2137CB82D123751B21CC0461A00734CD32CB470BC6278BA139EE13C1ECC3C2B1">>},
  {<<"observer_cli">>, <<"F3A47C22327424291758167DF3C4600DCF8BF34795A3787612019AA5B86A5C21">>},
- {<<"parse_trans">>, <<"1BAD3B959941CC53FFD6F4769A5D2666F9CDF179B2D62826636497D3FBAD9EC0">>}]}
+ {<<"parse_trans">>, <<"1BAD3B959941CC53FFD6F4769A5D2666F9CDF179B2D62826636497D3FBAD9EC0">>},
+ {<<"pooler">>, <<"898CD1FA301FC42D4A8ED598CE139B71CA85B54C16AB161152B5CC5FBDCFA1A8">>}]}
```
arcusfelis: https://github.com/esl/MongooseIM/pull/1875 pooler should not come BACK!

rebar.lock
```diff
@@ -37,7 +37,7 @@
  {<<"eini">>,{pkg,<<"eini">>,<<"1.2.4">>},1},
  {<<"eodbc">>,
   {git,"git://github.com/arcusfelis/eodbc.git",
-       {ref,"7dfed7e29cbfc9497e7969ea84aac2c96b3dcb81"}},
+       {ref,"528e5e9a51263e6938b7140970893b8128553abc"}},
```
arcusfelis: 528e5e9a51263e6938b7140970893b8128553abc is in master, so no changes in this file.

erszcz:
Looks good to me. Congrats for reaching a measurable speedup!

michalwski:
That's really big improvement to build times! I see one potential disadvantage  (at least for me while working on the code locally). Since now deps will be downloaded as zip archives from github, they'll not be git repositories. This means it will not be easy to see the git history of a dep by simply calling `git log` somewhere in the `_build/default/lib` dir. I happen to do it several times a week. If others are fine with this, we can merge and be happy about build speed up. Alternatively the optimised deps fetching could be used online on Travis (based on one of the envs exported by Travis for every build).

erszcz:
> Since now deps will be downloaded as zip archives from github, they'll not be git repositories.

If we're talking about local checkouts I don't like it too. We can either (auto?)detect whether a build is a Travis one and enable the plugin only there, not in local checkouts, or we can use a more sophisticated approach similar to kerl. I have a hunch it's less work to autodetect Travis.

For each git repository kerl accesses (e.g. different OTP forks, plugin repos, etc) [it mirrors the repository in a local cache](https://github.com/kerl/kerl/blob/9c615ea9cc2723cae5d0c0bf9cc9965350afebe7/kerl#L390-L404) (see side not below for `GIT` variable meaning). Then, for each build it runs, [it clones from the local mirror](https://github.com/kerl/kerl/blob/9c615ea9cc2723cae5d0c0bf9cc9965350afebe7/kerl#L408-L412). In our Travis build, the "local cache" would be the Travis cache.

Side note: [the `GIT` variable](https://github.com/kerl/kerl/blob/9c615ea9cc2723cae5d0c0bf9cc9965350afebe7/kerl#L390) is created in such a strange way to make sure URLs don't cause trouble when interpreted as paths in a local filesystem. It's just a checksum of the full repository address.

arcusfelis:
@michalwski I've updated the plugin. It prints now

```erlang
===> [rebar_faster_deps] Rewrite locks                           
===> [rebar_faster_deps] To clone using git command, run git deps with PREFER_GIT_CLONE=true
===> [rebar_faster_deps] Example: PREFER_GIT_CLONE=true ./rebar3 get-deps  
```

So, first time you run get-deps, only zips are downloaded.

But if you want git for deps, you run `PREFER_GIT_CLONE=true ./rebar3 get-deps` once.

After that, the deps become local git repositories and the plugin handles them the same way as rebar3 would treat them. Even if you do not pass `PREFER_GIT_CLONE` the next time you run `get-deps`.

But if run `make clean`, you would need to set `PREFER_GIT_CLONE` again.

Actually, you can set `PREFER_GIT_CLONE` in  bashrc, and the plugin would be disabled for you.

The reason why I still want to download zip-s by default is that it reduces building time and _build directory size for every person, who builds MongooseIM from source.

michalwski:
@arcusfelis that's fine by me. @erszcz what do you think?

erszcz:
@arcusfelis @michalwski It looks really good now!

PS: The printout probably should say "run get-deps with" instead of "run git deps with".

arcusfelis:
Updated message

```erlang
===> [rebar_faster_deps] To clone using git command, set env variable PREFER_GIT_CLONE=true and run rebar3 get-deps               
===> [rebar_faster_deps] Example: PREFER_GIT_CLONE=true ./rebar3 get-deps
```

PR #1897, opened 2018-05-30
Fix database deadlock handling in mod_mam_odbc_prefs

arcusfelis:
The old way of matching abortion type is error-prone.
Basically, we changed `{aborted, Reason}` to `{aborted, #{reason=>Reason, sql_query => ....}` and it broke the logic.

So, let's not much on Reason.


arcusfelis:
Adding test to reproduce the deadlock is tricky.

PR #1898, opened 2018-05-31
Ported User Leave and Join muc hooks from ejabberd

SamuelNichols:
This PR addresses lack of hook on users leaving and entering multi-user chats

Proposed changes include:
* Hooks for user enter/leave muc
* No tests added, not sure what to do in this. Will add tests if pointed in the right direction (if needed).
* Added hook to list of hooks in developer documentation hooks_description.md

### DIFF
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -1684,8 +1684,8 @@ remove_online_user(JID, StateData, Reason) ->
     Sessions = case is_last_session(Nick, StateData) of
         true ->
             add_to_log(leave, {Nick, Reason}, StateData),
-            run_leave_room_hook(JID, StateData),
             tab_remove_online_user(JID, StateData),
+            run_leave_room_hook(JID, StateData),
             dict:erase(Nick, StateData#state.sessions);
         false ->
             IsOtherLJID = fun(J) -> jid:to_lower(J) /= LJID end,
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -1693,7 +1693,7 @@ remove_online_user(JID, StateData, Reason) ->
     notify_users_modified(StateData#state{users = Users, sessions = Sessions}).
 
 -spec run_leave_room_hook(jid:jid(), state()) -> ok.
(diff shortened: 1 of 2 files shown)


src/mod_muc_room.erl
```diff
@@ -4334,6 +4334,9 @@ tab_add_online_user(JID, StateData) ->
     US = {LUser, LServer},
     Room = StateData#state.room,
     Host = StateData#state.host,
+    ServerHost = StateData#state.server_host,
+    MucJID = StateData#state.jid,
+    ejabberd_hooks:run(join_room, ServerHost, [ServerHost, Room, Host, JID, MucJID]),
```
arcusfelis: The function is called `tab_add_online_user/2`. So, it should update `muc_online_users` table and that's it.
A hook call can be done outside of the function, for example

```erlang
tab_add_online_user(JID, StateData),
run_join_room_hook(JID, StateData),
```

Than you need to put hook running logic inside a new function `run_join_room_hook`.

ludwikbukowski:
Can we add metrics for those hooks in this PR as well?

src/mod_muc_room.erl
SamuelNichols: So would you want me to add run_join_room_hook(JID, StateData) inside of this function?
https://github.com/esl/MongooseIM/blob/1bc8515f9c7f070ce157523a70910bef06f8b91c/src/mod_muc_room.erl#L1654 Then make a separate method for calling the hook?

SamuelNichols:
@ludwikbukowski Sorry, I'm somewhat new to this repo. What exactly do you mean by metrics for the hooks?

ludwikbukowski:
@SamuelNichols In Mongooseim we support range of metrics[1]. 
But I've just realised that for each new hook new metric will be added automatically, so no more changes to this PR needed to be done :)
[1] https://mongooseim.readthedocs.io/en/latest/operation-and-maintenance/Mongoose-metrics/

SamuelNichols:
So is there any else that needs to be done with this issue?

michalwski:
I just re-opened to test with current master branch. @arcusfelis could you check if your comments were addressed?

fenek:
Looks good to me :)

PR #1902, opened 2018-06-04
[clt-rabbit] Logs and metrics

mkacper:
Closes esl/tide#421.


include/mod_event_pusher_rabbit.hrl
```diff
@@ -0,0 +1,14 @@
+%%%===================================================================
+%%% Metrics
+%%%===================================================================
+
+-define(METRIC_MODULE, mod_event_pusher_rabbit).
```
mentels: Is the `METRIC_MODULE` name a standard one? `BACKEND` reads better in that context.

include/mod_event_pusher_rabbit.hrl
```diff
@@ -0,0 +1,14 @@
...
+-define(RABBIT_CONNECTIONS_METRIC,
+        [backends, ?METRIC_MODULE, rabbit_connections]).
+-define(MESSAGES_PUBLISH_METRIC, [backends, ?METRIC_MODULE, messages_publish]).
```
mentels: `published`? Or there's a reason for `publish`?

src/event_pusher/mod_event_pusher_rabbit.erl
```diff
@@ -53,10 +54,13 @@
 start(Host, _Opts) ->
     application:ensure_all_started(amqp_client),
     application:ensure_all_started(worker_pool),
+    initialize_metrics(Host),
+    mongoose_metrics:init_subscriptions(),
```
mentels: Are you **sure** you should have this call here? I mean isn't it intended to be called once per the whole system?

src/event_pusher/mod_event_pusher_rabbit.erl
```diff
@@ -177,6 +182,17 @@ exchanges(Host) ->
      opt(Host, groupchat_msg_exchange, ?DEFAULT_GROUP_CHAT_MSG_EXCHANGE)
     ].
 
+-spec initialize_metrics(Host :: jid:server()) -> ok.
+initialize_metrics(Host) ->
+    [mongoose_metrics:ensure_metric(Host, Name, Type)
+     || {Name, Type} <- [{?RABBIT_CONNECTIONS_METRIC, spiral},
```
mentels: Why not `gauge` for the number of connections?

src/mongoose_rabbit_worker.erl
```diff
@@ -132,18 +135,36 @@ publish_groupchat_msg_recv(Channel, EventData = #{to_jid := To,
...
+publish(Channel, Exchange, RoutingKey, Message, Host) ->
+    {PublishTime, _} = timer:tc(amqp_channel, call,
+                                [Channel,
+                                 #'basic.publish'{exchange = Exchange,
+                                                  routing_key = RoutingKey},
+                                 #amqp_msg{payload = Message}]),
+    case amqp_channel:wait_for_confirms(Channel) of
+        true ->
+            mongoose_metrics:update(Host, ?MESSAGES_PUBLISH_METRIC, 1),
+            mongoose_metrics:update(Host, ?MESSAGE_PUBLISH_TIME_METRIC,
```
mentels: Hmm this seem to be wrong for me as you put the time it takes to actually send the message not the time span between sending it and getting the confirmation.

src/mongoose_rabbit_worker.erl
```diff
@@ -168,9 +189,25 @@ make_chat_msg(From, To, UserMsg) ->
 is_user_online(online) -> true;
 is_user_online(offline) -> false.
 
--spec close_rabbit_connection(Connection :: pid(), Channel :: pid()) ->
+-spec establish_rabbit_connection(Opts :: #amqp_params_network{},
+                                  Host :: jid:server()) -> ok.
+establish_rabbit_connection(Opts, Host) ->
+    {ok, Connection} = amqp_connection:start(Opts),
+    {ok, Channel} = amqp_connection:open_channel(Connection),
+    mongoose_metrics:update(Host, ?RABBIT_CONNECTIONS_METRIC, 1),
```
mentels: Moreover, all the call to `mongoose_metrics:update/2` could be extracted into a separate local function.

src/mongoose_rabbit_worker.erl
```diff
@@ -132,18 +135,36 @@ publish_groupchat_msg_recv(Channel, EventData = #{to_jid := To,
...
-              Message :: binary()) -> term().
-publish(Channel, Exchange, RoutingKey, Message) ->
-    amqp_channel:call(Channel,
-                      #'basic.publish'{
-                         exchange = Exchange,
-                         routing_key = RoutingKey},
-                      #amqp_msg{payload = Message}).
+              Message :: binary(), Host :: jid:server()) -> term().
+publish(Channel, Exchange, RoutingKey, Message, Host) ->
+    {PublishTime, _} = timer:tc(amqp_channel, call,
```
mentels: Publish could folllow a pattern present in this module that all the calls to `amqp_*` are wrapped into small functions. That would result in calling two "locals": `call_channel` and `wait_for_confirm_on_channel` (sth along these lines)

src/event_pusher/mod_event_pusher_rabbit.erl
mkacper: I'll ask about that someone from MIM team.

src/event_pusher/mod_event_pusher_rabbit.erl
mkacper: Cause `spiral` simply fits here. Using `spiral` metric I can just increase a counter when a worker established a connection and decrease when it closed one. Moreover, we use `spiral` for the same type of metric in Amoc.

src/mongoose_rabbit_worker.erl
mkacper: It depends how we look at this. Do we want to measure time it takes to only publish a message or to publish and get the confirmation? But I think you're right and we should use the latter approach.

src/mongoose_rabbit_worker.erl
mkacper: I'm not sure that there is such a need. Isn't it readable enough?

mkacper:
@mentels please see my comments and introduced changes.

src/mongoose_rabbit_worker.erl
mentels: It's readable enough.

bszaf:
And I confirm it works well with load.

src/mongoose_rabbit_worker.erl
```diff
@@ -132,18 +135,34 @@ publish_groupchat_msg_recv(Channel, EventData = #{to_jid := To,
...
+        timer:tc(fun publish_message_and_wait_for_confirm/4,
+                 [Channel, Exchange, RoutingKey, Message]),
+    case Result of
+        true ->
+            mongoose_metrics:update(Host, ?MESSAGES_PUBLISHED_METRIC, 1),
+            mongoose_metrics:update(Host, ?MESSAGE_PUBLISH_TIME_METRIC,
+                                    PublishTime),
+            mongoose_metrics:update(Host, ?MESSAGE_PAYLOAD_SIZE_METRIC,
+                                    byte_size(Message)),
+            ?DEBUG("Message sent to RabbitMQ node: ~p", [Message]);
```
ludwikbukowski: Can we fit the logging format we usually use?
https://github.com/esl/MongooseIM/blob/logging-docs/doc/developers-guide/logging.md

mkacper:
@ludwikbukowski please see the fixed logging. Could you also address this comment: [click](https://github.com/esl/MongooseIM/pull/1902#pullrequestreview-126356686)?

mkacper:
@ludwikbukowski could you please take a look at https://github.com/esl/MongooseIM/pull/1902#discussion_r193025514 too?

ludwikbukowski:
I think we would need to update documentation regarding the metrics as well

src/mongoose_rabbit_worker.erl
```diff
@@ -132,18 +135,34 @@ publish_groupchat_msg_recv(Channel, EventData = #{to_jid := To,
...
+    {PublishTime, Result} =
+        timer:tc(fun publish_message_and_wait_for_confirm/4,
+                 [Channel, Exchange, RoutingKey, Message]),
+    case Result of
+        true ->
+            mongoose_metrics:update(Host, ?MESSAGES_PUBLISHED_METRIC, 1),
+            mongoose_metrics:update(Host, ?MESSAGE_PUBLISH_TIME_METRIC,
+                                    PublishTime),
+            mongoose_metrics:update(Host, ?MESSAGE_PAYLOAD_SIZE_METRIC,
+                                    byte_size(Message)),
```
ludwikbukowski: I think this metrics updates may be hidden in separate function e.g:
```
update_metrics_message_published(Host, Message, PublishTime) ->
   mongoose_metrics:update(Host, ?MESSAGES_PUBLISHED_METRIC, 1),
   mongoose_metrics:update(Host, ?MESSAGE_PUBLISH_TIME_METRIC, PublishTime),
   mongoose_metrics:update(Host, ?MESSAGE_PAYLOAD_SIZE_METRIC, byte_size(Message)).
```

ludwikbukowski: or you can even earlier:
```
handle_message_publish(Result, Host, Message, PublishTime) ->
 ...
```
and call it instead of matching on `Result`

src/mongoose_rabbit_worker.erl
```diff
@@ -168,9 +187,35 @@ make_chat_msg(From, To, UserMsg) ->
...
+
+-spec close_rabbit_connection(Connection :: pid(), Channel :: pid(),
+                              Host :: jid:server()) ->
     ok | term().
-close_rabbit_connection(Connection, Channel) ->
+close_rabbit_connection(Connection, Channel, Host) ->
+    mongoose_metrics:update(Host, ?RABBIT_CONNECTIONS_METRIC, -1),
     try amqp_channel:close(Channel)
     catch
         _Error:_Reason -> already_closed
```
ludwikbukowski: are we sure that "already closed" is really error here?

src/mongoose_rabbit_worker.erl
mkacper: You're right but I believe that it's only a matter of an approach a programmer chose :)

src/mongoose_rabbit_worker.erl
mkacper: Could you please explain what you meant here? We just try to close a channel and don't care about the result. We have to catch a possible exception cause if such occurs we won't be able to close a rabbit connection cause a worker process will die on this call.

mkacper:
@ludwikbukowski When I asked you about [the comment](https://github.com/esl/MongooseIM/pull/1902#pullrequestreview-126356686) I was thinking if we can leave the current solution as is or we should rewrite it as @bszaf suggested in the comment. What do you think?

Maybe because of the current solution regarding metrics I have to call `mongoose_metrics:init_subscriptions` again in the module to make them work?

ludwikbukowski:
Yes, we have api for metrics and I think we should be consistent
Maybe it could be even moved to separate module as its done for mam:
https://github.com/esl/MongooseIM/blob/master/src/mongoose_metrics_mam_hooks.erl

src/mongoose_rabbit_worker.erl
ludwikbukowski: hmm but why there might be exception?
If so, can we log it with `?ERROR( ...` ?
my point is that with this implementation, the error will be completely ignored

mkacper: There might be an exception because e.g. a Rabbit server closed a connection and we try to close the connection that is already closed, so there is no need to log this as error.

ludwikbukowski: IS it possible to distinguish this kind of exception ? _Error:_Reason matches on everything

mkacper: As far as I know this call causes an exception only if the channel we're trying to close is already closed or if some timeout is exceeded (it's `gen_server:call` underneath).

mkacper:
@mentels @ludwikbukowski I talk to @bszaf and we agreed that the current solution regarding metrics is good enough for now and eventually I changed it if necessary when we'll be preparing a final PR. Please see the PR once again. Thanks.

src/mongoose_rabbit_worker.erl
```diff
@@ -168,11 +182,91 @@ make_chat_msg(From, To, UserMsg) ->
...
     end,
     amqp_connection:close(Connection).
+
+-spec update_messages_published_metrics(Host :: jid:server(),
+                                        PublishTime :: non_neg_integer(),
+                                        Message :: binary()) -> any().
+update_messages_published_metrics(Host, PublishTime, Message) ->
```
mentels: Maybe make `mod_event_pusher_rabbit` provide an API for bumping the metrics?

Anyway, it a [circular dependency](https://en.wikipedia.org/wiki/Circular_dependency) which is [better to be avoided](http://learnyousomeerlang.com/modules).

Why don't you just put appropriate metrics in the worker module? You can always export a function that lists the required metrics and the `mod_event_push_rabbit` would call it to crate them in its `init_metrics/0`.

mkacper: @mentels please see the latest commit.

PR #1905, opened 2018-06-05
Strict control for starting supervised processes

arcusfelis:
This PR ensures that start_module would fail, if start_child fails.
It allows to catch badly started module at start time, not when we try to use it.

It helps with failing muc_http_api_SUITE, that fails with service unavailable reason sometimes.

```erlang
big_tests/tests/muc_http_api_SUITE.erl:    muc_helper:load_muc(muc_helper:muc_host()),
```

Proposed changes include:
* `ejabberd_sup:start_child/1` crashes when child already present
* I haven't touched places, when code expects to add child several times. i.e. ejabberd_users. It needs an extra refactoring effort.
* `muc_helper:load_muc/1` unloads already started module and proceeds instead of doing nothing.

### DIFF
--- a/src/ejabberd_users.erl
+++ b/src/ejabberd_users.erl
@@ -50,7 +50,7 @@ start(Host) ->
      5000,
      worker,
      [ejabberd_users]},
-    supervisor:start_child(ejabberd_sup, UserChildSpec),
+    ejabberd_sup:start_child(UserChildSpec),
     ejabberd_hooks:add(remove_user, Host, ?MODULE, remove_user, 50),
     ok.
 
--- a/src/ejabberd_users.erl
+++ b/src/ejabberd_users.erl
@@ -50,13 +50,15 @@ start(Host) ->
      5000,
      worker,
      [ejabberd_users]},
-    ejabberd_sup:start_child(UserChildSpec),
+    %% XXX lifecycle of this child is broken
+    supervisor:start_child(ejabberd_sup, UserChildSpec),
     ejabberd_hooks:add(remove_user, Host, ?MODULE, remove_user, 50),
     ok.
 
 
 -spec stop(jid:server()) -> 'ok'.
 stop(Host) ->
+    %% TODO properly remove ejabberd_sup child
     ejabberd_hooks:delete(remove_user, Host, ?MODULE, remove_user, 50),
     ok.
 
--- a/src/mod_mam_cache_user.erl
+++ b/src/mod_mam_cache_user.erl
@@ -97,6 +97,7 @@ writer_child_spec() ->
      [?MODULE]}.
 
 start_server(_Host) ->
+    %% TODO make per host server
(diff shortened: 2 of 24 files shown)


src/mod_mam_muc_cache_user.erl
```diff
@@ -75,6 +75,7 @@ writer_child_spec() ->
      [?MODULE]}.
 
 start_server(_Host) ->
+    %% XXX make per host server
```
michalwski: What does `XXX` mean here?

arcusfelis: can be TODO.
XXX signals that it's one of the places, where we don't follow pattern "each host has it's own servers". 
Potentially, can cause issues (cleaning? scalability?).

arcusfelis: changed :)

src/ejabberd_users.erl
```diff
@@ -50,13 +50,15 @@ start(Host) ->
      5000,
      worker,
      [ejabberd_users]},
+    %% XXX lifecycle of this child is broken
```
ludwikbukowski: can you elaborate more on that?

src/mod_mam_cache_user.erl
```diff
@@ -97,6 +97,7 @@ writer_child_spec() ->
      [?MODULE]}.
 
 start_server(_Host) ->
+    %% TODO make per host server
```
ludwikbukowski: Can we introduce it in this PR? It doesn't seem to be big change and its one "TODO" less :D

src/mod_mam_muc_cache_user.erl
```diff
@@ -75,6 +75,9 @@ writer_child_spec() ->
      [?MODULE]}.
 
 start_server(_Host) ->
+    %% TODO make per host server to be consistent with the way
```
ludwikbukowski: Can we introduce it in this PR? It doesn't seem to be big change and its one "TODO" less :D

src/ejabberd_users.erl
arcusfelis: it's not host specific. So, start/stop procedure would keep old data.
But otherwise we would have a table per host, which is not optimal too.

arcusfelis: ah, waaait.
It's host specific. But we don't check if it's already running

arcusfelis: So, I need mongoose_service here

arcusfelis: hmm, gen_mod deps support is broken, but would be required to convert this file into a module.
I am happy with this code for now, I don't want to change starting logic here YET.

src/mod_mam_muc_cache_user.erl
arcusfelis: nope, should be a separate PR to avoid making a giant change ;)

PR #1255, opened 2017-03-31
Change log formatter to logfmt

kzemek:
This PR addresses #1248

It changes lager formatter to logfmt-compatible one. The formatter works in conjunction with lager's metadata, so logged tags can be added per-process as in this pull request's ejabberd_c2s.erl . We still require discipline to write logs in `task=adding_things msg="what's going on"` format, though (discuss in #1248 ?).

PR #1906, opened 2018-06-05
Add some logging docs

arcusfelis:
This PR addresses "better docs".

https://github.com/esl/MongooseIM/blob/logging-docs/doc/developers-guide/logging.md


doc/developers-guide/logging.md
```diff
@@ -0,0 +1,103 @@
...
+-include("mongoose_logger.hrl").
+```
+
+or
+
+```erlang
+-include("mongoose.hrl").
+```
+
+# Logging macroses
```
arkgil: Should be "macros".

doc/developers-guide/logging.md
```diff
@@ -0,0 +1,103 @@
...
+                       "reason=~p:~p user=~ts stacktrace=~1000p",
+                       [Class, Reason, LUser, Stacktrace]),
+            erlang:raise(Class, Reason, Stacktrace)
+    end
+```
+
+`user=~ts` is often used too.
+
+
+A common way to name event is `event=function_name_failed`.
```
arkgil: Let's not advise it. Naming things is hard and our functions rarely have good names :P We can check which function has failed by enabling lager's `function` metadata.

arkgil:
What I don't like about it is that `event` and `reason` fields won't be caught by backends like https://github.com/esl/lager_graylog as a metadata, and thus won't interpreted by the logging aggregator service, Graylog in this case, as a separate field. Sure, you can always add a parser to the log aggregator so that it extracts those fields from the log message, but still, this is less then ideal. Maybe we could set the `event` and `reason` as a metadata? And then configure lager to include that metadata in the message.

doc/developers-guide/logging.md
arcusfelis: > A common way to name event is `event=function_name_failed`, but use common sense :)

arcusfelis:
@arkgil we were thinking about it, but nobody actually managed to make a code we can merge :). Actually, splunk parses KEY=VALUE just fine. ELK too after defining parser. It's also good to have a json lager backend for ELK.

arcusfelis:
@arkgil last trial I remember https://github.com/esl/MongooseIM/pull/1255/files

arkgil:
@arcusfelis I think that our logging standards require deeper discussion. I believe this PR is a good start! 🌮

arcusfelis:
@arkgil sure :) I usually write the same things in comments, so just put everything together.

doc/developers-guide/logging.md
```diff
@@ -0,0 +1,111 @@
...
+If an user sets log level 5, than he would see all messages in logs.
+
+Levels 3 or 2 are usually used for production systems.
+
+
+# Logging format
+
+We use modified [logfmt](https://brandur.org/logfmt) format.
+
+This format is Splunk and ELK friendly.
```
ludwikbukowski: [Splunk](https://www.splunk.com/en_us/solutions/solution-areas/log-management.html) and [ELK](https://www.elastic.co/elk-stack)

doc/developers-guide/logging.md
```diff
@@ -0,0 +1,111 @@
...
+WARNING_MSG("event=warning_event reason=~1000p", [Arg]),
+```
+
+Use them to log on corresponding log level.
+
+Don't use `ERROR_MSG` for cases that are not errors.
+
+# Logging levels
+
+A system operator can choose log level by setting loglevel in `mongooseim.cfg`.
```
ludwikbukowski: where is this file?

doc/developers-guide/logging.md
```diff
@@ -0,0 +1,111 @@
...
+%% Use
+?ERROR_MSG("event=check_password_failed reason=~p", [Reason])
+
+%% Don't use
+?ERROR_MSG("event=check_password_failed, reason=~p", [Reason])
+```
+
+# Filtering logs by module
+
+Setting loglevel in ejabberd.cfg to debug leads to a flood of messages in logs.
```
ludwikbukowski: What do you think?
```
Setting loglevel to `debug` leads to a flood of messages in logs.
To set different loglevel for just one module, call:
```

doc/developers-guide/logging.md
arcusfelis: still? :trollface:

doc/developers-guide/logging.md
```diff
@@ -0,0 +1,112 @@
...
+`user=~ts` is often used too.
+
+
+A common way to name an error event is `event=function_name_failed`.
+For example, `event=remove_user_failed`. Use the advice critically, it would
+not work well for any function. Counter example:
+
+```erlang
+handle_info(Info, State) ->
+    ?ERROR_MSG("issue=unexpected_info_received info=~1000p", [Info]),
```
fenek: I disagree, `event=unexpected_info_received` sounds equally well IMO.

doc/developers-guide/logging.md
```diff
@@ -0,0 +1,112 @@
...
+    {noreply, State}.
+```
+
+Log messages should not contain new lines.
+
+We usually use `~1000p` to log long datastructures.
+
+Use whitespace as a field separator:
```
fenek: What should be done about values that contain whitespaces?

doc/developers-guide/logging.md
```diff
@@ -0,0 +1,112 @@
...
+# Logging format
+
+We use modified [logfmt](https://brandur.org/logfmt) format.
+
+This format is [Splunk](https://www.splunk.com/en_us/solutions/solution-areas/log-management.html)
+and [ELK](https://www.elastic.co/elk-stack) friendly.
+
+`event=something_interesting` field is required.
+
+`reason=~1000p` field is commonly used.
```
fenek: I think we could use an info about default message length limit enforced by lager.

PR #1908, opened 2018-06-05
Fix travis scripts on macosx

arcusfelis:
This PR addresses compatibility issues with mac.

Proposed changes include:
- Partial support for deploying DB on the remote host (but our tests support only localhost anyway)
- sed/timeout/readarray usage on Mac.


tools/travis-common-vars.sh
```diff
@@ -4,6 +4,7 @@ TOOLS=`dirname $0`
 
 if [ `uname` = "Darwin" ]; then
     BASE=$(cd "$TOOLS/.."; pwd -P)
+    # Don't forget to install in with "brew install gnu-sed"
```
michalwski: `it` instead of `in`?

tools/wait_for_service.sh
```diff
@@ -6,9 +6,21 @@ if [ "$#" -ne 2 ]; then
...
-IP=$(/usr/bin/docker inspect -f {{.NetworkSettings.IPAddress}} "$CONTAINER")
+IP=$(docker inspect -f {{.NetworkSettings.IPAddress}} "$CONTAINER")
 echo "$CONTAINER IP is $IP"
 
-tools/wait-for-it.sh -h $IP -p $PORT
+if [ `uname` = "Darwin" ]; then
+  # Direct access to IPs is not supported on Mac
+  # https://docs.docker.com/docker-for-mac/networking/
+  # But we can run wait-for-it from another container
+  docker run --rm -d --name wait-helper ubuntu sleep infinity || echo "We can continue if the wait-helper exists"
```
michalwski: I see that here you used `ubuntu` image, but in other place (https://github.com/esl/MongooseIM/pull/1908/files#diff-4e96dc6d6584239b1e20b34ff2804857R52) you used `busybox`. Is it possible to have only one image?

michalwski: And why do we need docker container for this wait-for-it script? Isn't there other way?

tools/travis-common-vars.sh
```diff
@@ -45,3 +48,26 @@ function mktempdir
...
     mktemp -d "/tmp/$1.XXXXXXXXX"
 }
+
+function init_docker
+{
+    if [ "$REMOTE_RO_VOLUMES" = 'true' ]; then
+        export REMOTE_ROOT_DIR="$(mktempdir remote_mongooseim)"
+        docker rm -f mongooseim-uploader || echo "Skip mongooseim-uploader removal"
+        echo "Start remote uploader helper"
+        docker run --name mongooseim-uploader -v "$REMOTE_ROOT_DIR":/data -it busybox true
```
michalwski: Why exactly do we need this `mongooseim-uploader` container?

tools/wait_for_service.sh
arcusfelis: because mac suuu...

I mean it has old bash, no timeout command, it's hard to check tcp connection (ok, nc?), and wait_for_service.sh waits for port to appear on private IP (which is not available in Mac).

The ubuntu image is only used for Mac (i.e. once you download the image, it runs fast).

tools/travis-common-vars.sh
arcusfelis: We can't copy onto volume, that no container uses.
To copy we need to create a container with volume connected. Than we can copy onto it even if the container is not running.

We mount `$REMOTE_ROOT_DIR` twice, once in uploader container and once in the main service container.

tools/wait_for_service.sh
arcusfelis: You can test performance with

```erlang
DB=cassandra ./tools/travis-setup-db.sh
DB=mysql ./tools/travis-setup-db.sh
...
````

Comment (michalwski) [reply tools/wait_for_service.sh]: All right, can we use only one image? `ubuntu` or `busybox`?

Comment (michalwski) [reply tools/wait_for_service.sh]: Actually it'd be perfect to use `busybox` here and for the uploader. If that's not possible, we can live with this 2 I think.

File: tools/travis-common-vars.sh (hunk 3 above)
Comment (michalwski) [reply tools/travis-common-vars.sh]: So, can you copy to $REMOTE_ROOT_DIR (that's on local file system, right?) and use it only once in the main service container. I still don't understand why we need to copy via the additional container.

Comment (arcusfelis) [reply tools/travis-common-vars.sh]: REMOTE_ROOT_DIR is on remote system. This code is to support docker client on one machine and docker server on another.

Docker will just create an empty directory in case it runs on docker-machine. Again, we don't really need the code, so we can remove support for DEFAULT_REMOTE_RO_VOLUMES from this PR.

Comment (arcusfelis) [reply tools/travis-common-vars.sh]: REMOTE_RO_VOLUMES removed.
Can be resurrected back once we need docker-machine.

File: tools/wait_for_service.sh (hunk 2 above)
Comment (arcusfelis) [reply tools/wait_for_service.sh]: uploader was removed.
busybox does not have bash :D

File: tools/travis-common-vars.sh (hunk 3 above)
Comment (michalwski) [reply tools/travis-common-vars.sh]: ok, cool. Thanks. I just tried this branch to set up MSSQL on my MacOSX and it worked smoothly! Waiting for travis to pass and it can be merged.

File: tools/travis-setup-db.sh [hunk 4]
```diff
@@ -194,16 +198,20 @@ elif [ "$DB" = 'cassandra' ]; then
 
     opts="$(docker inspect -f '{{range .Config.Entrypoint}}{{println}}{{.}}{{end}}' cassandra:${CASSANDRA_VERSION})"
     opts+="$(docker inspect -f '{{range .Config.Cmd}}{{println}}{{.}}{{end}}' cassandra:${CASSANDRA_VERSION})"
-    readarray -t -s 1 init_opts <<< "$opts"
```
arcusfelis: readarry is not supported in mac

PR #1930, opened 2018-06-15
Build stages

michalwski:
This PR introduces build stages on travis-ci.

## How does it work?

For Erlang/OTP 19.3 (the one used in most jobs on travis) there is:
* a `Build` stage where dependencies are downloaded and the whole code is compiled. Also there is the `prod` release build mainly to produce all the ssl certificates and `small tests` are run
    * `_build`, `tools/ssl` and `big_tests/_build` directories are put in a `bzip2` archive, named `$TRAVIS_BUILD_IS-$TRAVIS_OTP_VERSION` and uploaded to S3 (the same bucket as for test results)
   * the whole stage takes around 11 minutes (depending how loaded travis-ci is)
   * artefact building takes on avg 2min 30s
   * artefact upload takes on avg 35s
* then there is the `Test` stage where there are 8 parallel jobs (the same presets as currently without stages)
    * 5 of them running on Erlang/OTP 19.3 reuse the build artefact and doesn't run small tests (as they were already run). This reduces these jobs time by around 7 to 9 minutes.
    * 3 of them on Erlang/OTP 20.0 and 18.3 are the same as currently (without any artefacts and with small tests run).
* then there is the `Post tests` stage where currently there are 2 jobs running pkg scripts.

## Advantages:
 * reduces total build time by around 30 minutes and can be optimise a bit more

## Disadvantages:
 * Build artefacts are not uploaded and download for PR from other repository (there is no travis secure evns for such builds, so we are not able to upload artefacts). This means that `Build` stage is useless as all other stages will be building MongooseIM as it happens now.
 * More complicated `.travis.yml` file and whole pipeline.
 * First test results are visible around 10 minutes later then it's now (due to `Build` and `Pre test` stages)

## TODO:
- [x] maybe combine `Build` and `Pre test` stage
- [ ] fix coverage report for the stage with only small tests run
- [x] maybe build `big tests` in `Build stage` and add it to the archive so that it doesn't have to be compiled on all the jobs (will safe up 2 minutes for every job on Erlang/OTP 19.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,18 +84,14 @@ jobs:
         - SKIP_REPORT_UPLOAD=1 SKIP_COV=1 SKIP_RELEASE=1 SKIP_BUILD_TESTS=1
       after_success:
         - tools/travis-upload-build-artifacts-to-s3.sh
-    - stage: test
-      script:
-        - ls -la _build
-        - ls -la tools/ssl
 
 stages:
   - build
   - test
 
-#env:
-#        - PRESET=internal_mnesia DB=mnesia REL_CONFIG=with-all TLS_DIST=yes
-#        - PRESET=mysql_redis DB=mysql REL_CONFIG="with-mysql with-redis"
+env:
+        - PRESET=internal_mnesia DB=mnesia REL_CONFIG=with-all TLS_DIST=yes
+        - PRESET=mysql_redis DB=mysql REL_CONFIG="with-mysql with-redis"
 #          #        - PRESET=odbc_mssql_mnesia DB=mssql REL_CONFIG=with-odbc
 #          #        - PRESET=ldap_mnesia DB=mnesia REL_CONFIG=with-none
 #          #        - PRESET=cassandra_mnesia DB=cassandra REL_CONFIG=with-cassandra CASSANDRA_VERSION=3.9
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,35 +73,56 @@ branches:
 
 otp_release:
         - 19.3
-env:
-        - PRESET=internal_mnesia DB=mnesia REL_CONFIG=with-all TLS_DIST=yes
-        - PRESET=mysql_redis DB=mysql REL_CONFIG="with-mysql with-redis"
-        - PRESET=odbc_mssql_mnesia DB=mssql REL_CONFIG=with-odbc
-        - PRESET=ldap_mnesia DB=mnesia REL_CONFIG=with-none
-        - PRESET=elasticsearch_and_cassandra_mnesia DB="elasticsearch cassandra"
-          REL_CONFIG="with-elasticsearch with-cassandra" TESTSPEC=mam.spec
-          ELASTICSEARCH_VERSION=5.6.9 CASSANDRA_VERSION=3.9
-          # In case you want to test with another ODBC driver, uncomment this
-      # - PRESET=odbc_pgsql_mnesia DB=pgsql REL_CONFIG=with-odbc
 
-matrix:
-    include:
-        - otp_release: 20.0
-          env: PRESET=dialyzer_only
-               SKIP_RELEASE=1 SKIP_BUILD_TESTS=1 SKIP_COV=1 SKIP_REPORT_UPLOAD=1
-        - otp_release: 20.0
-          env: PRESET=pgsql_mnesia DB=pgsql REL_CONFIG="with-pgsql with-jingle-sip"
-        - otp_release: 18.3
-          env: PRESET=riak_mnesia DB=riak REL_CONFIG=with-riak
-        - language: generic
-          env: PRESET=pkg pkg_PLATFORM=centos7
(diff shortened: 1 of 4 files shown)


mongoose-im:
[4925.7](https://travis-ci.org/esl/MongooseIM/jobs/392602132) / Erlang 19.3 / elasticsearch_and_cassandra_mnesia / 1c4ad5dc597e8091442bdbd8300d8df4fa9599ed
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/1930/4925/elasticsearch_and_cassandra_mnesia.19.3) / [big](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1930/4925/elasticsearch_and_cassandra_mnesia.19.3/big/ct_run.test%40travis-job-bd1f3f5c-a4a1-4601-b8a8-5420007bd5e7.2018-06-15_08.13.51/index.html) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1930/4925/elasticsearch_and_cassandra_mnesia.19.3/small/)
**OK: 445** / Failed: 0 / **User-skipped: 8** / Auto-skipped: 0

arcusfelis:
`small_tests` do not really need a separate stage, can be run together with other 8 jobs.

arcusfelis:
I would love to compare execution time without stages, but with small tests executed just for one job.
Because I can see time saving here not because of stages, but because of executing small tests once per build.

michalwski:
> small_tests do not really need a separate stage, can be run together with other 8 jobs

They can take up to 5 minutes some times. I think it's better to run them with the build stage instead of the stage with `big tests` only.

michalwski:
> I would love to compare execution time without stages, but with small tests executed just for one job.
> Because I can see time saving here not because of stages, but because of executing small tests once per build.

That's a good idea, I'll create such a PR to compare.

arcusfelis:
> 5 minutes some times
I mean a separate task for small tests, like the dialyzer task.
So, the only difference would be is that we wouldn't wait for small tests to pass to start executing big tests.

mongoose-im:
[4929.2](https://travis-ci.org/esl/MongooseIM/jobs/392691343) / Erlang 19.3 / internal_mnesia / 25fca99412255ebe9065fe88abc737c629a76892
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/1930/4929/internal_mnesia.19.3) / [big](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1930/4929/internal_mnesia.19.3/big/ct_run.test%40travis-job-c19c0542-ed97-4d56-a611-b81aa626a4b6.2018-06-15_13.16.58/index.html) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1930/4929/internal_mnesia.19.3/small/ct_run.mongooseim@localhost.2018-06-15_13.03.48/lib.mongooseim.logs/run.2018-06-15_13.03.48/suite.log.html)
**OK: 1068** / Failed: 0 / **User-skipped: 41** / Auto-skipped: 0

michalwski:
@arcusfelis the PR with only small tests moved to its own job is here: #1932. It's around 2 minutes longer so not much of a difference. Maybe this is better way to go since it's less complicated and works in the same way for both internal and external PRs.

.travis.yml
```diff
@@ -17,19 +17,22 @@ addons:
...
         - test 1 = "$SKIP_COMPILE" || ./rebar3 compile
+        - test 1 = "$SKIP_COMPILE" || make certs
         - test 1 = "$SKIP_RELEASE" || make devrel
         - test 1 = "$SKIP_BUILD_TESTS" || tools/travis-build-tests.sh
         - test 1 = "$SKIP_COV" || travis_retry pip install --user codecov
 before_script:
         - tools/travis-setup-db.sh
         - if [ $PRESET = 'ldap_mnesia' ]; then sudo tools/travis-setup-ldap.sh; fi
-script: KEEP_COVER_RUNNING=1 tools/travis-test.sh -p $PRESET
+script: KEEP_COVER_RUNNING=1 tools/travis-test.sh -p $PRESET -s $RUN_SMALL_TESTS
```
ludwikbukowski: I'm just not sure - how this `RUN_SMALL_TESTS` variable is comparable to `SMALL_TESTS` from `travis-test.sh` here https://github.com/esl/MongooseIM/blob/master/tools/travis-test.sh#L6

michalwski: It's passed as value of `-s` parameter to `travis-test.sh` script. Then in https://github.com/esl/MongooseIM/blob/master/tools/travis-test.sh#L10-L33 it get's parsed and overrides the value of `SMALL_TESTS` form line 6.

ludwikbukowski: Oh, now I get it. 
It would be easier to grep if we had the same var name in `travis-test.sh` :D

michalwski:
I'm closing this one since #1932 gives us almost the same improvement (when it comes to total build time on travis) without the disadvantages of these changes.

Issue #1910, opened 2018-06-06
MongooseIM's client REST API


Beisenbek:
MongooseIM version: 3.0.0
Installed from: source 
Erlang/OTP version: 18

I found that REST API for `One-to-one messages` returns partial info about messages. There are no **extension** part of messages.  Only body.
Request:
 ``` 
curl -X GET --header 'Accept: application/json' 'http://localhost:8089/api/messages?limit=10'
```
And there are no **archive_chat_markers** in result set. But it enabled in config:
```
{mod_mam_meta, [
    {backend, riak},
    {archive_chat_markers,true},
...
```

**Comment by michalwski** [comment]
Hi @Beisenbek,

Thanks for trying the HTTP API! We implemented basic functionality over HTTP and are open to extended it. How would you see the extended result? Do you expect the full xml stanzas in the response or more keys in the JSON object?

**Comment by Beisenbek** [comment]
Hi @michalwski 
It seems to me that it'll better to get full xml stanzas in the response and more keys in the JSON object. In my case I want to put messages through active tcp connection and get MAM messages through REST API. So, there is problem in getting identical result for json based REST API and standard xml based MAM stanza.

**Comment by michalwski** [comment]
I think we could start with the full xml added to the JSON object under `xml` key like below:
```json
[
    {
        "to": "UserA@{host}",
        "timestamp": 1528283702183,
        "id": "DeGu9-20159",
        "from": "UserB@{host}",
        "body": "test message"
        "xml": "<message id='some-id' to='UserA@{host}`><body>test message</body><other/></message>"
    }
]
```

The place in the code which generates the result is here: https://github.com/esl/MongooseIM/blob/33db205313b412e89fad3a3217b7a67808acfbb6/src/mongoose_client_api_messages.erl#L108-L114 would you be able to send us a PR? We are about to start new sprint so there will probably be no time to make the change in next 2 weeks by us.

Beisenbek:
ok, I'll try send PR as quick as possible.

Beisenbek:
Hi @michalwski  

I've faced with problem. I got error `invalid_ejson` when tried to convert xml object to json.

My steps:
1. got binary `X1` from `Msg` object with `exml:to_binary`
2. recursively parsed erlang tuple `X2` from `X1`. So `X2` equals to
```
 {<<"message">>,
[
    [
        {<<"@xml:lang">>,<<"en">>},
        {<<"@to">>,<<"UserA">>},
[...]

PR #1938, opened 2018-06-22
json result object extended with message extension properties

Beisenbek:
json result object extended with message extension properties if there are some properties in raw message.

This PR addresses #1910

Proposed changes include:
Basic idea is to add properties part of raw xml message to the result json. If there are properties then old json will be returned.

### DIFF
--- a/src/mongoose_client_api_messages.erl
+++ b/src/mongoose_client_api_messages.erl
@@ -106,12 +106,33 @@ make_json_msg(Msg, MAMId) ->
 
 -spec encode(exml:item(), integer()) -> map().
 encode(Msg, Timestamp) ->
+    RawProps = exml_query:path(Msg, [{element,<<"properties">>}]),
     BodyTag = exml_query:path(Msg, [{element, <<"body">>}]),
-    #{from => exml_query:attr(Msg, <<"from">>),
-      to => exml_query:attr(Msg, <<"to">>),
-      id => exml_query:attr(Msg, <<"id">>),
-      body => exml_query:cdata(BodyTag),
-      timestamp => Timestamp}.
+
+    CoreList = [{<<"from">>, exml_query:attr(Msg, <<"from">>)},
+                {<<"to">>, exml_query:attr(Msg, <<"to">>)},
+                {<<"id">>,exml_query:attr(Msg, <<"id">>)},
+                {<<"body">>,exml_query:cdata(BodyTag)},
+                {<<"timestamp">>,Timestamp}],
+
+    ExtensionList =
+      case RawProps of
+           #xmlel{name = Name
+              ,attrs = Attrs
+              ,children = Children} ->
+           Props = [convert_prop_child(Child) || Child <- Children],
+           [{<<"properties">>,maps:from_list(Props)}];
+       _ -> []
+      end,
+
+    maps:from_list(CoreList ++ ExtensionList).
+
+convert_prop_child(Child)->
+    Name = exml_query:path(Child,[{element,<<"name">>},cdata]),
+    Value = exml_query:path(Child,[{element,<<"value">>},cdata]),
+    {Name, Value}.
+
+
 
 maybe_jid(undefined) ->
     undefined;


ludwikbukowski:
In my opinion, we shouldn't rely on element's name (the "properties") but rather on the dedicated namespace at least.

Beisenbek:
Hi @ludwikbukowski 

I took this element from [Smack Android](https://github.com/igniterealtime/Smack/blob/master/documentation/extensions/properties.md#xml-format) and there is written that:
1) 

> The current XML format used to send property data is not a standard, so will likely not be recognized by clients not using Smack.

 Is it ok for MongooseIM if I'll add this specific library conditions within the its code?

2) `<properties xmlns="http://www.jivesoftware.com/xmlns/xmpp/properties">` can I use this namespace?

Ok, I'll fill docs regarding feature when it'll be clear how to work with such extra info in message stanza.

Thanks.

src/mongoose_client_api_messages.erl
```diff
@@ -106,12 +106,33 @@ make_json_msg(Msg, MAMId) ->
...
+      case RawProps of
+           #xmlel{name = Name
+              ,attrs = Attrs
+              ,children = Children} ->
+           Props = [convert_prop_child(Child) || Child <- Children],
+           [{<<"properties">>,maps:from_list(Props)}];
+       _ -> []
+      end,
+
+    maps:from_list(CoreList ++ ExtensionList).
```
fenek: It's more efficient to create `ExtensionList` first, then create full list with `[{<<"from">>,... | ExtensionList]`

src/mongoose_client_api_messages.erl
```diff
@@ -106,12 +106,33 @@ make_json_msg(Msg, MAMId) ->
...
+
+    CoreList = [{<<"from">>, exml_query:attr(Msg, <<"from">>)},
+                {<<"to">>, exml_query:attr(Msg, <<"to">>)},
+                {<<"id">>,exml_query:attr(Msg, <<"id">>)},
+                {<<"body">>,exml_query:cdata(BodyTag)},
+                {<<"timestamp">>,Timestamp}],
+
+    ExtensionList =
+      case RawProps of
+           #xmlel{name = Name
```
fenek: `Name` and `Attrs` are not used, right?

src/mongoose_client_api_messages.erl
```diff
@@ -106,12 +106,33 @@ make_json_msg(Msg, MAMId) ->
...
+              ,attrs = Attrs
+              ,children = Children} ->
+           Props = [convert_prop_child(Child) || Child <- Children],
+           [{<<"properties">>,maps:from_list(Props)}];
+       _ -> []
```
fenek: Broken indentation.

src/mongoose_client_api_messages.erl
```diff
@@ -106,12 +106,33 @@ make_json_msg(Msg, MAMId) ->
...
     BodyTag = exml_query:path(Msg, [{element, <<"body">>}]),
-    #{from => exml_query:attr(Msg, <<"from">>),
-      to => exml_query:attr(Msg, <<"to">>),
-      id => exml_query:attr(Msg, <<"id">>),
-      body => exml_query:cdata(BodyTag),
-      timestamp => Timestamp}.
+
+    CoreList = [{<<"from">>, exml_query:attr(Msg, <<"from">>)},
+                {<<"to">>, exml_query:attr(Msg, <<"to">>)},
+                {<<"id">>,exml_query:attr(Msg, <<"id">>)},
```
fenek: Spaces after commas.

Beisenbek:
Hi

@ludwikbukowski @michalwski  @fenek I've updated PR with your review comments:

> This PR needs:
> 1.Docs update.
> 2.Test(s).
> 3.I'd appreciate a small comment in the code that it's Smack specific.
> 4.Please do match namespace as well.

all changes in the new [PR](https://github.com/esl/MongooseIM/pull/1976)

Can you make review it.

PR #1941, opened 2018-06-22
Add logging to XML stream error cases.

mikebirmingham:
Under load it isn't realistic to turn on debug logging, nor is it always possible to get logs from client devices.  This ensures that we trace out when and why an XML stream has been unexpectedly torn down so we have a chance of debugging.

This PR addresses the fact that it is very difficult to attribute reported client problems to network problems/client bugs under load.

Proposed changes include:
* Adding logs to the unexpected stream termination cases
* No UTs or doc changes as this is just extra logs

### DIFF
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -248,21 +248,26 @@ get_subscribed(FsmRef) ->
 wait_for_stream({xmlstreamstart, _Name, _} = StreamStart, StateData) ->
     handle_stream_start(StreamStart, StateData);
 wait_for_stream(timeout, StateData) ->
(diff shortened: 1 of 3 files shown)


fenek:
Hi @mikebirmingham 

Thanks for the contribution! It's certainly a good idea to have additional tracing on these events. However, I don't like a fact that it might add some noise to average production system (using warning or error log level). I'd prefer these to be of debug or info level. In order to allow easy tracing, you may use lager's built-in support: https://github.com/erlang-lager/lager#tracing. You may add relevant attributes such as user JID or main event name (e.g. `invalid_xml`). Then, you'll be able to log these events separately to a new place if necessary.

What is more, please read https://mongooseim.readthedocs.io/en/latest/developers-guide/logging/#logging-format to learn about recommended logging format.

PR #1942, opened 2018-06-22
stabilise connect_SUITE

ludwikbukowski:
The story:
- in `connect_SUITE:starttls_noproc`[0] we setup monitor for c2s process and we catch DOWN's event:
```
{'DOWN', MonRef, process, C2sPid, Reason = {normal,
                                 {gen_server,call,
                                  [<6900.28922.0>,
                                   {starttls,
                                    [verify_none,
                                     {certfile,"priv/ssl/fake_server.pem"},
                                     {tls_module,just_tls},
                                     {protocol_options,"no_sslv3"}]}]}}}}
```
instead of `Reason` being just `normal`.

When you go through some OTP sources [1] [2] you can see that such a `{normal, Reason}` exception will be thrown when gen_server's process exits/crashes in the middle of call.

The major issue is that we suspend [3] the process just before the call [4] so `gen_server:call(Proc, {starttls ... `[5] shouldn't even be triggered - because after suspending process, it won't consume any message except the system's messages. If it went to `starttls` branches means that it consumed some xml-related messages.

Anyway, I believe it might be related to some race conditions between calling `suspend` and starting TLS. 
In this PR I add additional 100 ms sleep after suspending process and want to see if the problem still happens.

BTW I change all `noproc` names to `suspend` cause its not about killing process but suspending

[0] https://github.com/esl/MongooseIM/blob/master/big_tests/tests/connect_SUITE.erl#L345
[1] https://github.com/erlang/otp/blob/9d5af99762b3795c763fb62c1516247bd3f8e12f/lib/stdlib/src/gen_server.erl#L215
[2] https://github.com/erlang/otp/blob/9d5af99762b3795c763fb62c1516247bd3f8e12f/lib/stdlib/src/gen.erl#L177
[3] https://github.com/esl/MongooseIM/blob/master/big_tests/tests/connect_SUITE.erl#L332
[4] https://github.com/esl/MongooseIM/blob/master/big_tests/tests/connect_SUITE.erl#L335
[5] https://github.com/esl/MongooseIM/blob/master/src/ejabberd_receiver.erl#L87

### DIFF
--- a/big_tests/tests/connect_SUITE.erl
+++ b/big_tests/tests/connect_SUITE.erl
@@ -46,13 +46,13 @@ all() ->
     ].
 
 groups() ->
-    G = [ {c2s_suspend, [], [reset_stream_suspend,
-                            starttls_suspend,
-                            compress_suspend,
-                            bad_xml,
-                            invalid_host,
-                            invalid_stream_namespace,
-                            pre_xmpp_1_0_stream]},
+    G = [ {c2s_error_cases, [], [reset_stream_no_receiver,
+                                 starttls_no_receiver,
+                                 compress_no_receiver,
+                                 bad_xml,
+                                 invalid_host,
+                                 invalid_stream_namespace,
+                                 pre_xmpp_1_0_stream]},
           {starttls, [], test_cases()},
           {tls, [parallel], [auth_bind_pipelined_session,
                              auth_bind_pipelined_auth_failure |
@@ -71,7 +71,7 @@ groups() ->
     ct_helper:repeat_all_until_all_ok(G).
 
 all_groups()->
-    [{group, c2s_suspend},
+    [{group, c2s_error_cases},
      {group, starttls},
      {group, feature_order},
      {group, tls}].
@@ -111,7 +111,7 @@ end_per_suite(Config) ->
     restore_ejabberd_node(Config),
     escalus:end_per_suite(Config).
 
-init_per_group(c2s_suspend, Config) ->
+init_per_group(c2s_error_cases, Config) ->
     config_ejabberd_node_tls(Config,
                              fun mk_value_for_starttls_config_pattern/0),
     ejabberd_node_utils:restart_application(mongooseim),
@@ -294,7 +294,7 @@ clients_can_connect_with_advertised_ciphers(Config) ->
                          ciphers_working_with_ssl_clients(Config))).
 
 
-reset_stream_suspend(Config) ->
+reset_stream_no_receiver(Config) ->
     UserSpec = escalus_users:get_userspec(Config, alice),
     Steps = [start_stream, stream_features],
     {ok, Conn, _Features} = escalus_connection:start(UserSpec, Steps),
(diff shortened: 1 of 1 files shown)


mongoose-im:
[4958.2](https://travis-ci.org/esl/MongooseIM/jobs/395466689) / Erlang 19.3 / internal_mnesia / fde7883aae85ed59f9c56726d00afe11b6e9c933
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/1942/4958/internal_mnesia.19.3) / [big](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1942/4958/internal_mnesia.19.3/big/ct_run.test%40travis-job-c44accfe-d724-459e-b92d-643ea3320747.2018-06-22_14.36.20/index.html) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1942/4958/internal_mnesia.19.3/small/)
**OK: 1059** / Failed: 0 / **User-skipped: 41** / Auto-skipped: 0

big_tests/tests/connect_SUITE.erl
```diff
@@ -46,9 +46,9 @@ all() ->
     ].
 
 groups() ->
-    G = [ {c2s_noproc, [], [reset_stream_noproc,
-                            starttls_noproc,
-                            compress_noproc,
+    G = [ {c2s_suspend, [], [reset_stream_suspend,
+                            starttls_suspend,
```
michalwski: The indentation here look broken (at least on github). In most of other places the second element in the list (in this case `starttls_suspend`) starts in the same column as the first element above.

big_tests/tests/connect_SUITE.erl
```diff
@@ -303,6 +303,7 @@ reset_stream_noproc(Config) ->
     [RcvPid] = children_specs_to_pids(rpc(mim(), supervisor, which_children, [ejabberd_receiver_sup])),
     MonRef = erlang:monitor(process, C2sPid),
     ok = rpc(mim(), sys, suspend, [C2sPid]),
+    timer:sleep(100),
```
michalwski: Could we use for example, `erlang:process_info` to see if the process is in `suspend` state and only then proceed? Today `100ms` may be enough, the really thing we are wait for is that the process is in `suspend` state as far as I understand.

michalwski: You may want to reuse `mongoose_helper/wait_until/3` function to retry the check if needed.

ludwikbukowski: sounds like good idea.

mongoose-im:
[4977.2](https://travis-ci.org/esl/MongooseIM/jobs/397303428) / Erlang 19.3 / internal_mnesia / 8dd17522f8fa83e742128a0fa64ee7957b529fb3
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/1942/4977/internal_mnesia.19.3) / [big](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1942/4977/internal_mnesia.19.3/big/ct_run.test%40travis-job-911c1fce-85c4-490a-9fa1-e0b0cd3ca0b2.2018-06-27_12.46.59/index.html) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1942/4977/internal_mnesia.19.3/small/)
**OK: 1060** / Failed: 0 / **User-skipped: 41** / Auto-skipped: 0

Issue #1946, opened 2018-06-24
"Problem starting the module mod_jingle_sip"


streaps:
MongooseIM version: 3.0.0
Installed from: mongooseim_3.0.0-1\~debian\~stretch_amd64.deb
Erlang/OTP version: Erlang/OTP 19

I've installed MongooseIM to test the SIP/Jingle Proxy, but it doesn't load. I have no idea if it is configuration error or a bug.

```
2018-06-24 18:04:25.759 [info] <0.31.0> Application mnesia started on node mongooseim@localhost
2018-06-24 18:04:26.372 [critical] <0.372.0>@gen_mod:start_module_for_host:150 Problem starting the module mod_jingle_sip for host <<"localhost">>
 options: []
 exit: {noproc,{gen_server,call,
                           [nkservice_all_srvs_sup,
                            {terminate_child,bk2kkvw},
                            infinity]}}
[{gen_server,call,3,[{file,"gen_server.erl"},{line,212}]},
 {nkservice_srv_sup,stop_service,1,
                    [{file,"/home/vagrant/2018-05-23_15-33-44/scripts/mongooseim/deb/mongooseim/_build/default/lib/nkservice/src/nkservice_srv_sup.erl"},
                     {line,65}]},
 {nkservice,start,2,
            [{file,"/home/vagrant/2018-05-23_15-33-44/scripts/mongooseim/deb/mongooseim/_build/default/lib/nkservice/src/nkservice.erl"},
             {line,148}]},
 {mod_jingle_sip,start,2,
                 [{file,"/home/vagrant/2018-05-23_15-33-44/scripts/mongooseim/deb/mongooseim/_build/prod/lib/mongooseim/src/jingle_sip/mod_jingle_sip.erl"},
                  {line,58}]},
 {gen_mod,start_module_for_host,3,
          [{file,"/home/vagrant/2018-05-23_15-33-44/scripts/mongooseim/deb/mongooseim/_build/prod/lib/mongooseim/src/gen_mod.erl"},
           {line,123}]},
 {lists,foreach,2,[{file,"lists.erl"},{line,1338}]},
 {ejabberd_app,start,2,
               [{file,"/home/vagrant/2018-05-23_15-33-44/scripts/mongooseim/deb/mongooseim/_build/prod/lib/mongooseim/src/ejabberd_app.erl"},
                {line,72}]},
 {application_master,start_it_old,4,
                     [{file,"application_master.erl"},{line,273}]}]
[...]

PR #1960, opened 2018-07-05
Accept error already_started and make sure nksip app is running

michalwski:
This PR makes sure that `nksip` app is started while starting `mod_jingle_sip`. Also it allows to start the module for more than just one host.

### DIFF
--- a/src/jingle_sip/mod_jingle_sip.erl
+++ b/src/jingle_sip/mod_jingle_sip.erl
@@ -50,15 +50,26 @@
 
 -spec start(ejabberd:server(), list()) -> ok.
 start(Host, Opts) ->
+    start_nksip_service_or_error(Opts),
+    mod_jingle_sip_backend:init(Host, Opts),
+    ejabberd_hooks:add(hooks(Host)),
+    ok.
+
+start_nksip_service_or_error(Opts) ->
+    application:ensure_all_started(nksip),
     ListenPort = gen_mod:get_opt(listen_port, Opts, 5600),
     NkSipBasicOpts = #{sip_listen => "sip:all:" ++ integer_to_list(ListenPort),
                        callback => jingle_sip_callbacks,
                        plugins => [nksip_outbound, nksip_100rel]},
     NkSipOpts = maybe_add_udp_max_size(NkSipBasicOpts, Opts),
-    {ok, _SrvID} = nksip:start(?SERVICE, NkSipOpts),
-    mod_jingle_sip_backend:init(Host, Opts),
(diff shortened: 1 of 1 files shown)


mongoose-im:
[5049.1](https://travis-ci.org/esl/MongooseIM/jobs/400320764) / Erlang 19.3 / small_tests / 12435f8d87fd6e1cbaeba08b3cba01bd6f422956
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/1960/5049/small_tests.19.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1960/5049/small_tests.19.3/small/ct_run.mongooseim@localhost.2018-07-05_09.09.26/lib.mongooseim.logs/run.2018-07-05_09.09.27/suite.log.html)

src/jingle_sip/mod_jingle_sip.erl
```diff
@@ -71,7 +82,6 @@ maybe_add_udp_max_size(NkSipOpts, Opts) ->
 -spec stop(ejabberd:server()) -> ok.
 stop(Host) ->
     ejabberd_hooks:delete(hooks(Host)),
-    nksip:stop(?SERVICE),
```
fenek: Maybe part of Jingle/SIP extension could be extracted into `service_...`? Not stopping `nksip` seems a bit like a temporary solution.

michalwski: 1. I removed the stop here since there may be many xmpp hosts served by MongooseIM. This service is shared between them so stopping it with first host stopped is bad idea, it's better to let it operate on stop manually if needed.
2. I tried `service_` framework, I was not able to use it without modifications. Below is what I miss or makes things more difficult:
     1. currently if I want to have a service started for a module, it needs to be explicitly added in the config file in the `services` section. This complicates modules configuration since the config is split into 2 places - the service and the module. I think it be more useful to allow configuring the service via module's options.
    2. even I start nksip as a service, there is still no mechanism which will stop it when all modules using it are stopped.

Both items from point 2 are not very small changes, that's why I decided not to go with the `service` framework. I think nksip service is perfect fit for the `service` framework when it supports what's missing for me at this moment.

fenek: OK, understood, I agree it's not a big deal to let it running. Maybe some yet another alternative solution will emerge over time.

fenek:
According to user's report in #1946 the startup problem is fixed, so I'm merging it. Any further issues will be handled in separate PR.

PR #1964, opened 2018-07-05
Better packages versions

fenek:
This PR addresses changes the way packages are versioned. With these changes `generate_vsn.sh` script is used instead of `VERSION` file contents. It makes building easier for everyone who builds packages from tags or branches but doesn't want to update `VERSION` every time.

Bonus: `centos6` package is built with OTP 19.3 now.

### DIFF
--- a/tools/generate_vsn.sh
+++ b/tools/generate_vsn.sh
@@ -27,8 +27,8 @@ echo_git_vsn() {
 }
 
 echo_file_vsn() {
-	FILE_VSN=`cat ${DIR}/../VERSION`
-    
+    FILE_VSN=`cat ${DIR}/../VERSION`
+
     if [ "$VSN_TYPE" == "tag" ]; then
         echo $FILE_VSN
     elif [ "$VSN_TYPE" == "commits" ]; then
--- a/tools/generate_vsn.sh
+++ b/tools/generate_vsn.sh
@@ -1,9 +1,48 @@
 #!/bin/bash
+
+# generate_vsn.sh [type]
+# type = describe | tag | commits
+#
+# Certain 'type' values will produce following results:
+# - With "describe" output being "3.0.0-233-g984dd6df7"...
+# - "tag" will return "3.0.0"
+# - "commits" will return "233"
+# "describe" is the default value
+#
+# With git information missing, contents of VERSION file
+# will be returned for "describe" and "tag"; "0" will be returned
+# for "commits".
+
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-GIT_VSN=`git describe --always --tags 2>/dev/null`
+VSN_TYPE=${1:-"describe"}
+
+echo_git_vsn() {
+    if [ "$VSN_TYPE" == "tag" ]; then
+        echo $LAST_TAG
+    elif [ "$VSN_TYPE" == "commits" ]; then
+        echo `git rev-list --count $LAST_TAG..HEAD 2>/dev/null`
+    else
+        echo `git describe --always --tags 2>/dev/null`
+    fi
+}
+
+echo_file_vsn() {
+	FILE_VSN=`cat ${DIR}/../VERSION`
+    
+    if [ "$VSN_TYPE" == "tag" ]; then
+        echo $FILE_VSN
+    elif [ "$VSN_TYPE" == "commits" ]; then
+        echo "0"
+    else
+        echo $FILE_VSN
+    fi
+}
+
+LAST_TAG=`git describe --tags --abbrev=0 2>/dev/null`
 
 if [ $? -eq 0 ]; then
-	echo $GIT_VSN
+    echo_git_vsn
 else
-	echo `cat ${DIR}/../VERSION`
+    echo_file_vsn
 fi
+
--- a/tools/pkg/files/build
+++ b/tools/pkg/files/build
@@ -3,7 +3,7 @@
 set -e
 
 VER=$1
-REVISION=${2:-1}
+REVISION=$2
 
 if [ x"$VER" == x"" ]; then
     echo "$0: which version to build?"
@@ -11,5 +11,5 @@ if [ x"$VER" == x"" ]; then
 fi
 
 cd /buildfiles
-bash ./script.sh ${VER} ${REVISION}
+bash -x ./script.sh ${VER} ${REVISION}
 cp -i mongooseim*.??? /packages/
--- a/tools/pkg/files/build
+++ b/tools/pkg/files/build
@@ -11,5 +11,5 @@ if [ x"$VER" == x"" ]; then
 fi
 
 cd /buildfiles
-bash -x ./script.sh ${VER} ${REVISION}
+bash ./script.sh ${VER} ${REVISION}
 cp -i mongooseim*.??? /packages/
--- a/tools/pkg/platforms/centos6/files/build/mongooseim.spec
+++ b/tools/pkg/platforms/centos6/files/build/mongooseim.spec
@@ -13,7 +13,7 @@ License: %{_license}
 Group: "Development/Languages"
 BuildArch: @ARCH@
 BuildRoot: %{_topdir}/BUILDROOT/%{name}-%{version}.@ARCH@
-BuildRequires:  esl-erlang <= 18.3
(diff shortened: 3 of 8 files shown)


mongoose-im:
[5053.1](https://travis-ci.org/esl/MongooseIM/jobs/400663618) / Erlang 19.3 / small_tests / dc9983610b7c26e08be774efdf071ce42a35a308
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/1964/5053/small_tests.19.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1964/5053/small_tests.19.3/small/ct_run.mongooseim@localhost.2018-07-05_23.39.00/lib.mongooseim.logs/run.2018-07-05_23.39.01/suite.log.html)

mongoose-im:
[5056.1](https://travis-ci.org/esl/MongooseIM/jobs/400693451) / Erlang 19.3 / small_tests / ebac542350798dc72980757fb486e32354f8c161
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/1964/5056/small_tests.19.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/1964/5056/small_tests.19.3/small/ct_run.mongooseim@localhost.2018-07-06_01.50.26/lib.mongooseim.logs/run.2018-07-06_01.50.27/suite.log.html)

tools/generate_vsn.sh
```diff
@@ -1,9 +1,48 @@
 #!/bin/bash
+
+# generate_vsn.sh [type]
```
michalwski: 👍 for the doc!

tools/generate_vsn.sh
```diff
@@ -1,9 +1,48 @@
 #!/bin/bash
+
```
michalwski: I already suggested that some time ago and I'm going to repeat myself :) I think using at least

```bash
set -euo
```

will help to write better and easier to debug bash scripts. More details on the above and other flags can be found here: http://redsymbol.net/articles/unofficial-bash-strict-mode/

tools/generate_vsn.sh
```diff
@@ -1,9 +1,48 @@
...
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-GIT_VSN=`git describe --always --tags 2>/dev/null`
+VSN_TYPE=${1:-"describe"}
+
+echo_git_vsn() {
+    if [ "$VSN_TYPE" == "tag" ]; then
+        echo $LAST_TAG
+    elif [ "$VSN_TYPE" == "commits" ]; then
+        echo `git rev-list --count $LAST_TAG..HEAD 2>/dev/null`
```
michalwski: At first I thought "wait a sec, the $LAST_TAG var is undefined here". After some time I realised that we are inside a function and we use a global var. I know this is correct and we can live with this. On the other hand we could benefit from functional programming even in bash scripts.
This is just a comment, not a request for changes.

tools/pkg/platforms/centos6/files/build/mongooseim.spec
```diff
@@ -13,7 +13,7 @@ License: %{_license}
 Group: "Development/Languages"
 BuildArch: @ARCH@
 BuildRoot: %{_topdir}/BUILDROOT/%{name}-%{version}.@ARCH@
-BuildRequires:  esl-erlang <= 18.3
+BuildRequires:  esl-erlang >= 19.3
```
michalwski: 👍 I forgot about this when switch from supported version 18..20 to 19..21

tools/generate_vsn.sh
```diff
@@ -1,9 +1,48 @@
 #!/bin/bash
+
+# generate_vsn.sh [type]
+# type = describe | tag | commits
+#
+# Certain 'type' values will produce following results:
+# - With "describe" output being "3.0.0-233-g984dd6df7"...
+# - "tag" will return "3.0.0"
+# - "commits" will return "233"
```
michalwski: I think it'd be worth mentioning this will count commits since last tag.

tools/generate_vsn.sh
fenek: Thanks. I haven't updated any shell scripts in repo for quite a while so I've simply forgotten about this. :)

tools/generate_vsn.sh
fenek: Right. My basic skill in bash scripting instinctively told me to go for MVP. :D I'll improve it.

fenek:
1. Debian script always expect version and release number to be provided. Actually I suspect that RPM ones never have occasion to use own default values but I'm not comfortable enough with packaging stuff to be 100% certain.
2. You may be right. I think we can even make it a task for an intern from it, as this PR is a direct solution to a problem reported by a customer.

tools/pkg/files/build
```diff
@@ -3,7 +3,7 @@
 set -e
 
 VER=$1
-REVISION=${2:-1}
+REVISION=$2
```
erszcz: Do not use undefined variables. If we do not provide a default for `REVISION`, we should later check that it is defined and not empty or error with a meaningful message otherwise - the same way `VER` is handled.
It's way more convenient to just read a message, than to try to figure out why some script deeper down behaves erratically because of mangled input.

tools/pkg/platforms/centos7/files/build/script.sh
```diff
@@ -15,8 +14,9 @@ git clone ${MONGOOSEIM_REPO} mongooseim
 cd mongooseim
 git checkout ${COMMIT}
 
-VER=$(cat VERSION)
-
+VER=$(./tools/generate_vsn.sh tag)
+COMMITS_COUNT=$(./tools/generate_vsn.sh commits)
+REV=${2:-$((COMMITS_COUNT+1))}
```
erszcz: I'm not sure if we're specific enough here. If the intention is to simplify packaging side branches, then don't we lose information if the final package version is only `mongooseim-$VER-$REV`? In case of `3.0.0-456` there's probably a slim chance that we'll have two branches 456 commits ahead of 3.0.0, but in case of `3.0.0-3` there's a whole lot of possible ambiguity.

[Fedora Versioning Examples](https://fedoraproject.org/wiki/Package_Versioning_Examples) has examples on how to include a version control tag in a package version. Maybe we should take:

```
$ ./tools/generate_vsn.sh describe
3.0.0-236-g7b9afc984
```

and put the third part (`g7b9afc984`) in the package version? My comment also touches line 40 below, which is not directly affected by this PR.

erszcz:
The assumption of package versioning is that given a package we can tell with certainty which code commit it was created from.

With the `VERSION` file, the idea was that the file is checked into git and kept up to date, but then also packaged into an .rpm or .deb. By installing/unpacking a package we get the file, can read its contents and tell which git commit to look at. The commit-to-package-version relationship is maintained.

If we allow the `VERSION` file to be out-of-date and generate the version programatically, we have to give the package user some other way to tell what commit the package was built from. That's my reasoning behind https://github.com/esl/MongooseIM/pull/1964#discussion_r200606211. We should not lose the information on which exact commit the package was built from.

On a different note, in case of Debian package versioning `REV` corresponds to [`debian_revision`](https://www.debian.org/doc/debian-policy/#version), which is quite well defined. While I was looking for a similar definition for RPM / CentOS / Fedora, there doesn't seem to be one. I think that in case of Debian we should not use the `generate_vsn commits` value for `debian_revision` - instead, it should only be bumped up in case of changing package metadata or build scripts, while the "commit offset from tag" should go into the `upstream_version`.

Sorry if my description above is not detailed enough, but I'm a bit scarce on time at the moment.

fenek:
@erszcz 

The aim of this PR is to make it easier for MongooseIM forks to build packages with meaningful versions. Ignoring `VERSION` file content (if git version is available) is justified by a fact that `VERSION` is maintained by us and states the "base" MongooseIM version, while forks may introduce various improvements on top of base code and use tags such as `3.0.0-company-10` to name subsequent releases. For test environments using packages, it's also useful to have revision bumped automatically out of the box.

I agree it's a good idea to always add commit hash.

I'll take a closer look at Debian versioning in this PR.

fenek:
No longer necessary.

Issue #1963, opened 2018-07-05
conflict error on xmpp connection 


sandy5487:
MongooseIM version: 3.0
Installed from: source
Erlang/OTP version: same with the version

when we are trying to connect to MongooseIM server from our in-house Android app, its printing these error logs at server:

`2018-07-05 15:45:10.864 [error] <0.7348.1>@ejabberd_c2s:process_incoming_stanza_with_conflict_check:1322 event=conflict_check_failed jid=6-6001_36@im.myapp.com/A-6001-356029087277827 c2s_sid={{1530,805510,851980},<0.7348.1>} origin_sid={{1530,805510,19748},<0.7342.1>} acc=#{attrs => [{<<"id">>,<<"0LsMa-29">>},{<<"type">>,<<"result">>}],element => {xmlel,<<"iq">>,[{<<"id">>,<<"0LsMa-29">>},{<<"type">>,<<"result">>}],[{xmlel,<<"time">>,[{<<"xmlns">>,<<"urn:xmpp:time">>}],[{xmlel,<<"tzo">>,[],[{xmlcdata,<<"+00:00">>}]},{xmlel,<<"utc">>,[],[{xmlcdata,<<"2018-07-05T15:45:10Z">>}]}]}]},from => <<"6-6001_36@im.myapp.com/A-6001-356029087277827">>,from_jid => {jid,<<"6-6001_36">>,<<"im.myapp.com">>,<<"A-6001-356029087277827">>,<<"6-6001_36">>,<<"im.myapp.com">>,<<"A-6001-356029087277827">>},mongoose_acc => true,name => <<"iq">>,persistent_properties => [{origin_jid,{jid,<<"6-6001_36">>,<<"im.myapp.com">>,<<"A-6001-356029087277827">>,<<"6-6001_36">>,<<"im.myapp.com">>,<<"A-6001-356029087277827">>}},{origin_sid,{{1530,805510,19748},<0.7342.1>}}],ref => #Ref<0.3958627556.78118913.200497>,server => <<"im.myapp.com">>,timestamp => {1530,805510,852364},to => <<"im.myapp.com">>,to_jid => {jid,<<>>,<<"im.myapp.com">>,<<>>,<<>>,<<"im.myapp.com">>,<<>>},type => <<"result">>,user => <<"6-6001_36">>}`

P.S: the same android app is working fine with the ejabberd XMPP server with the similar configurations.

Please guide me what we are missing or where we are going wrong.

Thanks in advance

[...]

PR #1981, opened 2018-07-16
Adjust "conflict_check_failed" log severity

GalaxyGorilla:
See #1963. There are error logs occuring when a client attempts to
reconnect in a very short timespan and at the same time a stanza is
routed to the old invalid connection.

As this is just an application/client oddity the log severity should be
lowered to "warning". This is simply nothing to alert an operations guy
in the middle of the night.

There are actually a lot more logs like this but this one currently
stands out ;).

### DIFF
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -1319,10 +1319,10 @@ process_incoming_stanza_with_conflict_check(Name, From, To, Acc, StateName, Stat
         conflict -> %% A race condition detected
             %% Same jid, but different sids
             OriginSID = mongoose_acc:get_prop(origin_sid, Acc),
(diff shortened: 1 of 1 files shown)


GalaxyGorilla:
Hey @michalwski, I looked through all error logs and I think my main concern is about logs like those:

```
src/ejabberd_router.erl:525: ?ERROR_MSG("error when routing from=~ts to=~ts in 
src/gen_iq_handler.erl:144: ?ERROR_MSG("~p", [Reason]);
```

They work like a general fallback and whenever $something happens you get an error log. I would rather go here for a warning and use appropiate case handling and log levels within the mods. IIRC I usually got alerted here when some client sent some really weird payload that my code just did not expect and e.g. some parsing didn't work out.

PR #1983, opened 2018-07-17
Upgrade worker_pool to 3.1.1

michalwski:
This PR upgrades worker_pool dep to 3.1.1. In fact it uses worker_pool from branch `rel-3.1` (with small fix to function specs) since master branch works only with Erlang/OTP 21.


kanes115:
@michalwski why is [wpool:stop/1](https://github.com/esl/MongooseIM/blob/9968905d853a40b9a5253a94e9d1a5283df34c15/src/event_pusher/mod_event_pusher_sns.erl#L66) used here? Isn't it [private](https://github.com/inaka/worker_pool/blob/master/src/wpool.erl#L121) callback?

michalwski:
It is private yet still exported, a comment in the code doesn't make it private... Where exactly did you find it? My changes didn't introduce `wpool:stop/1`.

kanes115:
Of course it doesn't but at the same time this function does nothing. It is not introduced after the ref we're upgrading from, therefore nevermind.

PR #1995, opened 2018-07-23
deprecated erlang:get_stacktrace/0 function

getong:


michalwski:
I'd really love to see a short description to the PR so that all the changes are clearly summarised.

getong:
@michalwski New msg updated.

getong:
@michalwski 
As erlang/otp 21 is released, a few functions are deprecated and need to write some code.
I notice the MongooseIM use the [worker_pool](https://github.com/inaka/worker_pool), I also made a pr to it. But the worker_pool only supports erlang/otp 21, you can see the inaka/worker_pool#138 , that is made by me.
If you want to support erlang version from 17 to 21, you can fork the worker_pool repo, and I can make some code to it.

michalwski:
There is no need to fork worker_pool. We use `red-3.1` from the repo which compile and runs on all Erlang/OTP versions we support. Currently these are 19, 20 and 21.

getong:
@michalslaski Don' merge it, something needs to fix.

include/mongoose.hrl
```diff
@@ -35,4 +35,10 @@
 %% Logging mechanism
 -include("mongoose_logger.hrl").
 
-
+-ifdef(OTP_RELEASE). %% this implies 21 or higher
```
fenek: Where is it defined? Besides, it's not a good name, as it doesn't imply OTP 21 or higher.

src/auth/ejabberd_auth_odbc.erl
```diff
@@ -128,11 +128,10 @@ check_password(LUser, LServer, Password, Digest, DigestGen) ->
                        "reason=~p user=~ts", [Error, LUser]),
             false %% Typical error is that table doesn't exist
     catch
-        Class:Reason ->
-            Stacktrace = erlang:get_stacktrace(),
+        ?EXCEPTION(Class, Reason, CURRENT_STACKTRACE) ->
```
fenek: Can we make it just `Stacktrace`? After preprocessing it's an ordinary variable after all.

getong:
@fenek 
http://www.erlang.org/news/123
```
A new predifined macro `OTP_RELEASE`
```

getong:
@fenek  @michalwski 
I rewrite the code, feedback is needed.

include/mongoose.hrl
erszcz: From [_Erlang OTP 21.0 is released_](http://www.erlang.org/news/123) (original spelling):

> A new predifined macro `OTP_RELEASE` and preprocessor directives `-if` and `-elif`

If it was added in OTP 21 we can be pretty sure it implies OTP 21 or higher.

kzemek: It also [has a value](http://erlang.org/doc/reference_manual/macros.html) so we could make it more explicit

fenek:
Hi @getong 

Together with the team we've decided that a lack of warning is not a sufficient benefit at a cost of worse looking source code. If these deprecation warnings cause you any problems, maybe using `{nowarn_deprecated_function, {erlang, get_stacktrace, 0}}` compiler option in `mongoose.hrl` is a less invasive choice for now? The old method to get stack trace works in OTP 21 anyway.

getong:
@fenek 
I don't think so. 
[EEP 47: Add syntax in try/catch to retrieve the stacktrace directly](https://www.erlang.org/erlang-enhancement-proposals/eep-0047.html)
```
The main motivation for this feature is to be able to deprecate (and ultimately remove) erlang:get_stacktrace/0.

The problem with erlang:get_stacktrace/0 is that it forces the stacktrace from the latest exception in a process to be retained until another exception occurs or the process terminates. 
The stacktrace often includes the arguments for the last function call, BIF call, or (in OTP 21) operator that failed. 
The arguments can be of any size.
```
And the macro is copied from [meck](https://github.com/eproxus/meck)
You can see it at [the code](https://github.com/eproxus/meck/blob/master/src/meck_code_gen.erl#L186-L192)

getong:
The build failed reason is the spec wrong, I fix it and made a new pr, see #2010

fenek:
@getong 

Thank you, I wasn't aware of this difference. Then it is indeed an improvement.

fenek:
Thank you for your effort but we've decided to do a proper migration when MongooseIM stops supporting OTP versions older than 21. The optimisation described in EEP won't be possible until `get_stacktrace` is removed completely.

PR #1996, opened 2018-07-23
use gen fsm compat

getong:


michalwski:
Could you add a description to the PR so that we know what are the intentions behind the change?

getong:
@michalwski New msg updated.

fenek:
I'd prefer to replace `gen_fsm` with `gen_statem` instead of using a compatibility library. At least in the modules where `p1_fsm` is not used.

getong:
The `gen_statem` is much better than `gen_fsm`, but the `gen_statem` callback is different with `gen_fsm` callback. If using the `gen_statem` module, a lot of code need to rewrite. A lot of work.
The `gen_fsm_compat` is just the same with `gen_fsm`, only the module name is different.

fenek:
There are many tests failing. Can you please fix them? :)

getong:
@fenek 
I don’t know why the codecov/patch failed, but the other three are all passed. Any tips?

getong:
@fenek 
I don’t know why the codecov/patch failed, but the other three are all passed. Any tips?

fenek:
Codecov failed only because your changes affected parts of the code that are not covered with tests. Please note that "diff" check failed, not "project".

PR #2003, opened 2018-07-25
Elixir support

arcusfelis:
This PR addresses "experimental" stuff.

- [x] Make elixir optional
- [x] Hello world module in `elixir_libs/mongooseim_elixir/lib/mongooseim_elixir.ex`
- [x]  Elixir test example `elixir_libs/mongooseim_elixir/test/mongooseim_elixir_test.exs`
- [x] remote iex shell can be started using `_build/mim1/rel/mongooseim/bin/mongooseim iex_debug`

# Where should we store our elixir files?

`elixir_libs` is hardcoded in https://github.com/barrel-db/rebar3_elixir_compile/blob/master/src/rebar3_elixir_compile.erl

Alternative is to move `elixir_libs/mongooseim_elixir/` into just `elixir` or `ex` and make a patch for `rebar3_elixir_compile`. But I am fine with the current `elixir_libs/mongooseim_elixir/` during prototype stage of the feature.

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -56,12 +56,12 @@ end,
 
 RequiredApps = fun() -> [mongooseim, inets, tools] end,
 EnvApps = GetEnvApps(),
+EnvAppsToInclude = [ list_to_atom(App) || App <- string:tokens(EnvApps, " \n\r") ],
 
 SetupIncludedApps =
 fun(Config, EnvApps) ->
         RelxCfg = proplists:get_value(relx, Config),
         {release, Desc, _Apps} = lists:keyfind(release, 1, RelxCfg),
-        EnvAppsToInclude = [ list_to_atom(App) || App <- string:tokens(EnvApps, " \n\r") ],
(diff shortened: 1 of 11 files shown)


rebar.config.script
```diff
@@ -82,6 +82,39 @@ MaybeFIPSSupport = fun(Config) ->
...
+
+DisablePlugin = fun(Config, DisabledPluginName) ->
+    Plugins = proplists:get_value(plugins, Config),
+    Plugins2 = [Plugin || Plugin <- Plugins,
+                GetPluginName(Plugin) =/= DisabledPluginName],
+    lists:keyreplace(plugins, 1, Config, {plugins, Plugins2})
+end,
+
+%% If tools/configure has added elixir application, we keep config as-is
+%% Otherwise, we can remove rebar3_elixir_compile and it's hooks
```
erszcz: Typo: it's -> its

imVinayPandya:
any update on this? we will be having Elixir support in future?

fenek:
Hi @imVinayPandya 

Well, it's not our focus right now so although we don't say a definite "no", it is unlikely in near future.

arcusfelis:
too old.

PR #2005, opened 2018-07-25
Set user 'unavailable' after ping timeout

GalaxyGorilla:
If a user is timed out due to mod_ping then there is no presence
information broadcasted, the session is simply killed or hibernated
(when stream management is used). This patch lets mod_ping send an
'unavailable' presence instead of the user such that other peers
are informed about the timed out session.

### DIFF
--- a/src/mod_ping.erl
+++ b/src/mod_ping.erl
@@ -169,6 +169,11 @@ handle_cast({iq_pong, JID, timeout}, State) ->
             #jid{user = User, server = Server, resource = Resource} = JID,
             case ejabberd_sm:get_session_pid(User, Server, Resource) of
                 Pid when is_pid(Pid) ->
+                    Presence = #xmlel{name = <<"presence">>,
+                                      attrs = [{<<"type">>, <<"unavailable">>},
+                                               {<<"from">>, jid:to_binary(JID)}],
+                                      children = []},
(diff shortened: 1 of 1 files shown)


fenek:
Hi @GalaxyGorilla 

I'm not sure if this is the correct way of fixing this problem. In case of Stream Management, it is an expected behaviour to postpone unavailable presence broadcast until the session timeout. It is indirectly stated in XEP:

"It can happen that an XML stream is terminated unexpectedly (e.g., because of network outages). In this case, it is desirable to quickly resume the former stream rather than complete the tedious process of stream establishment, roster retrieval, **and presence broadcast.**"

When Stream Management is not enabled, user session should broadcast such presence upon termination: https://github.com/esl/MongooseIM/blob/master/src/ejabberd_c2s.erl#L1686-L1699

If it's not happening in your case, then are you certain that client has opened the session and has sent an initial available presence?

GalaxyGorilla:
@fenek since we use stream management those presences are not sent. I get your points here, I'd also stick to the XEPs for MIM and actually I was not aware of the line you quoted above. Personally, I don't understand the purpose of not using a repeated presence broadcast, that kinda makes presence information really unreliable in the mobile world. Still, thanks for review!

PR #2036, opened 2018-08-22
More error resilient test runner

arcusfelis:
This PR addresses some errors, because test runner assumes that someone run it before. And it expects beam files of SUITEs to be available for spec generation and autocompletion.

Be aware, that with this patch travis test would stop, if `redis-server` is running without a docker container. Use `--db` in this case.

Later we can add code, that would test database options sanity before running any tests (gen_tcp connect? actually probing connection with some library? idk).

### DIFF
--- a/big_tests/rebar.config
+++ b/big_tests/rebar.config
@@ -3,6 +3,8 @@
 
 {require_otp_vsn, "19|20|21"}.
 
+{src_dirs, ["src", "tests"]}.
+
 %% We use https:// instead of git://
 {deps, [{usec, ".*", {git, "https://github.com/esl/usec.git", {branch, "master"}}},
         {bbmustache, ".*", {git, "https://github.com/soranoba/bbmustache.git", "v1.4.0"}},
--- a/tools/test-runner.sh
+++ b/tools/test-runner.sh
@@ -9,6 +9,8 @@
 # - VERBOSE
 # - STOP_NODES
 
+# Fail on errors
+set -e
 
 USAGE=$(cat <<-END
 This script runs small and big tests for MongooseIM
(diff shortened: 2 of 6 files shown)


mongoose-im:
[5810.1](https://travis-ci.org/esl/MongooseIM/jobs/450839934) / Erlang 19.3 / small_tests / 7172d257551232f70443b6ff4c599be53900c71f
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2036/5810/small_tests.19.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2036/5810/small_tests.19.3/small/ct_run.mongooseim@localhost.2018-11-05_12.47.54/lib.mongooseim.logs/run.2018-11-05_12.47.55/suite.log.html)

michalwski:
@arcusfelis could you explain this paragraph:

> Be aware, that with this patch travis test would stop, if redis-server is running without a docker container. Use --db in this case.

I see that both big tests with preset mysql_redis and small tests' ejabberd_sm_SUITE with redis backend passed even though redis on travis is not run in docker.

arcusfelis:
@michalslaski More like:
Be aware, that with this patch travis-test.sh script would exit with error on database setup step because of port conflict, if it can't bind port (i.e. it's already in use, most likely by regular redis-server daemon). Basically, it tries to start redis in a docker container => failed to bind the port => complain about it (before it was just ignore the error).

Basically, the problem with the current behavior is it kinda confusing not to see redis in `docker ps`, if it's already running without a container.

if you provide --db option without redis in it, we would not try to start it from the test runner script. I.e. something like `--db mysql`. In such case we would just use the one you already have.


Alternative behavior WOULD be (need to be coded, if we want it) to skip starting some database, if the port is already used by something.

Issue #2041, opened 2018-08-23
Use eredis with Unix domain sockets support


igors:
MongooseIM version: master
Installed from: source
Erlang/OTP version: 20

I've submitted a [pull request](https://github.com/wooga/eredis/pull/108) adding Unix domain sockets support to eredis some half a year ago but haven't gotten any response so far - it doesn't sounds like eredis is being all that actively maintained. We've gotten some measurable (10-15%) performance improvement with unix domain sockets vs TCP sockets in some basic testing so it'd be nice to get this into mongooseIM, imho.

I'm going to make a pull request pointing to my own github eredis fork for the lack of any better options - not sure if it's acceptable; it's clearly not ideal...

fenek:
Merged. Thanks for the contribution! :)

PR #2042, opened 2018-08-23
Use eredis with Unix domain sockets support

igors:
This PR addresses #2041

Proposed changes include:
* Point to my fork of eredis with unix domain sockets support
* update documentation with usage example


kanes115:
Unix domain sockets change [link](https://github.com/igors/eredis/commit/e9688a1ec98e10acca310f19266bd95f976bc262)

doc/modules/mod_global_distrib.md
```diff
@@ -133,7 +133,7 @@ Global distribution modules expose several per-datacenter metrics that can be us
 
 #### Redis session storage options
 
-* **server** (string, default: `"127.0.0.1"`): Address of the Redis listener.
+* **server** (string, default: `"127.0.0.1"`): Address of the Redis listener. For Unix domain sockets use `{local, "/path/to/listener.sock"}` (available starting with OTP 19).
```
fenek: `(available starting with OTP 19)` is not necessary, as MongooseIM doesn't support OTP 18 or older. :)

igors: Fixed - thanks for noticing.

fenek:
@michalwski 

I think we should verify our dependencies for activity, as there might be more than just `eredis`, which could be worth forking.

PR #2051, opened 2018-08-27
Fix checks for matching configuration before reload_cluster.

kzemek:
Config files on nodes have to be matching in order to successfully reload cluster configuration. The check was broken, so the reload could have proceeded (loading incompatible configurations on nodes) and then fail during a check if the loaded configs are compatible.

### DIFF
--- a/src/config/mongoose_config_parser.erl
+++ b/src/config/mongoose_config_parser.erl
@@ -308,8 +308,7 @@ process_term(Term, State) ->
         {routing_modules, Mods} ->
             add_option(routing_modules, Mods, State);
         {loglevel, Loglevel} ->
-            ejabberd_loglevel:set(Loglevel),
-            State;
+            add_option(loglevel, Loglevel, State);
         {max_fsm_queue, N} ->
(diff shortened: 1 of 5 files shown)


mongoose-im:
[5450.1](https://travis-ci.org/esl/MongooseIM/jobs/421083858) / Erlang 19.3 / small_tests / 333254092db203cae50d408f6bb7280e63419fb1
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2051/5450/small_tests.19.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2051/5450/small_tests.19.3/small/ct_run.mongooseim@localhost.2018-08-27_13.00.25/lib.mongooseim.logs/run.2018-08-27_13.00.26/suite.log.html)

kzemek:
@michalwski Config files on nodes have to be matching in order to successfully reload cluster configuration. The check was broken, so the reload could have proceeded (loading incompatible configurations on nodes) and then fail during a check if the loaded configs are compatible.

PR #1308, opened 2017-05-04
Set Sec-Websocket-Protocol header according RFC

arcusfelis:
This PR addresses #1221 

Proposed changes include:
* This PR fixes compatibility issues with some Websocket libraries (chrome, some dot net stuff).
* Do not set `Sec-Websocket-Protocol` header in response if it's missing in request.
* Set `Sec-Websocket-Protocol` header with xmpp value, if it is specified in request.
* Do not set `Sec-Websocket-Protocol` header in respond if request does not specify xmpp.

Some libraries does not set Sec-Websocket-Protocol, but check that Sec-Websocket-Protocol exists in the result. It confuses the libraries and they disconnect from the server with "cryptic" error description.

Spec:

```
   The |Sec-WebSocket-Protocol| header field is used in the WebSocket
   opening handshake.  It is sent from the client to the server and back
   from the server to the client to confirm the subprotocol of the
   connection.  This enables scripts to both select a subprotocol and be
   sure that the server agreed to serve that subprotocol.

   The |Sec-WebSocket-Protocol| header field MAY appear multiple times
   in an HTTP request (which is logically the same as a single
   |Sec-WebSocket-Protocol| header field that contains all values).
   However, the |Sec-WebSocket-Protocol| header field MUST NOT appear
   more than once in an HTTP response.
```

More interesting part of the spec:

```
       /subprotocol/
          Either a single value representing the subprotocol the server
          is ready to use or null.  The value chosen MUST be derived
          from the client's handshake, specifically by selecting one of
          the values from the |Sec-WebSocket-Protocol| field that the
          server is willing to use for this connection (if any).  If the
          client's handshake did not contain such a header field or if
          the server does not agree to any of the client's requested
          subprotocols, the only acceptable value is null.  The absence
          of such a field is equivalent to the null value (meaning that
          if the server does not wish to agree to one of the suggested
          subprotocols, it MUST NOT send back a |Sec-WebSocket-Protocol|
          header field in its response).  The empty string is not the
          same as the null value for these purposes and is not a legal
          value for this field.  The ABNF for the value of this header
          field is (token), where the definitions of constructs and
          rules are as given in [RFC2616].

```

from here https://tools.ietf.org/html/rfc6455

PR #2055, opened 2018-08-29
Only respond with xmpp protocol when client requests it

tomaszwojcikowski:
Proposed changes include:
* Websocket handshake always include `Sec-WebSocket-Protocol` header which breaks old clients.
* We only want to send xmpp protocol when client explicitly requests it.

Relevant RFC https://tools.ietf.org/html/rfc6455#section-11.3.4

### DIFF
--- a/src/mod_websockets.erl
+++ b/src/mod_websockets.erl
@@ -58,7 +58,13 @@
 
 init(Req, Opts) ->
     Peer = cowboy_req:peer(Req),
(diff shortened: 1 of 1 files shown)


michalwski:
Looks like the work was also done in #1308 long time ago. @arcusfelis could you find some time to clean it?

michalwski:
The work is being continued (with tests) in #1308 byt @arcusfelis. I'm closing this one.

PR #2090, opened 2018-09-29
elvis

justinba1010:
This will add the plugin Elvis to rebar3 for Erlang linting.
It uses the repository [rebar3_lint](https://github.com/project-fifo/rebar3_lint)


fenek:
Thanks! This is useful indeed. Can you please add a new Makefile target so running this plugin will be a  bit more convenient? :)

PR #2105, opened 2018-10-12
Fix optargs issue

kanes115:
When you tried to define a command for action different than read/GET MIM would not take optional arguments into consideration.

It was because we checked actual number of arguments against `command arity` instead of `function arity` (which is not the same!!!).

This PR fixes this problem.

### DIFF
--- a/src/mongoose_api_admin.erl
+++ b/src/mongoose_api_admin.erl
@@ -171,7 +171,7 @@ from_json(Req, #http_api_state{command_category = Category,
             Cmds = mongoose_commands:list(admin, Category, method_to_action(Method), SubCategory),
             QVals = cowboy_req:parse_qs(Req),
             Arity = length(B) + length(Params) + length(QVals),
-            case [C || C <- Cmds, mongoose_commands:arity(C) == Arity] of
+            case [C || C <- Cmds, mongoose_commands:func_arity(C) == Arity] of
                 [Command] ->
                     process_request(Method, Command, {Params, Req}, State);
                 [] ->


varnerac:
I don't think this solves all cases. It has to handle the situation where some/all of the `optargs` are present and when they're not. For the `func_arity ` test to be valid `parse_request_body` would have to insert the default values for missing `optargs` into the proplist, and it doesn't.

I think you need some sort of argument that fills in the missing values, similar to [here](https://github.com/esl/MongooseIM/blob/4ece84ab1f5b89308729cb25018cb430e1efa75a/src/mongoose_api_common.erl#L197) before checking the arity.

varnerac:
I don't think this solves all cases. It has to handle the situation where some/all of the `optargs` are present and when they're not. For the `func_arity ` test to be valid `parse_request_body` would have to insert the default values for missing `optargs` into the proplist, and it doesn't.

I think you need some sort of argument that fills in the missing values, similar to [here](https://github.com/esl/MongooseIM/blob/4ece84ab1f5b89308729cb25018cb430e1efa75a/src/mongoose_api_common.erl#L197) before checking the arity.

varnerac:
I don't think this solves all cases. It has to handle the situation where some/all of the `optargs` are present and when they're not. For the `func_arity ` test to be valid `parse_request_body` would have to insert the default values for missing `optargs` into the proplist, and it doesn't.

I think you need some sort of argument that fills in the missing values, similar to [here](https://github.com/esl/MongooseIM/blob/4ece84ab1f5b89308729cb25018cb430e1efa75a/src/mongoose_api_common.erl#L197) before checking the arity.

varnerac:
I don't think this solves all cases. It has to handle the situation where some/all of the `optargs` are present and when they're not. For the `func_arity ` test to be valid `parse_request_body` would have to insert the default values for missing `optargs` into the proplist, and it doesn't.

I think you need some sort of argument that fills in the missing values, similar to [here](https://github.com/esl/MongooseIM/blob/4ece84ab1f5b89308729cb25018cb430e1efa75a/src/mongoose_api_common.erl#L197) before checking the arity.

varnerac:
I don't think this solves all cases. It has to handle the situation where some/all of the `optargs` are present and when they're not. For the `func_arity ` test to be valid `parse_request_body` would have to insert the default values for missing `optargs` into the proplist, and it doesn't.

I think you need some sort of argument that fills in the missing values, similar to [here](https://github.com/esl/MongooseIM/blob/4ece84ab1f5b89308729cb25018cb430e1efa75a/src/mongoose_api_common.erl#L197) before checking the arity.

varnerac:
test

kanes115:
@varnerac, of course you're right :) We also need tests in this PR that would test this untested till now code branch.

kanes115:
@varnerac, of course you're right :) We also need tests in this PR that would test this untested till now code branch.

kanes115:
@varnerac, of course you're right :) We also need tests in this PR that would test this untested till now code branch.

kanes115:
@varnerac, of course you're right :) We also need tests in this PR that would test this untested till now code branch.

kanes115:
@varnerac, of course you're right :) We also need tests in this PR that would test this untested till now code branch.

kanes115:
@varnerac, of course you're right :) We also need tests in this PR that would test this untested till now code branch.

kanes115:
@varnerac, of course you're right :) We also need tests in this PR that would test this untested till now code branch.

kanes115:
@varnerac, of course you're right :) We also need tests in this PR that would test this untested till now code branch. Thanks for the comment

kanes115:
@varnerac, of course you're right :) We also need tests in this PR that would test this untested till now code branch. Thanks for the comment

kanes115:
@varnerac, of course you're right :) We also need tests in this PR that would test this untested till now code branch. Thanks for the comment

kanes115:
@varnerac, of course you're right :) We also need tests in this PR that would test this untested till now code branch. Thanks for the comment

varnerac:
Sorry for comment spam. GitHub went sideways last night.

PR #2110, opened 2018-10-19
[clt-rabbit] Extend mod_event_pusher_rabbit configuration

mkacper:
Closes esl/tide#490, esl/tide#451.

### DIFF
--- a/big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
+++ b/big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
@@ -43,13 +43,13 @@
          {amqp_password, <<"guest">>},
          %% enables publisher one-to-one confirms; disabled by default
          %% {confirms_enabled, true},
-         {presence_exchange, ?PRESENCE_EXCHANGE},
-         {chat_msg_exchange, ?CHAT_MSG_EXCHANGE},
-         {chat_msg_sent_topic, ?CHAT_MSG_SENT_TOPIC},
-         {chat_msg_recv_topic, ?CHAT_MSG_RECV_TOPIC},
-         {groupchat_msg_exchange, ?GROUP_CHAT_MSG_EXCHANGE},
-         {groupchat_msg_sent_topic, ?GROUP_CHAT_MSG_SENT_TOPIC},
-         {groupchat_msg_recv_topic, ?GROUP_CHAT_MSG_RECV_TOPIC}
+         {presence_exchange, [{name, ?PRESENCE_EXCHANGE}]},
+         {chat_msg_exchange, [{name, ?CHAT_MSG_EXCHANGE},
+                              {sent_topic, ?CHAT_MSG_SENT_TOPIC},
+                              {recv_topic, ?CHAT_MSG_RECV_TOPIC}]},
+         {groupchat_msg_exchange, [{name, ?GROUP_CHAT_MSG_EXCHANGE},
+                                   {sent_topic, ?GROUP_CHAT_MSG_SENT_TOPIC},
+                                   {recv_topic, ?GROUP_CHAT_MSG_RECV_TOPIC}]}
         ]).
 -define(MOD_EVENT_PUSHER_CFG, [{backends,
                                 [{rabbit, ?MOD_EVENT_PUSHER_RABBIT_CFG}]}]).
@@ -162,11 +162,7 @@ exchanges_are_created_on_module_startup(Config) ->
                  {?CHAT_MSG_EXCHANGE, ExCustomType},
                  {?GROUP_CHAT_MSG_EXCHANGE, ExCustomType}],
     ConfigWithCustomExchangeType =
-        ?MOD_EVENT_PUSHER_RABBIT_CFG ++ [
-                                         {presence_exchange_type, ExCustomType},
-                                         {chat_msg_exchange_type, ExCustomType},
-                                         {groupchat_msg_exchange, ExCustomType}
-                                        ],
+        extend_config_with_exchange_type(ExCustomType),
     %% WHEN
     start_mod_event_pusher_rabbit(ConfigWithCustomExchangeType),
     %% THEN exchanges are created
@@ -714,3 +710,17 @@ room_address(Room, Nick) ->
 user_room_jid(RoomJID, UserJID) ->
     Nick = nick(UserJID),
     <<RoomJID/binary, "/", Nick/binary>>.
+
+extend_config_with_exchange_type(ExType) ->
+    ExConfigKeys = [presence_exchange, chat_msg_exchange, groupchat_msg_exchange],
+    ExtendExConfig =
+        fun(ExKey) ->
+                {ExKey, proplists:get_value(ExKey, ?MOD_EVENT_PUSHER_RABBIT_CFG) ++ [{type, ExType}]}
+        end,
+    ExtendedExConfigs = [ExtendExConfig(ExKey) || ExKey <- ExConfigKeys],
+    ReplaceExConfig =
+       fun(ExConfig = {ExKey, _}, Config) ->
+               lists:keyreplace(ExKey, 1, Config, ExConfig)
+       end,
+    lists:foldl(ReplaceExConfig, ?MOD_EVENT_PUSHER_RABBIT_CFG,
+                ExtendedExConfigs).
--- a/big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
+++ b/big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
@@ -713,15 +713,10 @@ user_room_jid(RoomJID, UserJID) ->
     <<RoomJID/binary, "/", Nick/binary>>.
 
 extend_config_with_exchange_type(ExType) ->
-    ExConfigKeys = [presence_exchange, chat_msg_exchange, groupchat_msg_exchange],
-    ExtendExConfig =
-        fun(ExKey) ->
-                {ExKey, proplists:get_value(ExKey, ?MOD_EVENT_PUSHER_RABBIT_CFG) ++ [{type, ExType}]}
-        end,
-    ExtendedExConfigs = [ExtendExConfig(ExKey) || ExKey <- ExConfigKeys],
-    ReplaceExConfig =
-       fun(ExConfig = {ExKey, _}, Config) ->
-               lists:keyreplace(ExKey, 1, Config, ExConfig)
-       end,
-    lists:foldl(ReplaceExConfig, ?MOD_EVENT_PUSHER_RABBIT_CFG,
-                ExtendedExConfigs).
+    lists:map(fun({Ex, Opts}) when
+                        Ex == presence_exchange orelse
+                        Ex == chat_msg_exchange orelse
+                        Ex == groupchat_msg_exchange ->
+                      {Ex, Opts ++ [{type, ExType}]};
+                 (Other) -> Other
+              end, ?MOD_EVENT_PUSHER_RABBIT_CFG).
--- a/big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
+++ b/big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
@@ -40,6 +40,8 @@
          {amqp_port, 5672},
          {amqp_username, <<"guest">>},
          {amqp_password, <<"guest">>},
+         %% enables publisher one-to-one confirms; disabled by default
+         %% {confirms_enabled, true},
          {presence_exchange, ?PRESENCE_EXCHANGE},
          {chat_msg_exchange, ?CHAT_MSG_EXCHANGE},
          {chat_msg_sent_topic, ?CHAT_MSG_SENT_TOPIC},
--- a/big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
+++ b/big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
@@ -143,7 +143,6 @@ end_per_testcase(exchanges_are_created_on_module_startup, Config) ->
     close_rabbit_connection(Config),
     Config;
 end_per_testcase(exchanges_are_deleted_on_module_stop, Config) ->
-    close_rabbit_connection(Config),
     Config;
 end_per_testcase(CaseName, Config) ->
     maybe_cleanup_muc(CaseName, Config),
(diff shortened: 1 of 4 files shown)


doc/modules/mod_event_pusher_rabbit.md
```diff
@@ -144,9 +148,26 @@ which means that we don't care if a message is delivered to a RabbitMQ server.
 If a message couldn't be delivered to the server for any reason the module
 just updates appropriate metrics and print some log messages.
 
+### Type of exchanges
+
+By default all the exchanges used are of type `topic`. Using topic exchanges
+gives a lot of flexibility when binding queues to such an exchange by using
+`#` and `*` in binding keys. But flexibility comes at the cost of performance -
+imagine a scenario when there is thousands of users and AMQP consumers use
```
mentels: there are

doc/modules/mod_event_pusher_rabbit.md
```diff
@@ -144,9 +148,26 @@ which means that we don't care if a message is delivered to a RabbitMQ server.
...
+bindig keys, for particular users, which looks like `user_N@host.#`. In such
```
mentels: skip coma

doc/modules/mod_event_pusher_rabbit.md
```diff
@@ -144,9 +148,26 @@ which means that we don't care if a message is delivered to a RabbitMQ server.
...
+case RabbitMQ has to go through all the users in order to find out where
+a message should be sent to. This operations is proved to be costly. In a load
+test with 100k users a delay caused by this operation was substantial (about an
+order of magnitue higher in compared to a load test with 60k users).
```
mentels: than compared to

doc/modules/mod_event_pusher_rabbit.md
```diff
@@ -24,9 +24,13 @@ bare jid (`user@domain`) and configurable topic e.g `alice@localhost.private_mes
 * **amqp_port** (integer, default: `5672`) - Defines RabbitMQ server AMQP port;
 * **amqp_username** (string, default: `<<"guest">>`) - Defines RabbitMQ server username;
 * **amqp_password** (string, default: `<<"guest">>`) - Defines RabbitMQ server password;
+* **confirms_enabled** (boolean, default: `false`) - Enables/disables one-to-one publishers confirms;
 * **presence_exchange** (string, default: `<<"presence">>`) - Defines RabbitMQ presence exchange name;
+* **presence_exchange_type** (string, default: `<<"topic">>`) - Defines RabbitMQ presence exchange type;
```
mentels: Hmm this configuration seems to be a little bit to complex. Why not simplifying it a bit and go for a format like:

```erlang
{Exchange :: atom(), {Name :: string(), Type :: atom()}}
%% or
{Exchange :: atom(), Name :: string()} %% then we have the default type
```

mentels: For example:

```erlang
{presence_exchange, <<"presence">>, direct}
```

mkacper: @mentels but what if one want to change only exchange name or type? What do you think about that?

```erlang
{presence_exchange, [{name, <<"presence">>, {type, <<"direct">>}]
```

mkacper: @mentels but what if one want to change only exchange name or type? What do you think about that?

```erlang
{presence_exchange, [{name, <<"presence">>, {type, <<"direct">>}]
```

mkacper: @mentels but what if one want to change only exchange name or type? What do you think about that?

```erlang
{presence_exchange, [{name, <<"presence">>, {type, <<"direct">>}]
```

mkacper: @mentels but what if one want to change only exchange name or type? What do you think about that?

```erlang
{presence_exchange, [{name, <<"presence">>, {type, <<"direct">>}]
```

mkacper: @mentels but what if one want to change only exchange name or type? What do you think about that?

```
{presence_exchange, [{name, <<"presence">>, {type, <<"direct">>}]
```

sebb7:
testowanie komentarzy - ESL github 22.10.2018 godzina 11:35

doc/modules/mod_event_pusher_rabbit.md
mkacper: @mentels but what if one want to change only exchange name or type? What do you think about that?

```
{presence_exchange, [{name, <<"presence">>, {type, <<"direct">>}]
```

mkacper: @mentels but what if one want to change only exchange name or type? What do you think about that?

```
{presence_exchange, [{name, <<"presence">>, {type, <<"direct">>}]
```

mkacper: @mentels but what if one want to change only exchange name or type? What do you think about that?

```
{presence_exchange, [{name, <<"presence">>, {type, <<"direct">>}]
```

mkacper: @mentels but what if one want to change only exchange name or type? What do you think about that?

```
{presence_exchange, [{name, <<"presence">>, {type, <<"direct">>}]
```

big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
```diff
@@ -698,3 +711,17 @@ room_address(Room, Nick) ->
 user_room_jid(RoomJID, UserJID) ->
     Nick = nick(UserJID),
     <<RoomJID/binary, "/", Nick/binary>>.
+
+extend_config_with_exchange_type(ExType) ->
```
mentels: That could be way more simple with:

```erlang
lists:map(fun({Exchange, Opts})
                   when Exchange == presence_exchange;
                   when Exchange == chat_msg_exchange;
                   when Exchange == groupchat_msg_exchange ->
                       {Exchange, Opts ++ [{type, ExType}]};
                 (Other) -> Other
         end).
```

doc/modules/mod_event_pusher_rabbit.md
```diff
@@ -10,10 +10,10 @@ receiver) along with the message body.
 * **group message sent/received** - Carries the user id and the room id
 (full jids by default) along with the message body.
 
-All these notifications are sent as JSON strings to RabbitMQ topic exchanges.
-Each type of the notifications is sent to it's dedicated exchange. There are
-three exchanges created on startup of the module, for presences, private
-messages and group chat messages related events.
+All these notifications are sent as JSON strings to RabbitMQ exchanges. Type
+of exchanges can be chosen as desired. Each type of the notifications is sent
+to it's dedicated exchange. There are three exchanges created on startup of the
```
mentels: Now I think if the exchange should be deleted on node shutdown. What if one node in a cluster goes down? I think they should not be deleted. You could have an API to delete them explicitly wired to mongooseimctl if needed.

src/event_pusher/mod_event_pusher_rabbit.erl
```diff
@@ -175,16 +186,31 @@ amqp_client_opts(Host) ->
...
      || {Name, Type} <- mongoose_rabbit_worker:list_metrics()].
 
+-spec exchange_opt(Host :: jid:lserver(), ExchangeKey :: atom(),
+                   Option :: atom(), Default :: term()) -> term().
+exchange_opt(Host, Exchange, Option, Default) ->
+    ExchangeOptions = opt(Host, Exchange, []),
+    proplists:get_value(Option, ExchangeOptions, Default).
+
 %% Getter for module options
 -spec opt(Host :: jid:lserver(), Option :: atom()) -> Value :: term().
```
mentels: If we just pass `undefined` I think this function is uncecessary.

src/event_pusher/mod_event_pusher_rabbit.erl
```diff
@@ -175,16 +186,31 @@ amqp_client_opts(Host) ->
 -spec exchanges(Host :: jid:server()) -> [binary()].
 exchanges(Host) ->
     [
-     opt(Host, presence_exchange, ?DEFAULT_PRESENCE_EXCHANGE),
-     opt(Host, chat_msg_exchange, ?DEFAULT_CHAT_MSG_EXCHANGE),
-     opt(Host, groupchat_msg_exchange, ?DEFAULT_GROUP_CHAT_MSG_EXCHANGE)
+     {exchange_opt(Host, presence_exchange, name,
```
mentels: Here's another implementation for that:

```erlang
exchanges(Host) ->
    [{
      exchange_opt(Host, Ex, name, DefName),
      exchange_opt(Host, Ex, type, DefType),
     }
     || {Ex, {DefName, DefType}} <-
            [{presence_exchange, ?DEFAULT_PRESENCE_EXCHANGE},
             {chat_msg_exchange, ?DEFAULT_CHAT_MSG_EXCHANGE},
             {groupchat_msg_exchange, ?DEFAULT_GROUP_CHAT_MSG_EXCHANGE}].
```

src/mongoose_rabbit_worker.erl
```diff
@@ -210,20 +208,28 @@ make_chat_msg(From, To, UserMsg) ->
...
 
--spec publish_message_and_wait_for_confirm(Channel :: pid(),
-                                           Exchange :: binary(),
+-spec publish_message_and_wait_for_confirm(Exchange :: binary(),
                                            RoutingKey :: binary(),
-                                           Message :: binary()) -> term().
-publish_message_and_wait_for_confirm(Channel, Exchange, RoutingKey, Message) ->
+                                           Message :: binary(),
+                                           Opts :: publish_opts()) -> term().
+publish_message_and_wait_for_confirm(Exchange, RoutingKey, Message,
```
mentels: The name should indicate that we *maybe* wait for confirm.

mentels: Also when confirms are enabled the client code talking the the worker needs a way to get info when confirms have arrived. Otherwise there's no way to implement the backpressure mechanisms based on the confirms. So either with confirm on these are calls (or are always calls) or you somehow communicate back to the client.

big_tests/tests/mod_event_pusher_rabbit_SUITE.erl
```diff
@@ -40,13 +41,15 @@
...
-         {groupchat_msg_exchange, ?GROUP_CHAT_MSG_EXCHANGE},
-         {groupchat_msg_sent_topic, ?GROUP_CHAT_MSG_SENT_TOPIC},
-         {groupchat_msg_recv_topic, ?GROUP_CHAT_MSG_RECV_TOPIC}
+         %% enables publisher one-to-one confirms; disabled by default
+         %% {confirms_enabled, true},
+         {presence_exchange, [{name, ?PRESENCE_EXCHANGE}]},
+         {chat_msg_exchange, [{name, ?CHAT_MSG_EXCHANGE},
+                              {sent_topic, ?CHAT_MSG_SENT_TOPIC},
+                              {recv_topic, ?CHAT_MSG_RECV_TOPIC}]},
+         {groupchat_msg_exchange, [{name, ?GROUP_CHAT_MSG_EXCHANGE},
```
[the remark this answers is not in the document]
mkacper: I was thinking about that but I decided to stick to `topic` word because routing keys we use take advantage of topic syntax `user_N@host.chat_msg_sent` and you can only set the latest part (after user jid). Does it make sense?

doc/modules/mod_event_pusher_rabbit.md
mkacper: Good spot. Maybe mongooseimctl API is to much for the first iteration? What should we do with that for now?

src/mongoose_rabbit_worker.erl
mkacper: But how can we implement backpressure here? Isn't it enough to accumulate events to be sent in worker processeses message queues? 

As far as I understand `mod_event_pusher_rabbit:push_event/3` is called in c2s process context. Here is the flow:
1. c2s "emits" appropriate hooks (e.g. [user_send_packet hook])
2. Event pusher adds a handler for this hook ([code](https://github.com/esl/MongooseIM/blob/master/src/event_pusher/mod_event_pusher_hook_translator.erl#L34))
3. [The handler calls](https://github.com/esl/MongooseIM/blob/master/src/event_pusher/mod_event_pusher_hook_translator.erl#L65) `mod_event_pusher:push_event/3` which in turn [calls configured backends](https://github.com/esl/MongooseIM/blob/master/src/event_pusher/mod_event_pusher.erl#L40) (`mod_event_pusher_rabbit`) in our case.

Having this in mind doing a synchronous call to Rabbit workers we put pushing events on a critical path. Is that what we want? How can we implement the backpressure you mentioned? I think it requires a separate discussion and eventually a separate PR :)

[user_send_packet hook]: https://github.com/esl/MongooseIM/blob/beea15b37cadaf29b64b112970e0877898936778/src/ejabberd_c2s.erl#L877

mkacper:
@mentels please see the comments.

doc/modules/mod_event_pusher_rabbit.md
mkacper: Maybe we can check if we're the last node in the cluster?

mentels: I think Mongoose should just check if the appropriate toplogy is in place, that's all. So that if you take down a cluster for Maintance RabbitMQ still retains it's messages etc. Then if you start the cluster up again all is configured. 


But I'd go for an API to delete the topology anyway.

src/mongoose_rabbit_worker.erl
mentels: Ahh I see. I think it's fine then... I though that even if we wait for confirms too long the worker process can simply get overloaded by having tons of messages in it's mailbox. As C2S (or other client code) has no way of knowing of this overload condition. So that should be noted somehow, ideally prevented from.

doc/modules/mod_event_pusher_rabbit.md
mkacper: You mean `mongooseimctl` API for that?

mentels: Not necessarily.

mkacper: I'll deal with that in the separate PR.

src/mongoose_rabbit_worker.erl
mkacper: @mentels What's the final conclusion on that issue?

mentels: Maybe for starters queue length monitoring would be enough? And that could be set to some value.

mkacper: Do you mean that a worker can monitor its queue length and drop requests if the value is too high?

mentels: C2S monitors its queue and probably crash if it's too long. Yeah we loose messages that way but at least have a safety valve.

PR #2115, opened 2018-10-28
Sending messages with Smack Stanza Properties by the REST API.

2nets:
We need to send a message with Smack Stanza Properties by the REST API.
`<message xml:lang='en' to='alice@localhost' id='123' type='chat'>
        <body xml:lang='en_US'>Hi!</body>
        <properties xmlns="http://www.jivesoftware.com/xmlns/xmpp/properties"
            <property>
                <name>some_number</name>
                <value type='integer'>123</value>
            <property>
            <property>
                <name>some_string</name>
                <value type='string'>abc</value>
            <property>
        </properties>
      </message>`
We can get properties from the Stanza right now.
We need to post them.
`		$post = array(
			'caller' => '1@host',
			'to' => '2@host',
			'body' => 'Hi!',
			'properties' => json_encode(array('a'=>'123','b'=>'321')),
		);
`
Thanks!


michalwski:
@2nets I'm wondering if you can use a master branch for now. We are planing new release soon so then you will be able to use a tagged version instead of a branch.

PR #2116, opened 2018-10-29
[clt-rabbit] Refactor RabbitMQ event pusher backend

mkacper:
Implements https://github.com/esl/tide/issues/418#issuecomment-432707318.

### DIFF
--- a/src/event_pusher/mod_event_pusher_rabbit.erl
+++ b/src/event_pusher/mod_event_pusher_rabbit.erl
@@ -100,15 +100,10 @@ initialize_metrics(Host) ->
 
 -spec create_exchanges(Host :: jid:server()) -> ok.
 create_exchanges(Host) ->
-    [wpool:call(pool_name(Host),
-                {amqp_call, mongoose_amqp:exchange_declare(ExName, Type)},
-                available_worker) || {ExName, Type} <- exchanges(Host)].
-
--spec delete_exchanges(Host :: jid:server()) -> ok.
-delete_exchanges(Host) ->
-    [wpool:call(pool_name(Host),
-                {amqp_call, mongoose_amqp:exchange_delete(ExName)},
-                available_worker) || {ExName, _Type} <- exchanges(Host)].
+    Res = [wpool:call(pool_name(Host),
+                      {amqp_call, mongoose_amqp:exchange_declare(ExName, Type)},
+                      available_worker) || {ExName, Type} <- exchanges(Host)],
+    verify_exchanges_were_created_or_crash(Res).
 
 -spec handle_user_presence_change(JID :: jid:jid(), Status :: atom()) -> ok.
 handle_user_presence_change(JID = #jid{lserver = Host}, Status) ->
@@ -250,3 +245,14 @@ verify_opts(Opts) ->
     lists:filter(fun({_Opt, undefined}) -> false;
                     (_Other) -> true
                  end, Opts).
+
+-spec verify_exchanges_were_created_or_crash(Res :: list()) -> ok | no_return().
+verify_exchanges_were_created_or_crash(Res) ->
+    case lists:all(fun(E) ->
+                           E == mongoose_amqp:exchange_declare_ok()
+                   end, Res) of
+        true ->
+            ok;
+        false ->
+            erlang:throw("Creating exchanges failed.")
+    end.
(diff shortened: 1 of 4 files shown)


src/mongoose_rabbit_worker.erl
```diff
@@ -241,20 +176,9 @@ establish_rabbit_connection(Opts, Host) ->
...
-                amqp_channel:call(Channel, #'confirm.select'{}),
-            true;
-        false ->
-            false
-    end.
-
 -spec close_rabbit_connection(Connection :: pid(), Channel :: pid(),
                               Host :: jid:server()) ->
-    ok | term().
+    ok | no_return().
```
bszaf: This function should not have `no_return()`
It can either return `already_closed` or `ok` or `Error :: term()`  (from amq_connection:close() )

src/event_pusher/mod_event_pusher_rabbit.erl
```diff
@@ -88,102 +93,65 @@ push_event(Acc, _, _) ->
 %%% Internal functions
 %%%===================================================================
 
+-spec initialize_metrics(Host :: jid:server()) -> ok.
```
bszaf: The spec is incorrect. It returns the list of results of `mongoose_metrics:ensure_metric/3`

src/event_pusher/mod_event_pusher_rabbit.erl
```diff
@@ -88,102 +93,65 @@ push_event(Acc, _, _) ->
...
+initialize_metrics(Host) ->
+    [mongoose_metrics:ensure_metric(Host, Name, Type)
+     || {Name, Type} <- mongoose_rabbit_worker:list_metrics()].
+
 -spec create_exchanges(Host :: jid:server()) -> ok.
 create_exchanges(Host) ->
-    wpool:call(pool_name(Host), {create_exchanges, exchanges(Host)},
-               available_worker).
+    [wpool:call(pool_name(Host),
```
bszaf: We are silently ignoring fails here. If mongoose is not able to create exchange, there is no warning/crash.

bszaf: However it is okay if the worker crashes. But still the `mod_event_pusher_rabbit` is considered healthy. I think it would be ideally to prevent MongooseIM from starting, if we cannot declare exchanges.

src/mongoose_rabbit_worker.erl
```diff
@@ -117,64 +113,28 @@ terminate(_Reason, #state{connection = Connection, channel = Channel,
...
-                                     to_jid := To,
-                                     msg := UserMsg,
-                                     exchange := Exchange,
-                                     host := Host}, Opts) ->
-    Message = make_chat_msg(From, To, UserMsg),
-    publish(Exchange, RoutingKey, Message, Opts#state{host = Host}).
+-spec handle_amqp_call(Method :: mongoose_amqp:method(), worker_opts()) ->
+    {ok, term()} | {error, term(), term()}.
+handle_amqp_call(Method, #state{channel = Channel}) ->
+    try amqp_channel:call(Channel, Method) of
```
bszaf: I am not sure, if catching is ok here. If exception is thrown, the channel is closed. If we catch it, the process is still participating in the pool as a worker and accepts the requests.  I think we should try to reopen connection/channel to avoid loosing messages from message queue and then crash without catching any exceptions if it fails.

src/mongoose_rabbit_worker.erl
```diff
@@ -185,44 +145,19 @@ publish(Exchange, RoutingKey, Message, Opts = #state{host = Host}) ->
...
+    true | no_return().
+publish_message_and_wait_for_confirm(Method, Payload,
                                      #state{channel = Channel,
-                                            confirms_enabled = IsConfirmOn}) ->
-    amqp_channel:call(Channel, #'basic.publish'{exchange = Exchange,
-                                                routing_key = RoutingKey},
-                      #amqp_msg{payload = Message}),
-    case IsConfirmOn of
+                                            confirms = IsConfirmEnabled}) ->
+    amqp_channel:call(Channel, Method, Payload),
```
bszaf: We are not handling the failures here, which may lead to loosing messages that are waiting in the process' message queue to be published.
I think we should apply similar retry mechanisms, which I described for `handle_amqp_call` function.

src/mongoose_rabbit_worker.erl
mkacper: When we try to close a connection that is already closed, according to [this]( https://github.com/rabbitmq/rabbitmq-erlang-client/blob/master/src/amqp_gen_connection.erl#L82), it ends up with an exception.

mkacper:
@bszaf thanks for the very valuable review!

I fixed "cosmetics" issues and improve dealing with exchanges creation/deletion.

When it comes to catching `amqp_channel:call/2/3` exceptions I noted this down ([here](https://github.com/esl/tide/issues/418#issuecomment-434607213)). I'll create a separate PR for that.

Please check this again and merge if everything looks OK to you :)

PR #2119, opened 2018-11-02
[clt-rabbit] Don't crash RMQ worker when amqp exception occurs

mkacper:
Closes https://github.com/esl/tide/issues/418#issuecomment-434607213.

### DIFF
--- a/src/mongoose_rabbit_worker.erl
+++ b/src/mongoose_rabbit_worker.erl
@@ -32,7 +32,8 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2]).
 
--record(state, {connection :: pid(),
+-record(state, {amqp_client_opts :: mongoose_amqp:network_params(),
+                connection :: pid(),
                 channel :: pid(),
                 host :: binary(),
                 confirms :: boolean()}).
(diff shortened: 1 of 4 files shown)


src/mongoose_rabbit_worker.erl
```diff
@@ -128,36 +133,55 @@ handle_amqp_publish(Method, Payload, Opts = #state{host = Host}) ->
...
-    amqp_channel:call(Channel, Method, Payload),
-    case IsConfirmEnabled of
-        true ->
-            amqp_channel:wait_for_confirms(Channel);
-        false ->
-            true
+    case try_to_apply(amqp_channel, call, [Channel, Method, Payload]) of
+        {ok, _} ->
+            maybe_wait_for_confirms(Channel, IsConfirmEnabled);
+        {error, _, _} -> channel_exception
```
bszaf: I do not like idea of ignoring an error - we are not returning error or logging it.

src/mongoose_rabbit_worker.erl
```diff
@@ -128,36 +133,55 @@ handle_amqp_publish(Method, Payload, Opts = #state{host = Host}) ->
...
+            {noreply, Opts};
         false ->
             update_messages_failed_metrics(Host),
             ?WARNING_MSG("event=rabbit_message_sent_failed reason=negative_ack",
-                         []);
+                         []),
+            {noreply, Opts};
+        channel_exception ->
+            update_messages_failed_metrics(Host),
+            ?WARNING_MSG("event=rabbit_message_sent_failed reason=channel_exception",
```
bszaf: We should have description of exception returned from `publish_message_and_wait_for_confirm/3`. Here, we know that something is wrong, but do not have information about it.

src/mongoose_rabbit_worker.erl
```diff
@@ -88,21 +89,30 @@ init(Opts) ->
...
-handle_call({amqp_call, Method}, _From, State) ->
-    Res = handle_amqp_call(Method, State),
-    {reply, Res, State}.
+handle_call({amqp_call, Method, Opts}, _From,
+            State = #state{channel = Channel}) ->
+    NewState = State#state{host = proplists:get_value(host, Opts)},
+    case try_to_apply(amqp_channel, call, [Channel, Method]) of
+        {ok, _} = Res ->
+            {reply, Res, NewState};
+        {error, _, _} = Err ->
```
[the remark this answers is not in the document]
mkacper: @bszaf  Do you think it's necessary? We return an error to a caller.

PR #2138, opened 2018-11-20
update deps

getong:


fenek:
Hi @getong 

Thanks for this update! Are there any particular fixes or improvements introduced in these updated libs? Or is it just a routine update?

getong:
routine update.

PR #2155, opened 2018-12-06
[skip ci] Add note about cache backend for pubsub_last_item

zofpolkowska:
Update `mod_pubsub` documentation.

### DIFF
--- a/doc/modules/mod_pubsub.md
+++ b/doc/modules/mod_pubsub.md
@@ -35,16 +35,24 @@ Node configuration still uses the default configuration defined by the node plug
 
 #### Note about RDBMS backend
 
-Current RDBMS backend replaces `pubsub_state` and `pubsub_item` Mnesia tables with RDBMS equivalents.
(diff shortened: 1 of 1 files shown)


doc/modules/mod_pubsub.md
```diff
@@ -35,16 +35,25 @@ Node configuration still uses the default configuration defined by the node plug
...
+Current RDBMS backend replaces `pubsub_node`, `pubsub_state` and `pubsub_item` Mnesia tables with RDBMS equivalents.
 Due to a fact that some data is still maintained in Mnesia, there is a certain risk of data becoming inconsistent.
 The schema used by this backend may change until it reaches stable status.
 
+#### Cache Backend
+
+Cache backend can be configured to use Mnesia or RDBMS, by default it is
+not used. It provides a possibility of storing `pubsub_last_item` table
+separately. For example you can configure `backend` to use `rdbms` and
+`last_item_cache` to use `mnesia`.
```
goddammit: Although it's not a default setting, the cache backend can be configured to use Mnesia or RDBMS. 
It allows storing the `pubsub_last_item` table separately. 
For example you can configure `backend` to use `rdbms` and`last_item_cache` to use `mnesia`.

PR #2201, opened 2019-01-25
Fix `mongooseimctl debug` command

erszcz:
This approaches fixing `mongooseimctl debug` for cases of no hostname provided with `-sname`/`-name` in `vm.args`. The details are in respective commit log messages.

One remaining caveat is the case of the Mac example - running the script on a Mac is not that worrying, since it's not going to happen in production, but it means that we might run into the problem somewhere in production due to incomplete knowledge of how the Erlang VM expands `-name a` into `-name a@FQDN`.

This fix is necessary for remote shell access in a Kubernetes pod.

### DIFF
--- a/rel/files/mongooseim
+++ b/rel/files/mongooseim
@@ -35,6 +35,14 @@ if [ -z "$NAME_ARG" ]; then
 fi
 NAME=$(echo $NAME_ARG | awk '{print $1}')
 NODE=$(echo $NAME_ARG | awk '{print $2}')
+HOST=$(echo $NODE | awk -F@ '{ print $2 }')
+if [ -z "$HOST" ]; then
+    case $NAME in
+        "-sname") HOST=$(hostname -s) ;;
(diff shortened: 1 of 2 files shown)


mongoose-im:
[6129.1](https://travis-ci.org/esl/MongooseIM/jobs/484352102) / Erlang 19.3 / small_tests / 95711281c32b394a334dce3914feb536bb655872
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2201/6129/small_tests.19.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2201/6129/small_tests.19.3/small/ct_run.mongooseim@localhost.2019-01-25_13.03.57/lib.mongooseim.logs/run.2019-01-25_13.03.59/suite.log.html)

michalwski:
Thanks for the patch!

Issue #2210, opened 2019-02-03
HTTP/2 SSL error with BOSH due to uppercase characters in headers


cogentParadigm:
MongooseIM version: 3.2.0
Installed from: source
Erlang/OTP version: 21.1

I pointed my javascript client using Strophe.js to the HTTPS BOSH endpoint in Chrome and the connection failed with the error:

```
 net::ERR_SPDY_PROTOCOL_ERROR
```

I went to chrome://net-export/ to record a network log and uploaded it to https://netlog-viewer.appspot.com for inspection. I saw the following error:

![screen shot 2019-02-02 at 5 28 32 pm](https://user-images.githubusercontent.com/78707/52171023-5d511400-2712-11e9-9f2c-3b67e283ad34.png)

It looks like the HTTP/2 specification requires lowercase header names. Excerpt from https://tools.ietf.org/html/rfc7540#section-8.1.2 below:

```
Just as in HTTP/1.x, header field names are strings of ASCII
characters that are compared in a case-insensitive fashion.  However,
header field names MUST be converted to lowercase prior to their
encoding in HTTP/2.  A request or response containing uppercase
header field names MUST be treated as malformed
```

michalwski:
Thanks for the report @cogentParadigm, it's a good catch. We didn't test MongooseIM's BOSH implementation with HTTP/2 client yet.

michalwski:
A fix for this issue was merged (#2211 ). I added task to our backlog to extend tests with an HTTP/2 client.

PR #2211, opened 2019-02-03
lowercase HTTP headers in mod_bosh for HTTP/2 compliance

cogentParadigm:
This PR addresses #2210 

Proposed changes include:
* make HTTP headers lowercase to avoid HTTP/2 connection errors


michalwski:
Thanks for the patch!

PR #2230, opened 2019-02-27
Introduce XMPP Ping response (aka pong) time metric

erszcz:
This PR introduces a new hook called for successful XMPP ping responses (aka pongs). It can, for example, be used to build a response time histogram therefore roughly gauge connection quality of client devices. This, in turn, allows for `mod_ping` parameter fine-tuning.


michalwski:
The work is being continued in #2527

PR #2247, opened 2019-03-26
fix docs about mongoose_api_admin authorization

cgrtrifork:
This PR addresses #
Fixes documentation about `mongoose_api_admin` authorization. The example given does not work, User and Password need to be binaries.

Proposed changes include:
* Make explicit that User and Password are binaries by using << >> in the example given.


.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,7 +1,5 @@
 This PR addresses #
+Fixes documentation about `mongoose_api_admin` authorization. The example given does not work, User and Password need to be binaries.
```
fenek: File modified by accident. :)

PR #2259, opened 2019-04-09
Add cookie security guide

fenek:
This PR adds a guide on securing a MongooseIM deployment.

### DIFF
--- a/doc/Erlang-cookie-security.md
+++ b/doc/Erlang-cookie-security.md
@@ -1,0 +1,12 @@
+In order for MongooseIM nodes to communicate with each other, they have to share a common secret - i.e. a cookie - which is a feature of the underlying Erlang VM.
+The cookie itself is an UTF8 string that is up to 255 characters in size.
+Thanks to the cookie, MongooseIM nodes can determine if they are allowed to communicate with each other and with no cookie no communication would flow between the nodes - a feature especially useful when you are running more than one applications on a single machine. 
+
+For ease of deployment and staging, each MongooseIM node is configured with a predefined erlang cookie.
+However, one should remember that for production environments this cookie should be reconfigured to a new secret cookie, as this will secure your system from intrusion.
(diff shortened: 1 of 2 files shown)


doc/Erlang-cookie-security.md
```diff
@@ -0,0 +1,12 @@
...
+The cookie itself is an UTF8 string that is up to 255 characters in size.
+Thanks to the cookie, MongooseIM nodes can determine if they are allowed to communicate with each other and with no cookie no communication would flow between the nodes - a feature especially useful when you are running more than one applications on a single machine. 
+
+For ease of deployment and staging, each MongooseIM node is configured with a predefined erlang cookie.
+However, one should remember that for production environments this cookie should be reconfigured to a new secret cookie, as this will secure your system from intrusion.
+You can change the cookie by changing the parameters of the `-setcookie` parameter in the `vm.args` file.
+
+Nonetheless, one should remember that communication between Erlang nodes is unencrypted by default, hence, the cookie is vulnerable to sniffing.
+If one has access to a MongooseIM cookie and figures out the hostname of a node, one can execute shell commands remotely on that node.
+Therefore, one should either provide privacy at the network layer (strongly recommended) or disable port 4369 for ultimate security.
```
rslota: Locking of port 4369 (epmd) doesn't help with anything. The port on which Erlang VM is listening is still open (by default it's randomized) and answers to external calls. Yes, you'll need a cookie to actually do something, but with very simple IP spoofing it's quite easy to trick other nodes to send a request with cookie on it. There simply should not be any assumptions on security unless either local network is considered "secure", all Erlang VM ports are fixed and locked or the communication is based on TLS with "Certificate Pinning".

rslota: Oh, I guess I was late, as this has been already merged :(

fenek: Yes, it does because without `epmd` the attacker would need to learn node names as well. However, I know it's not a perfect protection.

PR #2260, opened 2019-04-12
gdpr logs collection

DenysGonchar:
currently **get_logs_from_node/4** receives zipped logs from other nodes as rpc call return value. as agreed that is ok for the first version, if logs are too big this should be changed to some other delivery method (e.g. FTP) 

for testing run:
```
tools/test-runner.sh --db --preset internal_mnesia --skip-small-tests --skip-cover -- gdpr:retrieve_personal_data:retrieve_logs
```

### DIFF
--- a/big_tests/tests/gdpr_SUITE.erl
+++ b/big_tests/tests/gdpr_SUITE.erl
@@ -256,13 +256,13 @@ retrieve_inbox(Config) ->
 retrieve_logs(Config) ->
     escalus:fresh_story(Config, [{alice, 1}],
         fun(Alice) ->
-            User = escalus_client:username(Alice),
-            Domain = escalus_client:server(Alice),
-            JID = binary_to_list(User) ++ "@" ++ binary_to_list(Domain),
+            User = string:lowercase(escalus_client:username(Alice)),
+            Domain = string:lowercase(escalus_client:server(Alice)),
+            JID = string:uppercase(escalus_client:short_jid(Alice)),
             MIM2Node = distributed_helper:mim2(),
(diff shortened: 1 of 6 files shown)


src/ejabberd_config.erl
```diff
@@ -54,6 +54,9 @@
 -export([config_state/0]).
 -export([config_states/0]).
 
+-export([all_cluster_nodes/0,
```
fenek: `all_cluster_nodes/0` not used in the PR eventually

big_tests/tests/gdpr_SUITE.erl
```diff
@@ -254,12 +254,20 @@ retrieve_inbox(Config) ->
...
-                                   ["event=disturbance_in_the_force, jid=sith@localhost", []]),
-    Dir = request_and_unzip_personal_data(<<"sith">>, <<"localhost">>, Config),
-    Filename = filename:join(Dir, "logs.txt"),
-    {ok, Content} = file:read_file(Filename),
-    {match, _} = re:run(Content, "disturbance_in_the_force").
+    escalus:fresh_story(Config, [{alice, 1}],
+        fun(Alice) ->
+            User = escalus_client:username(Alice),
+            Domain = escalus_client:server(Alice),
+            JID = binary_to_list(User) ++ "@" ++ binary_to_list(Domain),
```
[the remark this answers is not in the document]
fenek: Also I would convert this to lower case to ensure that nobody introduces a regression (like removing `-i` from grep) in the future.

src/ejabberd_config.erl
fenek: Still not resolved.

PR #2262, opened 2019-04-12
Register command with automatic jid creation

kanes115:
This PR modifies `register` command in the cli as well as adds `register_identified` command.

`register` - it no longer accepts username argument and instead generates random and monotonic user uid.
`register_identified` - it behaves the same way as `register` used to behave.

TODO:
- [ ] We have to think if the random string suffix is necessary (added so that we don't have collisions with MUCs.
- [ ] Think of a schema of jids. Do we want to have a special format of user id? Should it conatin information like whether it is a private user jid or room jid (and this way avoid collisions with mucs from point 1.?

### DIFF
--- a/big_tests/tests/ejabberdctl_SUITE.erl
+++ b/big_tests/tests/ejabberdctl_SUITE.erl
@@ -172,6 +172,10 @@ end_per_group(Rosters, Config) when (Rosters == roster) or (Rosters == roster_ad
 end_per_group(_GroupName, Config) ->
     Config.
 
+get_registered_users() ->
+    Host = ct:get_config({hosts, mim, domain}),
+    Users = rpc(mim(), ejabberd_auth, get_vh_registered_users, [Host]).
+
 init_per_testcase(CaseName, Config)
   % these cases are incompatible with domainless rdbms schema
   when CaseName == delete_old_users_vhost
@@ -305,7 +309,7 @@ delete_old_users_vhost(Config) ->
     Now = Mega*1000000+Secs,
     set_last(AliceName, Domain, Now-86400*30),
 
-    {_, 0} = ejabberdctl("register", [KateName, SecDomain, KatePass], Config),
+    {_, 0} = ejabberdctl("register_identified", [KateName, SecDomain, KatePass], Config),
     {_, 0} = ejabberdctl("check_account", [KateName, SecDomain], Config),
     {_, 0} = ejabberdctl("delete_old_users_vhost", [SecDomain, "10"], Config),
     {_, 0} = ejabberdctl("check_account", [AliceName, Domain], Config),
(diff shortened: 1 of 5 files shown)


mongoose-im:
[6310.1](https://travis-ci.org/esl/MongooseIM/jobs/519302689) / Erlang 19.3 / small_tests / 5cf6b0937dea033ba37a1704b0eee577b8ac936e
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2262/6310/small_tests.19.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2262/6310/small_tests.19.3/small/ct_run.mongooseim@localhost.2019-04-12_15.16.47/lib.mongooseim.logs/run.2019-04-12_15.16.48/suite.log.html)

src/ejabberd_admin.erl
```diff
@@ -333,6 +346,11 @@ register(User, Host, Password) ->
             {ok, io_lib:format("User ~s@~s successfully registered", [User, Host])}
     end.
 
+generate_user() ->
+    mongoose_bin:join([mongoose_bin:gen_from_timestamp(),
```
aleklisi: How about adding `<<"private_jid_">>` at the beginning or/and at the end for easier search??
@mattsaidwhat what do you think?

mattsaidwhat: I would not do that preemptively.

big_tests/tests/ejabberdctl_SUITE.erl
```diff
@@ -907,10 +911,17 @@ simple_register(Config) ->
     Domain = ct:get_config({hosts, mim, domain}),
     {Name, Password} = {<<"tyler">>, <<"durden">>},
     %% when
-    {_, 0} = ejabberdctl("register", [Name, Domain, Password], Config),
+    {R1, 0} = ejabberdctl("registered_users", [Domain], Config),
+    {match, ResList1} = re:run(R1, "(.+)", [global]),
```
fenek: Isn't it less obscure to just use `string:tokens`? :)

fenek:
There is no chance of collision, because MUCs exist in different subdomain. So even if room username is equal to user username, then it's fine. Todos from description are not necessary.

big_tests/tests/ejabberdctl_SUITE.erl
kanes115: Indeed

mongoose-im:
[6346.1](https://travis-ci.org/esl/MongooseIM/jobs/523833576) / Erlang 19.3 / small_tests / c622065283f44214b566ac9df8c7fc9c385287c2
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2262/6346/small_tests.19.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2262/6346/small_tests.19.3/small/ct_run.mongooseim@localhost.2019-04-24_07.14.10/lib.mongooseim.logs/run.2019-04-24_07.14.11/suite.log.html)

Issue #2269, opened 2019-04-16
MAM not advertised in disco#info


jzskca:
MongooseIM version: 3.2.0
Installed from: source
Erlang/OTP version: 19.2.1

According to [XEP-0313](https://xmpp.org/extensions/xep-0313.html#support), MAM must be advertised when running `disco#info` against archiving JIDs. Unfortunately, this does not seem to be happening. This causes e.g. Smack to assume that there is no MAM archive for the JID. (`disco#items` does seem to advertise MAM.)

Sample conversation:
```
<iq to='room-light-3d9d3009-9e0b-4bcc-ada0-8e1a27202b8b@muclight.example.com' id='R6fvl-114' type='get'>
  <query xmlns='http://jabber.org/protocol/disco#info'></query>
</iq>
​
<iq from='room-light-3d9d3009-9e0b-4bcc-ada0-8e1a27202b8b@muclight.example.com' to='user1@example.com/3Y3TVk' id='R6fvl-114' type='result'>
  <query xmlns='http://jabber.org/protocol/disco#info'>
    <identity category='conference' type='text' name='MUC Light'/>
    <feature var='urn:xmpp:muclight:0'/>
  </query>
</iq>
```

FWIW, we use the following to resolve this:
```
--- src/mam/mod_mam.erl
+++ src/mam/mod_mam.erl
@@ -50,6 +50,7 @@
          remove_user/3,
          filter_packet/1,
          determine_amp_strategy/5,
+         get_sm_features/5,
          sm_filter_offline_message/4]).
 
 %%private
@@ -202,6 +203,8 @@ start(Host, Opts) ->
     ejabberd_hooks:add(anonymous_purge_hook, Host, ?MODULE, remove_user, 50),
     ejabberd_hooks:add(amp_determine_strategy, Host, ?MODULE, determine_amp_str
ategy, 20),
     ejabberd_hooks:add(sm_filter_offline_message, Host, ?MODULE, sm_filter_offl
ine_message, 50),
+    ejabberd_hooks:add(disco_sm_features, Host, ?MODULE, get_sm_features, 50),
+    ejabberd_hooks:add(disco_local_features, Host, ?MODULE, get_sm_features, 50
),
     mongoose_metrics:ensure_metric(Host, [backends, ?MODULE, lookup], histogram
),
     mongoose_metrics:ensure_metric(Host, [Host, modMamLookups, simple], spiral),
[...]

PR #2272, opened 2019-04-16
Advertise MAM in disco#info

jzskca:
This PR addresses #2269.

Proposed changes include:
 * Advertise MAM for user JIDs.
 * Advertise MAM for MUC Light rooms.

As currently written, this unconditionally advertises MAM for MUC Light rooms, since MAM is a requirement for MUC Light.

### DIFF
--- a/src/muc_light/mod_muc_light_codec_modern.erl
+++ b/src/muc_light/mod_muc_light_codec_modern.erl
@@ -61,7 +61,7 @@ encode({#msg{} = Msg, AffUsers}, Sender, {RoomU, RoomS} = RoomUS, HandleFun) ->
       end, AffUsers);
 encode(OtherCase, Sender, RoomUS, HandleFun) ->
     {RoomJID, RoomBin} = jids_from_room_with_resource(RoomUS, <<>>),
-    case encode_iq(OtherCase, RoomJID, RoomBin, HandleFun) of
+    case encode_iq(OtherCase, Sender, RoomJID, RoomBin, HandleFun) of
         {reply, ID} ->
             IQRes = make_iq_result(RoomBin, jid:to_binary(Sender), ID, <<>>, undefined),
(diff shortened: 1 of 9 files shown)


src/muc_light/mod_muc_light_codec_modern.erl
```diff
@@ -268,7 +268,12 @@ encode_iq({get, #disco_info{ id = ID }}, _RoomJID, _RoomBin, _HandleFun) ->
                        attrs = [{<<"category">>, <<"conference">>},
                                 {<<"type">>, <<"text">>},
                                 {<<"name">>, <<"MUC Light">>}]},
-                #xmlel{name = <<"feature">>, attrs = [{<<"var">>, ?NS_MUC_LIGHT}]}],
+                #xmlel{name = <<"feature">>, attrs = [{<<"var">>, ?NS_MUC_LIGHT}]},
+                % MUC Light requires MAM
+                #xmlel{name = <<"feature">>, attrs = [{<<"var">>, ?NS_MAM_03}]},
```
fenek: These shouldn't be advertised if MAM is not enabled.

jzskca:
Updated.

I removed the changes to `mod_mam` and just addressed advertising for MUC, since the source of that problem seems clear (`disco#info` is handled by the MUC modules directly instead of by `mod_disco`). For `disco#info` queries to users, although MAM is registered with `mod_disco`, `mod_disco:get_sm_features` does not return registered features, unlike `mod_disco:get_local_features`. I don't understand the distinction between the two well enough to offer a solution.

I haven't figured out yet why test steps 3, 7, and 8 are failing the `disco_features` tests. Any advice would be appreciated.

aaabbbccc-123:
just FYI, src/mod_muc_room.erl need to be update with similar change to make converse.js show MAM support for the room. Reading from the specs, that should be the correct interpretation i think.

PR #2276, opened 2019-04-18
send messages from inbox directly to the requesting resource

bartekgorny:
If messages from inbox are sent to bare jid, then if a requestor has another connection open at the same time with higher priority, then he is going to receive iq result, but messages would go to the other connection. This PR fixes it and adds a regression test.

### DIFF
--- a/big_tests/tests/inbox_SUITE.erl
+++ b/big_tests/tests/inbox_SUITE.erl
@@ -28,6 +28,7 @@
 -export([msg_sent_stored_in_inbox/1,
          user_has_empty_inbox/1,
          user_has_two_unread_messages/1,
+         other_resources_do_not_interfere/1,
          reset_unread_counter/1,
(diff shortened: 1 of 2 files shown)


mongoose-im:
[6331.1](https://travis-ci.org/esl/MongooseIM/jobs/521789137) / Erlang 19.3 / small_tests / 38700aede7f5c583da2ff51e5d4e64d9ea375b80
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2276/6331/small_tests.19.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2276/6331/small_tests.19.3/small/ct_run.mongooseim@localhost.2019-04-18_16.01.28/lib.mongooseim.logs/run.2019-04-18_16.01.29/suite.log.html)

PR #2280, opened 2019-04-23
Gdpr retreive personal data from mod vcard

aleklisi:
This PR allows to retrieve personal data form `mod_vcard` backends.

### DIFF
--- a/src/gdpr.erl
+++ b/src/gdpr.erl
@@ -1,18 +1,18 @@
 -module(gdpr).
 
 -export_type(
-    [table_name/0,
+    [table/0,
     schema/0,
     entities/0,
     username/0,
     domain/0]).
 
--type table_name() :: atom().
+-type table() :: atom().
 -type entity() :: [term()] | tuple() | string().
 -type entities() :: [entity()].
--type schema() :: [binary() | string()].
+-type schema() :: [string()].
 -type domain() :: binary().
 -type username() :: binary().
 
 -callback get_personal_data(username(), domain()) ->
-    [{table_name(), schema(), entities()}].
+    [{table(), schema(), entities()}].
--- a/src/gdpr.erl
+++ b/src/gdpr.erl
@@ -1,18 +1,18 @@
 -module(gdpr).
 
 -export_type(
-    [binary_table_name/0,
+    [table_name/0,
     schema/0,
     entities/0,
     username/0,
     domain/0]).
 
--type binary_table_name() :: binary().
--type entity() :: [term()].
+-type table_name() :: atom().
+-type entity() :: [term()] | tuple() | string().
 -type entities() :: [entity()].
--type schema() :: [binary()].
+-type schema() :: [binary() | string()].
 -type domain() :: binary().
 -type username() :: binary().
 
 -callback get_personal_data(username(), domain()) ->
-    [{binary_table_name(), schema(), entities()}].
+    [{table_name(), schema(), entities()}].
--- a/src/mod_vcard_mnesia.erl
+++ b/src/mod_vcard_mnesia.erl
@@ -23,17 +23,16 @@
 -export([get_personal_data/2]).
 
 -spec get_personal_data(gdpr:username(), gdpr:domain()) ->
-    [{gdpr:binary_table_name(), gdpr:schema(), gdpr:entities()}].
+    [{gdpr:table_name(), gdpr:schema(), gdpr:entities()}].
 
 get_personal_data(Username, Server) ->
     LUser = jid:nodeprep(Username),
-    Table = vcard,
-    Schema = ["vcard"],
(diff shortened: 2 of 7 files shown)


src/gdpr.erl
```diff
@@ -1,18 +1,18 @@
 -module(gdpr).
 
 -export_type(
-    [binary_table_name/0,
+    [table_name/0,
```
fenek: I'm actually wondering if we should call it "table" because the returned schema or data doesn't have to necessarily reflect actual, specific table in DB. Maybe a name like "group" would be more accurate (like "data group). Or "type".

src/gdpr.erl
```diff
@@ -1,18 +1,18 @@
...
     username/0,
     domain/0]).
 
--type binary_table_name() :: binary().
--type entity() :: [term()].
+-type table_name() :: atom().
+-type entity() :: [term()] | tuple() | string().
 -type entities() :: [entity()].
--type schema() :: [binary()].
+-type schema() :: [binary() | string()].
```
fenek: Why can't we decide one single type?

src/mod_vcard_mnesia.erl
```diff
@@ -23,17 +23,16 @@
...
-    Schema = ["vcard"],
+    Schema = ["jid", "vcard"],
     US = {LUser, Server},
-    Trans = fun() -> mnesia:read({Table, US}) end,
+    Trans = fun() -> mnesia:read({vcard, US}) end,
     {atomic, Records} = mnesia:transaction(Trans),
-    SerialzedRecords = lists:map(fun({T, U, Xeml}) -> {T,U,exml:to_binary(Xeml)} end, Records),
-    [{Table, Schema, SerialzedRecords}].
+    SerialzedRecords = lists:map(fun({_T, U, Xeml}) -> {U,exml:to_binary(Xeml)} end, Records),
+    [{vcard, Schema, SerialzedRecords}].
```
fenek: `SerializedRecords`

src/mod_vcard_rdbms.erl
```diff
@@ -49,16 +49,20 @@
...
-    [{Table, Schema, [[SerialzedRecord]]}].
+    Jid = jid:to_binary({LUser, LServer}),
+    Schema = ["jid", "vcard"],
+    Entitis = case get_vcard(LUser, LServer) of
+        {ok, Record} ->
+            SerialzedRecord = exml:to_binary(Record),
+            [{Jid, SerialzedRecord}];
+         _ -> []
+        end,
+        [{vcard, Schema, Entitis}].
```
fenek: `Entities`

src/mod_vcard_rdbms.erl
```diff
@@ -49,16 +49,20 @@
...
-    Schema = ["vcard"],
-    {ok, Record} = get_vcard(LUser, LServer),
-    SerialzedRecord = exml:to_binary(Record),
-    [{Table, Schema, [[SerialzedRecord]]}].
+    Jid = jid:to_binary({LUser, LServer}),
+    Schema = ["jid", "vcard"],
+    Entitis = case get_vcard(LUser, LServer) of
+        {ok, Record} ->
+            SerialzedRecord = exml:to_binary(Record),
+            [{Jid, SerialzedRecord}];
```
fenek: `SerializedRecord`

src/mod_vcard_riak.erl
```diff
@@ -42,16 +42,20 @@
...
     LUser = jid:nodeprep(Username),
     LServer = jid:nameprep(Server),
-    Table = vcard,
-    Schema = ["vcard"],
-    {ok, Record} = get_vcard(LUser, LServer),
-    SerialzedRecord = exml:to_binary(Record),
-    [{Table, Schema, [[SerialzedRecord]]}].
+    Jid = jid:to_binary({LUser, LServer}),
+    Schema = ["jid", "vcard"],
+    Entitis = case get_vcard(LUser, LServer) of
```
fenek: `Entities`

src/mod_vcard_riak.erl
```diff
@@ -42,16 +42,20 @@
...
+        {ok, Record} ->
+            SerialzedRecord = exml:to_binary(Record),
```
fenek: `SerializedRecord`

src/gdpr.erl
aleklisi: I'll rename it to `table`

src/gdpr.erl
aleklisi: I'll replace it with 
```erlang
-type schema() :: [string()].
```

src/gdpr.erl
aleklisi: I'll rename it in all 3 PRs #2280, #2281, #2282

fenek:
Merged manually in #2285

PR #2281, opened 2019-04-23
Gdpr retreive personal data from mod roster

aleklisi:
This PR allows to retrieve personal data form `mod_roster` backends.

### DIFF
--- a/src/mod_roster.erl
+++ b/src/mod_roster.erl
@@ -67,6 +67,7 @@
          ]).
 
 -export([remove_test_user/2, transaction/2, process_subscription_transaction/6]). % for testing
+-export([record_to_list_without_first/1]).
 
 -include("mongoose.hrl").
 -include("jlib.hrl").
@@ -1033,3 +1034,5 @@ item_to_map(#roster{} = Roster) ->
     #{jid => ContactJid, name => ContactName, subscription => Subs,
       groups => Groups, ask => Ask}.
 
+record_to_list_without_first(Record) ->
+    [ element(I,Record) || I <- lists:seq(2,tuple_size(Record)) ].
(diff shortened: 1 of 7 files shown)


src/mod_roster.erl
```diff
@@ -1033,3 +1034,5 @@ item_to_map(#roster{} = Roster) ->
     #{jid => ContactJid, name => ContactName, subscription => Subs,
       groups => Groups, ask => Ask}.
 
+record_to_list_without_first(Record) ->
+    [ element(I,Record) || I <- lists:seq(2,tuple_size(Record)) ].
```
ludwikbukowski: is this function really related to `mod_roster`? I think it more about data format for GDPR

fenek: I'd even say it's a record abuse. What if record definition changes at some point? Such function should extract desired fields explicitly.

aleklisi: @fenek please provide counterexample of change in record structure that will make this code not work.

aleklisi: Correct me if I am wrong:
According to http://erlang.org/doc/reference_manual/records.html#internal-representation-of-records record is something like `{RecordName, F_1, F_2, ..., F_n}` where RecordName is an atom, n > 0 and `F_1`, `F_2`, ..., `F_n` are Erlang terms.
`record_to_list_without_first/1` function is actually `{RecordName, F_1, F_2, ..., F_n} -> [F1, F2, ..., F_n]` where  `F_1, F_2, ..., F_n` is record structure, so it does not matter what record structure will be given, the function will still work as expected.

fenek: Let's say we have a record `#rec{ a, b, c }` and we are interested in returning a list `[a, b, c]`. OK, `record_to_list_without_first` does what we need. Now someone changes the record definition to `#rec{ a, b, z, c }` because one more field is required. Now this function produces `[a, b, z, c]` while we are still interested in `[a, b, c]`. Without really careful testing, such problem can go unnoticed. If record syntax is used without such hacks, then no extra verification is needed (i.e. we do `#rec{ a = A, b = B, c = C}, [A, B, C]`).
What is more, exploiting the fact that records are just tagged tuples is considered a bad practice (e.g. it's more difficult to find all places in the code where the record is used).

PR #2282, opened 2019-04-23
Gdpr retreive personal data from mod inbox

aleklisi:
This PR allows to retrieve personal data form `mod_inbox` backends.

### DIFF
--- a/big_tests/tests/gdpr_SUITE.erl
+++ b/big_tests/tests/gdpr_SUITE.erl
@@ -19,6 +19,7 @@
          retrieve_pubsub/1,
          retrieve_private_xml/1,
          retrieve_inbox/1,
+         retrieve_inbox_for_multipe_messages/1,
          retrieve_logs/1
         ]).
 -export([
@@ -54,6 +55,7 @@ groups() ->
                                    retrieve_pubsub,
                                    retrieve_private_xml,
                                    retrieve_inbox,
+                                   retrieve_inbox_for_multipe_messages,
                                    retrieve_logs
                                   ]},
     {data_is_not_retrieved_for_missing_user, [],
@@ -114,8 +116,6 @@ end_per_testcase(CN, Config) ->
 
 inbox_required_modules() ->
     [
-     {mod_muc_light, [{host, binary_to_list(muclight_domain())},
-                      {backend, rdbms}]},
      {mod_inbox, inbox_opts()}
     ].
 
@@ -124,10 +124,6 @@ inbox_opts() ->
      {remove_on_kicked, true},
      {groupchat, [muclight]},
      {markers, [displayed]}].
-
-muclight_domain() ->
-    Domain = inbox_helper:domain(),
-    <<"muclight.", Domain/binary>>.
 
 pick_backend_for_mam() ->
     BackendsList = [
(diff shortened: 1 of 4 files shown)


big_tests/tests/gdpr_SUITE.erl
```diff
@@ -113,7 +113,21 @@ end_per_testcase(CN, Config) ->
     escalus:end_per_testcase(CN, Config).
 
 inbox_required_modules() ->
-    [{mod_inbox, []}].
+    [
+     {mod_muc_light, [{host, binary_to_list(muclight_domain())},
```
ludwikbukowski: why is it really added?

big_tests/tests/gdpr_SUITE.erl
```diff
@@ -240,14 +254,17 @@ retrieve_private_xml(Config) ->
...
+            ExpectedHeader = ["jid", "content", "unread_count", "timestamp"],
             ExpectedItems = [
-                             #{ "content" => Body, "jid" => BobJid }
+                             #{ "content" => [{contains, Body}],
+                                "jid" => [{contains, BobS},
+                                          {contains, BobU}],
+                                "unread_count" => "1" }
                             ],
             retrieve_and_validate_personal_data(
               Alice, Config, "inbox", ExpectedHeader, ExpectedItems)
```
ludwikbukowski: Can we assert the inbox of `Bob` as well?

ludwikbukowski:
What do you think for testing scenarios like retrieve inbox for muclight or retrieve inbox for different values of `unread count`?

src/inbox/mod_inbox.erl
```diff
@@ -78,6 +78,33 @@
 
 -export_type([get_inbox_params/0]).
 
+%%--------------------------------------------------------------------
+%% gdpr callbacks
+%%--------------------------------------------------------------------
+
+-behaviour(gdpr).
```
fenek: It breaks the convention where we have behaviour and exports at the top of the file.

src/inbox/mod_inbox.erl
```diff
@@ -78,6 +78,33 @@
...
+    LUser = jid:nodeprep(Username),
+    LServer = jid:nameprep(Server),
+    Schema = ["jid", "content", "unread_count", "timestamp"],
+    InboxParams = #{
+        start => {0,0,0},
+        'end' => erlang:timestamp(),
+        order => asc,
+        hidden_read => false
+       },
+    Record = mod_inbox_backend:get_inbox(LUser, LServer, InboxParams),
```
fenek: Definitely not "record". Either "records" or even better - "entries".

big_tests/tests/gdpr_SUITE.erl
```diff
@@ -290,11 +307,11 @@ retrieve_and_validate_personal_data(Alice, Config, FilePrefix, ExpectedHeader, E
              })
     end.
 
-csv_to_maps(ExpectedHeader, [HeaderRow | [Rows]]) ->
+csv_to_maps(ExpectedHeader, [ExpectedHeader | Rows]) ->
```
fenek: Such fixes should be handled separately, or even directly in base branch, as it affects the logic of all other branches.

big_tests/tests/gdpr_SUITE.erl
aleklisi: works without so I'll remove it

big_tests/tests/gdpr_SUITE.erl
aleklisi: yes, we can

src/inbox/mod_inbox.erl
aleklisi: I'll move to the top of the file.

src/inbox/mod_inbox.erl
aleklisi: I'll rename this variable to `Entries`

PR #2322, opened 2019-05-28
Add `clusterState` metric

mkacper:
This PR addresses **MIM-261**.

I have some doubts regarding this line: https://github.com/esl/MongooseIM/compare/add-cluster-state-metric?expand=1#diff-54e29ccb6566fec2bcf8aad705f9d93eR83, specifically this `tagged` atom. Do you know what it exactly does?

TODO:
* ~~update documentation~~

### DIFF
--- a/big_tests/tests/metrics_api_SUITE.erl
+++ b/big_tests/tests/metrics_api_SUITE.erl
@@ -338,7 +338,8 @@ get_diffs(L1, L2) ->
     lists:zip(L1, L2).
 
 ensure_nodes_not_clustered(Config) ->
-    Nodes = rpc(mim(), mnesia, system_info, [running_db_nodes]),
+    Nodes1 = rpc(mim(), mnesia, system_info, [running_db_nodes]),
+    Nodes = [Node || Node <- Nodes1, Node =/= mim()],
     [distributed_helper:remove_node_from_cluster(N, Config) || N <- Nodes],
     Config ++ [{nodes_clustered, Nodes}].
 
--- a/big_tests/tests/metrics_api_SUITE.erl
+++ b/big_tests/tests/metrics_api_SUITE.erl
@@ -57,7 +57,8 @@ groups() ->
          {metrics, [], ?METRICS_CASES},
          {all_metrics_are_global, [], ?METRICS_CASES},
          {global, [], [session_counters,
-                       node_uptime]}
+                       node_uptime,
+                       cluster_state]}
         ],
     ct_helper:repeat_all_until_all_ok(G).
 
@@ -77,9 +78,15 @@ init_per_group(GroupName, Config) ->
 end_per_group(GroupName, Config) ->
     metrics_helper:finalise_by_all_metrics_are_global(Config, GroupName =:= all_metrics_are_global).
 
+init_per_testcase(cluster_state = CN, Config) ->
+    Config1 = ensure_nodes_not_clustered(Config),
+    escalus:init_per_testcase(CN, Config1);
 init_per_testcase(CaseName, Config) ->
     escalus:init_per_testcase(CaseName, Config).
 
+end_per_testcase(cluster_state = CN, Config) ->
+    Config1 = ensure_nodes_clustered(Config),
+    escalus:end_per_testcase(CN, Config1);
 end_per_testcase(CaseName, Config) ->
     escalus:end_per_testcase(CaseName, Config).
 
@@ -220,6 +227,21 @@ node_uptime(Config) ->
       Y = fetch_global_incrementing_gauge_value(nodeUpTime, Config),
       ?assert_equal_extra(true, Y > X, [{counter, nodeUpTime}, {first, X}, {second, Y}]).
 
+cluster_state(Config) ->
+      SingleNodeClusterState =
+            fetch_global_incrementing_gauge_value(clusterState, Config),
+      ?assert_equal(1, SingleNodeClusterState),
+
+      distributed_helper:add_node_to_cluster(Config),
+      TwoNodesClusterState =
+            fetch_global_incrementing_gauge_value(clusterState, Config),
+      ?assert_equal(2, TwoNodesClusterState),
+
+      distributed_helper:remove_node_from_cluster(Config),
+      SingleNodeClusterState2 =
+            fetch_global_incrementing_gauge_value(clusterState, Config),
+      ?assert_equal(1, SingleNodeClusterState2).
+
 %%--------------------------------------------------------------------
 %% Helpers
 %%--------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


mongoose-im:
[6615.1](https://travis-ci.org/esl/MongooseIM/jobs/538238094) / Erlang 19.3 / small_tests / fb56c593cfc288ba434b82311999360bcc13746d
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2322/6615/small_tests.19.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2322/6615/small_tests.19.3/small/ct_run.mongooseim@localhost.2019-05-28_13.40.27/lib.mongooseim.logs/run.2019-05-28_13.40.28/suite.log.html)

big_tests/tests/metrics_api_SUITE.erl
```diff
@@ -312,3 +334,14 @@ assert_counter_inc(_Name, Inc, Counter1, Counter2) when Counter1 + Inc =:= Count
 
 get_diffs(L1, L2) ->
     lists:zip(L1, L2).
+
+ensure_nodes_not_clustered(Config) ->
+    Nodes = mnesia:system_info(running_db_nodes),
```
erszcz: I think we should not assume that Mnesia will be running in the test environment - it's a separate VM than the servers, why would it have to be running?

I'd rather fetch this information from `test.config`, using something similar to:

```erlang
Hosts = [ {Name, maps:from_list(Spec)}
          || {Name, Spec} <- distributed_helper:get_or_fail(hosts),
             {cluster, mim} <- [lists:keyfind(cluster, 1, Spec)] ],
ct:pal("hosts: ~p", [Hosts]).
```

mkacper: @erszcz You're, of course, right. What do you think about running `mnesia:system_info/1` but via rpc on one of mim nodes?

erszcz: > My intention here was to save the current state of MIM cluster, remove all nodes from the cluster for the sake of this particular test case and return to the same cluster state once the test case is finished.
Definitely makes sense :+1:

mongoose-im:
[6662.1](https://travis-ci.org/esl/MongooseIM/jobs/542137696) / Erlang 19.3 / small_tests / d27316afefda0340852f683f1c4d042ccb732ddc
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2322/6662/small_tests.19.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2322/6662/small_tests.19.3/small/ct_run.mongooseim@localhost.2019-06-06_08.02.34/lib.mongooseim.logs/run.2019-06-06_08.02.35/suite.log.html)

src/mongoose_metrics_definitions.hrl
```diff
@@ -77,6 +77,9 @@
            eval, ?EX_EVAL_SINGLE_VALUE}},
          {nodeUpTime,
           {function, mongoose_metrics, get_up_time, [],
+           tagged, [value]}},
+         {clusterState,
```
rslota: IMHO, we should report two values here: one for `length(nodes())` and the other one for `length(mnesia:system_info(running_db_nodes))`. It's really hard to say which value is more useful as it depends on situation. 
Also, naming. I'm not really sure what to expected from metric named `clusterState`. What about `nodesDetected` for `length(nodes())` and `nodesConnected` for `length(mnesia:system_info(running_db_nodes))`?
cc @fenek

erszcz: > I'm not really sure what to expected from metric named `clusterState`

Indeed, I'd opt for `clusterSize`, since _state_ is somewhat vague. Anyway, this is a superficial point - whatever it's called won't change its usefulness.

> IMHO, we should report two values here: one for `length(nodes())` and the other one for `length(mnesia:system_info(running_db_nodes))`. It's really hard to say which value is more useful as it depends on situation.

We're not talking about an arbitrary Erlang system, but about MongooseIM - MongooseIM doesn't work properly if Mnesia is not clustered. That was the point of the ticket, let's not change scope once the PR is ready.

To give more context, the operation procedure to tell if Mnesia is clustered is to look at the metric graph and see if all cluster members report the same number for `running_db_nodes`. Specifically, we don't want `mongooseimctl debug` nodes to show up on that metric, which could (I'm not 100% sure, since these might be hidden) if we used `length(nodes())`.
I'm not saying that we should not report `length(nodes())`. I'm saying that I don't think it's as useful and even if it is, we should do it as a separate ticket. This one has already taken a month (sic!).

rslota: The usefulness of `length(nodes())` comes when we we do have inconsistency in original metric to distinguish between network issues vs Mnesia issues. 
This is however more tricky then I initially expected due to possible not-hidden debug nodes, so I do agree that this should be part of separate story/task. 

+1 for `clusterSize` :)

arcusfelis: I am for mnesia_running_db_nodes ;)

NelsonVides:
> I have some doubts regarding this line: https://github.com/esl/MongooseIM/compare/add-cluster-state-metric?expand=1#diff-54e29ccb6566fec2bcf8aad705f9d93eR83, specifically this `tagged` atom. Do you know what it exactly does?
That is information for `exometer`, to tell him that reads of this metric should return tagged tuples, with the atom to tag it being `value`.

mongoose-im:
[6751.1](https://travis-ci.org/esl/MongooseIM/jobs/553731478) / Erlang 20.3 / small_tests / 9ca05f2bc0d255ad5abfdb293fefbf8269c3b401
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2322/6751/small_tests.20.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2322/6751/small_tests.20.3/small/ct_run.mongooseim@localhost.2019-07-03_12.30.23/lib.mongooseim.logs/run.2019-07-03_12.30.24/suite.log.html)

big_tests/tests/metrics_api_SUITE.erl
```diff
@@ -77,9 +80,15 @@ init_per_group(GroupName, Config) ->
...
     metrics_helper:finalise_by_all_metrics_are_global(Config, GroupName =:= all_metrics_are_global).
 
+init_per_testcase(cluster_size = CN, Config) ->
+    Config1 = ensure_nodes_not_clustered(Config),
+    escalus:init_per_testcase(CN, Config1);
 init_per_testcase(CaseName, Config) ->
     escalus:init_per_testcase(CaseName, Config).
 
+end_per_testcase(cluster_size = CN, Config) ->
+    Config1 = ensure_nodes_clustered(Config),
```
fenek: Do we need them clustered after the test case?

NelsonVides: This is as Kacper explained before:

> My intention here was to save the current state of MIM cluster, remove all nodes from the cluster for the sake of this particular test case and return to the same cluster state once the test case is finished.

PR #2330, opened 2019-06-04
Gdpr mam muc schema riak

NelsonVides:
This PR addresses riak schema change for gdpr mam muc retrieval

PR #2336, opened 2019-06-13
GDPR data retrival for mod_mam_muc @ Riak DBMS

rslota:
There are two new fields that are indexed in Riak schema:
 * `owner_jid` - (essentially JID of person that "owns" a
   message - for PM that's all messages that include this person,
   for MUC those are all messages that were sent by this person).
 * `mam_type` - we store information about what MAM module has been used
   to save the message. This is not "really" necessary, as we could
   filter messages for GDPR outside database, or prepare different GDPR
   implementations for PM and MUC, but this would result in either
   slower or more complicated code.

The implementation of GDPR for `mod_mam` / `mod_mam_muc` on Riak is now
the same - there are no real differences between those two, other then
`mam_type` field in the query.

This PR closes #2330 as the solution here is effectively alternative approach to the same issue.

### DIFF
--- a/src/mam/mod_mam_riak_timed_arch_yz.erl
+++ b/src/mam/mod_mam_riak_timed_arch_yz.erl
@@ -481,9 +481,9 @@ search_text_filter(SearchText) ->
     <<"search_text_register:", NormText/binary, "~1">>.
 
 jid_filters(LocalJid, undefined) ->
-    <<"_yz_rk:", LocalJid/binary, "*">>;
+    <<"_yz_rk:", LocalJid/binary, "/*/*">>;
 jid_filters(LocalJid, RemoteJid) ->
-    <<"_yz_rk:", LocalJid/binary, "/", RemoteJid/binary, "*">>.
+    <<"_yz_rk:", LocalJid/binary, "/", RemoteJid/binary, "/*">>.
 
 id_filters(undefined, undefined) ->
     undefined;
--- a/src/mam/mod_mam_riak_timed_arch_yz.erl
+++ b/src/mam/mod_mam_riak_timed_arch_yz.erl
@@ -41,7 +41,7 @@
 -export([key/3]).
 
 %% For tests only
--export([create_obj/5, read_archive/7, bucket/1,
+-export([create_obj/6, read_archive/7, bucket/1,
          list_mam_buckets/0, remove_bucket/1]).
 
 -export([get_mam_muc_gdpr_data/2, get_mam_pm_gdpr_data/2]).
@@ -122,9 +122,12 @@ stop_muc_archive(Host) ->
     ejabberd_hooks:delete(mam_muc_remove_archive, Host, ?MODULE, remove_archive, 50),
     ok.
 
+%% LocJID - archive owner's JID
+%% RemJID - interlocutor's JID
+%% SrcJID - "Real" sender JID
 archive_message(_Result, Host, MessId, _UserID, LocJID, RemJID, SrcJID, _Dir, Packet) ->
     try
-        archive_message(Host, MessId, LocJID, RemJID, SrcJID, Packet, pm)
+        archive_message(Host, MessId, LocJID, RemJID, SrcJID, LocJID, Packet, pm)
     catch _Type:Reason ->
             ?WARNING_MSG("Could not write message to archive, reason: ~p",
                          [{Reason, erlang:get_stacktrace()}]),
@@ -132,10 +135,13 @@ archive_message(_Result, Host, MessId, _UserID, LocJID, RemJID, SrcJID, _Dir, Pa
             {error, Reason}
     end.
 
-archive_message_muc(_Result, Host, MessId, _UserID, LocJID, _FromJID, SrcJID, _Dir, Packet) ->
+%% LocJID - MUC/MUC Light room's JID
+%% FromJID - "Real" sender JID
+%% SrcJID - Full JID of user within room (room@domain/user)
+archive_message_muc(_Result, Host, MessId, _UserID, LocJID, FromJID, SrcJID, _Dir, Packet) ->
     RemJIDMuc = maybe_muc_jid(SrcJID),
     try
-        archive_message(Host, MessId, LocJID, RemJIDMuc, SrcJID, Packet, muc)
+        archive_message(Host, MessId, LocJID, RemJIDMuc, SrcJID, FromJID, Packet, muc)
     catch _Type:Reason ->
         ?WARNING_MSG("Could not write MUC message to archive, reason: ~p",
                      [{Reason, erlang:get_stacktrace()}]),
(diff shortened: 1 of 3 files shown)


mongoose-im:
[6681.1](https://travis-ci.org/esl/MongooseIM/jobs/545219334) / Erlang 20.3 / small_tests / 4b3bacf4f1dc5fb8b948e54d54b0a748aef33263
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2336/6681/small_tests.20.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2336/6681/small_tests.20.3/small/ct_run.mongooseim@localhost.2019-06-13_12.26.02/lib.mongooseim.logs/run.2019-06-13_12.26.03/suite.log.html)

mongoose-im:
[6684.1](https://travis-ci.org/esl/MongooseIM/jobs/545252327) / Erlang 20.3 / small_tests / b6534f5a6bfa0a742af67ee924d5d0883ac1f273
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2336/6684/small_tests.20.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2336/6684/small_tests.20.3/small/ct_run.mongooseim@localhost.2019-06-13_15.31.46/lib.mongooseim.logs/run.2019-06-13_15.31.46/suite.log.html)

src/mam/mod_mam_riak_timed_arch_yz.erl
```diff
@@ -203,22 +209,34 @@ remove_bucket(Bucket) ->
...
+%%  * LocJID - From/To JID depending on direction of message (single message is saved in two directions)
+%%  * RemJID - To/From JID depending on direction of message (single message is saved in two directions)
+%%  * SrcJID - "Real" sender JID
+%%  * OwnerJID - Same as LocJID
+%% MUC / MUC Light:
+%%  * LocJID - MUC/MUC Light room's JID
+%%  * RemJID - Nickname of JID of destination
+%%  * SrcJID - Full JID of user within room (room@domain/user)
+%%  * OwnerJID - "Real" sender JID (not room specific)
+archive_message(Host, MessID, LocJID, RemJID, SrcJID, OwnerJID, Packet, Type) ->
```
ludwikbukowski: I have a feeling that if we need to explain what variable means, depending on some context, we could decompose the logic. However, I understand thats implication of working with legacy code

fenek: This would definitely be solved by a refactor (in the future), where we could replace these arguments with parameter map. Indeed now it's confusing as ***.

src/mam/mod_mam_riak_timed_arch_yz.erl
```diff
@@ -122,20 +122,26 @@ stop_muc_archive(Host) ->
     ejabberd_hooks:delete(mam_muc_remove_archive, Host, ?MODULE, remove_archive, 50),
     ok.
 
+%% LocJID - From/To JID depending on direction of message (single message is saved in two directions)
```
fenek: If we're going to explain these here, I think it's more accurate to say:
* `LocJID` - archive owner's JID
* `RemJID` - interlocutor's JID

src/mam/mod_mam_riak_timed_arch_yz.erl
```diff
@@ -122,20 +122,26 @@ stop_muc_archive(Host) ->
...
             ?WARNING_MSG("Could not write message to archive, reason: ~p",
                          [{Reason, erlang:get_stacktrace()}]),
             ejabberd_hooks:run(mam_drop_message, Host, [Host]),
             {error, Reason}
     end.
 
-archive_message_muc(_Result, Host, MessId, _UserID, LocJID, _FromJID, SrcJID, _Dir, Packet) ->
+%% LocJID - MUC/MUC Light room's JID
+%% FromJID - "Real" sender JID
+%% SrcJID - Full JID of user within room (room@domain/user)
```
fenek: `room@domain/nickname`, because `user` may be confused with username

src/mam/mod_mam_riak_timed_arch_yz.erl
```diff
@@ -231,6 +249,10 @@ create_obj(Host, MsgId, SourceJID, Packet, Type) ->
             fun(R) -> riakc_register:set(MsgId, R) end},
            {{<<"source_jid">>, register},
             fun(R) -> riakc_register:set(SourceJID, R) end},
+           {{<<"owner_jid">>, register},
```
fenek: `owner_jid` is kind of ambiguous in MAM context, because it may mean both "message owner" or "archive owner". I think that we should come up with a better name, like `sender_jid` or `real_sender_jid`.

src/mam/mod_mam_riak_timed_arch_yz.erl
rslota: I actually didn't want to use `nickname` as this may also be JID (I think) in case of MUC Light. I agree with the fact that`user` is not the best choice, but `nickname` doesn't feel generic enough.

src/mam/mod_mam_riak_timed_arch_yz.erl
rslota: Yes, indeed this is very confusing. But changing this is this module only will only make things worse, as other MAM backends use exactly the same wording. 
@ludwikbukowski I actually tried to provide descriptions without context, because those arguments seem to be abstract enough (in the end, whole logic of this function doesn't do anything specific depending on context), but I simply failed to figure out the proper wording. Still, I believe that having those definitions here even context-specific is better then not having it and guessing each time we develop an extension.

src/mam/mod_mam_riak_timed_arch_yz.erl
rslota: The problem here is that, this is not "sender's" JID. In case of MUC, yeah, sure. But for PM this archive owner. Would `archive_owner_jid` be better? I guess it's not "correct" for MUC in such case. I feel that, there should be a different "schema" for PM and MUC, it would be much less confusing...

rslota: Actually `msg_owner_jid` feels quite nice. @fenek What do you think?

mongoose-im:
[6686.1](https://travis-ci.org/esl/MongooseIM/jobs/545754088) / Erlang 20.3 / small_tests / f0e193674296317c065296cac622eb21d8a8e61a
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2336/6686/small_tests.20.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2336/6686/small_tests.20.3/small/ct_run.mongooseim@localhost.2019-06-14_13.39.49/lib.mongooseim.logs/run.2019-06-14_13.39.50/suite.log.html)

rslota:
This closes #2330 as the solution here is effectively alternative approach to the same issue.

src/mam/mod_mam_riak_timed_arch_yz.erl
fenek: Yup, `msg_owner_jid` will be accurate IMO. :)

src/mam/mod_mam_riak_timed_arch_yz.erl
fenek: OK, fair enough. :)

mongoose-im:
[6690.1](https://travis-ci.org/esl/MongooseIM/jobs/546618227) / Erlang 20.3 / small_tests / 0bd977ba1f3451c8fbfa89fde1cbd13ac0453f6f
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2336/6690/small_tests.20.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2336/6690/small_tests.20.3/small/ct_run.mongooseim@localhost.2019-06-17_07.53.15/lib.mongooseim.logs/run.2019-06-17_07.53.16/suite.log.html)

rslota:
@fenek Any other comments, or shall we just merge it?

PR #2346, opened 2019-06-25
Add known issues for rel 3.4.0 to docs

fenek:
This PR updates docs with known issues as of release 3.4.0.


doc/operation-and-maintenance/known-issues.md
```diff
@@ -0,0 +1,12 @@
+This document provides a list of all known issues with MongooseIM operation and configuration.
+You may also find proposed workarounds if any is available.
```
DenysGonchar: any are

goddammit: both work

PR #2374, opened 2019-07-17
Metrics/tcp connections rework

NelsonVides:
Taking the work on exometer probes, this is a reimplementation of the open TCP connections metric that risked being overloaded and breaking exometer.


mongoose-im:
[6796.1](https://travis-ci.org/esl/MongooseIM/jobs/559969316) / Erlang 22.0 / small_tests / fd9fbb4d3ba70409d942872bff612d3eea467113
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2374/6796/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2374/6796/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-07-17_15.06.43/lib.mongooseim.logs/run.2019-07-17_15.06.43/suite.log.html)

src/metrics_probes.erl
```diff
@@ -0,0 +1,144 @@
...
+    ref :: reference() | undefined
+}).
+
+-spec behaviour() -> exometer:behaviour().
+behaviour() ->
+    probe.
+
+probe_init(_Name, _Type, Opts) ->
+    FunSampling = proplists:get_value(sampling, Opts, error_no_sampling_given),
+    DataPoints = proplists:get_value(datapoints, Opts, error_no_datapoints_given),
```
rslota: Why won't we crash instead? This code enables us to enter incorrect state.

src/metrics_probes.erl
```diff
@@ -0,0 +1,144 @@
...
+         probe_reset/1,
+         probe_sample/1,
+         probe_setopts/3,
+         probe_handle_msg/2,
+         probe_code_change/3
+        ]).
+
+%% Samples
+-export([
+         do_sample_queue_lengths/0,
```
rslota: How about moving those functions to separate module in order to keep this one just a generic probe calling given MFA?

NelsonVides: I agree with this idea. Two questions: what name should I give to this module, and, should I think about moving the numerous modules related to metrics to a separate `metrics/` folder inside of `src/`? Just to organise the code a bit, or does it make no difference?

src/metrics_probes.erl
NelsonVides: I thought about that for a while, that's why I have weird default values, to see them clearly in the logs. Also, `exometer_probe` will call `probe_sample` as soon as it returns from init, so the crash will come very soon.
How should I make it crash here? `case of badfun -> throw`? I thought of that being defensive, dunno, tell me your opinion or some other idea please :)

src/metrics_probes.erl
rslota: What about `metrics_probes` and `metrics_probes_server`? And, yes, I all for moving all those into a sub-directory. `mongoose_metrics` is already very bloated and I would really like to see it split soon, so that would generate even more files.

src/metrics_probes.erl
rslota: I'm not sure how to tackle it, but I would really suggest somehow crashing here. Exometer in it's API doesn't allow this function to return anything other then `{ok, any()}`, so I guess we need to `error(no_datapoints)` etc.

Also, sampling function could a be function reference here, not an arbitrary atom.

mongoose-im:
[6797.1](https://travis-ci.org/esl/MongooseIM/jobs/560410100) / Erlang 22.0 / small_tests / d9a966a2f1ad4741ba417e16dec607f0cb2d3abe
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2374/6797/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2374/6797/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-07-18_10.26.20/lib.mongooseim.logs/run.2019-07-18_10.26.21/suite.log.html)

big_tests/tests/metrics_api_SUITE.erl
```diff
@@ -247,9 +247,11 @@ cluster_size(Config) ->
       ?assert_equal(1, SingleNodeClusterState2).
 
 tcp_connections_increases_with_user_connection(Config) ->
+    rpc(mim(), exometer, repair, [[global, tcpPortsUsed]]),
```
fenek: Why do we need these?

src/metrics/mongoose_metrics_definitions.hrl
```diff
@@ -82,8 +82,15 @@
...
-          {function, mongoose_metrics, tcp_ports, [],
-           tagged, [value]}}
+          {probe,
+           [{sampling, fun metrics_probes:do_sample_tcp_connections/0},
+            {datapoints, [value]},
+            {sample_interval, timer:seconds(30)}]}},
+         {processQueueLengths,
+          {probe,
+           [{sampling, fun metrics_probes:do_sample_queue_lengths/0},
+            {datapoints, [regular, fsm, total]},
```
fenek: IMO it's not the place to define datapoints because they are hardcoded in the actual sampling implementations anyway.

big_tests/tests/metrics_api_SUITE.erl
NelsonVides: To force exometer to launch the sample function again, otherwise exometer waits for the timeout.

mongoose-im:
[6803.1](https://travis-ci.org/esl/MongooseIM/jobs/562005907) / Erlang 22.0 / small_tests / 40824e884570c62227046180f3ddd2ba3fa71edc
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2374/6803/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2374/6803/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-07-22_09.35.33/lib.mongooseim.logs/run.2019-07-22_09.35.34/suite.log.html)

test/mongooseim_metrics_SUITE.erl
```diff
@@ -91,6 +105,15 @@ tcp_connections_detected(_C) ->
     {ok, [{value, X}]} = mongoose_metrics:get_metric_value(global, tcpPortsUsed),
     ?assert(X > 0).
 
+queued_messages_increase(_C) ->
+    Pids = [spawn(fun() -> receive die -> ok end end) || _ <- lists:seq(1,5)],
+    lists:foreach(fun(Pid) -> Pid ! undefined end, Pids),
+    timer:sleep(60),
```
fenek: I still suggest a synchronous approach:

```erlang
Parent = self(),
Pids = [spawn(fun F() -> receive die -> ok; ack -> Parent ! ack, F() end end) || _ <- lists:seq(1,5)],
lists:foreach(fun(Pid) -> Pid ! undefined, Pid ! ack, receive ack -> ok after 1000 -> ct:fail(timeout) end, Pids)
```

test/mongooseim_metrics_SUITE.erl
```diff
@@ -91,6 +105,15 @@ tcp_connections_detected(_C) ->
...
+    {ok, L} = mongoose_metrics:get_metric_value(global, processQueueLengths),
+    [Pid ! die || Pid <- Pids],
+    X = proplists:get_value(regular, L),
+    ?assert(X >= 5).
```
fenek: We need to validate `fsm` and `total` as well.

test/mongooseim_metrics_SUITE.erl
fenek: Instead of single sleep, let's wrap the verification in `async_helper:wait_until/2`.

test/mongooseim_metrics_SUITE.erl
```diff
@@ -83,6 +84,19 @@ end_per_group(_Name, C) ->
...
 
+init_per_testcase(queued_messages_increase, C) ->
+    exometer:setopts([global, processQueueLengths], [{sample_interval, 50}]),
+    exometer:repair([global, processQueueLengths]),
+    C;
+init_per_testcase(_N, C) ->
+    C.
+
+end_per_testcase(queued_messages_increase, C) ->
+    exometer:setopts([global, processQueueLengths], [{sample_interval, 30000}]),
```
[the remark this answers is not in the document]
NelsonVides: Not necessarily. The situation is a bit tricky: exometer will call `sample` every `sample_interval` time. Then our module implementing the behaviour will call `probe_sample`, which motinor_spawns the sample process _and store the `Ref` for this process in the state_. Next time `probe_sample` is called, it checks if there's a `Ref` assigned: if so, then the previous sample hasn't finished yet and there's no sense in yet launching another one, we already know that the sample has turned into an expensive computation; if `Ref` is undefined, then we can launch another sample. The answer from the sample is given in `handle_msg`, which in turn clears the `Ref` in the probe state.

Now, when we do set the `sample_interval`, exometer launches the sample, but if the previous hasn't finished, then is a noop. Then exometer lanches this sample again next time the timer times off, which in this case is something bigger, so it's fine. We need the `repair` it in the `init_per_testcase` just to ensure that the sample_interval is enforced short when we need it, later on we don't care for the purpose of the test.

src/metrics/mongoose_metrics_probe_queues.erl
```diff
@@ -0,0 +1,45 @@
...
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%==============================================================================
+-module(mongoose_metrics_probe_queues).
+-behaviour(mongoose_metrics_probe).
+
+-include("mongoose_logger.hrl").
```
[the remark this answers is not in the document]
rslota: It should be though. @NelsonVides could you please add log message when metrics probe exits with unknown reason? I know I said I'm gonna add that, but apparently I haven't got change to get to it.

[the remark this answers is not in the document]
NelsonVides: @rslota that's not in this module, this one is only the callback, the logging is done in the behaviour.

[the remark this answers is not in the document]
rslota: Oh, sorry, my bad :)

mongoose-im:
[6844.1](https://travis-ci.org/esl/MongooseIM/jobs/562978471) / Erlang 22.0 / small_tests / 31d1f966326d8c2511e4dad0a6a0f36a5c0c013b
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2374/6844/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2374/6844/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-07-24_12.47.04/lib.mongooseim.logs/run.2019-07-24_12.47.05/suite.log.html)

test/mongooseim_metrics_SUITE.erl
```diff
@@ -91,6 +105,15 @@ tcp_connections_detected(_C) ->
     {ok, [{value, X}]} = mongoose_metrics:get_metric_value(global, tcpPortsUsed),
     ?assert(X > 0).
 
+queued_messages_increase(_C) ->
+    Pids = [spawn(fun() -> receive die -> ok end end) || _ <- lists:seq(1,5)],
+    lists:foreach(fun(Pid) -> Pid ! undefined end, Pids),
+    {ok, {ok, _L}} = async_helper:wait_until(
```
fenek: No need to match for a specific value here as `wait_until` will simply crash on failure.

test/mongooseim_metrics_SUITE.erl
```diff
@@ -91,6 +105,15 @@ tcp_connections_detected(_C) ->
...
+                fun() ->  mongoose_metrics:get_metric_value(global, processQueueLengths) end,
+                {ok, [{fsm, 0}, {regular, 5}, {total, 5}]}
```
fenek: 1. I'm not sure if we should match on the exact order of the fields. I'd apply `lists:sort` on the expected and returned value.
2. It's easy to validate `fsm` value as well - spawned processes have to simply set `$internal_queue_len` dictionary value to something higher than 0.

mongoose-im:
[6891.1](https://travis-ci.org/esl/MongooseIM/jobs/563433253) / Erlang 22.0 / small_tests / 70237a4273854cd62643792be40ab39064da799c
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2374/6891/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2374/6891/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-07-25_07.41.20/lib.mongooseim.logs/run.2019-07-25_07.41.20/suite.log.html)

PR #2402, opened 2019-08-07
Reduce exometer memory usage for dev builds

arcusfelis:
This PR addresses high memory usage of exometer.

Proposed changes include:
* extra_metrics_options argument
* ask exometer to not set its own min_heap_size value.


arcusfelis:
Just use most recent exometer_core, it has https://github.com/Feuerlabs/exometer_core/commit/6615c85c1bee03bf260acb01de8a7adaf71c0514

PR #2409, opened 2019-08-08
Improve mongoose control script

arcusfelis:
This PR addresses "mongoosectl can be better".

Proposed changes include:
* Waiting for status is done by calling "erl -remsh" each second. Travis don't like loading new VMs (slow and cpu intensive for some reason (?))
* Added "mongooseim force_stop", that would kill node, if it's not responding after we asked it to stop.
* Write pid and status files, which can be used instead of calling ps or calling "mongooseim status".
* When testing, we would print `erlang.log`, if starting fails. Useful for debugging case when a node fails to start.


mongoose-im:
[6985.1](https://travis-ci.org/esl/MongooseIM/jobs/569461028) / Erlang 22.0 / small_tests / 68b73b2ab517958a354028fccb1211dd92616b08
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2409/6985/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2409/6985/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-08-08_17.43.27/lib.mongooseim.logs/run.2019-08-08_17.43.28/suite.log.html)

PR #2416, opened 2019-08-12
Storage backends update

Nyco:
Note sure this is very helpful, as it is repeating:
* Added a list of possible RDBMS
* Added Cassandra and ElasricSearch in persistent data
* Added user data

This PR addresses a way to find relevant info


arcusfelis:
Thanks :)

PR #2431, opened 2019-08-23
Ci/tcp metrics

NelsonVides:
Sadly, my tcp_metrics test was unstable. Very rarely, it was failing. This test tries to assert that, on a new user connection, the number of tcp connections increases exactly one. First thing is, it was failing mostly on the preset `mysql_redis`, and never in `internal_mnesia` for example.

What was happening? => _Worker Pools!_

Turns out, they can connect and disconnect at will, so the metric read from before and after the user connection are not stable whatsoever. The whole test was assuming too much, and testing stuff with too many dependencies.

With advise from @chrzaszcz, I decided to remove it from `big_tests` and test what I was supposed to test, the metric: that the metric does detect changes in the number of tcp ports. It was a task for `small_tests`

Next step was, to sync with the async nature of exometer probes, so it was obviously a task for `async_helper:wait_until`. Except for the fact that that function requires exact expectations and it has no support for pattern matching, nor extracting what it was actually returned. So I expanded this function to take a "validator" function object that will validate if retrying or returning the given module, with a default validator of `=:=` to keep compatibility with everything that was using it before.

Once done, I rewrote the old small test to use it and ensure that it returns a non-empty metric, and a new test to ensure that the metric detects with precision the connection and disconnection of a single tcp connection.

### DIFF
--- a/test/mongooseim_metrics_SUITE.erl
+++ b/test/mongooseim_metrics_SUITE.erl
@@ -25,6 +25,7 @@ all_metrics_list() ->
      no_skip_metric,
      subscriptions_initialised,
      tcp_connections_detected,
+     tcp_metric_varies_with_tcp_variations,
      up_time_positive,
      queued_messages_increase
     ].
@@ -84,6 +85,11 @@ end_per_group(_Name, C) ->
     meck:unload(),
     C.
 
+init_per_testcase(CN, C) when tcp_connections_detected =:= CN;
+                              tcp_metric_varies_with_tcp_variations =:= CN ->
+    exometer:setopts([global, tcpPortsUsed], [{sample_interval, 50}]),
+    exometer:repair([global, tcpPortsUsed]),
+    C;
 init_per_testcase(queued_messages_increase, C) ->
     exometer:setopts([global, processQueueLengths], [{sample_interval, 50}]),
     exometer:repair([global, processQueueLengths]),
@@ -99,7 +105,6 @@ init_per_testcase(_N, C) ->
 
 end_per_testcase(queued_messages_increase, C) ->
     [Pid ! die || Pid <- ?config(pids, C)],
-    exometer:setopts([global, processQueueLengths], [{sample_interval, 30000}]),
     C;
 end_per_testcase(_N, C) ->
     C.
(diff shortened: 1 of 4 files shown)


mongoose-im:
[7054.1](https://travis-ci.org/esl/MongooseIM/jobs/575882162) / Erlang 22.0 / small_tests / 29e512d2fc1fe6b45ba086e19f9c113b82618d8e
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2431/7054/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2431/7054/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-08-23_15.24.13/lib.mongooseim.logs/run.2019-08-23_15.24.13/suite.log.html)

test/mongooseim_metrics_SUITE.erl
```diff
@@ -108,9 +113,24 @@ up_time_positive(_C) ->
...
+    X.
+
 tcp_connections_detected(_C) ->
-    {ok, [{value, X}]} = mongoose_metrics:get_metric_value(global, tcpPortsUsed),
-    ?assert(X > 0).
+    get_new_tcp_metric_value({ok, []}).
+
+tcp_metric_varies_with_tcp_variations(_C) ->
+    X = get_new_tcp_metric_value({ok, []}),
+    {ok, Socket} = gen_tcp:listen(1805, []),
```
michalwski: I'm not sure the metric works as documented. In the doc I can see:

>  A number of open tcp connections. This should relate to the number of connected sessions and databases, in order to detect connection leaks. 

And in this test we start listening on a port which also increments the counter. How does it correlated to the what's in the doc?

NelsonVides: That both sessions and databases connect through tcp ports, for example the sessions through `ejabberd_c2s` and `ejabberd_receiver`, and the databases through worker pools (maybe @erszcz can say something more about it?).

michalwski: Yes, and there is more:
* s2s incoming connections
* s2s outgoing connections
* connections for HTTP request MongooseIM sends
* connections for HTTP request MongooseIM serves

Basically any incoming or outgoing TCP connection as well as TCP ports MongooseIM listens on (without any connection at a particular moment). If this is what we want, that's fine and I think the doc could be improved as currently it may be misleading.

PR #2392, opened 2019-07-30
"fix" test in global_distrib_SUITE

zofpolkowska:
First issue:
After sending initial presence to the server's we were not waiting till it is sent back the the client.

Solving the above issue makes the race condition described below even MORE unlikely to happen.
Still, `mod_global_distrib` does not support the feature which is tested in `test_pm_with_ungraceful_reconnection_to_different_server`  and it (the feature) is not trivial to implement. 

When race condition happens:

1. Eve’s process is killed on cluster 1.
2. Alice sends a message to Eve.
3. Eve connects to cluster 2. Entry in Redis from C1 is overwritten by C2.
4. The session/routing table is refreshed on C1. Since Eve’s session is still present there, entry from C2 is overwritten by C1.
5. The message is not delivered to C2, as it is supposed to (it goes to C1 instead).

It fails only if refresh of routing/session table of C1 happens just after the client reconnects.
If we modify `refresh_after` parameter and set it to be significantly longer on C2 than on C1 then the test always fails. 

We should decide what to do with the test (turn it off/delete/leave it) @fenek

PR #2437, opened 2019-09-04
mod_global_distrib_SUITE: ignore known failure, fail otherwise

NelsonVides:
If the test fails with the known reason, simply exit the escalus story and return `{skip, Reason}`. This way, we don't get CI in red any more for the known issue, but if it ever fails with a different one, then CI will get red and we will know about it.

Unfortunately, `escalus:story` ignores the return value of the function given, so the only way to move the control flow is by throwing: it exits the function and gives us a "return value".

Yes, this is kicking the ball down the road. But it's there, and we'll have a chance to work on it if we ever need to come back to global distribution.

I also fixed the suite by ~removing the user that is created in the `ungraceful` test, which was not being removed and then leaving the environment dirty; and~ refactoring the tests by moving not relevant behaviour out of it, into `init_per_testcase` and `end_per_testcase`.

### DIFF
--- a/big_tests/tests/mod_global_distrib_SUITE.erl
+++ b/big_tests/tests/mod_global_distrib_SUITE.erl
@@ -909,9 +909,8 @@ refresh_node(NodeName, Config) ->
 
 connect_from_spec(UserSpec, Config) ->
     {ok, User} = escalus_client:start(Config, UserSpec, <<"res1">>),
-    escalus_story:send_initial_presence(User),
-    escalus:wait_for_stanza(User),
     escalus_connection:set_filter_predicate(User, fun(S) -> not escalus_pred:is_presence(S) end),
+    escalus_story:send_initial_presence(User),
     User.
 
 chat_with_seqnum(To, Text) ->
--- a/big_tests/tests/mod_global_distrib_SUITE.erl
+++ b/big_tests/tests/mod_global_distrib_SUITE.erl
@@ -564,9 +564,10 @@ test_pm_with_disconnection_on_other_server(Config) ->
 
 test_pm_with_graceful_reconnection_to_different_server(Config) ->
     EveSpec = ?config(evespec_mim, Config),
-    escalus:fresh_story(
-      Config, [{alice, 1}],
-      fun(Alice) ->
+    try
+        escalus:fresh_story(
+          Config, [{alice, 1}],
+          fun(Alice) ->
               Eve = connect_from_spec(EveSpec, Config),
 
               escalus_client:send(Eve, escalus_stanza:chat_to(Alice, <<"Hi from Asia!">>)),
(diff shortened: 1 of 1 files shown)


mongoose-im:
[7095.1](https://travis-ci.org/esl/MongooseIM/jobs/580791765) / Erlang 22.0 / small_tests / c164cd58953ca865ca3bec399a23fd9277a7693e
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2437/7095/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2437/7095/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-09-04_15.50.58/lib.mongooseim.logs/run.2019-09-04_15.50.58/suite.log.html)

mongoose-im:
[7096.1](https://travis-ci.org/esl/MongooseIM/jobs/580835167) / Erlang 22.0 / small_tests / 42faf3016c5ead28c31fad99756578047c2d1eec
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2437/7096/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2437/7096/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-09-04_17.33.12/lib.mongooseim.logs/run.2019-09-04_17.33.13/suite.log.html)

big_tests/tests/mod_global_distrib_SUITE.erl
```diff
@@ -909,9 +942,8 @@ refresh_node(NodeName, Config) ->
 
 connect_from_spec(UserSpec, Config) ->
     {ok, User} = escalus_client:start(Config, UserSpec, <<"res1">>),
-    escalus_story:send_initial_presence(User),
-    escalus:wait_for_stanza(User),
```
michalwski: Why don't we wait for the presence reply anymore?

NelsonVides: Because in global distribution they can arrive out of sync and makes our tests a lot more complicated to take that into consideration, when in the tests that are using `connect_from_spec` they're not relevant anyway.
`wait_for_stanza` was actually added recently when these tests were attempted to be fixed before (https://github.com/esl/MongooseIM/pull/2392/files) as a mean to sync a bit the processes, but in the end it didn't bring any improvements, as this adds a bit of sync for `ungraceful` but makes `graceful` a lot more likely to fail. So I decided to get rid of it altogether and focus the tests on what they're trying to test, the messages.

fenek: Are you sure about this? Why would this arrive out of sync, since we're waiting as the same user, as the one who sent the initial presence in the first place?

NelsonVides: https://mongooseim.readthedocs.io/en/latest/modules/mod_global_distrib/#message-routing and as an example http://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/branch/CI/mod_global_distrib/7057/internal_mnesia.22.0/big/ct_run.test@travis-job-5a57dae3-a595-4f45-a4eb-e4b0300aaafe.2019-08-26_10.02.52/big_tests.tests.mod_global_distrib_SUITE.logs/run.2019-08-26_10.02.54/mod_global_distrib_suite.test_pm_with_graceful_reconnection_to_different_server.67.html
It's all in the routing, tables might not be fresh and messages can be routed in circles until they arrive.

fenek:
I don't like hiding this issue but GD has burnt enough of ours.
Besides, most likely it's the test, not the code, as the actual GD user doesn't report any issues.

PR #2461, opened 2019-09-18
Fix IPv6 issue

NelsonVides:
Try if that fixes the IPv6 issue in circleci


arcusfelis:
Merged here https://github.com/esl/MongooseIM/pull/2462

PR #2472, opened 2019-09-23
SM/update_return_old_h

NelsonVides:
In the context of stream resumption, the new XEP (1.6) says as following:

In the **_POSITIVE_** case, which hasn’t changed and is in line with our implementation:
> If the server CAN resume the former stream, it MUST return a `<resumed/>` element, which MUST include a 'previd' attribute set to the SM-ID of the former stream and MUST also include an 'h' attribute set to the sequence number of the last handled stanza sent over the former stream from the client to the server (in the unlikely case that the server never received any stanzas, it would set 'h' to zero).
>Example 12. Stream resumed
```xml
<resumed xmlns='urn:xmpp:sm:3' h='some-number' previd='some-long-sm-id'/>
```

But, in the **_NEGATIVE_** case, there’s a suggested change:

> If the server DOES NOT SUPPORT session resumption, it MUST return a `<failed/>` element, which SHOULD include an error condition of `<feature-not-implemented/>`.
> If the server DOES NOT RECOGNIZE THE 'previd' as an earlier session (e.g., because the former session has timed out), it MUST return a `<failed/>` element, which SHOULD include an error condition of `<item-not-found/>`.
>If the server **_DOES RECOGNIZE_** the 'previd' as an earlier session that has **_TIMED OUT_** the server MAY also include a 'h' attribute indicating the number of stanzas received before the timeout. (Note: For this to work the server has to store the SM-ID/sequence number tuple past the time out of the actual session.)
>Example 13. Stream timed out
```xml
<failed xmlns='urn:xmpp:sm:3' h='some-number'>
    <item-not-found xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/>
</failed>
```
> In both of these failure cases, the server SHOULD allow the client to bind a resource at this point rather than forcing the client to restart the stream negotiation process and re-authenticate.

**_This all means that, when returning the failure for an unsuccessful resumption, we might as well notify of the old h value. That is effectively the only change._**

---

To implement this, now when requesting `mod_stream_management` for the SID of the previous session, given the `SMID`, `mod_stream_management` might return the found `SID` if we’re still within the timeout, just the old `h` if we passed the timeout or lost the information (the XEP asks only for a best-effort policy!), or nothing found.

Accordingly, `ejabberd_c2s` has to either proceed with the resumption if the `SID` is given, or return the error and proceed with the binding if not. This error will have the `h` attribute depending on it being given or not.

So `ejabberd_c2s` basically needs one more pattern matching, and `mod_stream_management` one more table.

---

For this functionality, I needed to pass important data around (`h` and `SMID`, but luckily — or unfortunately, but I think luckily — the functions implicated are all running through hooks, run_folds, and passing Accumulators. Hence, the logic is implemented around these, the Accs.

---

Maybe TODO: implement some `stream_mgmt_stale_h` eviction policy (TTL, garbage collection, timestamps, something like that). <- please give me some feedback!

### DIFF
--- a/big_tests/tests/sm_SUITE.erl
+++ b/big_tests/tests/sm_SUITE.erl
@@ -70,6 +70,7 @@ parallel_manual_ack_test_cases() ->
     [client_acks_more_than_sent,
      too_many_unacked_stanzas,
      resend_unacked_after_resume_timeout,
+     resume_expired_session_returns_correct_h,
      resume_session_state_send_message,
      resume_session_state_stop_c2s,
      server_requests_ack_after_session,
@@ -554,6 +555,37 @@ resend_unacked_after_resume_timeout(Config) ->
     escalus_connection:stop(Bob),
     escalus_connection:stop(NewAlice).
 
+resume_expired_session_returns_correct_h(Config) ->
+    %% connect bob and alice
+    {Bob, _, _, _} =
+        get_stream_resumption_enabled_fresh_user_smid_and_h(Config, bob),
+    {Alice, AliceSpec, SMID, SMH} =
+        get_stream_resumption_enabled_fresh_user_smid_and_h(Config, alice),
+    escalus:assert(is_sm_ack_request, escalus_connection:get_stanza(Alice, ack)),
+    %% Bob sends a message to Alice, and Alice receives it but doesn't acknowledge
+    escalus_connection:send(Bob, escalus_stanza:chat_to(common_helper:get_bjid(AliceSpec), <<"msg-1">>)),
+    escalus:wait_for_stanza(Alice),
+    %% kill alice connection
+    escalus_connection:kill(Alice),
+    %% ensure there is no session
+    wait_until_disconnected(AliceSpec),
+    0 = length(get_user_alive_resources(AliceSpec)),
+    %% alice comes back, but too late, so resumption doesn't work,
+    %% but she receives the previous h = 1 anyway
+    {ok, NewAlice, _} = escalus_connection:start(AliceSpec, connection_steps_to_authenticate()),
+    escalus_connection:send(NewAlice, escalus_stanza:resume(SMID, SMH)),
+    FailedResumption = escalus_connection:get_stanza(NewAlice, failed_resumption),
+    <<"1">> = exml_query:attr(FailedResumption, <<"h">>),
+    %% And we can continue with bind and session
+    escalus_session:session(escalus_session:bind(NewAlice)),
+    escalus_connection:send(NewAlice, escalus_stanza:presence(<<"available">>)),
+    Stanzas = [escalus_connection:get_stanza(NewAlice, msg),
+               escalus_connection:get_stanza(NewAlice, msg)],
+    escalus_new_assert:mix_match([is_presence, is_chat(<<"msg-1">>)], Stanzas),
+    escalus_connection:stop(Bob),
+    escalus_connection:stop(NewAlice).
+
+
 resume_session_state_send_message(Config) ->
     ConnSteps = connection_steps_to_session(),
 
@@ -968,6 +1000,7 @@ messages_are_properly_flushed_during_resumption(Config) ->
         RecvMsg = escalus:wait_for_stanza(Alice2),
         escalus:assert(is_chat_message, [MsgBody], RecvMsg)
       end).
+
 
 %%--------------------------------------------------------------------
 %% Helpers
--- a/big_tests/tests/sm_SUITE.erl
+++ b/big_tests/tests/sm_SUITE.erl
@@ -367,7 +367,8 @@ h_non_given_closes_stream_gracefully(ConfigIn) ->
         escalus:assert(is_stream_error,
                        [<<"policy-violation">>, <<>>],
                        escalus:wait_for_stanza(Alice)),
-        escalus:assert(is_stream_end, escalus_connection:get_stanza(Alice, stream_end))
+        escalus:assert(is_stream_end, escalus_connection:get_stanza(Alice, stream_end)),
+        true = escalus_connection:wait_for_close(Alice,timer:seconds(5))
     end).
 
 client_acks_more_than_sent(Config) ->
--- a/big_tests/tests/sm_SUITE.erl
+++ b/big_tests/tests/sm_SUITE.erl
@@ -186,6 +186,7 @@ end_per_testcase(CN, Config) when CN =:= resume_expired_session_returns_correct_
                                   CN =:= gc_repeat_after_timeout_does_clean
                                    ->
     dynamic_modules:stop(domain(), ?MOD_SM),
+    rpc(mim(), ejabberd_sup, stop_child, [stream_management_stale_h]),
     dynamic_modules:restore_modules(domain(), Config),
     escalus:end_per_testcase(CN, Config);
 end_per_testcase(server_requests_ack_freq_2 = CN, Config) ->
--- a/big_tests/tests/sm_SUITE.erl
+++ b/big_tests/tests/sm_SUITE.erl
@@ -704,7 +704,7 @@ resume_session_with_wrong_h_does_not_leak_sessions(Config) ->
         escalus:assert(is_stream_error, [<<"undefined-condition">>, <<>>], Resumed),
 
         [] = get_user_present_resources(AliceSpec),
-        [] = get_sid_by_stream_id(SMID),
+        smid_not_found = get_sid_by_stream_id(SMID),
         escalus_connection:wait_for_close(Alice, timer:seconds(5))
     end).
 
--- a/big_tests/tests/sm_SUITE.erl
+++ b/big_tests/tests/sm_SUITE.erl
@@ -98,8 +98,9 @@ stream_management_with_stale_h(RepeatAfter, Geriatric) ->
     [
      {ack_freq, 1},
      {resume_timeout, ?SHORT_RESUME_TIMEOUT},
-     {stale_h, {true, [{gc_repeat_after, RepeatAfter},
-                       {gc_geriatric, Geriatric}]}}]}].
+     {stale_h, [{enabled, true},
+                {stale_h_repeat_after, RepeatAfter},
+                {stale_h_geriatric, Geriatric}]}]}].
 
 %%--------------------------------------------------------------------
 %% Init & teardown
@@ -643,7 +644,7 @@ resume_expired_session_returns_correct_h(Config) ->
     escalus_connection:stop(NewAlice).
 
 gc_repeat_after_never_means_no_cleaning(Config) ->
-    true = rpc(mim(), ?MOD_SM, set_gc_repeat_after, [never]),
+    true = rpc(mim(), ?MOD_SM, set_stale_h_repeat_after, [?BIG_BIG_BIG_TIMEOUT]),
     [{SMID1, _}, {SMID2, _}, {SMID3, _}] = ?config(smid_test, Config),
     {stale_h, 1} = rpc(mim(), ?MOD_SM, get_session_from_smid, [SMID1]),
     {stale_h, 2} = rpc(mim(), ?MOD_SM, get_session_from_smid, [SMID2]),
--- a/big_tests/tests/sm_SUITE.erl
+++ b/big_tests/tests/sm_SUITE.erl
@@ -17,6 +17,7 @@
 
 -import(escalus_stanza, [setattr/3]).
 
+-define(BIG_BIG_BIG_TIMEOUT, 30000000000).
 -define(SHORT_RESUME_TIMEOUT, 3).
 -define(SMALL_SM_BUFFER, 3).
 
@@ -157,12 +158,12 @@ register_some_smid_h(Config) ->
     [{smid_test, [S1, S2, S3]} | Config].
 
 init_per_testcase(resume_expired_session_returns_correct_h = CN, Config) ->
-    Config2 = set_gc_parameters(never, never, Config),
+    Config2 = set_gc_parameters(?BIG_BIG_BIG_TIMEOUT, ?BIG_BIG_BIG_TIMEOUT, Config),
     rpc(mim(), ?MOD_SM, set_resume_timeout, [?SHORT_RESUME_TIMEOUT]),
     true = rpc(mim(), ?MOD_SM, set_ack_freq, [1]),
     escalus:init_per_testcase(CN, Config2);
 init_per_testcase(gc_repeat_after_never_means_no_cleaning = CN, Config) ->
-    Config2 = set_gc_parameters(never, 1, Config),
+    Config2 = set_gc_parameters(?BIG_BIG_BIG_TIMEOUT, ?SHORT_RESUME_TIMEOUT, Config),
     Config3 = register_some_smid_h(Config2),
     escalus:init_per_testcase(CN, Config3);
 init_per_testcase(gc_repeat_after_timeout_does_clean = CN, Config) ->
--- a/big_tests/tests/sm_SUITE.erl
+++ b/big_tests/tests/sm_SUITE.erl
@@ -736,7 +736,7 @@ resume_session_with_wrong_h_does_not_leak_sessions(Config) ->
         escalus:assert(is_stream_error, [<<"undefined-condition">>, <<>>], Resumed),
 
         [] = get_user_present_resources(AliceSpec),
-        smid_not_found = get_sid_by_stream_id(SMID),
+        {error, smid_not_found} = get_sid_by_stream_id(SMID),
         escalus_connection:wait_for_close(Alice, timer:seconds(5))
     end).
 
--- a/big_tests/tests/sm_SUITE.erl
+++ b/big_tests/tests/sm_SUITE.erl
@@ -148,14 +148,14 @@ set_gc_parameters(RepeatAfter, Geriatric, Config) ->
       domain(), stream_management_with_stale_h(RepeatAfter, Geriatric)),
     Config2.
 
+register_smid(IntSmidId) ->
+    S = {SMID = make_smid(), IntSmidId},
+    ok = rpc(mim(), ?MOD_SM, register_stale_smid_h, [SMID, IntSmidId]),
+    S.
+
 register_some_smid_h(Config) ->
-    S1 = {SMID1 = make_smid(), 1},
-    S2 = {SMID2 = make_smid(), 2},
-    S3 = {SMID3 = make_smid(), 3},
-    ok = rpc(mim(), ?MOD_SM, register_stale_smid_h, [SMID1, 1]),
-    ok = rpc(mim(), ?MOD_SM, register_stale_smid_h, [SMID2, 2]),
-    ok = rpc(mim(), ?MOD_SM, register_stale_smid_h, [SMID3, 3]),
-    [{smid_test, [S1, S2, S3]} | Config].
+    TestSmids = lists:map(fun register_smid/1, lists:seq(1, 3)),
+    [{smid_test, TestSmids} | Config].
 
 init_per_testcase(resume_expired_session_returns_correct_h = CN, Config) ->
     Config2 = set_gc_parameters(?BIG_BIG_BIG_TIMEOUT, ?BIG_BIG_BIG_TIMEOUT, Config),
@@ -179,8 +179,8 @@ init_per_testcase(replies_are_processed_by_resumed_session = CN, Config) ->
 init_per_testcase(CaseName, Config) ->
     escalus:init_per_testcase(CaseName, Config).
 
-end_per_testcase(CN, Config) when CN =:= resume_expired_session_returns_correct_h,
-                                  CN =:= gc_repeat_after_never_means_no_cleaning,
+end_per_testcase(CN, Config) when CN =:= resume_expired_session_returns_correct_h;
+                                  CN =:= gc_repeat_after_never_means_no_cleaning;
                                   CN =:= gc_repeat_after_timeout_does_clean
                                    ->
     dynamic_modules:stop(domain(), ?MOD_SM),
--- a/big_tests/tests/sm_SUITE.erl
+++ b/big_tests/tests/sm_SUITE.erl
@@ -27,12 +27,14 @@
 all() ->
     [{group, parallel},
      {group, parallel_manual_ack_freq_1},
+     {group, stale_h},
      server_requests_ack_freq_2
      ].
 
 groups() ->
     G = [{parallel, [parallel], parallel_test_cases()},
          {parallel_manual_ack_freq_1, [parallel], parallel_manual_ack_test_cases()},
+         {stale_h, [], stale_h_test_cases()},
          {manual_ack_freq_long_session_timeout, [parallel], [preserve_order]}],
     ct_helper:repeat_all_until_all_ok(G).
 
@@ -70,7 +72,6 @@ parallel_manual_ack_test_cases() ->
     [client_acks_more_than_sent,
      too_many_unacked_stanzas,
      resend_unacked_after_resume_timeout,
-     resume_expired_session_returns_correct_h,
      resume_session_state_send_message,
      resume_session_state_stop_c2s,
      server_requests_ack_after_session,
@@ -78,9 +79,26 @@ parallel_manual_ack_test_cases() ->
      server_requests_ack
      ].
 
+stale_h_test_cases() ->
+    [
+     resume_expired_session_returns_correct_h,
+     gc_repeat_after_never_means_no_cleaning,
+     gc_repeat_after_timeout_does_clean
+    ].
 
 suite() ->
     require_rpc_nodes([mim]) ++ escalus:suite().
+
+domain() ->
+    ct:get_config({hosts, mim, domain}).
+
+stream_management_with_stale_h(RepeatAfter, Geriatric) ->
+  [{mod_stream_management,
+    [
+     {ack_freq, 1},
+     {resume_timeout, ?SHORT_RESUME_TIMEOUT},
+     {stale_h, {true, [{gc_repeat_after, RepeatAfter},
+                       {gc_geriatric, Geriatric}]}}]}].
 
 %%--------------------------------------------------------------------
 %% Init & teardown
@@ -99,12 +117,15 @@ end_per_suite(Config) ->
     escalus_fresh:clean(),
     escalus:end_per_suite(NewConfig1).
 
+
 init_per_group(manual_ack_freq_long_session_timeout, Config) ->
     true = rpc(mim(), ?MOD_SM, set_ack_freq, [1]),
     escalus_users:update_userspec(Config, alice, manual_ack, true);
 init_per_group(parallel_manual_ack_freq_1, Config) ->
     true = rpc(mim(), ?MOD_SM, set_ack_freq, [1]),
     rpc(mim(), ?MOD_SM, set_resume_timeout, [?SHORT_RESUME_TIMEOUT]),
+    escalus_users:update_userspec(Config, alice, manual_ack, true);
+init_per_group(stale_h, Config) ->
     escalus_users:update_userspec(Config, alice, manual_ack, true);
 init_per_group(_GroupName, Config) ->
     Config.
@@ -119,6 +140,35 @@ end_per_group(parallel_manual_ack_freq_1, Config) ->
 end_per_group(_GroupName, Config) ->
     Config.
 
+
+set_gc_parameters(RepeatAfter, Geriatric, Config) ->
+    Config2 = dynamic_modules:save_modules(domain(), Config),
+    dynamic_modules:ensure_modules(
+      domain(), stream_management_with_stale_h(RepeatAfter, Geriatric)),
+    Config2.
+
+register_some_smid_h(Config) ->
+    S1 = {SMID1 = make_smid(), 1},
+    S2 = {SMID2 = make_smid(), 2},
+    S3 = {SMID3 = make_smid(), 3},
+    ok = rpc(mim(), ?MOD_SM, register_stale_smid_h, [SMID1, 1]),
+    ok = rpc(mim(), ?MOD_SM, register_stale_smid_h, [SMID2, 2]),
+    ok = rpc(mim(), ?MOD_SM, register_stale_smid_h, [SMID3, 3]),
+    [{smid_test, [S1, S2, S3]} | Config].
+
+init_per_testcase(resume_expired_session_returns_correct_h = CN, Config) ->
+    Config2 = set_gc_parameters(never, never, Config),
+    rpc(mim(), ?MOD_SM, set_resume_timeout, [?SHORT_RESUME_TIMEOUT]),
+    true = rpc(mim(), ?MOD_SM, set_ack_freq, [1]),
+    escalus:init_per_testcase(CN, Config2);
+init_per_testcase(gc_repeat_after_never_means_no_cleaning = CN, Config) ->
+    Config2 = set_gc_parameters(never, 1, Config),
+    Config3 = register_some_smid_h(Config2),
+    escalus:init_per_testcase(CN, Config3);
+init_per_testcase(gc_repeat_after_timeout_does_clean = CN, Config) ->
+    Config2 = set_gc_parameters(?SHORT_RESUME_TIMEOUT, ?SHORT_RESUME_TIMEOUT, Config),
+    Config3 = register_some_smid_h(Config2),
+    escalus:init_per_testcase(CN, Config3);
 init_per_testcase(server_requests_ack_freq_2 = CN, Config) ->
     true = rpc(mim(), ?MOD_SM, set_ack_freq, [2]),
     escalus:init_per_testcase(CN, Config);
@@ -128,6 +178,13 @@ init_per_testcase(replies_are_processed_by_resumed_session = CN, Config) ->
 init_per_testcase(CaseName, Config) ->
     escalus:init_per_testcase(CaseName, Config).
 
+end_per_testcase(CN, Config) when CN =:= resume_expired_session_returns_correct_h,
+                                  CN =:= gc_repeat_after_never_means_no_cleaning,
+                                  CN =:= gc_repeat_after_timeout_does_clean
+                                   ->
+    dynamic_modules:stop(domain(), ?MOD_SM),
+    dynamic_modules:restore_modules(domain(), Config),
+    escalus:end_per_testcase(CN, Config);
 end_per_testcase(server_requests_ack_freq_2 = CN, Config) ->
     true = rpc(mim(), ?MOD_SM, set_ack_freq, [never]),
     escalus:end_per_testcase(CN, Config);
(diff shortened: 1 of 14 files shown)


NelsonVides:
My ideas for the eviction policy thing, if we want it.
* The simplest one: put timestamps on the records in mnesia, and have a process in the background removing records older than a given timelapse, launching this "garbage collection" every other timelapse, where both parameters can be configured in `mongooseim.cfg` (in such case, please suggest default values).
* A more tricky one, kinda cool: Keep a fixed amount of ETS tables with a fixed capacity for records (both also configured in `mongooseim.cfg`), and simply fill them. When the last one is filled, in order to add the next record then simply clear the oldest table entirely and put it at the front of the table list, and keep filling that one. Might have better performance when it comes to memory management.
Both options are very easily tweak-able, so please add all sorts of inputs, add a third option, or vote for no eviction. Waiting for all your thoughts 😃

src/ejabberd_sm.erl
```diff
@@ -364,17 +366,19 @@ unset_presence(Acc, SID, User, Server, Resource, Status, Info) ->
...
-      Reason :: close_reason().
-close_session_unset_presence(SID, User, Server, Resource, Status, Reason) ->
-    close_session(SID, User, Server, Resource, Reason),
+      Reason :: close_reason(),
+      Acc1 :: mongoose_acc:t().
+close_session_unset_presence(Acc, SID, User, Server, Resource, Status, Reason) ->
+    close_session(Acc, SID, User, Server, Resource, Reason),
     LServer = jid:nameprep(Server),
-    ejabberd_hooks:run(unset_presence_hook, LServer,
+    ejabberd_hooks:run_fold(unset_presence_hook, LServer, Acc,
```
michalwski: Do we want to pass the `Acc1` we get from calling `close_session/6` two lines above?

src/ejabberd_sm_mnesia.erl
```diff
@@ -103,7 +103,12 @@ cleanup(Node) ->
                          ['$_']}]),
                 lists:foreach(fun(#session{ usr = {U, S, R}, sid = SID }) ->
                                       mnesia:delete({session, SID}),
-                                      ejabberd_hooks:run(session_cleanup, S, [U, S, R, SID])
+                                      ejabberd_hooks:run_fold(session_cleanup, S,
+                                                              mongoose_acc:new(
```
michalwski: The code could be a bit more readable here if we pass the `Acc` as a variable and construct it just belowe `mnesia:delete`.

src/ejabberd_sm_redis.erl
```diff
@@ -125,7 +125,12 @@ cleanup(Node) ->
                           %% Add possible removed ":" from encoded SID
                           SID = binary_to_term(ejabberd_binary:join(SIDEncoded, <<":">>)),
                           delete_session(SID, U, S, R),
-                          ejabberd_hooks:run(session_cleanup, S, [U, S, R, SID])
+                          ejabberd_hooks:run_fold(session_cleanup, S,
+                                                  mongoose_acc:new(
```
michalwski: Similar not as for `ejabberd_sm_mnesia`

src/mod_stream_management.erl
```diff
@@ -59,19 +73,43 @@ sm() ->
...
+      Acc :: mongoose_acc:t(),
+      SID :: ejabberd_sm:sid(),
+      JID :: undefined | jid:jid(),
+      Info :: undefined | [any()],
+      Reason :: undefined | ejabberd_sm:close_reason(),
+      Acc1 :: mongoose_acc:t().
 remove_smid(Acc, SID, _JID, _Info, _Reason) ->
-    case mnesia:dirty_index_read(sm_session, SID, #sm_session.sid) of
+    H = mongoose_acc:get(stream_mgmt, h, undefined, Acc),
+    MaybeSMID = case mnesia:dirty_index_read(sm_session, SID, #sm_session.sid) of
```
michalwski: It doesn't look good to me. If there is no such session we use value `ok` otherwise use the real `SMID`. How about converting this code into a function and returning `{error, not_found}` and `{ok, SMID}` and setting these in the Acc?

src/mod_stream_management.erl
```diff
@@ -119,6 +157,34 @@ set_resume_timeout(ResumeTimeout) ->
 %% API for `ejabberd_c2s'
 %%
 
+-spec make_smid() -> smid().
+make_smid() ->
+    base64:encode(crypto:strong_rand_bytes(21)).
+
+%% Getters
+-spec get_session_from_smid(SMID :: smid()) ->
+    ejabberd_sm:sid() | non_neg_integer() | smid_not_found.
```
michalwski: The return value is quite confusing. As far as I understand we need sth like this:
`{ok, ejabberd_sm:sid()} | {smid_not_found, maybe_stale_h()}`

NelsonVides:
> Thanks for the effort @NelsonVides! I've put some comments to the code. Here are my notes regarding stale h / eviction:
> 
>     1. I'd make this functionality configurable and disabled by default. As far as I get it, it is no mandatory for the server to implement this.
> 
>     2. If I'm correct we need this info on all nodes in the cluster, this tells me it will be easier to implement on top of mnesia rather than bare ETS tables.
> 
>     3. IMO this stale h parameter is needed shortly after the waiting session times out. If so I think we don't need to store this for longer than 30-60 minutes.

Great review, @michalslaski, thank you, will apply everything tomorrow morning from work! :)

About the stale h value, I suppose you're then for the option nr1 in https://github.com/esl/MongooseIM/pull/2472#issuecomment-534128074? Put a timestamp on the records and have a backgroun garbage collector running on timeouts? When I said about ETS tables, yeah, mnesia tables is granted much easier, but for that matter the idea is still the same: keep a finite circular list of tables and evict entire tables at once. If we want to nevertheless garbage-collect on timeouts, we can still do, simply evict all tables whose youngest value is older than the configured threshold. The idea is that implementing a GC on timeouts seems very very easy to me, but, does it have the best performance? Or it doesn't matter?

mongoose-im:
[7236.1](https://travis-ci.org/esl/MongooseIM/jobs/588614237) / Erlang 22.0 / small_tests / 7deac6eaa25ec73d477a725c20a690e26507b752
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2472/7236/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2472/7236/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-09-23_20.57.55/lib.mongooseim.logs/run.2019-09-23_20.57.56/suite.log.html)

src/ejabberd_sm.erl
NelsonVides: It's not strictly necessary, but it makes it uglily inconsistent not to. Thanks for the catch!

src/mod_stream_management.erl
arcusfelis: Or something like: `#{sid => Sid} | #{stale_h => Stale} | #{}`

src/mod_stream_management.erl
```diff
@@ -128,12 +194,16 @@ register_smid(SMID, SID) ->
...
-get_sid(SMID) ->
-    case mnesia:dirty_read(sm_session, SMID) of
-        [#sm_session{sid = SID}] ->
-            [SID];
+register_stale_smid_h(SMID, H) ->
+    mnesia:sync_dirty(fun mnesia:write/1,
+                      [#stream_mgmt_stale_h{smid = SMID, h = H}]).
+
+remove_stale_smid_h(SMID) ->
+    case mnesia:dirty_read(stream_mgmt_stale_h, SMID) of
```
arcusfelis: `mnesia:dirty_delete({Tab, Key})` can be used here instead

src/mod_stream_management.erl
```diff
@@ -59,19 +73,43 @@ sm() ->
...
+            SMID
     end,
-    Acc.
+    mongoose_acc:set(stream_mgmt, smid, MaybeSMID, Acc).
 
 -spec session_cleanup(Acc :: map(), LUser :: jid:luser(), LServer :: jid:lserver(),
                       LResource :: jid:lresource(), SID :: ejabberd_sm:sid()) -> any().
 session_cleanup(Acc, _LUser, _LServer, _LResource, SID) ->
-    remove_smid(Acc, SID, undefined, undefined, undefined).
+    Acc1 = remove_smid(Acc, SID, undefined, undefined, undefined),
```
arcusfelis: You can make a separate non-hook handler function do_remove_smid with less arguments and call it instead from here:

```erlang
remove_smid(Acc, SID, _JID, _Info, _Reason) ->
    do_remove_smid(Acc, SID).

do_remove_smid(Acc, SID) ->
     ....
```

src/mod_stream_management.erl
NelsonVides: Great idea 🙂

mongoose-im:
[7241.1](https://travis-ci.org/esl/MongooseIM/jobs/588851221) / Erlang 22.0 / small_tests / 5d4ce2f3d1b98201f316c07bf6f62a876b59506e
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2472/7241/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2472/7241/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-09-24_09.01.31/lib.mongooseim.logs/run.2019-09-24_09.01.31/suite.log.html)

michalwski:
@NelsonVides, yes I'm for option nr1. This data will not be the most frequently accused, only if the resumption fails because the previous session timed out. A background process or `timer:apply_interval` traversing these table with combination of `mnesia:first`, `mnesia:next` will probably not harm.

src/mod_stream_management.erl
```diff
@@ -59,19 +73,48 @@ sm() ->
...
+    do_remove_smid(Acc, SID).
 
 -spec session_cleanup(Acc :: map(), LUser :: jid:luser(), LServer :: jid:lserver(),
                       LResource :: jid:lresource(), SID :: ejabberd_sm:sid()) -> any().
 session_cleanup(Acc, _LUser, _LServer, _LResource, SID) ->
-    remove_smid(Acc, SID, undefined, undefined, undefined).
+    Acc1 = do_remove_smid(Acc, SID),
+    case mongoose_acc:get(stream_mgmt, smid, Acc1) of
+        {error, smid_not_found} -> ok;
+        {ok, SMID} -> remove_stale_smid_h(SMID)
```
fenek: The aim of `session_cleanup` is to remove session entries bound to a MIM node that died/crashed/shutdown. Is it even possible to have stale SMID h for a session that is currently alive?

src/mod_stream_management.erl
```diff
@@ -20,13 +20,23 @@
          set_resume_timeout/1]).
 
 %% API for `ejabberd_c2s'
--export([register_smid/2,
-         get_sid/1]).
+-export([
```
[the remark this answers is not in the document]
NelsonVides: Yeah kind of. But they might be useful to keep exported when I have to build the test for the garbage collecting thing. I'm leaving this comment here for now, we'll see later.

mongoose-im:
[7274.1](https://travis-ci.org/esl/MongooseIM/jobs/589494795) / Erlang 22.0 / small_tests / 68aea291493b981c106d26053f390a4204fb11b2
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2472/7274/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2472/7274/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-09-25_15.16.02/lib.mongooseim.logs/run.2019-09-25_15.16.02/suite.log.html)

mongoose-im:
[7275.1](https://travis-ci.org/esl/MongooseIM/jobs/589528190) / Erlang 22.0 / small_tests / ec851b3ff24450804812bc38f78da4ba8685cce7
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2472/7275/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2472/7275/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-09-25_16.13.29/lib.mongooseim.logs/run.2019-09-25_16.13.29/suite.log.html)

src/mod_stream_management.erl
```diff
@@ -128,17 +293,74 @@ register_smid(SMID, SID) ->
...
+    end.
+
+remove_stale_smid_h(SMID) ->
+    case gen_mod:get_module_opt(?MYNAME, ?MODULE, stale_h, {false, []}) of
+        {false, []} -> ok;
+        {true, _} -> mnesia:dirty_delete(stream_mgmt_stale_h, SMID)
     end.
 
+%%
+%% gen_server
```
michalwski: Two things regarding the gen_server:
1. I think it'd be better to move it to a separate module.
2. In cluster setup we'll have such a process on every node in the cluster. Did you check what happens if we run the `select_delete` in the same time? I know the chances are rather low.

src/mod_stream_management.erl
```diff
@@ -128,17 +293,74 @@ register_smid(SMID, SID) ->
...
+    {noreply, State, RepeatAfter};
+handle_info(_Info, #state{gc_repeat_after = RepeatAfter,
+                          gc_geriatric = _GeriatricAge} = State) ->
+    {noreply, State, RepeatAfter}.
+
 %%
 %% Helpers
 %%
+-spec now_to_seconds(erlang:timestamp()) -> non_neg_integer().
+now_to_seconds({MegaSecs, Secs, _MicroSecs}) ->
```
aleklisi: How about renaming `now_to_seconds` to `timestamp_to_seconds`?
Btw it looks very similar to http://erlang.org/doc/man/calendar.html#time_to_seconds-1 ;)

src/mod_stream_management.erl
```diff
@@ -128,17 +293,74 @@ register_smid(SMID, SID) ->
...
-            [SID];
-        [] ->
-            []
+register_stale_smid_h(SMID, H) ->
+    case gen_mod:get_module_opt(?MYNAME, ?MODULE, stale_h, {false, []}) of
+        {false, []} -> ok;
+        {true, _} ->
+            try
+                Stamp = now_to_seconds(erlang:timestamp()),
+                mnesia:sync_dirty(fun mnesia:write/1,
```
aleklisi: How about just using:
http://erlang.org/doc/man/mnesia.html#dirty_write-1
to simplify the code a bit?

big_tests/tests/sm_SUITE.erl
```diff
@@ -127,6 +179,13 @@ init_per_testcase(replies_are_processed_by_resumed_session = CN, Config) ->
 init_per_testcase(CaseName, Config) ->
     escalus:init_per_testcase(CaseName, Config).
 
+end_per_testcase(CN, Config) when CN =:= resume_expired_session_returns_correct_h,
```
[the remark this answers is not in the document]
NelsonVides: Very strange, in the case of this function not being called, the tests should have then failed, as we're not stopping the module and so on. Cool xD

src/ejabberd_c2s.erl
```diff
@@ -2879,10 +2881,13 @@ stream_mgmt_enabled(ExtraAttrs) ->
            attrs = [{<<"xmlns">>, ?NS_STREAM_MGNT_3}] ++ ExtraAttrs}.
 
 stream_mgmt_failed(Reason) ->
+    stream_mgmt_failed(Reason, []).
+
+stream_mgmt_failed(Reason, Attrs) ->
```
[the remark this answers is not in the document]
NelsonVides: Yeap, there are old calls to arity one, and there's one new call that I'm adding with arity 2, in https://github.com/esl/MongooseIM/blob/bb2bf711e4320667d416f37d0c8bfb31ffecf31b/src/ejabberd_c2s.erl#L3076

src/mod_stream_management.erl
```diff
@@ -115,10 +202,88 @@ get_resume_timeout(Default) ->
...
+    end.
+
+
 %%
 %% API for `ejabberd_c2s'
 %%
 
+-spec make_smid() -> smid().
+make_smid() ->
+    base64:encode(crypto:strong_rand_bytes(21)).
```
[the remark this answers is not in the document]
NelsonVides: I took that function just as it was from `ejabberd_c2s`. You might notice on the file changes that that piece of code is now missing from `ejabberd_c2s` and was moved to `mod_stream_management`, precisely because I think it is the module responsible for generating these kind of things. Whether it is the best or not, well, it is just as it was. Maybe optimising/refactoring is out of the scope of this already super big PR 🤔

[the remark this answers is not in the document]
NelsonVides: About the `make_ref()`, not either, because this produced value is given to the client, and the client will save it and later reuse it when he requests resumption -> that's how resumption works, when the client requests so, he provides this "unique stream management identifier".

On the other hand, `base64:encode(erlang:ref_to_list(make_ref()))` does indeed generate a pretty usable random string. The problem is that it doesn't promise really unicity. Actually, `make_ref()` can reuse references if it can ensure that the previous one was garbage collected and not used any more, in which case, it's legit. Anyway, I don't know how that behaves when it comes to nodes on clusters and so on. We can think about that if we ever get a complain about `crypto:strong_rand_bytes(21)` duplicating keys 😛

src/mod_stream_management.erl
```diff
@@ -59,19 +122,43 @@ sm() ->
...
+
+-spec do_remove_smid(Acc, SID) -> Acc1 when
+      Acc :: mongoose_acc:t(),
+      SID :: ejabberd_sm:sid(),
+      Acc1 :: mongoose_acc:t().
+do_remove_smid(Acc, SID) ->
+    H = mongoose_acc:get(stream_mgmt, h, undefined, Acc),
+    MaybeSMID = case mnesia:dirty_index_read(sm_session, SID, #sm_session.sid) of
+        [] -> {error, smid_not_found};
+        [#sm_session{smid = SMID}] ->
```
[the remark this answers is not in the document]
NelsonVides: Nope, they're generated by `crypto:strong_rand_bytes(21)`. If we get a second one then `crypto` is most likely broken. We'll see about that when it happens xD

src/mod_stream_management.erl
```diff
@@ -128,17 +293,74 @@ register_smid(SMID, SID) ->
...
+handle_cast(_Msg, State) ->
+    {noreply, State}.
+
+handle_info(timeout, #state{gc_repeat_after = RepeatAfter,
+                            gc_geriatric = GeriatricAge} = State) ->
+    TimeToDie = now_to_seconds(erlang:timestamp()) + GeriatricAge,
+    ets:select_delete(
+      stream_mgmt_stale_h,
+      ets:fun2ms(
+        fun(#stream_mgmt_stale_h{stamp=S}) when S < TimeToDie -> true end
```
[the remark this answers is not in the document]
NelsonVides: The matchspec is created at compile time, that fun2ms is given to the compiler on a parse transform, and the compiler rewrites the code accordingly. So no worries 🙂

src/mod_stream_management.erl
NelsonVides: Rename, cool. Similarity, that's saying about time from midnight, which doesn't work, it would break if we generate a key at 23:59 and garbage collect at 00:01, but having a geriatric age of two hours 😛

src/mod_stream_management.erl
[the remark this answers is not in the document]
NelsonVides: The matchspec is created at compile time, that fun2ms is given to the compiler on a parse transform, and the compiler rewrites the code accordingly. So no worries 🙂

mongoose-im:
[7277.1](https://travis-ci.org/esl/MongooseIM/jobs/589929356) / Erlang 22.0 / small_tests / 091ba180a4610dc9648089d604c2899a7ee48860
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2472/7277/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2472/7277/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-09-26_12.23.50/lib.mongooseim.logs/run.2019-09-26_12.23.50/suite.log.html)

src/ejabberd_c2s.erl
```diff
@@ -1503,6 +1505,7 @@ terminate(_Reason, StateName, StateData) ->
                             children = [StatusEl]},
             Acc0 = element_to_origin_accum(Packet, StateData),
             ejabberd_sm:close_session_unset_presence(
+              Acc,
```
arcusfelis: why not to use Acc0 here instead?

arcusfelis: maybe for simplicity :) 
fine now

src/stream_management/mod_stream_management_stale_h.erl
```diff
@@ -0,0 +1,136 @@
+-module(mod_stream_management_stale_h).
+-behaviour(gen_server).
+
+-include("mongoose.hrl").
+-include_lib("stdlib/include/ms_transform.hrl").
+
+-record(state,
```
arcusfelis: Try to avoid introducing new state records.
Basically, it's good to do prefixes to state records i.e. `#smsh_state{}`.

Reasoning: sometimes we need to load records into shell, duplicates are confusing.

doc/modules/mod_stream_management.md
```diff
@@ -1,21 +1,100 @@
...
 ### Options
 
 * `buffer_max` (default: 100): Buffer size for messages yet to be acknowledged.
-* `ack_freq` (default: 1): Frequency of ack requests sent from the server to the client, e.g. 1 means a request after each stanza, 3 means a request after each 3 stanzas.
-* `resume_timeout` (default: 600): Timeout for the session resumption. Sessions will be removed after the specified number of seconds.
+* `ack_freq` (default: 1): Frequency of ack requests sent from the server to the client, e.g. 1
+  means a request after each stanza, 3 means a request after each 3 stanzas.
+* `resume_timeout` (default: 600): Timeout for the session resumption. Sessions will be removed
+  after the specified number of seconds.
+* `stale_h`: enable keeping old server's `<h>` values after the resumption timed out. Defaults to
```
arcusfelis: We try to avoid non-tagged options (i.e. don't use tuples for options, because it would make harder implementing better config).

Good alternatives for you to choose from:
- use flat options: i.e. stale_h_enabled, stale_h_gc_repeat_after, stale_h_gc_geriatric.
- use {stale_h_opts, [{enabled, true}, {stale_h_gc_repeat_after, ...}, ...}]}
- use stale_h_opts and stale_h_enabled
- assume if stale_h_opts are defined, stale_h is enabled.

doc/modules/mod_stream_management.md
```diff
@@ -1,21 +1,100 @@
...
+        }).
+```
+
+Where `smid` is a unique identifier — in this case a random binary, and `sid` is an opaque session
+identifier from `ejabberd_sm`, which is needed to find the previous session we want to resume from.
+This module implements hooks that run on connection removals and session cleanups, in order to clean
+records from a dying session; and it also implements registration callbacks, used in `ejabberd_c2s`
+when a session is registered for resumption.
+
+In order to be compliant with the XEP version 1.6, where a server can do the best effort to give the
```
goddammit: XEP version 1.6 requires the server to attempt giving the user the value of the server's `<h>` when a session timed out and cannot be resumed anymore.
To be compliant with it, there's a second optional table:

alt:

XEP version 1.6 requires the server to give the user the value of the server's `<h>` when a session timed out and cannot be resumed anymore.
To be compliant with it, there's a second optional table:

src/ejabberd_c2s.erl
```diff
@@ -679,8 +679,7 @@ maybe_do_compress(El = #xmlel{name = Name, attrs = Attrs}, NextState, StateData)
     end.
 
 check_compression_auth(El, NextState, StateData) ->
```
arcusfelis: You've triggered meaningless changes. We prefer to match records in function clauses:

```erlang
check_compression_auth(El, NextState, StateData = #state{authenticated = false}) ->
    send_element_from_server_jid(StateData, compress_setup_failed()),
    fsm_next_state(NextState, StateData);
check_compression_auth(El, NextState, StateData) ->
    check_compression_method(El, NextState, StateData).
```

doc/modules/mod_stream_management.md
```diff
@@ -1,21 +1,100 @@
...
+         h :: non_neg_integer(),
+         stamp :: non_neg_integer()
+        }).
+```
+
+This table is created, together with a `gen_server` responsible for cleaning up the tables, when
+`stale_h` is set to true with the proper garbage collection configuration. Then, when removing a
+record from the `sm_session` table (which happens when the state of the previous session is also
+dropped), a new record is added to this new table with the `smid` and `h` values of the dropped
+session, together with a timestamp. Then, when a new session attempting resumption queries
```
goddammit:
```suggestion
session, together with a timestamp. Next, when a new session attempting resumption queries
```

doc/modules/mod_stream_management.md
```diff
@@ -1,21 +1,100 @@
...
+`mod_stream_management` for the data behind a `smid`, `mod_stream_management` can answer one of the
```
goddammit:
```suggestion
`mod_stream_management` for the data behind a `smid`, `mod_stream_management` can answer with one of the
```

src/stream_management/mod_stream_management_stale_h.erl
```diff
@@ -0,0 +1,136 @@
...
+        end
+    catch exit:_Reason ->
+              {error, smid_not_found}
+    end.
+
+-spec write_stale_h(SMID :: mod_stream_management:smid(), H :: non_neg_integer()) ->
+    ok | {error, any()}.
+write_stale_h(SMID, H) ->
+    try
+        Stamp = timestamp_to_seconds(erlang:timestamp()),
```
arcusfelis: you can ask erlang directly for seconds, without asking for more precise timestamp and casting.

You can use new timestamp API:
```erlang
erlang:system_time(seconds).
```

But we use compat module still:
```erlang
p1_time_compat:system_time(seconds).
```

src/stream_management/mod_stream_management_stale_h.erl
```diff
@@ -0,0 +1,136 @@
...
+        mnesia:dirty_write(
```
arcusfelis: fits on one line `mnesia:dirty_write(#stream_mgmt_stale_h{smid = SMID, h = H, stamp = Stamp})`

src/stream_management/mod_stream_management_stale_h.erl
```diff
@@ -0,0 +1,136 @@
...
+    {reply, ok, State}.
+
+handle_cast(Msg, State) ->
+    ?WARNING_MSG("unexpected message ~p", [Msg]),
+    {noreply, State}.
+
+handle_info(timeout, #state{gc_repeat_after = RepeatAfter,
+                            gc_geriatric = GeriatricAge} = State) ->
+    TimeToDie = timestamp_to_seconds(erlang:timestamp()) + GeriatricAge,
+    ets:select_delete(
```
arcusfelis: Before:
```
    ets:select_delete(
      stream_mgmt_stale_h,
      ets:fun2ms(
        fun(#stream_mgmt_stale_h{stamp=S}) when S < TimeToDie -> true end
       )
     ) % We don't use new lines for parenthesis 
```
After:
```
    MS = ets:fun2ms(fun(#stream_mgmt_stale_h{stamp=S}) when S < TimeToDie -> true end),
    ets:select_delete(stream_mgmt_stale_h, MS)
```

src/stream_management/mod_stream_management_stale_h.erl
```diff
@@ -0,0 +1,136 @@
...
+handle_info(Info, #state{gc_repeat_after = RepeatAfter,
+                          gc_geriatric = _GeriatricAge} = State) ->
+    ?WARNING_MSG("unexpected message ~p", [Info]),
+    {noreply, State, RepeatAfter}.
+
+%%
+%% Helpers
+%%
+-spec timestamp_to_seconds(erlang:timestamp()) -> non_neg_integer().
+timestamp_to_seconds({MegaSecs, Secs, _MicroSecs}) ->
```
arcusfelis: this would be removed

src/stream_management/mod_stream_management_stale_h.erl
```diff
@@ -0,0 +1,136 @@
...
+
+handle_call(Msg, _From, State) ->
+    ?WARNING_MSG("unexpected message ~p", [Msg]),
+    {reply, ok, State}.
+
+handle_cast(Msg, State) ->
+    ?WARNING_MSG("unexpected message ~p", [Msg]),
+    {noreply, State}.
+
+handle_info(timeout, #state{gc_repeat_after = RepeatAfter,
```
arcusfelis: Put this code into a separate function, maaaybe export it, so it can be triggered separately for debugging or for ops (or for tests).

src/mod_stream_management.erl
```diff
@@ -115,10 +163,84 @@ get_resume_timeout(Default) ->
...
+get_gc_geriatric(Default) ->
+    MaybeModOpts = gen_mod:get_module_opt(?MYNAME, ?MODULE, stale_h, undefined),
+    case MaybeModOpts of
+        {true, GCOpts} ->
+            proplists:get_value(gc_geriatric, GCOpts, Default);
+        _ -> Default
+    end.
+
+-spec set_gc_geriatric(pos_integer()) -> boolean().
+set_gc_geriatric(ResumeTimeout) ->
```
arcusfelis: And this is the exact reason for using flat options - you don't need extra helpers.

Btw, can we name it gc_ttl?

Anyway, you don't need three copies of the same function, parametrize it and make something like:

set_stale_h_option(...., StaleHOptName, StaleHOptValue)

src/ejabberd_c2s.erl
NelsonVides: Meh that was just my linter complaining about old code and I just changed it in place, I don't think it makes much of a difference but ok, I can do that xD

mongoose-im:
[7279.1](https://travis-ci.org/esl/MongooseIM/jobs/591164274) / Erlang 22.0 / small_tests / fd9444cf2b87c628ae840f64e7afda0531ac47a9
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2472/7279/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2472/7279/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-09-29_17.11.07/lib.mongooseim.logs/run.2019-09-29_17.11.07/suite.log.html)

src/stream_management/mod_stream_management_stale_h.erl
michalwski: There is one important thing we need to remember when using `erlang:system_time`. It's not monotonically increasing. For instance it can go back or forward in time when the system clock changes (sync or daylight saving changes). It may result in some records removed too quickly or staying in the database for too long. In this particular case, I'm not sure this causes much trouble , though.

Regarding the `p1_time_compat:system_time`. I think we no longer need it. It was need when we were in transition period between Erlang/OTP 17, 18 and 19.

big_tests/tests/sm_SUITE.erl
```diff
@@ -27,14 +28,16 @@
...
      server_requests_ack_freq_2
      ].
 
 groups() ->
     G = [{parallel, [parallel], parallel_test_cases()},
          {parallel_manual_ack_freq_1, [parallel], parallel_manual_ack_test_cases()},
+         {stale_h, [], stale_h_test_cases()},
          {manual_ack_freq_long_session_timeout, [parallel], [preserve_order]}],
-    ct_helper:repeat_all_until_all_ok(G).
+    G.
```
michalwski: I may be not aware of this so I ask. Did we decide to stop using the `repeat_all_until_all_ok`?

doc/modules/mod_stream_management.md
```diff
@@ -1,21 +1,101 @@
 ### Module Description
-Enables [XEP-0198: Stream Management](http://xmpp.org/extensions/xep-0198.html). 
-It is implemented mostly in `ejabberd_c2s`. 
-This module is just a "starter", to supply the configuration values to new client connections.
-It also provides a basic session table API and adds a new stream feature.
+
+Enables [XEP-0198: Stream Management](http://xmpp.org/extensions/xep-0198.html).
```
michalwski: Thanks for the detailed documentation @NelsonVides! I have one suggestion related to content organization. It seams to me that the most important goal of this documentation is to provide information what this module do, how to configure it and what are all the possible configuration options. In my opinion, it is more efficient for the reader if this kind of information is at the top. Technical details could be moved below. What do you think?

big_tests/tests/sm_SUITE.erl
NelsonVides: Oh damn, no, I just git-staged all instead of by patches. I personally dislike strongly that `repeat_all_until_all_ok`, first thing I do when I start working on any suite is _removing it_ and then just don't add it to the commit. This time I added it by mistake, but I'd love a discussion on why is that thing any good 🙁

doc/modules/mod_stream_management.md
NelsonVides: Cool, that's a nice idea 👍

src/stream_management/mod_stream_management_stale_h.erl
NelsonVides: Actually, there's `erlang:monotonic_time(seconds)` [function](http://erlang.org/doc/man/erlang.html#monotonic_time-1). It perfect: it is monotonic, from the erlang standard library, returns specifically seconds; seems like just what we wanted 😃

mongoose-im:
[7282.1](https://travis-ci.org/esl/MongooseIM/jobs/592009868) / Erlang 22.0 / small_tests / 53c363880443849a8024b41b1263d08cfac4e9ee
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2472/7282/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2472/7282/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-10-01_13.04.37/lib.mongooseim.logs/run.2019-10-01_13.04.38/suite.log.html)

mongoose-im:
[7284.1](https://travis-ci.org/esl/MongooseIM/jobs/592221882) / Erlang 22.0 / small_tests / 6ce506596996ed646f8cfd1b1e8b1b86aa7074c7
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2472/7284/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2472/7284/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-10-01_20.13.11/lib.mongooseim.logs/run.2019-10-01_20.13.12/suite.log.html)

PR #2479, opened 2019-09-25
[skip ci] Lists are inline

Nyco:
There is a need for a carriage return on ReadTheDocs, forgot that, sorry...

Capture d’écran 2019-09-25 à 15 03 23


NelsonVides:
@Nyco the `[skip ci]` tag works on the last commit's title, not on the PR name 🙂

Issue #1988, opened 2018-07-18
make rel 


allireno:
MongooseIM version: 3.0.0
Installed from: source
Erlang/OTP version: 18.3

Using Ubuntu 16.04

I am trying to follow the directions to complile MongooseIM. I have cloned the git repository using the following command:
`git clone https://github.com/esl/MongooseIM.git`

michalwski:
HI @allireno,

That's a good start! Were you able to build MongooseIM successfully?

michalwski:
I think I know what may be the issue. Current MongooseIM as in master branch doesn't work with Erlang/OTP 18.3. The minimum version is 19.3.

allireno:
@michalwski Updating the erlang version seemed to fix the problem. It is currently still compiling (very slowly)

allireno:
I am trying to do this on two computers. My other computer got this error after updating erlang to version 21
This one cloned the git repository using:
`git clone https://github.com/esl/mongooseim` 
(not sure if that makes a difference)
![ad636m_1531912058056](https://user-images.githubusercontent.com/41064213/42886788-f6f01cf8-8a93-11e8-80a6-eba9bbc92fbe.png)

michalwski:
@allireno what machine do you have that the compilation takes 2 hours? Even on travis-ci where the machines are very limit MongooseIM is compile in less then 5 minutes.

Regarding the issue with erlang 21. It looks like you don't have permissions (or the disk was full) to write to location `~/.cache/rebar3/hex`. Rebar3, the tool used to compile MongooseIM tries to cache some dependencies in the mentioned directory. Writing to the location failed so the compilation was aborted.

AdarshMaurya:
@ferd : Thanks for looking providing this link. Cannot built the MongooseIM with /usr/local/Cellar/erlang/21.0.3/lib/erlang/lib.

```
===> Resolving OTP Applications from directories:
          /Users/adarshmaurya/Downloads/MongooseIM-master/_build/prod/lib
          /Users/adarshmaurya/Downloads/MongooseIM-master/ebin
          /usr/local/Cellar/erlang/21.0.3/lib/erlang/lib
=ERROR REPORT==== 22-Jul-2018::16:36:07.486477 ===
Error in process <0.2081.0> with exit value:
{function_clause,
    [{rlx_app_info,parse_version,
         [{cmd,"tools/generate_vsn.sh"}],
         [{file,
[...]

PR #2488, opened 2019-10-10
tools: generate_vsn: remove bashism and use POSIX /bin/sh

kaniini:
This PR addresses #1988

Proposed changes include:
* removal of dependency on `/bin/bash` for generating the version number, use POSIX `/bin/sh`

### DIFF
--- a/tools/generate_vsn.sh
+++ b/tools/generate_vsn.sh
@@ -1,5 +1,5 @@
-#!/bin/bash
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+#!/bin/sh
+DIR="$( cd "$( dirname "$0" )" && pwd )"
 GIT_VSN=`git describe --always --tags 2>/dev/null`
 
 if [ $? -eq 0 ]; then


tools/generate_vsn.sh
```diff
@@ -1,5 +1,5 @@
-#!/bin/bash
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+#!/bin/sh
+DIR="$( cd "$( dirname "$0" )" && pwd )"
 GIT_VSN=`git describe --always --tags 2>/dev/null`
```
arcusfelis: fine.

But both versions assume, that we are running from one of directories of the repo.
Would still work differently for *both* versions if called from repo and from outside of repo.

arcusfelis: would replace with something like:

```bash
#!/bin/sh

# Go to repo root
cd `dirname "$0"`/../

GIT_VSN=`git describe --always --tags 2>/dev/null`
if [ $? -eq 0 ]; then
    echo $GIT_VSN
else
    cat VERSION
fi
```

michalwski: @arcusfelis thanks for the tip. I applied it and merge the PR into master.

michalwski:
@kaniini, thanks for you contribution. I merged your PR!

PR #2502, opened 2019-10-18
Workaround transport delivery problems

arcusfelis:
Sometimes we get timeout when waiting for stream_end.

For example:

```erlang
----------------------------------------------------

2019-10-16 10:13:34.899

{h_non_given_closes_stream_gracefully,parallel}

{thrown,{timeout,stream_end}}

Testing big_tests.tests.sm_SUITE: *** FAILED test case 15 ***
```

https://travis-ci.org/arcusfelis/MongooseIM/jobs/598572297

This test can fail, because we just hit the stream end timeout of 5 seconds or that we close socket on server side without waiting for all data received. I've failed to reproduce the last condition on my Mac though. Though there are some stackoverlow ideas:
- https://stackoverflow.com/questions/36533028/how-to-send-a-message-before-closing-tcp-connection-posix
- Or OTP docs:

>       close(Socket) -> ok
             Types:
                Socket = socket()
             Closes a TCP socket.
             Note  that in most implementations of TCP, doing a close 
             does not guarantee that any data sent is delivered to the 
             recipient before the close is detected at the remote
             side. If you want to guarantee delivery of the data to the recipient 
             there are two common ways to achieve this.
               * Use gen_tcp:shutdown(Sock, write) to signal that no more data is 
                  to be sent and wait for the read side of the socket to be closed.
               * Use the socket option {packet, N} (or something similar) to make it 
                  possible for the receiver to close the connection when it knowns it 
                  has received all the  data.

Proposed changes include:
* The cute function from old Erlang/OTP ssl, that does the best we can to flush `stream-end` https://github.com/erlang/otp/blob/OTP_17.0-rc2/lib/ssl/src/tls_connection.erl#L959

### DIFF
--- a/src/ejabberd_receiver.erl
+++ b/src/ejabberd_receiver.erl
@@ -290,7 +290,7 @@ terminate(_Reason, #state{parser = Parser,
         undefined -> ok;
         _ -> gen_fsm_compat:send_event(C2SPid, closed)
     end,
-    catch (State#state.sock_mod):close(State#state.socket),
+    catch workaround_transport_delivery_problems(State#state.socket, State#state.sock_mod),
     ok.
 
 %%--------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mongoose-im:
[7356.1](https://travis-ci.org/esl/MongooseIM/jobs/599667395) / Erlang 22.0 / small_tests / 6ccdb049f1bf208c5071024489e7790f6b5271dc
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2502/7356/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2502/7356/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-10-18_16.08.08/lib.mongooseim.logs/run.2019-10-18_16.08.08/suite.log.html)

src/ejabberd_receiver.erl
```diff
@@ -445,3 +445,19 @@ hibernate() ->
 -spec maybe_hibernate(state()) -> hibernate | infinity | pos_integer().
 maybe_hibernate(#state{hibernate_after = 0}) -> hibernate();
 maybe_hibernate(#state{hibernate_after = HA}) -> HA.
+
+%% It's workaround taken from ssl application
+workaround_transport_delivery_problems(Socket, gen_tcp) ->
```
michalwski: I think we can rename the function to sth like `ensure_data_sent_before_close`

arcusfelis: Renamed to shutdown_socket_and_wait_for_peer_to_close.
It does not ensure data is sent, it's *trying* to ensure that data is sent. It also closes socket (ok, peer closes).

PR #2531, opened 2019-11-14
Rewrite mod_ping to execute logic in C2S context

fenek:
WIP

### DIFF
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -149,7 +149,6 @@ store_session_info(FsmRef, JID, KV) ->
 remove_session_info(FsmRef, JID, Key) ->
     FsmRef ! {remove_session_info, JID, Key, self()}.
 
-
 %%%----------------------------------------------------------------------
 %%% Callback functions from gen_fsm
 %%%----------------------------------------------------------------------
@@ -948,14 +947,9 @@ resume_session(resume, From, SD) ->
 %%          {stop, Reason, NewStateData}
 %%----------------------------------------------------------------------
 
-handle_event({add_info_handler, Tag, M, F, Extra}, StateName, StateData0) ->
-    StateData1 = ejabberd_c2s_info_handler:add_to_state(Tag, M, F, Extra, StateData0),
-    case ejabberd_c2s_info_handler:safe_call(Tag, init, M, F, Extra, StateData1) of
-        {error, _} ->
-            fsm_next_state(StateName, StateData0);
-        StateData2 ->
-            fsm_next_state(StateName, StateData2)
-    end;
+handle_event({add_info_handler, Tag, InitialState}, StateName, StateData0) ->
+    StateData1 = ejabberd_c2s_info_handler:add_to_state(Tag, InitialState, StateData0),
+    fsm_next_state(StateName, StateData1);
 handle_event({remove_info_handler, Tag}, StateName, StateData0) ->
     StateData1 = ejabberd_c2s_info_handler:remove_from_state(Tag, StateData0),
     fsm_next_state(StateName, StateData1);
@@ -1147,18 +1141,18 @@ handle_incoming_message({broadcast, Type, From, Packet}, StateName, StateData) -
     lists:foreach(fun(USR) -> ejabberd_router:route(From, jid:make(USR), Packet) end,
         lists:usort(Recipients)),
     fsm_next_state(StateName, StateData);
-handle_incoming_message({Tag, Data} = MaybeCustomInfo, StateName, StateData) ->
+handle_incoming_message({call_info_handler, Tag, Data} = MaybeCustomInfo, StateName, StateData) ->
     NStateData =
     case ejabberd_c2s_info_handler:get_for(Tag, StateData) of
-        {M, F, HandlerState} ->
-            case ejabberd_c2s_info_handler:safe_call(Tag, Data, M, F, HandlerState, StateData) of
+        {Tag, HandlerState} ->
+            case ejabberd_c2s_info_handler:safe_call(Tag, Data, HandlerState, StateData) of
                 {error, _} ->
                     StateData;
                 NStateData0 ->
                     NStateData0
             end;
         _ ->
-            ?INFO_MSG("Unexpected info: ~p", [MaybeCustomInfo]),
+            ?INFO_MSG("No handler for: ~p", [MaybeCustomInfo]),
             StateData
     end,
     fsm_next_state(StateName, NStateData);
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -1158,7 +1158,7 @@ handle_incoming_message({Tag, Data} = MaybeCustomInfo, StateName, StateData) ->
                     NStateData0
             end;
         _ ->
-            ?ERROR_MSG("Unexpected info: ~p", [MaybeCustomInfo]),
+            ?INFO_MSG("Unexpected info: ~p", [MaybeCustomInfo]),
             StateData
     end,
     fsm_next_state(StateName, NStateData);
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -947,6 +947,18 @@ resume_session(resume, From, SD) ->
 %%          {next_state, NextStateName, NextStateData, Timeout} |
 %%          {stop, Reason, NewStateData}
 %%----------------------------------------------------------------------
+
+handle_event({add_info_handler, Tag, M, F, Extra}, StateName, StateData0) ->
+    StateData1 = ejabberd_c2s_info_handler:add_to_state(Tag, M, F, Extra, StateData0),
+    case ejabberd_c2s_info_handler:safe_call(Tag, init, M, F, Extra, StateData1) of
+        {error, _} ->
+            fsm_next_state(StateName, StateData0);
+        StateData2 ->
+            fsm_next_state(StateName, StateData2)
+    end;
+handle_event({remove_info_handler, Tag}, StateName, StateData0) ->
+    StateData1 = ejabberd_c2s_info_handler:remove_from_state(Tag, StateData0),
+    fsm_next_state(StateName, StateData1);
 handle_event(keep_alive_packet, session_established,
              #state{server = Server, jid = JID} = StateData) ->
     mongoose_hooks:user_sent_keep_alive(Server, JID),
@@ -1135,6 +1147,21 @@ handle_incoming_message({broadcast, Type, From, Packet}, StateName, StateData) -
     lists:foreach(fun(USR) -> ejabberd_router:route(From, jid:make(USR), Packet) end,
         lists:usort(Recipients)),
     fsm_next_state(StateName, StateData);
+handle_incoming_message({Tag, Data} = MaybeCustomInfo, StateName, StateData) ->
+    NStateData =
+    case ejabberd_c2s_info_handler:get_for(Tag, StateData) of
+        {M, F, HandlerState} ->
+            case ejabberd_c2s_info_handler:safe_call(Tag, Data, M, F, HandlerState, StateData) of
+                {error, _} ->
+                    StateData;
+                NStateData0 ->
+                    NStateData0
+            end;
+        _ ->
+            ?ERROR_MSG("Unexpected info: ~p", [MaybeCustomInfo]),
+            StateData
+    end,
+    fsm_next_state(StateName, NStateData);
 handle_incoming_message(Info, StateName, StateData) ->
     ?ERROR_MSG("Unexpected info: ~p", [Info]),
     fsm_next_state(StateName, StateData).
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -1141,7 +1141,7 @@ handle_incoming_message({broadcast, Type, From, Packet}, StateName, StateData) -
     lists:foreach(fun(USR) -> ejabberd_router:route(From, jid:make(USR), Packet) end,
         lists:usort(Recipients)),
     fsm_next_state(StateName, StateData);
-handle_incoming_message({call_info_handler, Tag, Data} = MaybeCustomInfo, StateName, StateData) ->
+handle_incoming_message({call_info_handler, Tag, Data}, StateName, StateData) ->
     NStateData =
     case ejabberd_c2s_info_handler:get_for(Tag, StateData) of
         {Tag, HandlerState} ->
@@ -1151,13 +1151,14 @@ handle_incoming_message({call_info_handler, Tag, Data} = MaybeCustomInfo, StateN
                 NStateData0 ->
                     NStateData0
             end;
-        _ ->
-            ?INFO_MSG("No handler for: ~p", [MaybeCustomInfo]),
+        no_handler ->
+            ?INFO_MSG("event=no_info_handler tag=~p host=~p", [Tag, ejabberd_c2s_state:server(StateData)]),
             StateData
     end,
     fsm_next_state(StateName, NStateData);
 handle_incoming_message(Info, StateName, StateData) ->
-    ?ERROR_MSG("Unexpected info: ~p", [Info]),
+    ?ERROR_MSG("event=unexpected_info info=~p state_name=~p host=~p",
+               [Info, StateName, ejabberd_c2s_state:server(StateData)]),
     fsm_next_state(StateName, StateData).
 
 process_incoming_stanza_with_conflict_check(From, To, Acc, StateName, StateData) ->
--- a/src/ejabberd_c2s_info_handler.erl
+++ b/src/ejabberd_c2s_info_handler.erl
@@ -1,0 +1,56 @@
+-module(ejabberd_c2s_info_handler).
+
+-include("ejabberd_c2s.hrl").
+-include("mongoose.hrl").
+
+% API for other modules
+-export([add/5, remove/2]).
+
+% API for ejabberd_c2s
+-export([get_for/2,
+         add_to_state/5,
+         remove_from_state/2,
+         safe_call/3,
+         safe_call/6]).
+
+add(C2S, Tag, M, F, HandlerState) ->
+    p1_fsm_old:send_all_state_event(C2S, {add_info_handler, Tag, M, F, HandlerState}).
+
+remove(C2S, Tag) ->
+    p1_fsm_old:send_all_state_event(C2S, {remove_info_handler, Tag}).
+
+get_for(Tag, #state{ info_handlers = InfoHandlers }) ->
+    maps:get(Tag, InfoHandlers, undefined).
+
+add_to_state(Tag, M, F, HandlerState, #state{ info_handlers = InfoHandlers0 } = C2SState) ->
+    C2SState#state{ info_handlers = InfoHandlers0#{ Tag => {M, F, HandlerState} } }.
+
+remove_from_state(Tag, #state{ info_handlers = InfoHandlers0 } = C2SState) ->
+    C2SState#state{ info_handlers = maps:remove(Tag, InfoHandlers0) }.
+
+safe_call(Tag, Data, C2SState) ->
+    case get_for(Tag, C2SState) of
+        {M, F, HandlerState} ->
+            case safe_call(Tag, Data, M, F, HandlerState, C2SState) of
+                {error, _} ->
+                    C2SState;
+                NC2SState0 ->
+                    NC2SState0
+            end;
+        _ ->
+            undefined
+    end.
+
+safe_call(Tag, Data, M, F, HandlerState, C2SState) ->
+    try M:F(Data, HandlerState, C2SState) of
+        {NewHandlerState, NewC2SState0} ->
+            % Perhaps may be optimized?
+            add_to_state(Tag, M, F, NewHandlerState, NewC2SState0)
+    catch
+        C:R:S ->
+            ?ERROR_MSG("event=custom_c2s_info_handler_crash,tag=~p,data=~p,m=~p,f=~p,"
+                       "extra=~p,class=~p,reason=~p,stacktrace=~1000p",
+                       [Tag, Data, M, F, HandlerState, C, R, S]),
+            {error, {C, R, S}}
+    end.
+
--- a/src/ejabberd_c2s_info_handler.erl
+++ b/src/ejabberd_c2s_info_handler.erl
@@ -42,10 +42,12 @@ safe_call(Tag, Data, C2SState) ->
     end.
 
 safe_call(Tag, Data, M, F, HandlerState, C2SState) ->
-    try M:F(Data, HandlerState, C2SState) of
-        {NewHandlerState, NewC2SState0} ->
+    Jid = ejabberd_c2s_state:jid(C2SState),
+    Server = ejabberd_c2s_state:server(C2SState),
+    try M:F(Data, HandlerState, #{jid => Jid, server => Server}) of
+        NewHandlerState ->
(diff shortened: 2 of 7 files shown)


mongoose-im:
[7447.1](https://travis-ci.org/esl/MongooseIM/jobs/611818015) / Erlang 22.0 / small_tests / f604d9c713a9afde6aa4b7324d7a9af94e50645b
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2531/7447/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2531/7447/small_tests.22.0/small/ct_run.mongooseim@localhost.2019-11-14_11.31.37/lib.mongooseim.logs/run.2019-11-14_11.31.38/suite.log.html)

src/mod_ping.erl
```diff
@@ -1,130 +1,101 @@
...
-stop(Host) ->
-    Proc = gen_mod:get_module_proc(Host, ?MODULE),
-    Pid = erlang:whereis(Proc),
-    gen_server:call(Proc, stop),
-    wait_for_process_to_stop(Pid),
-    supervisor:delete_child(?SUPERVISOR, Proc).
+cancel_timer(undefined) ->
+    do_nothing;
+cancel_timer(TRef) ->
+    erlang:cancel_timer(TRef).
```
erszcz: This function is redundant, let's remove it:

> If Result is false, a timer corresponding to `TimerRef` could not be found. This can be either because the timer had expired, already had been canceled, or because TimerRef never corresponded to a timer.
-- http://erlang.org/doc/man/erlang.html#cancel_timer-2

```erlang
1> R = make_ref().
#Ref<0.2521718886.3805282306.126943>
2> erlang:cancel_timer(R).
false
```

src/mod_ping.erl
```diff
@@ -1,130 +1,101 @@
...
+    PingReqTimeout = gen_mod:get_module_opt(Server, ?MODULE, ping_req_timeout,
+                                            ?DEFAULT_PING_REQ_TIMEOUT),
+    IQ = #iq{type = get,
+             sub_el = [#xmlel{name = <<"ping">>,
+                              attrs = [{<<"xmlns">>, ?NS_PING}]}]},
+    Pid = self(),
+    F = fun(_, _, _, timeout) ->
+               Pid ! {mod_ping, timeout};
+           (_From, _To, Acc, _Response) ->
+                Acc
```
erszcz:
```suggestion
               Acc
```

src/mod_ping.erl
```diff
@@ -1,130 +1,101 @@
...
+    {add_ping_timer(HandlerState, C2SState), C2SState};
+handle_info(send_ping, HandlerState, C2SState) ->
+    JID = ejabberd_c2s_state:jid(C2SState),
+    Server = ejabberd_c2s_state:server(C2SState),
+    route_ping_iq(JID, Server),
+    {add_ping_timer(HandlerState, C2SState), C2SState};
+handle_info(timeout, HandlerState, C2SState) ->
+    JID = ejabberd_c2s_state:jid(C2SState),
+    Server = ejabberd_c2s_state:server(C2SState),
+    ejabberd_hooks:run(user_ping_timeout, Server, [JID]),
```
michalwski: This needs to be rewritten to use the new `mongoose_hooks` module. Sth like the following:
```erlang
mongoose_hooks:user_ping_timeout(Server, JID)
```

src/mod_ping.erl
```diff
@@ -1,130 +1,101 @@
...
+    IQ = #iq{type = get,
+             sub_el = [#xmlel{name = <<"ping">>,
+                              attrs = [{<<"xmlns">>, ?NS_PING}]}]},
+    Pid = self(),
+    F = fun(_, _, _, timeout) ->
+               Pid ! {mod_ping, timeout};
+           (_From, _To, Acc, _Response) ->
+                Acc
+        end,
+    From = jid:make(<<"">>, Server, <<"">>),
```
michalwski: I think it's safe to use
```suggestion
    From = jid:make_noprep(<<"">>, Server, <<"">>),
```

since the value in `Server` is already stringprepped.

src/mod_ping.erl
michalwski: @erszcz how about this:
```erlang
erlang:cancel_timer(undefined).
** exception error: bad argument
     in function  erlang:cancel_timer/1
        called as erlang:cancel_timer(undefined)
```

I think we still need to handle the case when `TRef` is `undefined` unless we change the code and allow only references as value for `TRef`.

src/ejabberd_c2s_info_handler.erl
```diff
@@ -0,0 +1,56 @@
...
+         safe_call/6]).
+
+add(C2S, Tag, M, F, HandlerState) ->
+    p1_fsm_old:send_all_state_event(C2S, {add_info_handler, Tag, M, F, HandlerState}).
+
+remove(C2S, Tag) ->
+    p1_fsm_old:send_all_state_event(C2S, {remove_info_handler, Tag}).
+
+get_for(Tag, #state{ info_handlers = InfoHandlers }) ->
+    maps:get(Tag, InfoHandlers, undefined).
```
erszcz: I'd not return `undefined` from any function whatsoever, as it makes troubleshooting a pain in the neck. The `undefined` weaves through following code and makes it hard to debug, when it actually a crash. [`undefined` is the `NULL` of Erlang](https://www.lucidchart.com/techblog/2015/08/31/the-worst-mistake-of-computer-science/).  Instead, I think the `maps` module has the sanest policy for fetching values:

- there's either `maps:get(Key, Map)` which throws an exception if `Key` is not found
- or `maps:get(Key, Map, Default)` which returns `Default` in case `Key` is not in `Map`

this allows us to either fail-fast and catch the exception or get a useful default if the failure is not critical and we know what to do if without the value. Returning `undefined` is the worst possible case as we neither fail-fast, nor have a usable value to deal with in subsequent code.

src/ejabberd_c2s_info_handler.erl
```diff
@@ -0,0 +1,56 @@
...
+    case get_for(Tag, C2SState) of
+        {M, F, HandlerState} ->
+            case safe_call(Tag, Data, M, F, HandlerState, C2SState) of
+                {error, _} ->
+                    C2SState;
+                NC2SState0 ->
+                    NC2SState0
+            end;
+        _ ->
+            undefined
```
erszcz: `undefined` strikes again.

src/mod_ping.erl
erszcz: @michalwski Indeed, I didn't catch all the call sites :| My bad.

src/ejabberd_c2s_info_handler.erl
michalwski: I fully agree with you @erszcz.
In this particular case I thought about using `maps:find` function which returs `{ok, Value}` or `error`. This is not ideal as well since the `error` can be considered as harmful as `undefined`.

src/mod_ping.erl
michalwski: Now, considering your other [comment](https://github.com/esl/MongooseIM/pull/2531/files#r392120978) regarding avoiding `undefined`, in this case we would need to come up with other data for the initial state of the `c2s_info_handler` which doesn't crash the `erlang:cancel_timer`.

src/mod_ping.erl
```diff
@@ -150,132 +121,28 @@ terminate(_Reason, #state{host = Host}) ->
...
-    Acc.
+iq_ping(_From, _To, Acc, #iq{type = get, sub_el = #xmlel{name = <<"ping">>}} = IQ) ->
+    {Acc, IQ#iq{type = result, sub_el = []}};
+iq_ping(_From, _To, Acc, #iq{sub_el = SubEl} = IQ) ->
+    {Acc, IQ#iq{type = error, sub_el = [SubEl, mongoose_xmpp_errors:feature_not_implemented()]}}.
 
-user_offline(Acc, _SID, JID, _Info, _Reason) ->
-    stop_ping(JID#jid.lserver, JID),
+user_online(Acc, {_, Pid} = _SID, _JID, _Info) ->
+    ejabberd_c2s_info_handler:add(Pid, mod_ping, ?MODULE, handle_info, undefined),
```
michalwski: @fenek how about we start the timer when adding the handler? In this case we avoid the `undefined` situation and timer is started anyway when we call the `init` function of the handler.

src/ejabberd_c2s_info_handler.erl
```diff
@@ -0,0 +1,56 @@
...
+    end.
+
+safe_call(Tag, Data, M, F, HandlerState, C2SState) ->
+    try M:F(Data, HandlerState, C2SState) of
+        {NewHandlerState, NewC2SState0} ->
```
bartekgorny: I somehow don't like allowing other modules to mess with internal state of c2s process. They have to be able to read it, that's for sure, but if they need to modify it they have their own state for that. For example, in a while I'll be refactoring mod_roster to use it, and my gut feeling is that if it is mod_roster that is responsible for handling roster, then all roster-related functions should go there (like current `ejabberd_c2s:roster_change/4`), and the whole roster should be stored in the handler state, made available via a hook handler.

arcusfelis: If handler state is available for this handler only, then it makes extensibility harder.
It's good idea to pass the whole c2s state and have getters and setters to get and set specific handler state.
So we would still have incapsulation when we need and we can violate the encapsulation when we _really_ need it (to access state of one handler from another, for example).

fenek: The primary thought behind this code was to allow custom handlers to read & write "aux fields" from C2S state but it may indeed be replaced by new aux fields API and passing only them to the handler instead of C2S state (or like you've mentioned: pass it as read-only).

NelsonVides: I think that passing only an auxiliary field of the c2s state would be fine from the encapsulation perspective, and if we need to modify this we can just rebind it to the c2s_state outside of the handlers. If a handler needs to rebind other fields on the c2s state... Dunno, I'm having in mind something similar to what we do with `mongoose_acc`, like,
```erlang
    try M:F(Data, HandlerState) of
        NewHandlerState ->
              OtherStuff = extract_important_variables_from_handlers(NewHandlerState),
              NewC2SState = merge_states(C2SState, OtherStuff);
        ...
```

bartekgorny: To rephrase - IMO:
* a handler MUST be able to read c2s process state
* a handler MUST NOT be able to modify c2s process state
* a handler MUST NOT be able to modify other handlers' state
now, there is an alternative:
EITHER
* a handler can read state of other handlers
* the c2s process can read state of all handlers
OR
* handler state is available to this handler only, but the handler MAY expose its state (or part of it) by creating a hook handler

src/ejabberd_c2s.erl
```diff
@@ -1163,6 +1175,21 @@ handle_incoming_message({broadcast, Type, From, Packet}, StateName, StateData) -
...
+    case ejabberd_c2s_info_handler:get_for(Tag, StateData) of
+        {M, F, HandlerState} ->
+            case ejabberd_c2s_info_handler:safe_call(Tag, Data, M, F, HandlerState, StateData) of
+                {error, _} ->
+                    StateData;
+                NStateData0 ->
+                    NStateData0
+            end;
+        _ ->
+            ?ERROR_MSG("Unexpected info: ~p", [MaybeCustomInfo]),
```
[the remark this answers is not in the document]
fenek: Right, mea culpa. :)

src/ejabberd_c2s_info_handler.erl
erszcz: I think we're better off keeping encapsulation stricter for now. Maybe we can make handlers only use `HandlerState` by default, but opt into getting the complete `C2SState` for some specific handlers that require it? But keeping "explicit is better than implicit" in mind.

src/ejabberd_c2s_info_handler.erl
```diff
@@ -0,0 +1,56 @@
...
+            % Perhaps may be optimized?
```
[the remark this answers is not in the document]
fenek: Different structure or implementation that doesn't involve copying whole c2s state on virtually every incoming message. Especially when `NewHandlerState == HandlerState` and `NewC2SState0 == C2SState`.

src/ejabberd_c2s_info_handler.erl
fenek: I disagree. It doesn't matter whether it's `error`, `undefined` or some other atom returned. In case of this function the contract is that it either returns 3-element tuple or `undefined`. Perhaps it may be renamed to `find_for` to be consistent with e.g. maps or lists API.
I understand the concern behind `undefined` being Erlang's `NULL` but the alternative is to crash and use `try` in `c2s`. Replacing all `undefined` with `try` blocks sounds like making Erlang another Java.

src/ejabberd_c2s_info_handler.erl
fenek: But in this case `undefined` is clearly a very bad choice for return value. However - is this function actually used? I think it may be a leftover.

src/ejabberd_c2s_info_handler.erl
erszcz: @fenek In places where you don't want to crash/`catch`, use the variant which accepts the `Default`. That's why the proposed API is the most flexible - it allows you to follow any path you wish.

erszcz: The whole purpose of such an API is to write *less* code in any case.

erszcz: In yet other words, an access function should have _fail fast_ crashing variant and a non-crashing variant.

For the second one it could return `undefined`, but then the handling code gets clunky, as we have to `case ... of` around it. The best design is to make the function itself take a default  value and return it, as that way we move the `case` inside the getter instead of the call site.

src/ejabberd_c2s_info_handler.erl
michalwski: I'm in favor of passing only the `HandlerState` to the handler. If a handler needs access to the stat of the C2S process it can be added with an explicit demand for it, like @erszcz suggested.
I also agree with the statement that a handler is not allowed to modify the C2S state, only it's own state.

src/ejabberd_c2s_info_handler.erl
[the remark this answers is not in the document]
arcusfelis: that's one of the reasons to pass state :) 
So, module can decide, if it wants to update state or not.

mongoose-im:
[8019.1](https://travis-ci.org/esl/MongooseIM/jobs/663493402) / Erlang 22.0 / small_tests / c2812df19bd37447d6e1a631a950974ea3e46898
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2531/8019/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2531/8019/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-03-17_13.10.58/lib.mongooseim.logs/run.2020-03-17_13.10.59/suite.log.html)

mongoose-im:
[8020.1](https://travis-ci.org/esl/MongooseIM/jobs/663519111) / Erlang 22.0 / small_tests / 335eb0d67cc9065ec187945367f53838df98c4c7
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2531/8020/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2531/8020/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-03-17_14.14.51/lib.mongooseim.logs/run.2020-03-17_14.14.51/suite.log.html)

mongoose-im:
[8021.1](https://travis-ci.org/esl/MongooseIM/jobs/663594290) / Erlang 22.0 / small_tests / d572042f3d5e16eb42980a2d908ebbb2b208ccdb
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2531/8021/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2531/8021/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-03-17_16.59.48/lib.mongooseim.logs/run.2020-03-17_16.59.49/suite.log.html)

mongoose-im:
[8025.1](https://travis-ci.org/esl/MongooseIM/jobs/663900483) / Erlang 22.0 / small_tests / 866df2dc6a3622d6d6ff1df0a77bd5c4e2977376
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2531/8025/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2531/8025/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-03-18_12.10.27/lib.mongooseim.logs/run.2020-03-18_12.10.27/suite.log.html)

src/mod_ping.erl
```diff
@@ -1,226 +1,166 @@
...
+        _ -> ok
+    end,
+    HandlerState.
+
+start_ping_timer(HandlerState, #{server := Server}) ->
+    start_ping_timer(HandlerState, Server);
+start_ping_timer(HandlerState, Server) ->
+    cancel_timer(HandlerState),
+    PingInterval = gen_mod:get_module_opt(Server, ?MODULE, ping_interval, ?DEFAULT_PING_INTERVAL),
+    {ok, Tref} = timer:apply_after(PingInterval * 1000,
```
NelsonVides: Uff... I'm way less convinced about that one. That goes through a `gen_server` called `timer_server`, hence it might again route all timeouts through a single process 🙁

src/ejabberd_c2s.erl
```diff
@@ -1135,8 +1141,24 @@ handle_incoming_message({broadcast, Type, From, Packet}, StateName, StateData) -
...
+                    NStateData0
+            end;
+        no_handler ->
+            ?INFO_MSG("event=no_info_handler tag=~p host=~p", [Tag, ejabberd_c2s_state:server(StateData)]),
+            StateData
+    end,
+    fsm_next_state(StateName, NStateData);
 handle_incoming_message(Info, StateName, StateData) ->
-    ?ERROR_MSG("Unexpected info: ~p", [Info]),
+    ?ERROR_MSG("event=unexpected_info info=~p state_name=~p host=~p",
```
michalwski: I noticed we have a small inconsistency in the key=value log formatting. Here we separate `key=value` pairs with space only, and in other modules with just a `,` and no space. I don't have a strong opinion which formatting to use other than use the one which is most common in the entire codebase.

src/ejabberd_c2s_info_handler.erl
```diff
@@ -0,0 +1,70 @@
...
+    Jid = ejabberd_c2s_state:jid(C2SState),
+    Server = ejabberd_c2s_state:server(C2SState),
+    try Tag:handle_c2s_info(Data, HandlerState, #{jid => Jid, server => Server}) of
+        HandlerState ->
+            C2SState;
+        NewHandlerState ->
+            add_to_state(Tag, NewHandlerState, C2SState)
+    catch
+        C:R:S ->
+            ?ERROR_MSG("event=custom_c2s_info_handler_crash,tag=~p,data=~p,"
```
michalwski: Here's the log where `key=value` pairs are separated with `,` only.

src/ejabberd_c2s_state.erl
```diff
@@ -0,0 +1,12 @@
+-module(ejabberd_c2s_state).
```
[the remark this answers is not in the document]
michalwski: Before @bartekgorny changes, these functions were used from `mod_ping` as far as I remember. And if I'm correct, the purpose of the module was to have accessor to the c2s state so that there is no need to include header file with c2s state definition.

src/mod_ping.erl
```diff
@@ -1,226 +1,166 @@
...
+handle_c2s_info(timeout, HandlerState, #{jid := JID, server := Server}) ->
+    mongoose_hooks:user_ping_timeout(Server, ok, JID),
+    case gen_mod:get_module_opt(Server, ?MODULE, timeout_action, none) of
+        kill -> ejabberd_c2s:stop(self());
+        _ -> ok
+    end,
+    HandlerState.
+
+start_ping_timer(HandlerState, #{server := Server}) ->
+    start_ping_timer(HandlerState, Server);
```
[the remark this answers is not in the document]
michalwski: Sounds reasonable, unless there is a reason why it's done this way.

src/mod_ping.erl
michalwski: Yes, usage of the `timer` module is discouraged by the [Erlang Efficiency Guide](http://erlang.org/doc/efficiency_guide/commoncaveats.html#timer-module)

michalwski:
@chrzaszcz This PR is key for some other changes we may want to introduce to MongooseIM and broadcasting mechanism. I think both the team and the project would benefit from your review!

src/ejabberd_c2s_info_handler.erl
```diff
@@ -0,0 +1,70 @@
+-module(ejabberd_c2s_info_handler).
+
+-include("ejabberd_c2s.hrl").
+-include("mongoose.hrl").
+-include("jlib.hrl").
+
+% API for other modules
+-export([add/3, remove/2, call/3]).
+
+-callback handle_c2s_info(Args :: term(), HandlerState :: term(), C2SParams :: map()) ->
```
chrzaszcz: It looks like we could use the `handler_state()` type here.

src/ejabberd_c2s_state.erl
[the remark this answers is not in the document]
bartekgorny: It is true that since we decided to not pass c2s state to handlers, this modules becomes redundant.

src/ejabberd_c2s.erl
bartekgorny: space wins with coma, the score is 51:26.

mongoose-im:
[8033.1](https://travis-ci.org/esl/MongooseIM/jobs/665822593) / Erlang 22.0 / small_tests / b618829a02c449efe99df90aef984d3ef3c1416d
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2531/8033/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2531/8033/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-03-23_11.48.51/lib.mongooseim.logs/run.2020-03-23_11.48.52/suite.log.html)

src/ejabberd_c2s_info_handler.erl
```diff
@@ -39,6 +41,17 @@ call(FsmRef, Tag, Data) ->
     FsmRef ! {call_info_handler, Tag, Data},
     ok.
 
+-spec call_after(integer(), jid:jid() | pid(), atom(), term()) -> reference().
```
chrzaszcz: `session_not_found` not found... in the spec

michalwski: Right. @BG, looks like the following would work:
```suggestion
-spec call_after(integer(), jid:jid() | pid(), atom(), term()) -> session_not_found | reference().
```

bartekgorny:
I'd suggested postponing merge of this PR. There is some work ahead on applying this mechanism to a number of modules, which may result in some major changes to this code too.

michalwski:
As discussed during the meeting today, we are merging this one and change the implementation / naming when needed before the next release.

PR #2543, opened 2019-11-27
integration with cowboy_swagger

vkatsuba:
Hi MongooseIM team. I propose `cowboy_swagger` integration by reason that this is very helpful both for testing REST API for QA and integration with client side. Examples was added as in https://mongooseim.readthedocs.io/en/2.0.0/swagger/index.html?client=true

Proposed changes include:
* Add `cowboy_swagger`
* Add `cowboy-trails`
* Create base template for `cowboy_swagger`
* Update `mongooseim.cfg` for adding custom modules with Swagger documentation

**How to:**
* Build & Run
```
$ make rel
$ _build/prod/rel/mongooseim/bin/mongooseim live
```
* Open In browser: https://localhost:8089/api-docs/index.html in case if used default config

**P.S.** Covered only REST API `/api/messages/[:with]`. If MongooseIM team is will be agree with this changes I will continue working on covered APIs by `cowboy_swagger` and will be provided documentation about - how to create custom Swagger documentation for REST API:
```sh
# REST APIs without `cowboy_swagger`
/api/sse
/api/contacts/[:jid]
/api/rooms/[:id]
/api/rooms/[:id]/config
/api/rooms/:id/users/[:user]
/api/rooms/[:id]/messages
```
If MongooseIM team will be not agree with changes in this PR - this PR will be closed.

Regards,
--V


src/ejabberd_cowboy.erl
```diff
@@ -282,3 +283,22 @@ maybe_insert_max_connections(TransportOpts, Opts) ->
...
+%% @end
+%% -------------------------------------------------------------------
+trails_store() ->
+    try
+      Host = hd(ejabberd_config:get_global_option(hosts)),
+      Config = ejabberd_config:get_local_option(swagger, Host),
+      Mods = proplists:get_value(modules, Config),
+      Trails = trails:trails(Mods),
+      trails:store(Trails)
+    catch _:_ ->
```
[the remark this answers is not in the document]
vkatsuba: Done. Thanks.

rel/files/app.config
```diff
@@ -43,7 +43,23 @@
...
+  %% Swagger spec
+  {cowboy_swagger,
+    [
+      {global_spec,
+        #{swagger => "2.0",
+          info => #{
+            title => "MongooseIM REST API",
+            description =>
+              "This is documentation of `MongooseIM REST API`.
+              REST API for client developers [MongooseIM REST API](https://mongooseim.readthedocs.io/en/2.0.0/REST-API/)",
```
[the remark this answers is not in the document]
vkatsuba: Done. Thanks.

rebar.config.script
```diff
@@ -43,7 +43,7 @@ fun() ->
         end
 end,
 
-RequiredApps = fun() -> [mongooseim, inets, tools] end,
+RequiredApps = fun() -> [mongooseim, inets, tools, trails, cowboy_swagger] end,
```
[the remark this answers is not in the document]
vkatsuba: Actually I was spend some time for understand that if I will be use approach what was used I will be need update `dialyzer` section in `rebar.config`. But your approach is more better. Thanks. Done.

rebar.config
```diff
@@ -78,7 +78,11 @@
   {eredis, "1.2.0"},
   {gen_fsm_compat, "0.3.0"},
   {erlang_pmp, "0.1.1"},
-  {erl_csv_generator, ".*", {git, "https://github.com/aleklisi/erl_csv_generator.git"}}
+  {erl_csv_generator, ".*", {git, "https://github.com/aleklisi/erl_csv_generator.git"}},
+
+  %% trails and cowboy_swagger for generating of Swagger documentation
+  {trails, ".*", {git, "https://github.com/inaka/cowboy-trails.git", {tag, "2.1.0"}}},
```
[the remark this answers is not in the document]
vkatsuba: Done. Thanks.

rebar.config
```diff
@@ -78,7 +78,11 @@
...
+  {cowboy_swagger, ".*", {git, "https://github.com/inaka/cowboy_swagger.git", {tag, "2.1.0"}}}
```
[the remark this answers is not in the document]
vkatsuba: Done. Thanks.

vkatsuba:
> That's awesome. We were thinking about doing this some time ago and here comes a PR. Thank you very much!
> I put some comments to the code. Also I have one more comment related to how the swagger is generated. Your solution proposes to add list of modules to a config file. I was thinking about moving this to an optional callback in a new behavior (let's say `mongoose_api`) to the handlers.
> Let's look at the example configuration:
> 
> ```erlang
>   { 8089 , ejabberd_cowboy, [
>       {num_acceptors, 10},
>       {transport_options, [{max_connections, 1024}]},
>       {protocol_options, [{compress, true}]},
>       {ssl, [{certfile, "priv/ssl/fake_cert.pem"}, {keyfile, "priv/ssl/fake_key.pem"}, {password, ""}]},
>       {modules, [
>           {"_", "/api/sse", lasse_handler, [mongoose_client_api_sse]},
>           {"_", "/api/messages/[:with]", mongoose_client_api_messages, []},
>           {"_", "/api/contacts/[:jid]", mongoose_client_api_contacts, []},
>           {"_", "/api/rooms/[:id]",    mongoose_client_api_rooms, []},
>           {"_", "/api/rooms/[:id]/config",    mongoose_client_api_rooms_config, []},
>           {"_", "/api/rooms/:id/users/[:user]",    mongoose_client_api_rooms_users, []},
>           {"_", "/api/rooms/[:id]/messages",    mongoose_client_api_rooms_messages, []}
> ]}]}
> ```
> 
> Now when `ejabberd_cowboy` computes the cowboy routes it can also go through list of modules in the configuration above and call an optional callback which would return everything which is needed to build the swager doc.
> What do you think about this approach?

Hi @michalwski,

Nice to hear that I provided PR in time. You see some time ago I was start think about adding FCM/GCM/APNS push notifications into but when but I got time to implement I was see that you are already added this functionality and I had no choice but to unload light version to [GitHub epns](https://github.com/vkatsuba/epns) :-)

About your approach, you see I chose this approach for a reason that this approach is flexible. You see I integrated `swagger_cowboy` into few projects and for first time I was implemented similar approach what you described. What did this lead to? This led to modules becoming incredibly large and by this reason I was move all documentation into a separate module in second project and this approach kept clean the REST API modules. But in any case - I can update according to the requirements of the MongooseIM  team if this will be needed. Many thanks for your time.

Best Regards,
--V

michalwski:
@vkatsuba 

> This led to situation when modules becoming incredibly large and by this reason I was move all documentation into a separate module in second project and this approach kept clean the REST API modules.

I think we can still have this separation with my approach with some boilerplate. The documentation can be defined in a separate module than the handler. Then assuming the optional callback is named `trails` and the documentation module for `mongoose_client_api_messages` is in `mongoose_client_api_messages_doc` all we need to do in the handler is this:

```erlang
trails() -> 
    mongoose_client_api_messages_doc:trails()
end
```

What do you think?

vkatsuba:
> I think we can still have this separation with my approach with some boilerplate. The documentation can be defined in a separate module than the handler. Then assuming the optional callback is named `trails` and the documentation module for `mongoose_client_api_messages` is in `mongoose_client_api_messages_doc` all we need to do in the handler is this:
> 
> ```erlang
> trails() -> 
>     mongoose_client_api_messages_doc:trails()
> end
> ```
> 
> What do you think?

Hi @michalwski,

I believe that this is will be great. Lets I try to do it and lets take a look if we all on the same page. Before I started - maybe you have other proposes or abstract vision how this should looks like by architecture side?

Regards,
--V

michalwski:
In terms of calling an optional callback, you get an inspiration here:
https://github.com/esl/MongooseIM/blob/21fa934f9e26bf5107e1b939ba071009d87932f5/src/gen_mod.erl#L446-L454

In terms of the behaviour, I think you can add the `-callback` and `-optional_callbacks` definition to module `mongoose_api_client`. Another point of inspiration is here: https://github.com/esl/MongooseIM/blob/21fa934f9e26bf5107e1b939ba071009d87932f5/src/wpool/mongoose_wpool.erl#L51-L59

src/ejabberd_cowboy.erl
```diff
@@ -300,5 +300,5 @@ trails_store() ->
       Trails = trails:trails(Mods),
       trails:store(Trails)
     catch Class:Exception ->
-        ?WARNING_MSG("Trails Call: [~p:~p/0] catched ~p:~p", [?MODULE, trails_store, Class, Exception])
+        io:format("WARNING! Trails Call: [~p:~p/0] catched ~p:~p~n", [?MODULE, trails_store, Class, Exception])
```
michalwski: The first approach was better. Let's use the macro instead of `io:format`. Did you notice any issues with the macro?

vkatsuba: Yep, the issue was in CI when I was use macros. CI was failed with error ` {{badmatch,{error,lager_not_running}}`

michalwski: Oh, it's because the `mod_websocket_SUITE` didn't have lager application started. You can start it by `application:ensure_all_started(lager)` in the `setup()` function in the suite and then stop it in the `teardown()`.

vkatsuba: Yes, thanks, done.

vkatsuba:
Hi @michalwski,

The dynamic trails store approach was added. Now users can use two different approaches:
1 - User can create documentation inside in REST API module
2 - User can create documentation in a separate module and try do call with export `trails/0` in REST API module:
```erlang
-export([trails/0]).

trails() ->
  separate_module_of_swagger_docs:trails().
```

Regards,
--V

src/mongoose_client_api_messages.erl
```diff
@@ -15,11 +15,16 @@
...
 
+-export([trails/0]).
+
 -include("mongoose.hrl").
 -include("jlib.hrl").
 -include("mongoose_rsm.hrl").
 -include_lib("exml/include/exml.hrl").
 
+trails() ->
+  mongooseim_swagger_docs:trails().
```
michalwski: Could you rename the `mongooseim_swagger_doc` module to `mongoose_client_api_messages_doc`?

rebar.config
```diff
@@ -110,7 +114,10 @@
                    {template, "rel/files/vm.dist.args",     "etc/vm.dist.args"},
 
                    {copy, "priv/logo.txt", "priv/logo.txt"},
-                   {copy, "VERSION", "priv/VERSION"}
+                   {copy, "VERSION", "priv/VERSION"},
+
+                   %% cowboy_swagger for generating documentation
+                   {copy, "./_build/default/lib/cowboy_swagger/priv/swagger", "priv/swagger"}
```
michalwski: I wonder if we need this at all. Maybe function `code:priv_dir(cowboy_swagger)` could be used and change the `static_files` parameter of `cowboy_swagger` application when the REST API listener starts? What do you think?

src/mongoose_client_api_messages.erl
vkatsuba: yes, but I was thinking - when we finished with architecture I will continue cover other REST APIs in module of `mongooseim_swagger_doc.erl`, actually this name is due to the fact that I intended to collect in one place and cover by Swagger documentation the all REST APIs. But maybe I shouldn't do this, since this module can be serve as a template for build Swagger documentation for other modules as of MongooseIM as for custom REST APIs

vkatsuba: Maybe create some sub folder eg `swagger_docs` then rename  `mongooseim_swagger_doc` to `mongoose_client_api_messages_doc` and move this module to new sub folder? then we can add new Swagger documentations in separate modules inside of  `swagger_docs` folder. What do you think about it?

rebar.config
vkatsuba: You see, I use approach from example of `swagger_cowboy` - https://github.com/inaka/cowboy_swagger/blob/master/example/rebar.config#L35 and I not sure if we can apply other approach

src/mongoose_client_api_messages.erl
vkatsuba: Done. Please, take a look if this approach is ok.

rebar.config
vkatsuba: Done.

PR #2573, opened 2019-12-18
Fix for clause error of ejabberd_c2s:wait_for_feature_before_auth/2

vkatsuba:
When user try sends packet:
```xml
<open xmlns='urn:ietf:params:xml:ns:xmpp-framing' to='localhost' version='1.0'/>
```
Will be provided response:
```xml
<features xmlns='http://etherx.jabber.org/streams'><mechanisms xmlns='urn:ietf:params:xml:ns:xmpp-sasl'><mechanism>PLAIN</mechanism><mechanism>DIGEST-MD5</mechanism><mechanism>SCRAM-SHA-1</mechanism></mechanisms><register xmlns='http://jabber.org/features/iq-register'/><sm xmlns='urn:xmpp:sm:3'/></features>
```
When user try send again:
```xml
<open xmlns='urn:ietf:params:xml:ns:xmpp-framing' to='localhost' version='1.0'/>
```
Will be provided error:
```sh
[error] gen_fsm <0.1259.0> in state wait_for_feature_before_auth terminated with reason: no function clause matching ejabberd_c2s:wait_for_feature_before_auth({xmlstreamstart,<<"stream:stream">>,[{<<"xmlns:stream">>,<<"http://etherx.jabber.org/streams">>},...]}, {state,{websocket,<0.1256.0>,{{127,0,0,1},43960},undefined},mod_websockets,#Ref<0.3703933498.3988783107.134323>,...}) line 471
[error] CRASH REPORT Process <0.1259.0> with 0 neighbours exited with reason: no function clause matching ejabberd_c2s:wait_for_feature_before_auth({xmlstreamstart,<<"stream:stream">>,[{<<"xmlns:stream">>,<<"http://etherx.jabber.org/streams">>},...]}, {state,{websocket,<0.1256.0>,{{127,0,0,1},43960},undefined},mod_websockets,#Ref<0.3703933498.3988783107.134323>,...}) line 471 in p1_fsm_old:terminate/7 line 729
[error] Supervisor ejabberd_c2s_sup had child undefined started with {ejabberd_c2s,start_link,undefined} at <0.1259.0> exit with reason no function clause matching ejabberd_c2s:wait_for_feature_before_auth({xmlstreamstart,<<"stream:stream">>,[{<<"xmlns:stream">>,<<"http://etherx.jabber.org/streams">>},...]}, {state,{websocket,<0.1256.0>,{{127,0,0,1},43960},undefined},mod_websockets,#Ref<0.3703933498.3988783107.134323>,...}) line 471 in context child_terminated
```

### DIFF
--- a/big_tests/tests/metrics_c2s_SUITE.erl
+++ b/big_tests/tests/metrics_c2s_SUITE.erl
@@ -49,7 +49,8 @@ groups() ->
          {errors, [sequence], [error_total,
                                error_mesg,
                                error_iq,
-                               error_presence]},
+                               error_presence,
+                               error_double_open_stream_must_close_connect]},
          {count, [sequence], [stanza_count]}],
     ct_helper:repeat_all_until_all_ok(G).
 
@@ -267,3 +268,13 @@ error_iq(Config) ->
     Alice = escalus_users:get_user_by_name(alice, Users),
     escalus_users:create_user(Config, Alice),
     wait_for_counter(Errors + 1, xmppErrorIq).
+
+error_double_open_stream_must_close_connect(Config) ->
+    AliceSpecs = escalus_fresh:create_fresh_user(Config, alice),
+    Steps = [start_stream, stream_features],
+    {ok, Alice, _} = escalus_connection:start(AliceSpecs, Steps),
+    try
+        escalus_connection:start_stream(Alice)
+    catch throw:{timeout, stream_start} ->
+        escalus_connection:wait_for_close(Alice, timer:seconds(5))
+    end.
--- a/big_tests/tests/metrics_c2s_SUITE.erl
+++ b/big_tests/tests/metrics_c2s_SUITE.erl
@@ -275,6 +275,6 @@ error_double_open_stream_must_close_connect(Config) ->
     {ok, Alice, _} = escalus_connection:start(AliceSpecs, Steps),
     try
         escalus_connection:start_stream(Alice)
-    catch throw:{timeout, stream_start} ->
+    catch error:"Not a valid stream start" ->
         escalus_connection:wait_for_close(Alice, timer:seconds(5))
     end.
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -568,9 +568,7 @@ wait_for_sasl_response({xmlstreamend, _Name}, StateData) ->
     send_trailer(StateData),
     {stop, normal, StateData};
 wait_for_sasl_response({xmlstreamerror, _}, StateData) ->
-    send_element_from_server_jid(StateData, mongoose_xmpp_errors:xml_not_well_formed()),
-    send_trailer(StateData),
-    {stop, normal, StateData};
+    c2s_stream_error(mongoose_xmpp_errors:xml_not_well_formed(), StateData);
 wait_for_sasl_response(closed, StateData) ->
     {stop, normal, StateData}.
 
@@ -622,9 +620,7 @@ wait_for_feature_after_auth({xmlstreamend, _Name}, StateData) ->
     {stop, normal, StateData};
 
 wait_for_feature_after_auth({xmlstreamerror, _}, StateData) ->
-    send_element_from_server_jid(StateData, mongoose_xmpp_errors:xml_not_well_formed()),
-    send_trailer(StateData),
-    {stop, normal, StateData};
+    c2s_stream_error(mongoose_xmpp_errors:xml_not_well_formed(), StateData);
 
 wait_for_feature_after_auth(closed, StateData) ->
     {stop, normal, StateData}.
@@ -666,9 +662,7 @@ wait_for_session_or_sm({xmlstreamend, _Name}, StateData) ->
     {stop, normal, StateData};
 
 wait_for_session_or_sm({xmlstreamerror, _}, StateData) ->
-    send_element_from_server_jid(StateData, mongoose_xmpp_errors:xml_not_well_formed()),
-    send_trailer(StateData),
-    {stop, normal, StateData};
+    c2s_stream_error(mongoose_xmpp_errors:xml_not_well_formed(), StateData);
 
 wait_for_session_or_sm(closed, StateData) ->
     {stop, normal, StateData}.
@@ -819,9 +813,7 @@ session_established({xmlstreamelement, El}, StateData) ->
     % Check 'from' attribute in stanza RFC 3920 Section 9.1.2
     case check_from(El, FromJID) of
         'invalid-from' ->
-            send_element_from_server_jid(StateData, mongoose_xmpp_errors:invalid_from()),
-            send_trailer(StateData),
-            {stop, normal, StateData};
+            c2s_stream_error(mongoose_xmpp_errors:invalid_from(), StateData);
         _NewEl ->
             NewState = maybe_increment_sm_incoming(StateData#state.stream_mgmt,
                                                    StateData),
@@ -846,13 +838,9 @@ session_established({xmlstreamend, _Name}, StateData) ->
 
 session_established({xmlstreamerror, <<"child element too big">> = E}, StateData) ->
     PolicyViolationErr = mongoose_xmpp_errors:policy_violation(StateData#state.lang, E),
-    send_element_from_server_jid(StateData, PolicyViolationErr),
-    send_trailer(StateData),
-    {stop, normal, StateData};
+    c2s_stream_error(PolicyViolationErr, StateData);
 session_established({xmlstreamerror, _}, StateData) ->
-    send_element_from_server_jid(StateData, mongoose_xmpp_errors:xml_not_well_formed()),
-    send_trailer(StateData),
-    {stop, normal, StateData};
+    c2s_stream_error(mongoose_xmpp_errors:xml_not_well_formed(), StateData);
 session_established(closed, StateData) ->
     ?DEBUG("Session established closed - trying to enter resume_session", []),
     maybe_enter_resume_session(StateData#state.stream_mgmt_id, StateData).
@@ -1066,12 +1054,10 @@ handle_info(system_shutdown, StateName, StateData) ->
     case StateName of
         wait_for_stream ->
             send_header(StateData, ?MYNAME, <<"1.0">>, <<"en">>),
-            send_element_from_server_jid(StateData, mongoose_xmpp_errors:system_shutdown()),
-            send_trailer(StateData),
+            c2s_stream_error(mongoose_xmpp_errors:system_shutdown(), StateData),
             ok;
         _ ->
-            send_element_from_server_jid(StateData, mongoose_xmpp_errors:system_shutdown()),
-            send_trailer(StateData),
+            c2s_stream_error(mongoose_xmpp_errors:system_shutdown(), StateData),
             ok
     end,
     {stop, normal, StateData};
@@ -1101,9 +1087,7 @@ handle_info(check_buffer_full, StateName, StateData) ->
         true ->
             Err = mongoose_xmpp_errors:stream_resource_constraint((StateData#state.lang),
                                            <<"too many unacked stanzas">>),
-            send_element_from_server_jid(StateData, Err),
-            send_trailer(StateData),
-            {stop, normal, StateData};
+            c2s_stream_error(Err, StateData);
         false ->
             fsm_next_state(StateName,
                            StateData#state{stream_mgmt_constraint_check_tref = undefined})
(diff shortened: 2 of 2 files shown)


NelsonVides:
@vkatsuba that's great, such a simple change and it fixes a big thing 😄 
Just one thing, can I ask you for a regression test? Something that fails on the current master and shows what this commit fixes, that way we'll know if we break it in the future 🙂

vkatsuba:
> @vkatsuba that's great, such a simple change and it fixes a big thing smile
> Just one thing, can I ask you for a regression test? Something that fails on the current master and shows what this commit fixes, that way we'll know if we break it in the future slightly_smiling_face

@NelsonVides sure,  but I cannot find best place where I can open eg, WebSocket connection and send stream twice like
```erlang
XML = #xmlel{name = <<"open">>,
    attrs = [
        {<<"xmlns">>, <<"urn:ietf:params:xml:ns:xmpp-framing">>}, 
        {<<"to">>, <<"localhost">>}, {<<"version">>, <<"1.0">>}]}.
```
and detect if the connection was closed as expected.
Can you help me with this?

NelsonVides:
> > @vkatsuba that's great, such a simple change and it fixes a big thing smile
> > Just one thing, can I ask you for a regression test? Something that fails on the current master and shows what this commit fixes, that way we'll know if we break it in the future slightly_smiling_face
> 
> @NelsonVides sure, but I cannot find best place where I can open eg, WebSocket connection and send stream twice like
> 
> ```erlang
> XML = #xmlel{name = <<"open">>,
>     attrs = [
>         {<<"xmlns">>, <<"urn:ietf:params:xml:ns:xmpp-framing">>}, 
>         {<<"to">>, <<"localhost">>}, {<<"version">>, <<"1.0">>}]}.
> ```
> 
> and detect if the connection was closed as expected.
> Can you help me with this?

Hmmm... So, I'm not an expert on websockets, but I'll try a few ideas :)

In [`big_tests/test.config`](https://circleci.com/gh/esl/MongooseIM/11150), there's some users configured to connect through `escalus_ws`, which is the websockets implementation in escalus, so the first thing would be to use one of them. Then, taking as an example the `connection_steps_to_xxx` functions in [`big_tests/tests/sm_SUITE.erl`](https://github.com/esl/MongooseIM/blob/c53886a1d135230fe58f69ecc7dc322d5ccf9bb8/big_tests/tests/sm_SUITE.erl#L866), we could control the steps that geralt makes when connecting, and repeat the one you mean, and see how it fails. That's my first guess, hope it helps. If you have any question feel free to ask 🙂

vkatsuba:
> > > @vkatsuba that's great, such a simple change and it fixes a big thing smile
> > > Just one thing, can I ask you for a regression test? Something that fails on the current master and shows what this commit fixes, that way we'll know if we break it in the future slightly_smiling_face
> > 
> > 
> > @NelsonVides sure, but I cannot find best place where I can open eg, WebSocket connection and send stream twice like
> > ```erlang
> > XML = #xmlel{name = <<"open">>,
> >     attrs = [
> >         {<<"xmlns">>, <<"urn:ietf:params:xml:ns:xmpp-framing">>}, 
> >         {<<"to">>, <<"localhost">>}, {<<"version">>, <<"1.0">>}]}.
> > ```
> > 
> > 
> > and detect if the connection was closed as expected.
> > Can you help me with this?
> 
> Hmmm... So, I'm not an expert on websockets, but I'll try a few ideas :)
> 
> In [`big_tests/test.config`](https://github.com/esl/MongooseIM/blob/c53886a1d135230fe58f69ecc7dc322d5ccf9bb8/big_tests/test.config#L118), there's some users configured to connect through `escalus_ws`, which is the websockets implementation in escalus, so the first thing would be to use one of them. Then, taking as an example the `connection_steps_to_xxx` functions in [`big_tests/tests/sm_SUITE.erl`](https://github.com/esl/MongooseIM/blob/c53886a1d135230fe58f69ecc7dc322d5ccf9bb8/big_tests/tests/sm_SUITE.erl#L866), we could control the steps that geralt makes when connecting, and repeat the one you mean, and see how it fails. That's my first guess, hope it helps. If you have any question feel free to ask slightly_smiling_face

Unfortunately this info don't help me. I still don't know where the best place where I can add this tests without breaking or conflict other tests. If you have clear vision about how it should looks like, you can provide me this info.

NelsonVides:
> > > > @vkatsuba that's great, such a simple change and it fixes a big thing smile
> > > > Just one thing, can I ask you for a regression test? Something that fails on the current master and shows what this commit fixes, that way we'll know if we break it in the future slightly_smiling_face
> > > 
> > > 
> > > @NelsonVides sure, but I cannot find best place where I can open eg, WebSocket connection and send stream twice like
> > > ```erlang
> > > XML = #xmlel{name = <<"open">>,
> > >     attrs = [
> > >         {<<"xmlns">>, <<"urn:ietf:params:xml:ns:xmpp-framing">>}, 
> > >         {<<"to">>, <<"localhost">>}, {<<"version">>, <<"1.0">>}]}.
> > > ```
> > > 
> > > 
> > > and detect if the connection was closed as expected.
> > > Can you help me with this?
> > 
> > 
> > Hmmm... So, I'm not an expert on websockets, but I'll try a few ideas :)
> > In [`big_tests/test.config`](https://github.com/esl/MongooseIM/blob/c53886a1d135230fe58f69ecc7dc322d5ccf9bb8/big_tests/test.config#L118), there's some users configured to connect through `escalus_ws`, which is the websockets implementation in escalus, so the first thing would be to use one of them. Then, taking as an example the `connection_steps_to_xxx` functions in [`big_tests/tests/sm_SUITE.erl`](https://github.com/esl/MongooseIM/blob/c53886a1d135230fe58f69ecc7dc322d5ccf9bb8/big_tests/tests/sm_SUITE.erl#L866), we could control the steps that geralt makes when connecting, and repeat the one you mean, and see how it fails. That's my first guess, hope it helps. If you have any question feel free to ask slightly_smiling_face
> 
> Unfortunately this info don't help me. I still don't know where the best place where I can add this tests without breaking or conflict other tests. If you have clear vision about how it should looks like, you can provide me this info.

Hey there, sorry it took me a few days to answer, I've been busy with some other stuff at work.

So, the stanza like the one you mention:
```erlang
XML = #xmlel{name = <<"open">>,
     attrs = [
         {<<"xmlns">>, <<"urn:ietf:params:xml:ns:xmpp-framing">>}, 
         {<<"to">>, <<"localhost">>}, {<<"version">>, <<"1.0">>}]}.
```
There's one like that in escalus: https://github.com/esl/escalus/blob/c53b5b09cd92fc49f3acd757041f6447ac7881d9/src/escalus_stanza.erl#L163-L166. Now, the thing with the steps, this architecture is quite smart but tricky: you know that escalus is our XMPP client simulator, right? In escalus, an xmpp connection abstraction is that of `escalus_connection`, which uses a given callback module as the backend. Usually, it is `escalus_tcp`, for the default tcp connections most often used, but there are other backends, like `escalus_ws`, and that is where I mentioned the users configured in `test.config`, they're configured to used the necessary backend.

So, having a look backwards, I see that `escalus_stanza:ws_open/1` is used in `escalus_ws:stream_start_req/``, which is used in `escalus_connection:start_stream/1`, which is one of the default steps in `escalus_connection:default_connection_steps/0` :)

So I'm imagining a test like this:
```erlang
test_it(Config) ->
    GeraltSpecs = escalus_fresh:create_fresh_user(Config, geralt),
    Steps = [start_stream, stream_features],
    {ok, Geralt, _} = escalus_connection:start(GeraltSpecs, Steps),
    escalus_connection:start_stream(Geralt). % <- here is where that step is repeated.
    %% Assert something
```
Where in the last line we need to check for the condition. Should the connection be terminated? Then some assertion regarding some functionality in `escalus_connection` or in `escalus_ws` will probably help. Do we expect some error stanza? Then we can wait to receive so and then assert it is the error we want.

If you tell me a bit more about websockets and what should happen in this situation (as I said, I have no idea about websockets in xmpp), I can find out more details :)

vkatsuba:
Hi @NelsonVides,

Many thanks for feedback, now I see how it can be implemented more clear. You see with `escalus` I have no a lot of experience. I will try create CT ASAP.

Best Regards,
--V

vkatsuba:
@NelsonVides,

Done. Please take a look, if CT looks like as expected.

Regards,
--V

NelsonVides:
> @NelsonVides,
> 
> Done. Please take a look, if CT looks like as expected. Also, sorry for small confusion, the error
> 
> ```erlang
> [error] gen_fsm <0.1272.0> in state wait_for_feature_before_auth terminated with reason: no function clause matching ejabberd_c2s:wait_for_feature_before_auth({xmlstreamstart,<<"stream:stream">>,[{<<"to">>,<<"localhost">>},{<<"version">>,<<"1.0">>},{<<"xml...">>,...},...]}, {state,{socket_state,gen_tcp,#Port<0.38>,<0.1271.0>},ejabberd_socket,#Ref<0.1827860675.2565603329.85556>,...}) line 471
> ```
> 
> manifests itself not only for WebSockets.
> 
> Regards,
> --V

Riiiiiiiiight, it's not just a websockets problem, I see what's going on now!

So, a client is sending the stream start stanza twice, which is an outright violation of the protocol. This is actually intended on all the states of the c2s process. Don't know how familiar are you with c2s, so maybe I'll say boring stuff now, but in case it helps you: the c2s process is a finite state machine, that starts with the `wait_for_stream` state and advances one by one until all states of the initialisation are done and we reach the `session_established`, which is where we really do all the cool messaging.
```erlang
         wait_for_stream/2,
         wait_for_feature_before_auth/2,
         wait_for_feature_after_auth/2,
         wait_for_session_or_sm/2,
         wait_for_sasl_response/2,
         session_established/2, session_established/3,
```
If the protocol is violated, we can just crash along the way, connection is lost and the issue is logged, so from that perspective, the behaviour of mim is correct, and I haven't found any directives for what to exactly do in the case of duplicated `<stream>` stanzas on the RFC6120.

A different question is, if we want to change the way we deal with this protocol violation. Maybe we want to catch the error explicitly and answer to the user with a stream error stanza, or close the process gracefully instead of crashing it, I don't know, we'd need more opinions I think 🤔

bartekgorny:
> A different question is, if we want to change the way we deal with this protocol violation. Maybe we want to catch the error explicitly and answer to the user with a stream error stanza, or close the process gracefully instead of crashing it, I don't know, we'd need more opinions I think 🤔

Sending stream error seems to be the right way to go, this is actually what is done in some cases. Like, if you telnet to your 5222 port and type some garbage you receive a stream error and the connection is closed. So, if you want a clean shutdown, you can call `c2s_stream_error(Error, State)`. Then there is a question, which error to choose - I'd say `bad-format`, since we don't know why exactly we reject it and we don't really care. To facilitate debugging, I'd suggest writing the offender to logs and/or adding some descriptive text to error stanza.

vkatsuba:
Hi @NelsonVides, @bartekgorny,

PR updated. Please, take a look if logic looks like as expected.

Regards,
--V

src/ejabberd_c2s.erl
```diff
@@ -1065,12 +1054,10 @@ handle_info(system_shutdown, StateName, StateData) ->
...
-            send_trailer(StateData),
+            c2s_stream_error(mongoose_xmpp_errors:system_shutdown(), StateData),
             ok;
         _ ->
-            send_element_from_server_jid(StateData, mongoose_xmpp_errors:system_shutdown()),
-            send_trailer(StateData),
+            c2s_stream_error(mongoose_xmpp_errors:system_shutdown(), StateData),
             ok
     end,
     {stop, normal, StateData};
```
NelsonVides: In this case of, regardless of the branch, we'll return `{stop, normal, StateData}`, which is precisely what `c2s_stream_error` returns, so you could just
```erlang
    case StateName of
        wait_for_stream ->
            send_header(StateData, ?MYNAME, <<"1.0">>, <<"en">>),
            c2s_stream_error(mongoose_xmpp_errors:system_shutdown(), StateData);
        _ ->
            c2s_stream_error(mongoose_xmpp_errors:system_shutdown(), StateData),
    end;
```

bartekgorny:
> First, that test suite is for metrics, and this test is for an issue during login, so I think this is not the correct test suite for this testcase, but I actually don't know which other is better. 

connect_SUITE, maybe?

src/ejabberd_c2s.erl
NelsonVides: What the heck, or even shorter:
```erlang
    case StateName of
        wait_for_stream -> send_header(StateData, ?MYNAME, <<"1.0">>, <<"en">>);
        _ -> ok
    end,
    c2s_stream_error(mongoose_xmpp_errors:system_shutdown(), StateData);
```
or also
```erlang
handle_info(system_shutdown, wait_for_stream, StateData) ->
    send_header(StateData, ?MYNAME, <<"1.0">>, <<"en">>);
    c2s_stream_error(mongoose_xmpp_errors:system_shutdown(), StateData);
handle_info(system_shutdown, _StateName, StateData) ->
    c2s_stream_error(mongoose_xmpp_errors:system_shutdown(), StateData);
```
Not sure which one is better, maybe the first, but it's mostly just opinion here 😛

src/ejabberd_c2s.erl
```diff
@@ -520,11 +520,12 @@ wait_for_feature_before_auth({xmlstreamend, _Name}, StateData) ->
...
 wait_for_feature_before_auth({xmlstreamerror, _}, StateData) ->
-    send_element_from_server_jid(StateData, mongoose_xmpp_errors:xml_not_well_formed()),
-    send_trailer(StateData),
-    {stop, normal, StateData};
+    c2s_stream_error(mongoose_xmpp_errors:xml_not_well_formed(), StateData);
 wait_for_feature_before_auth(closed, StateData) ->
-    {stop, normal, StateData}.
+    {stop, normal, StateData};
+wait_for_feature_before_auth(Stream, StateData) ->
```
NelsonVides: I talked about this with my colleagues, and we believe that this catch-all-return-c2s_stream_error should happen in all the steps of the fsm, and not only on `wait_for_feature_before_auth`. Maybe tests for this could be made as well, do you think you can do that? Should be just mechanical changes, let me know what do you think.

big_tests/tests/metrics_c2s_SUITE.erl
```diff
@@ -267,3 +268,13 @@ error_iq(Config) ->
...
+
+error_double_open_stream_must_close_connect(Config) ->
+    AliceSpecs = escalus_fresh:create_fresh_user(Config, alice),
+    Steps = [start_stream, stream_features],
+    {ok, Alice, _} = escalus_connection:start(AliceSpecs, Steps),
+    try
+        escalus_connection:start_stream(Alice)
+    catch error:"Not a valid stream start" ->
+        escalus_connection:wait_for_close(Alice, timer:seconds(5))
+    end.
```
NelsonVides: As @bartekgorny suggested, maybe `connect_SUITE` is a better suite to place this test. You can make a CT group that run its testcases in parallel, with all the testcases for all the steps of the FSM for example.

michalwski:
Similar changes were made and merged in #2667

Issue #2596, opened 2020-01-16
RDBS password is leaked in the logs.


Varmtech:
MongooseIM version: 3.5.0
Installed from: pkg
Erlang/OTP version: 22

When the log level is set info the and rdbs is configured then the rdbs credentials are leaked in the logs when the node starts.

2020-01-16 13:30:45.414 [notice] <0.98.0>@lager_file_backend:141 Changed loglevel of /var/log/mongooseim/ejabberd.log to info
2020-01-16 13:30:45.555 [info] <0.562.0>@mongoose_wpool_mgr:handle_call:93 event=starting_pool, pool=rdbms, host=global, tag=default, pool_opts=[{workers,10}], conn_opts=[{server,{pgsql,"xxxxxx.xxxxx.eu-central-1.rds.amazonaws.com",5432,"mongooseim","postgres","**Passformangooseim**"}}]

gustawlippa:
Thanks for letting us know! We're working on the fix.

DenysGonchar:
connection options are not disclosed in `info` logs any more. However, as secretes passed as usual function parameters there's no way to guaranty that they won't be disclosed in case of failures/crashes.

In the further releases we will try to figure out some MIM generic solution to avoid such disclosure on failures from our code base. But keep in mind that we still have to pass secretes as parameters to the 3rd party libraries (e.g. epgsql, mysql-otp, etc). and as it's in the nature of the erlang, most of them work asynchronously. so we have no means to catch/handle all the exceptions in that libraries, and for sure we don't want to disable standard lager logging of crashes.

Taking all this into account, our general recommendation would  be to avoid public sharing of MIM logs and restrict DB host access from the outside.

erszcz:
@DenysGonchar 
[`epgsql` is secure](https://github.com/epgsql/epgsql/blob/devel/src/commands/epgsql_cmd_connect.erl#L100-L117) with regard to storing the password or printing it due to a crash, since it hides it in a fun. Could we apply the same technique in our code, check if other DB drivers do the same, and possibly PR patches if they do not?
Once we have that, I wouldn't remove logging connection options completely - the password wouldn't be visible anymore anyway, while the rest is useful for troubleshooting.
cc @michalwski

michalwski:
Thanks for the update @erszcz. This sounds like a good approach to take wrt passwords in logs.

PR #2597, opened 2020-01-16
Remove logging connection options

gustawlippa:
This PR addresses issue #2596 by removing logging the connection options.

### DIFF
--- a/src/rdbms/mongoose_rdbms.erl
+++ b/src/rdbms/mongoose_rdbms.erl
@@ -673,7 +673,7 @@ connect(Settings, Retry, RetryAfter, MaxRetryDelay) ->
         Error ->
             SleepFor = rand:uniform(RetryAfter),
             Backend = mongoose_rdbms_backend:backend_name(),
-            ?ERROR_MSG("connection attempt to ~s resulted in ~p."
+            ?ERROR_MSG("Connection attempt to ~s resulted in ~p."
                        " Retrying in ~p seconds.", [Backend, Error, SleepFor]),
             timer:sleep(timer:seconds(SleepFor)),
             NextRetryDelay = RetryAfter * RetryAfter,
--- a/src/rdbms/mongoose_rdbms.erl
+++ b/src/rdbms/mongoose_rdbms.erl
@@ -672,8 +672,9 @@ connect(Settings, Retry, RetryAfter, MaxRetryDelay) ->
             Error;
         Error ->
             SleepFor = rand:uniform(RetryAfter),
-            ?ERROR_MSG("Database connection attempt with ~p resulted in ~p."
-                       " Retrying in ~p seconds.", [Settings, Error, SleepFor]),
+            Backend = mongoose_rdbms_backend:backend_name(),
+            ?ERROR_MSG("connection attempt to ~s resulted in ~p."
+                       " Retrying in ~p seconds.", [Backend, Error, SleepFor]),
             timer:sleep(timer:seconds(SleepFor)),
             NextRetryDelay = RetryAfter * RetryAfter,
             connect(Settings, Retry - 1, min(MaxRetryDelay, NextRetryDelay), MaxRetryDelay)
--- a/src/wpool/mongoose_wpool_mgr.erl
+++ b/src/wpool/mongoose_wpool_mgr.erl
@@ -70,13 +70,7 @@ start_link(Type) ->
 
 start(Type, Host, Tag, PoolOpts, ConnOpts) ->
     ok = ensure_started(Type),
-    try
-        gen_server:call(name(Type), {start_pool, Host, Tag, PoolOpts, ConnOpts})
-    catch
-        _:_ ->
-            %% hide connection options
-            exit({"failed to start pool", Type, Host, Tag, PoolOpts})
-    end.
+    gen_server:call(name(Type), {start_pool, Host, Tag, PoolOpts, ConnOpts}).
 
 stop(Type, Host, Tag) ->
     gen_server:call(name(Type), {stop_pool, Host, Tag}).
--- a/src/wpool/mongoose_wpool_mgr.erl
+++ b/src/wpool/mongoose_wpool_mgr.erl
@@ -70,7 +70,13 @@ start_link(Type) ->
 
 start(Type, Host, Tag, PoolOpts, ConnOpts) ->
     ok = ensure_started(Type),
-    gen_server:call(name(Type), {start_pool, Host, Tag, PoolOpts, ConnOpts}).
+    try
+        gen_server:call(name(Type), {start_pool, Host, Tag, PoolOpts, ConnOpts})
+    catch
+        _:_ ->
+            %% hide connection options
+            exit({"failed to start pool", Type, Host, Tag, PoolOpts})
+    end.
 
 stop(Type, Host, Tag) ->
     gen_server:call(name(Type), {stop_pool, Host, Tag}).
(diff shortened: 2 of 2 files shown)


mongoose-im:
[7776.1](https://travis-ci.org/esl/MongooseIM/jobs/637988565) / Erlang 22.0 / small_tests / 83b214c4deb8f92b61833ac3343023098a628560
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2597/7776/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2597/7776/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-01-16_15.38.15/lib.mongooseim.logs/run.2020-01-16_15.38.16/suite.log.html)

mongoose-im:
[7780.1](https://travis-ci.org/esl/MongooseIM/jobs/638395426) / Erlang 22.0 / small_tests / a6ecc9c29c2eab92931f8c361a1a70ee2ebf7392
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2597/7780/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2597/7780/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-01-17_11.49.28/lib.mongooseim.logs/run.2020-01-17_11.49.29/suite.log.html)

src/rdbms/mongoose_rdbms.erl
```diff
@@ -672,8 +672,9 @@ connect(Settings, Retry, RetryAfter, MaxRetryDelay) ->
             Error;
         Error ->
             SleepFor = rand:uniform(RetryAfter),
-            ?ERROR_MSG("Database connection attempt with ~p resulted in ~p."
-                       " Retrying in ~p seconds.", [Settings, Error, SleepFor]),
+            Backend = mongoose_rdbms_backend:backend_name(),
+            ?ERROR_MSG("connection attempt to ~s resulted in ~p."
```
michalwski:
```suggestion
            ?ERROR_MSG("Connection attempt to ~s resulted in ~p."
```

src/wpool/mongoose_wpool_mgr.erl
```diff
@@ -70,7 +70,13 @@ start_link(Type) ->
 
 start(Type, Host, Tag, PoolOpts, ConnOpts) ->
     ok = ensure_started(Type),
-    gen_server:call(name(Type), {start_pool, Host, Tag, PoolOpts, ConnOpts}).
+    try
+        gen_server:call(name(Type), {start_pool, Host, Tag, PoolOpts, ConnOpts})
+    catch
+        _:_ ->
+            %% hide connection options
```
michalwski: I'm afraid we are hiding to much here. How do we know what was the reason and how to debug the issue if we only say that the pool failed to start?

DenysGonchar: first of all, in the most cases `gen_server:call()` fails with one of 2 reasons:
- timeout 
- gen_server dies before responding

in both cases you will rather want to debug the corresponding gen_server. not the caller.

but if you really need the exact error you can still get it using dbg w/o changing the code, see the example:
```
Eshell V10.6  (abort with ^G)
1> dbg:tracer().
{ok,<0.80.0>}
2> dbg:p(all,call).
{ok,[{matched,nonode@nohost,40}]}
3> dbg:tp(gen_server, call,[{[some_server,'_'],[],[{exception_trace}]}]).
{ok,[{matched,nonode@nohost,2},{saved,1}]}
4> catch gen_server:call(some_server,test).
(<0.78.0>) call gen_server:call(some_server,test)
(<0.78.0>) exception_from {gen_server,call,2} {exit,{noproc,{gen_server,call,[some_server,test]}}}
{'EXIT',{noproc,{gen_server,call,[some_server,test]}}}
5> catch gen_server:call(another_server,test).
{'EXIT',{noproc,{gen_server,call,[another_server,test]}}}
6>
```

DenysGonchar: unfortunately gen_server doesn't declare the format of the error, so there's no warranty that it stays unchanged from one erlang release to another

michalwski: @DenysGonchar 

> but if you really need the exact error you can still get it using dbg w/o changing the code, see the example

That's the point. In many situations we don't have access to the node, so we are not able to run `dbg` yet we are expected to help people.

src/wpool/mongoose_wpool_mgr.erl
```diff
@@ -70,7 +70,13 @@ start_link(Type) ->
...
+            exit({"failed to start pool", Type, Host, Tag, PoolOpts})
```
rslota: I'm sure we know how the password or any other secret field is actually called, so why don't we just log `maps:without(ConnOpts, [secret_1, secret_2...])`? Also, this approach looses stacktrace and converts possible `error` into `exit` (we wildcard the type of exception). 
One another thing to mention here is that using strings in error tuples makes them a bit useless, as it's quite hard to match on those if needed. 

That all being said, don't we have a better approach? How about passing secrets as functions, like "PasswordGetter" or something like that? It wouldn't come up in any logs, regardless what they are.

DenysGonchar: actually we don't know. it's passed as provided in config file. and it's different for various backends.

rslota: Actually it's hard to say where code can crash and whether some secrets are at this time somewhere in the call stack. With OTP reports, this means that the only way to make sure that secrets are not logged, ever, is to not pass them at all. Since this is not really possible, we can minimize this by e.g., like I mentioned, passing a functions instead. This makes the call stack that actually carries the secret limited to this one call that almost cannot crash and to whatever library that expects it does.

DenysGonchar: we know that these changes are not perfect, they are addressing the exact reported issue. there will be some discussion regarding the further changes regarding this topic.

rslota: > actually we don't know. it's passed as provided in config file. and it's different for various backends.

Then backends know that. I know that this is a bit out of scope, but I do think that each backend should define the configuration schema and as part of this definition, fields could be marked as "secret". 
As temporary solution, I wouldn't mind adding an optional callback to return list of possible secret fields. I do realize that this is more work, but current solution feels more like brute force solution employed in order to cover up the issue.

DenysGonchar: that will not be implemented without preliminary discussion and so will not be implemented in this PR. there is no clean opinion on this topic within the team. We anyway cannot guaranty that external libraries are not exposing secrets on crashes, so it might be that we will just create some recommendation to avoid sharing logs externally.

src/wpool/mongoose_wpool_mgr.erl
DenysGonchar: change rolled back, the further work in this area to be scheduled for the next release(s).

src/wpool/mongoose_wpool_mgr.erl
DenysGonchar: change rolled back, the further work in this area to be scheduled for the next release(s).

mongoose-im:
[7796.1](https://travis-ci.org/esl/MongooseIM/jobs/639394354) / Erlang 22.0 / small_tests / cd5935d153eed6be005612da03a982a89d1dc971
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2597/7796/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2597/7796/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-01-20_11.07.31/lib.mongooseim.logs/run.2020-01-20_11.07.32/suite.log.html)

PR #2611, opened 2020-01-27
Docs/push notifications

NelsonVides:
Here are the so much awaited docs to the new push notifications.

Have mercy if I got anything wrong 😅


doc/modules/mod_event_pusher_push.md
```diff
@@ -1,120 +1,51 @@
...
+  Read more about it [here](#plugin-module)
+* **virtual_pubsub_hosts** (list of strings, default: `[]`) - a list of "simulated"
+  Publish-Subscribe domains. See the [relevant section](#virtual-pubsub-hosts) for more details. You
+  may use the `@HOST@` pattern in the domain name. It will automatically be replaced by a respective
+  XMPP domain (e.g. `localhost`)
 
-* Push service handled by another cluster via XEP-0357
-* Push service handled locally via XEP-0357
-* PubSub-less
+## Virtual PubSub hosts
```
DenysGonchar: please remove `e.g. for publishing E2E encryption keys` statement, cause it's completely incorrect. especially when it's only `<<"push">>` plugin enabled in the config right below this statement. we may want to keep PubSub configured to ensure that old versions of the applications still can create (unique) push node, and they won't crash.

DenysGonchar: i know it's not you added that, but please remove it.

doc/modules/mod_event_pusher_push.md
```diff
@@ -1,120 +1,51 @@
...
+* **backend** (atom, default: `mnesia`) - Backend to use for storing the registrations.
+ Possible options are `mnesia` and `rdbms`.
+* **wpool** (list, default: `[]`) - List of options that will be passed to the `worker_pool` library that handles all the requests.
+ Please refer to the [Project Site](https://github.com/inaka/worker_pool) for more details.
+* **plugin_module** (atom, default: `mod_event_pusher_push_plugin_defaults`) - module implementing
+  `mod_event_pusher_push_plugin` behaviour, used for dynamic configuration of push notifications.
+  Read more about it [here](#plugin-module)
+* **virtual_pubsub_hosts** (list of strings, default: `[]`) - a list of "simulated"
+  Publish-Subscribe domains. See the [relevant section](#virtual-pubsub-hosts) for more details. You
+  may use the `@HOST@` pattern in the domain name. It will automatically be replaced by a respective
```
DenysGonchar: to ensure complete shadowing of all the PubSub subdomain we must use `@HOSTS@` pattern, otherwise only the subdomain of the user is shadowed.

doc/modules/mod_event_pusher_push.md
```diff
@@ -141,23 +72,47 @@ Please note the value of `virtual_pubsub_hosts` option.
...
+XMPP standpoint is to use the data provided with the [enable stanza][enabling]. Because of that
+each device should (re)enable the push notifications at the beginning of each and every connection.
+
+### Custom plugins
+
+A custom module implementing the optional callbacks of `mod_event_pusher_push_plugin`
+may be used as a plugin to change the default behaviour. In case of not implemented callbacks
+the defaults are used instead.
+
+### Unacknowledged message hook
```
DenysGonchar: this part must be moved to `/docs/push_notifications/doc/developers-guide/Hooks-and-handlers.md`

NelsonVides: I think you meant to `hooks_descriptions.md`, which is the one that describes a few hooks. Moved there.

Translation logic, I think that the code is way clearer than any docs, so that exact point I wasn't planning to explain, unless somebody asks me to 🤷‍♂

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,175 @@
...
+do so, the server would respond with an error stanza, hence the incompatibility with [XEP-0357][].
+They only have to provide the virtual PubSub address.
+
+```Erlang
+{mod_event_pusher, [
+    {backends, [
+        {push, [
+            {backend, mnesia}, % optional
+            {wpool, [{workers, 200}]}, % optional
+            {plugin_module, mod_event_pusher_push_plugin_defaults}, % optional, read more [here][mod_event_pusher_push_plugin]
```
gustawlippa: Are you sure the link will work in the code block?

NelsonVides: Damn, good catch, it won't! 😅

gustawlippa:
I think mkdocs.yml needs to be updated to reflect the changes to the structure

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,178 @@
...
+#### Drawbacks
+* It still all depends on how the federated server is configured. It can be a transparent custom
+  service, or another MongooseIM node with any of the previous configurations.
+
+If MongooseIM is used as the remote PubSub node, the architecture then looks like this:
+ ![](PushNotifications-MIM_as_remote_pubsub.png)
+
+The components that make push notifications possible in MongooseIM in the PubSub-full setup
+add up to the following architecture:
+![](component-architecture-with-PubSub.png)
```
gustawlippa: There is a typo in the file the link refers to. It contains `achitecture` instead of `architecture`, so the link does not work.

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,178 @@
...
+  private deployments — they need to be adapted
+* Not suitable for deployments that require connections from 3rd party client applications.
+  Notifications won't work at all in such case
+
+### XEP-0357 compliant + virtual hosts
+
+This is the most versatile setup. Allows clients to use all push notifications configurations:
+
+The components that make push notifications possible in MongooseIM add up to the following architecture:
+![](component-architecture-without-PubSub.png)
```
gustawlippa: This file has a typo with `achitecture` as well.

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,178 @@
...
+[our pubsub-less enhancement][pubsub-less].
+
+You might also want to read about the push notification's [client side configuration][client-side].
+
+All push notification mechanisms use [mod_event_pusher_push][] as a backend implementation,
+read [this][mod_event_pusher] to know more about it.
+
+While the whole setup can be incredibly extensible, we see the following straightforward uses of it.
+
+### XEP-0357 compliant with local PubSub
```
DenysGonchar: please arrange the sections in the next order:

1. XEP-0357 compliant with remote PubSub XMPP server
   - note here that the format of the push IQ is can be adjusted via the mod_event_pusher_push custom-made plugin module.
   - also mention that all the provided plugins are generating IQ that that is specific to our own `push PubSub node` implementation
2. XEP-0357 compliant with MIM used as remote PubSub XMPP server
   - state here one more time that we expect push notification IQ to have specific (not declared in XEP) custom fields, e.g. device token. So if someone wants to use with 3rd party XMPP server, he must ensure that this 3rd party server generates push IQ in that specific format. 
3. XEP-0357 compliant when PubSub is configured at the same MIM cluster and IQ is routed internally.
   - MB mention here that this was the most common use-case. but it's senseless and ineffective to generate extra routing, and in the most cases PubSub was the bottleneck of the whole setup. so we came up to an idea of PubSub-less setup.
4. PubSub-less configuration.

doc/user-guide/push-notifications/Push-notifications-client-side.md
```diff
@@ -0,0 +1,158 @@
...
+      <field var='silent'><value>false</value></field>
+      <field var='topic'><value>some_apns_topic</value></field>
+      <field var='priority'><value>some_priority</value></field>
+    </x>
+  </enable>
+</iq>
+```
+
+We have now enabled push notifications to be send to the `pubsub.mypubsub.com` domain
+on the node `punsub_node_for_my_private_iphone` created previously.
```
DenysGonchar: mention here that node name must be unique for the device. so for the PubSub-less setup the device token received from FCM/APNS can be used to generate (or mb even as is) the unique name

doc/modules/mod_event_pusher_push.md
```diff
@@ -1,128 +1,59 @@
...
 
-This module is a backend of [mod_event_pusher] that implements [XEP-0357: Push Notifications](https://xmpp.org/extensions/xep-0357.html).
-It provides push notification data to the service that delivers actual notifications to a client device.
+This module is a backend for [mod_event_pusher][] that implements
+[XEP-0357: Push Notifications](https://xmpp.org/extensions/xep-0357.html).
+It provides push notification data to the service that delivers actual notifications
+to a client device.
 
-We've prepared [a detailed tutorial](../user-guide/Push-notifications.md) about a proper push notifications setup both on a client and a server side.
+We've prepared [a detailed tutorial](../user-guide/Push-notifications.md) for a proper push
```
gustawlippa: Link changed to `../user-guide/push-notifications/Push-notifications.md`

doc/modules/mod_event_pusher_push.md
```diff
@@ -1,128 +1,59 @@
...
-* Push service handled by another cluster via XEP-0357
-* Push service handled locally via XEP-0357
-* PubSub-less
+* **backend** (atom, default: `mnesia`) - Backend to use for storing the registrations.
+ Possible options are `mnesia` and `rdbms`.
+* **wpool** (list, default: `[]`) - List of options that will be passed to the `worker_pool` library that handles all the requests.
+ Please refer to the [Project Site](https://github.com/inaka/worker_pool) for more details.
+* **plugin_module** (atom, default: `mod_event_pusher_push_plugin_defaults`) - module implementing
+  `mod_event_pusher_push_plugin` behaviour, used for dynamic configuration of push notifications.
+  Read more about it [here](#plugin-module)
```
gustawlippa: I personally like the `relevant section` approach used in line 39 instead of `here` with a link.

doc/user-guide/push-notifications/Push-notifications-client-side.md
```diff
@@ -0,0 +1,158 @@
+## Using push notifications on client side
```
DenysGonchar: I highly recommend to restructure the document:
* first provide the diagram and description of the XEP compliant PubSub based configuration.
* then provide the PubSub less diagram and describe the difference only.

src/event_pusher/mod_event_pusher_push_plugin.erl
```diff
@@ -25,15 +25,21 @@
          publish_notification/5]).
 
 
+%% callback used for filtering push notifications. A push notification is triggered for a given a
```
DenysGonchar: to many `a`s

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,178 @@
...
+```
+
+#### Virtual PubSub hosts
+
+These domains will shadow any identical domain configured for pubsub, stealing any notification
+published to it. It enables easy migration from PubSub-full deployments to PubSub-less variants.
+Read more in [mod_event_pusher_push_virtual_hosts][].
+
+#### Advantages
+* Versatility: pubsub-less and pubsub-full mechanisms can be configured with different domains and
```
gustawlippa: Minor detail, and I believe most files in the documentation lack consistency with this word, but in this line we have `pubsub-less` and in the line 142 there is `PubSub-less`. It just looks weird spelled differently in the same paragraph.

doc/user-guide/push-notifications/Push-notifications-client-side.md
```diff
@@ -0,0 +1,158 @@
...
+#### Creating a new push node
+In this example `mypubsub.com` is a domain of the MongooseIM server that has [mod_pubsub][] enabled
+with the `push` node support.  The client sends the following stanza to the server:
+
+```xml
+<iq type='set'
+    to='pubsub.mypubsub.com'
+    id='create1'>
+  <pubsub xmlns='http://jabber.org/protocol/pubsub'>
+    <create node='punsub_node_for_my_private_iphone' type='push'/>
```
gustawlippa: This is just a bad pun. I know it would work (and is probably not a real pun) and it was used throughout this file but I just noticed it. `punsub`, really?

doc/modules/mod_event_pusher_push.md
NelsonVides: That's cool 👍

doc/user-guide/push-notifications/Push-notifications.md
NelsonVides: Fun fact, the XEP is also mixing both spellings in the same paragraphs: https://xmpp.org/extensions/xep-0357.html 😂 
But I'll try going to the camel-case version anyway xD

doc/user-guide/push-notifications/Push-notifications-client-side.md
NelsonVides: It was there already, I just copied it as it was, also got confused by that "`punsub`", and then realised it is all consistently written everywhere, so 🤷‍♂ 😂

NelsonVides: It was there already, I just copied it as it was, also got confused by that "`punsub`", and then realised it is all consistently written everywhere, so 🤷‍♂ 😂

DenysGonchar:
component-architecture diagrams are not used anywhere. it makes sense to add them to `docs/push_notifications/doc/user-guide/push-notifications/With-MongoosePush.md` with explanation of the internal push notification flow.

doc/user-guide/push-notifications/Push-notifications-client-side.md
NelsonVides: @DenysGonchar what do you think about saying that in line 67, where we already mention that the node should be unique to the device?

gustawlippa:
Actually, there are still a few links that are not correct because of the changed structure. E. g. at the end of the file `user-guide/Features-and-supported-standards.md` there is a dead link to `user-guide/Push-notifications.md`
It is way easier to find those errors when you simply run `mkdocs serve` locally, as it prints out warnings. I was not aware of that, but it makes finding links to nowhere trivial.

src/event_pusher/mod_event_pusher_push_plugin.erl
```diff
@@ -25,15 +25,21 @@
          publish_notification/5]).
 
 
+%% callback used for filtering push notifications. A push notification is triggered for a given
```
gustawlippa: I've seen EDoc tags scattered throughout the MIM codebase. These three comments look edoc-ish, so if there is possible profit, perhaps a `@doc` tag would be ok here. However, I don't know what's the standard for us.

doc/user-guide/push-notifications/Push-notifications.md
DenysGonchar: actually the first and the second point can be replaced with some general XEP overview with an abstract diagram.

NelsonVides: Done in 818f62403

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,217 @@
...
+### MongooseIM as a PubSub-less XMPP server
+
+ ![](PushNotifications-no_pubsub.png)
+
+If compatibility with client libraries is not an issue, this is a really good approach.
+PubSub is completely bypassed and clients don't need to create a push node — if they attempt to
+do so, the server would respond with an error stanza, hence the incompatibility with [XEP-0357][].
+They only have to provide the virtual PubSub address in the enable stanza, and node name can be
+anything unique. In order to ensure uniqueness the APNS/FCM token can be used. Note that the token
+must be provided [as a publish options](https://xmpp.org/extensions/xep-0357.html#example-9) anyway.
```
gustawlippa: `as a publish option` or `as one of the publish options`? Or is it intentional?

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,217 @@
...
+
+A direct connection to a push service (e.g. MongoosePush) must be configured on the same MongooseIM
+node, checkout [this][WithMongoosePush] tutorial on how to setup [MongoosePush][].
```
gustawlippa: `check out` instead of `checkout`

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,217 @@
...
+#### Advantages
+* No need to use PubSub at all
+* More efficient (PubSub has a considerable impact on heavily loaded systems)
+* Simpler client-side usage
+
+#### Drawbacks
+* It is only partially XEP-0357 compliant, so 3rd party clients most probably won't work even in
+  private deployments — they need to be adapted
+* Not suitable for deployments that require connections from 3rd party client applications.
+  Notifications won't work at all in such case
```
gustawlippa: I believe it should be `in such a case`?

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,217 @@
...
+
+### XEP-0357 compliant + virtual hosts
+
+This is the most versatile setup, a compromise of both worlds.
```
gustawlippa: `compromise between` sounds better in my opinion

doc/user-guide/push-notifications/Push-notifications.md
gustawlippa: Actually this sentence appears three times in this file.

doc/user-guide/push-notifications/Push-notifications.md
NelsonVides: Doesn't really matter I think
https://english.stackexchange.com/questions/198941/in-such-case-and-in-such-a-case

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,217 @@
...
+]}
+```
+
+#### Advantages
+* No need to use PubSub at all
+* More efficient (PubSub has a considerable impact on heavily loaded systems)
+* Simpler client-side usage
+
+#### Drawbacks
+* It is only partially XEP-0357 compliant, so 3rd party clients most probably won't work even in
```
DenysGonchar: that's not true, XEP doesn't declare creation of the node as mandatory step. it declares only that PushService is represented by the `push PubSub node`. see [Business Rules](https://xmpp.org/extensions/xep-0357.html#sect-idm46544864749808) and [Enabling Notifications](https://xmpp.org/extensions/xep-0357.html#enabling) paragraphs

```
The App Server allocates or reuses a node on the App Server's XMPP Push Service
```

DenysGonchar: also according to the ASCII diagram in XEP, it doesn't have to be done via XMPP client.

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,217 @@
+## How to set up Push Notifications
+
+Push notifications are the bread and butter of the modern mobile experience, and MongooseIM has
+support for them. When used together with [MongoosePush][], you get out-of-the-box push
+notifications for FCM (**F**irebase **C**loud **M**essaging) and APNS (**A**pple **P**ush
+**N**otification **S**ervice) providers. And it's easy to extend it to any other protocols of your
+choice.
+
+You might also want to read about the push notification's [client side configuration][client-side].
```
gustawlippa: I believe it's my last complaint, but this file is only available through this one link, and when you enter it, you are nowhere in the docs tree (it is not listed in the files structure on the left). I don't have any idea how to fix this problem.
Perhaps a link in line 105 would at least make the chance of finding this file a little higher.

doc/user-guide/push-notifications/With-MongoosePush.md
```diff
@@ -0,0 +1,88 @@
...
+        {pool_name, mongoose_push_http},
+        {api_version, "v3"}]},
+    (...)
+    ]
+}.
+```
+
+Here, we assume that [MongoosePush][] will be available on the localhost on port 8443, which is the
+default one — note the `server` option in the outgoing pool definition.
+Next we enable [mod_push_service_mongoosepush][]. The first option is the name of the HTTP pool to
```
michalwski: Looks like the `mod_push_service_mongoosepush` link is not defined. When viewing the file on github (https://github.com/esl/MongooseIM/blob/docs/push_notifications/doc/user-guide/push-notifications/With-MongoosePush.md) I can see the raw text here instead of hyperlink.

doc/user-guide/push-notifications/Push-notifications-client-side.md
```diff
@@ -0,0 +1,162 @@
...
+There are just a few things the XMPP client application needs to receive the push notifications.
+One picture is worth a thousand words, hence, here you have two diagrams to follow:
+
+#### PubSub-full:
+
+ ![](pubsub_full_push_notifications_example_message_flow.png)
+
+#### PubSub-less:
+
+ ![](pubsub_less_push_notifications_example_message_flow.png)
```
michalwski: This diagram, even though the source image is of the same size as the one above, is rendered as a much bigger. This makes an impression (at least for me) that the PubSub-less configuration is more difficult then PubSub-full.

doc/user-guide/push-notifications/Push-notifications-client-side.md
DenysGonchar: or at least avoid 2 big diagrams followed one by one right at the beginning of the document.

doc/user-guide/push-notifications/With-MongoosePush.md
```diff
@@ -0,0 +1,89 @@
...
+
+If your _FCM_ app token is `MY_FCM_SECRET_TOKEN` and you have the `priv` directory with all
+certificates in the current directory, start MongoosePush with the following command:
+
+```bash
+docker run -v `pwd`/priv:/opt/app/priv \
+  -e PUSH_FCM_APP_KEY="MY_FCM_SECRET_TOKEN" \
+  -e PUSH_HTTPS_CERTFILE="/opt/app/priv/ssl/rest_cert.pem" \
+  -e PUSH_HTTPS_KEYFILE="/opt/app/priv/ssl/rest_key.pem" \
+  -it --rm mongooseim/mongoose-push:latest
```
michalwski: We already have MongoosePush 2.0.0, let's use the tag instead of `latest`.

doc/modules/mod_event_pusher_push.md
```diff
@@ -1,163 +1,122 @@
...
-        ]}
-    ]}
-]}
-```
-
-## Plugin module
+* `mod_event_pusher_push_plugin_defaults`, which implements an older behaviour. It does not notify
+  the user of unacknowledged messages immediately after detecting a lost connection to the user.
+* `mod_event_pusher_push_plugin_enhanced`, which pushes notifications as soon as the server detects
+  that the client has disconnected and waits for stream resumption (by an `unack_msg_event` event
```
goddammit:
```suggestion
  that the client has disconnected and waits for a stream resumption (by an `unack_msg_event` event
```

doc/modules/mod_event_pusher_push.md
```diff
@@ -1,163 +1,122 @@
...
+  being generated through the `unacknowledged_message` hook). The immediate notification prevents
```
goddammit:
```suggestion
   generated through the `unacknowledged_message` hook). The immediate notification prevents
```

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,155 @@
...
+
+#### Advantages
+* No need to use PubSub at all
+* More efficient (PubSub has a considerable impact on heavily loaded systems)
+* Simpler client-side usage — Read about the [client side configuration here][client-side].
+
+#### Drawbacks
+* If the client application is build to create the push PubSub node, this might require a migration
+  for such client — as he attempts to create the node, the server will answer with an IQ error
+  stanza if the domain is not shadowed (using `virtual_pubsub_hosts`). If migrating the client side
```
DenysGonchar: this is completely incorrect and must be changed. is pubsub is not configured server responds on node creation request with IQ error even if domain is shadowed! shadowing is not solving the this problem hence the need for migration tricks!

doc/modules/mod_event_pusher_push.md
```diff
@@ -1,163 +1,122 @@
...
+  the unneeded suspension of the client's application, if there are any not yet unacknowledged
```
goddammit: any not yet unacknowledged messages? something is wrong here.

doc/modules/mod_event_pusher_push.md
```diff
@@ -1,163 +1,122 @@
...
+  messages. This allows to create more power efficient mobile applications.
 
-A plugin module handles dynamic configuration of push notifications. It implements `mod_event_pusher_push_plugin` behaviour which
-requires two callbacks:
+In order for the enhanced plugin to work, each device (an entity that may receive push
+notifications) should be uniquely identified. The only correct way to identify a device from the
+XMPP standpoint is to use the data provided with the [enable stanza][enabling]. Because of that
```
goddammit:
```suggestion
XMPP standpoint is to use the data provided with the [enable stanza][enabling]. Because of that,
```

doc/user-guide/push-notifications/Push-notifications-client-side.md
```diff
@@ -0,0 +1,166 @@
+## Using push notifications on the client side
+
+There are just a few things the XMPP client application needs to receive the push notifications.
+Depending on whether you plan to use PubSub-full or PubSub-less configuration some of the steps may be unnecessary.
```
goddammit:
```suggestion
Depending on whether you plan to use PubSub-full or PubSub-less configuration, some of the steps may be unnecessary.
```

doc/user-guide/push-notifications/Push-notifications-client-side.md
```diff
@@ -0,0 +1,166 @@
...
+by either using it directly or within some custom node name generation.
+It is also important from the security perspective to configure the node with:
+
+* `access_model` set to `whitelist` so only affiliated users can access the node.
+* `publish_model` set to `publishers` so only users with `publisher` or `publisher_only` role
+can publish notifications.
+
+#### Adding the server's JID to allowed publishers
+
+Push notifications to the push node are addressed from your server JID.
```
goddammit:
```suggestion
Push notifications to the push node are addressed from your server's JID.
```
?

doc/user-guide/push-notifications/Push-notifications-client-side.md
```diff
@@ -0,0 +1,166 @@
...
+      <field var='priority'><value>some_priority</value></field>
+    </x>
+  </enable>
+</iq>
+```
+
+We have now enabled push notifications to be send to the `pubsub.mypubsub.com` domain
+on the node `punsub_node_for_my_private_iphone` created previously.
+In `publish-options` we have passed the service name we are using (`apns` or `fcm`)
+and the device token (here: `your_pns_device_token`) that you received from you push notification
```
goddammit:
```suggestion
and the device token (here: `your_pns_device_token`) that you received from your push notification
```

doc/user-guide/push-notifications/Push-notifications-client-side.md
```diff
@@ -0,0 +1,166 @@
...
+
+  * `mode` - which may be either `prod` or `dev` (default to `prod`). Decides which connection pool
+    type on [MongoosePush][] shall be used. This may be used when _APNS_ on [MongoosePush][] is
+    configured to work with both production and development certificate.
+  * `click_action` - action to perform when notification is clicked on the device. `activity` on
+    _Android_ and `category` on _iOS_. Please refer to your platform / push notification service
+    provider for more info.
+  * `topic` - currently only used with _APNS_. The value is passed to _APNS_ as `topic` header.  For
+    more information please refer to _APNS_ documentation.
+  * `silent` - if set to `true`, all notifications will be "silent". This means that only data
```
goddammit:
```suggestion
  * `silent` - if set to `true`, all notifications will be "silent". This means that only the data
```

doc/user-guide/push-notifications/Push-notifications-client-side.md
```diff
@@ -0,0 +1,166 @@
...
+    payload will be send to push notifications provider with no notification. The data payload will
```
goddammit:
```suggestion
    payload will be send to the push notifications provider with no notification. The data payload will
```

doc/user-guide/push-notifications/Push-notifications-client-side.md
```diff
@@ -0,0 +1,166 @@
...
+</iq>
+```
+
+You may skip the `node='punsub_node_for_my_private_iphone'` to globally disable push notifications
+on all nodes that are registered with your `JID`.
+This may be used to disable push notifications on all your devices.
+
+### Communication overview
+
+One picture is worth a thousand words, hence, here you have two diagrams
```
goddammit:
```suggestion
One picture is worth a thousand words, so here are two diagrams
```

doc/user-guide/push-notifications/Push-notifications-client-side.md
```diff
@@ -0,0 +1,166 @@
...
+to show the typical communication when using push notifications:
```
goddammit:
```suggestion
showing the typical communication when using push notifications:
```

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,155 @@
...
+that push notifications being a PubSub service — [we can do better than that][pubsub-less].
+
+If you're already familiar with the workings of [XEP-0357][], make sure to have a look at
+[our PubSub-less enhancement][pubsub-less].
+
+As it is always said, one picture is worth a thousand words:
+
+ ![](PushNotifications-XEP-overview.png)
+
+Who does what is highly configurable. You may use MongooseIM as the XMPP server clients connect to,
```
goddammit:
```suggestion
You may use MongooseIM as the XMPP server clients connect to
```

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,155 @@
...
+### XEP-0357 compliant with local PubSub
+
+ ![](PushNotifications-MIM_with_local_pubsub.png)
+
+This is, historically, the most common setup.
+It allows your clients to enable push notifications via a local PubSub,
+and the IQ stanza is routed internally.
+
+A direct connection to a push service (e.g. MongoosePush) must be configured on the same MongooseIM
+node, check out [this][WithMongoosePush] tutorial on how to setup [MongoosePush][].
```
goddammit:
```suggestion
node. Check out [this][WithMongoosePush] tutorial on how to setup [MongoosePush][].
```

doc/user-guide/push-notifications/Push-notifications-client-side.md
DenysGonchar: no, it's not good idea. there a statement above:
` If you're running a PubSub-less configuration, skip to this point`
and the line 67 is bypassed!

doc/user-guide/push-notifications/Push-notifications.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+#### Advantages
+* No need to use PubSub at all
+* More efficient (PubSub has a considerable impact on heavily loaded systems)
+* Simpler client-side usage — Read about the [client side configuration here][client-side].
+
+#### Drawbacks
+* If the client application is build to create the push PubSub node, this might require a migration
+  for such client — as he attempts to create the node, the server will answer with an IQ error
```
goddammit:
```suggestion
  for such client — as they attempt to create the node, the server will answer with an IQ error
```

doc/modules/mod_event_pusher_push.md
NelsonVides: Damn copy&pasting 😂

PR #2615, opened 2020-01-29
[skip ci] fix docs

NelsonVides:
Apparently readthedocs didn't build the client-side doc if it is not on the mkdocs.yml file, this should hopefully fix that


michalwski:
I think this needs to be open to the rel-3.6 branch, doesn't it?

michalwski:
Ah, no it doesn't forget about my comment.

PR #2637, opened 2020-02-18
Change ejabberd_sm hooks to use mongoose_hooks

gustawlippa:
This PR makes `ejabberd_sm` use newly written hooks in `mongoose_hooks` module. It also fixes some inconsistencies found in specs for handlers.

### DIFF
--- a/src/mongoose_hooks.erl
+++ b/src/mongoose_hooks.erl
@@ -1,6 +1,56 @@
 -module(mongoose_hooks).
 
--export([privacy_check_packet/6]).
+-export([failed_to_store_message/4,
+         offline_groupchat_message_hook/5,
+         offline_message_hook/5,
+         privacy_check_packet/6,
+         privacy_get_user_list/3,
+         roster_in_subscription/7,
+         set_presence_hook/5,
+         sm_broadcast/6,
+         sm_filter_offline_message/5,
+         sm_register_connection_hook/4,
+         sm_remove_connection_hook/6,
+         unset_presence_hook/5,
+         xmpp_bounce_message/2]).
+
+-spec failed_to_store_message(LServer, Acc, From, Packet) -> Result when
+    LServer :: jid:lserver(),
+    Acc :: mongoose_acc:t(),
+    From :: jid:jid(),
+    Packet :: exml:element(),
+    Result :: mongoose_acc:t().
+failed_to_store_message(LServer, Acc, From, Packet) ->
+    ejabberd_hooks:run_fold(failed_to_store_message,
+                            LServer,
+                            Acc,
+                            [From, Packet]).
+
+-spec offline_groupchat_message_hook(LServer, Acc, From, To, Packet) -> Result when
+    LServer :: jid:lserver(),
+    Acc :: map(),
+    From :: jid:jid(),
+    To :: jid:jid(),
+    Packet :: exml:element(),
+    Result :: map().
+offline_groupchat_message_hook(LServer, Acc, From, To, Packet) ->
+    ejabberd_hooks:run_fold(offline_groupchat_message_hook,
+                            LServer,
+                            Acc,
+                            [From, To, Packet]).
+
+-spec offline_message_hook(LServer, Acc, From, To, Packet) -> Result when
+    LServer :: jid:lserver(),
+    Acc :: map(),
+    From :: jid:jid(),
+    To :: jid:jid(),
+    Packet :: exml:element(),
+    Result :: map().
+    offline_message_hook(LServer, Acc, From, To, Packet) ->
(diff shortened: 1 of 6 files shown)


src/mongoose_hooks.erl
```diff
@@ -17,3 +67,98 @@ privacy_check_packet(LServer, Acc, User, PrivacyList, FromToNameType, Dir) ->
...
+    Result :: boolean().
+sm_filter_offline_message(LServer, Drop, From, To, Packet) ->
+    ejabberd_hooks:run_fold(sm_filter_offline_message, LServer,
+                            Drop, [From, To, Packet]).
+
+-spec sm_register_connection_hook(LServer, SID, JID, Info) -> Result when
+    LServer :: jid:lserver(),
+    SID :: 'undefined' | ejabberd_sm:sid(),
+    JID :: jid:jid(),
+    Info :: 'undefined' | [any()],
```
michalwski: I think type definition could be changed to
```suggestion
    Info :: 'undefined' | ejabberd_sm:info(),
```

src/mongoose_hooks.erl
```diff
@@ -17,3 +67,98 @@ privacy_check_packet(LServer, Acc, User, PrivacyList, FromToNameType, Dir) ->
...
+    Result :: mongoose_acc:t().
+sm_remove_connection_hook(LServer, Acc, SID, JID, Info, Reason) ->
+    ejabberd_hooks:run_fold(sm_remove_connection_hook, LServer, Acc,
+                            [SID, JID, Info, Reason]).
+-spec unset_presence_hook(LServer, Acc, LUser, LResource, Status) -> Result when
+    LServer :: jid:lserver(),
+    Acc :: mongoose_acc:t(),
+    LUser :: jid:luser(),
+    LResource :: jid:lresource(),
+    Status :: any(),
```
michalwski: As far as I know the status is a binary representing the status a user put in his presence stanza.

src/mongoose_hooks.erl
```diff
@@ -17,3 +67,98 @@ privacy_check_packet(LServer, Acc, User, PrivacyList, FromToNameType, Dir) ->
                              PrivacyList,
                              FromToNameType,
                              Dir]).
+
+-spec privacy_get_user_list(Server, UserList, User) -> Result when
+    Server :: jid:server(),
+    UserList :: any(),
```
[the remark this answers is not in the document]
gustawlippa: Yes, in all places where it's called the type is `mongoose_privacy:userlist()` but in one place when this hook is handled the input is discarded, so I wasn't sure whether to use this knowledge.

src/mongoose_hooks.erl
```diff
@@ -17,3 +67,98 @@ privacy_check_packet(LServer, Acc, User, PrivacyList, FromToNameType, Dir) ->
...
+    ejabberd_hooks:run_fold(privacy_get_user_list, Server,
+                                          UserList, [User, Server]).
+
+-spec roster_in_subscription(LServer, Acc, User, Server, From, Type, Reason) -> Result when
+    LServer :: jid:lserver(),
+    Acc :: mongoose_acc:t(),
+    User :: jid:user(),
+    Server :: jid:server(),
+    From :: jid:jid(),
+    Type :: any(),
```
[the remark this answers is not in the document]
gustawlippa: Yes, it seems it is. I was mistaken because handler `check_in_subscription` in `ejabberd_sm` expects `any()` there. This is a general question: If we know the specific type but the values are discarded, should we include this knowledge in specs? I. e. should I change spec of `check_in_subscription` to reflect the real type, even though it is not strictly connected to the logic of this function itself?

src/mongoose_hooks.erl
gustawlippa: It seems to be also correct, I will change corresponding specs in `ejabberd_sm`.

gustawlippa:
I've removed a spec for `undefined` passed to `ejabberd_sm` functions as `Info` (i.e. `open_session`), as it does not seem to be done anywhere (I've only seen an empty list passed as no info), it breaks the `session` record type as well as does not confirm to types expected by functions defined in `ejabberd_gen_sm`.
About this `session` type - why is it declared in `ejabberd_sm` instead of putting constrains on record field types in `session.hrl`?

michalwski:
>hy is it declared in ejabberd_sm instead of putting constrains on record field types in session.hrl?

Sometimes we want to allow, for instance 'undefined' or '_' as a value of a record's field but not in the type based on the record. This is useful mainly in situations when records are used to create mnesia or ets select queries. I don't have the the code at hand now, so I'm not sure if this is the case here.

Also we prefer to define types in modules and export them, instead of .hrl files which needs to be included and the name of a type becomes global (no module prefix).

Issue #2209, opened 2019-02-01
Different arities for c2s_update_presence hook


varnerac:
MongooseIM version: `master`
Installed from: source
Erlang/OTP version: 20.3.8.6

It appears that the `c2s_update_presence` hook gets called in two places with two different arities.

https://github.com/esl/MongooseIM/blob/75146781ac5561da012008a0d3906c2d5c6a90b8/src/ejabberd_c2s.erl#L875

and 

https://github.com/esl/MongooseIM/blob/75146781ac5561da012008a0d3906c2d5c6a90b8/src/ejabberd_c2s.erl#L1053-L1057

michalwski:
Hi @varnerac, thanks for the report. Linse 1053-1057 are not covered with our tests so we simply didn't catch the issue when refactoring hooks and handlers. Did you notice that issue in real life or just come across it when analyzing the code?

varnerac:
We were writing a mod that was going to use that hook. When we first wrote the callback, we wrote it with an arity of `2`. Later we realize that the arity `0` version was called. We’ll end up using another hook for our mod.

michalwski:
Thanks for the update. I'll add this issue to our backlog.

fenek:
Fixed by #2639

PR #2639, opened 2020-02-20
Change ejabberd_c2s to use mongoose_hooks

gustawlippa:
This PR updates `ejabberd_c2s` to use new `mongoose_hooks` module. I have tried to write specs for the functions from `mongoose_hooks` as correctly as possible. I have noted hooks that are good candidates for removal, because they are not handled as far as I can see:
-  `c2s_update_presence` - it was not only not handled but also called with different number of arguments. One place when this hook is fired is in `handle_info({force_update_presence, LUser}...` in `c2s`. This message is sent only in `ejabberd_sm` and is exposed in API as `force_update_presence/1`. Is there a reason for such functionality? It seems that nothing executes this code right now.
- `forbidden_session_hook`
- ` c2s_loop_debug` which is also executed with different arguments but does not seem to be handled
- `session_opening_allowed_for_user`
- `check_bl_c2s` which is supposed to be used by "c2s blacklist plugins", but I don't know if such plugins exist.

### DIFF
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -1970,10 +1970,7 @@ presence_update_to_available(false, Acc, OldPriority, NewPriority, From, Packet,
                      State :: state()) -> {mongoose_acc:t(), state()}.
 presence_track(Acc, StateData) ->
     To = mongoose_acc:to_jid(Acc),
-    From = mongoose_acc:from_jid(Acc),
     LTo = jid:to_lower(To),
-    User = StateData#state.user,
-    Server = StateData#state.server,
     case mongoose_acc:stanza_type(Acc) of
         <<"unavailable">> ->
             Acc1 = check_privacy_and_route(Acc, StateData),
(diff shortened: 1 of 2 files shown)


michalwski:
Answering to questioned hooks:

> c2s_update_presence - it was not only not handled but also called with different number of arguments. One place when this hook is fired is in handle_info({force_update_presence, LUser}... in c2s. This message is sent only in ejabberd_sm and is exposed in API as force_update_presence/1. Is there a reason for such functionality? It seems that nothing executes this code right now.

This is needed by our customers. There is an issue about it #2209 

> forbidden_session_hook

As far as I remember this was used in a functionality developed for a customer

> c2s_loop_debug which is also executed with different arguments but does not seem to be handled

It was used for debugging purposes. I think we can remove it.

> session_opening_allowed_for_user

Again, needed by a functionality developed for a customer.

> check_bl_c2s which is supposed to be used by "c2s blacklist plugins", but I don't know if such plugins exist.

There is no in open source, I'm not aware of customers code using this hook. In my opinion it's worth keeping

src/ejabberd_c2s.erl
```diff
@@ -2000,31 +1995,31 @@ presence_track(Acc, StateData) ->
             {Acc1, StateData#state{pres_i = I,
                                    pres_a = A}};
         <<"subscribe">> ->
-            Acc2 = ejabberd_hooks:run_fold(roster_out_subscription,
+            Acc2 = mongoose_hooks:roster_out_subscription(
```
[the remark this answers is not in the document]
gustawlippa: I've changed it, if you could take a look if that's what you had in mind

src/ejabberd_c2s.erl
```diff
@@ -1988,33 +1985,17 @@ presence_track(Acc, StateData) ->
...
                                    pres_a = A}};
         <<"subscribe">> ->
-            Acc2 = mongoose_hooks:roster_out_subscription(
-                                           Server,
-                                           Acc,
-                                           User, To, subscribe),
-            Acc3 = check_privacy_and_route(Acc2, jid:to_bare(From), StateData),
-            {Acc3, StateData};
+            Acc1 = process_presence_subscription_and_route(Acc, subscribe, StateData),
+            {Acc1, StateData};
```
michalwski: How about returning the tuple from inside the `process_presence_subscribed_and_route` function? Do you think it's better to keep it outside?

gustawlippa: I just wanted to keep consistency with `check_privacy_and_route`, `check_privacy_and_route_or_ignore` and `check_privacy_and_route_probe` as they all seem to be pretty similar. But sure, this function could also return this tuple.

michalwski: Ok, fair enough. Thanks.

PR #2641, opened 2020-02-21
Change mongoose_hooks api

gustawlippa:
This PR continues work done in #2639.
Its aim is to standardize APIs exposed from `mongoose_hooks` module. Change introduced in comparison to `ejabberd_hooks` is the fact that hook caller needs to pass `Server` explicitly, even if he doesn't care about XMPP domain (in such cases an atom `global` used to be inserted). There is still a possibility to run a hook without an accumulator in some cases(an equivalent of `ejabberd_hooks:run`), but such `mongoose_hooks` always execute the same `mongoose_hook` with arity higher by 1. This way, when writing a handler one will know what is the number of arguments they should accept (one less then the biggest arity for a given `mongoose_hook`).

### DIFF
--- a/src/mongoose_hooks.erl
+++ b/src/mongoose_hooks.erl
@@ -1,37 +1,39 @@
 -module(mongoose_hooks).
 
--export([auth_failed/2,
-         c2s_broadcast_recipients/6,
-         c2s_filter_packet/6,
+-export([auth_failed/3,
+         c2s_broadcast_recipients/7,
+         c2s_filter_packet/7,
          c2s_preprocessing_hook/3,
          c2s_presence_in/5,
-         c2s_stream_features/1,
-         c2s_unauthenticated_iq/4,
+         c2s_stream_features/2,
+         c2s_stream_features/3,
+         c2s_unauthenticated_iq/5,
          c2s_update_presence/2,
-         check_bl_c2s/2,
+         check_bl_c2s/3,
          failed_to_store_message/4,
          forbidden_session_hook/3,
          offline_groupchat_message_hook/5,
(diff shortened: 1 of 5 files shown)


src/mongoose_hooks.erl
```diff
@@ -1,76 +1,81 @@
...
          user_available_hook/3,
          user_receive_packet/6,
          user_sent_keep_alive/2,
          user_send_packet/5,
          xmpp_bounce_message/2,
          xmpp_send_element/3]).
 
--spec auth_failed(Server, Username) -> Result when
-    Server :: jid:server(),
+-spec auth_failed(HookServer, Username, Server) -> Result when
```
DenysGonchar: is it really possible to have `Server` non-equal to `HookServer`?

michalwski: In theory it's possible. This change aims to provide more clarity on what arguments are passed to the handlers. Currently the idea is that the first argument which is always `HookServer` is used to find the hook handlers. Then the second argument is most often the accumulator and the rest is passed to the handler.

Initially there was only one `Server` but we thought it makes writing handlers less clear.

erszcz:
I've looked at the example calls and it doesn't seem to me there's a single case of the both servers not being equal. This leads me to believe we don't have to make it a rule that `ServerHook` and `Server` ought to be separate, unless a specific hook requires that.

gustawlippa:
We've chosen to go with servers separated explicitly in `mongoose_hooks` only when hooks may require two different domains. Moreover, some accumulator has to be provided, even if caller does not care about the return value

PR #2582, opened 2019-12-26
Cleaning code

NelsonVides:
This PR is a mix of various code cleaning things, mainly guided by the idea of getting rid of redundant calls to the `jid` module, specially of calls to `stringprep` over something that was already `stringprep`-ed, for example in regards to `ejabberd_sm`. There're also some other functions that have been cleaned to use a given BIF like in `cyrsasl_plain`.

PR #2648, opened 2020-02-26
Perf/mod roster refactor

NelsonVides:
In line with https://github.com/esl/MongooseIM/pull/2582, here I continue removing a bit amount of extra stringprepping, this time from roster related functionality.

Actually the most tricky part was to reason what happens when `jid:make/3` returns `error`. Sometimes the old code was assuming that `jid:nameprep/1` or `jid:nodeprep/1` was not returning so, and it would crash on a pattern match somewhere with this assumption, so my new code assumes as well that `jid:make/3` will return a `#jid{}` record and fail in the same manner than old code anyway. When this assumption was not the case, I had to add some pattern-matching for `jid:make/3` returning error and not crash in such case.

For example we can try following `service_admin_extra_roster:add_roster_item/7`. It is a command, so it will take just a bunch of strings. After making jids for the local and the remote user, `ejabberd_auth:is_user_exists/2` would return false after stringprepping and matching on any error in `ejabberd_auth:does_user_exists/2`, hence my new `ejabberd_auth:is_user_exists/1` should pattern match on `jid:make/3` returning `error`. BUT, then, the old `service_admin_extra_roster:subscribe/8` passes the remote's user and server to `service_admin_extra_roster:build_roster_item/3`, which does `jid:to_binary(jid:make(U, S, <<"">>))`, and if `jid:make/3` returns an error, `jid:to_binary/1` would crash with a `function_clause` error. So I don't check for that and let the new `service_admin_extra_roster:build_roster_item/2` crash in the same way.

Same reasoning in `src/mongoose_client_api/mongoose_client_api_contacts.erl` for example.

By the way, in this same `service_admin_extra_roster:add_roster_item/7`, try following everything that could be called in the old implementation: we stringprep `LocalUser` I think like five times, and `User` (the remote) another two 😄

### DIFF
--- a/src/auth/ejabberd_auth.erl
+++ b/src/auth/ejabberd_auth.erl
@@ -46,6 +46,7 @@
          get_password_s/2,
          get_passterm_with_authmodule/2,
          does_user_exist/1,
+         is_user_exists/1,
          is_user_exists/2,
          is_user_exists_in_other_modules/3,
          remove_user/2,
@@ -461,6 +462,14 @@ do_get_passterm_with_authmodule(LUser, LServer) ->
 
 %% @doc Returns true if the user exists in the DB or if an anonymous user is
 %% logged under the given name
+-spec is_user_exists(JID :: jid:jid()) -> boolean().
+is_user_exists(#jid{luser = <<>>}) ->
+    false;
+is_user_exists(#jid{luser = LUser, lserver = LServer}) ->
+    do_does_user_exist(LUser, LServer);
+is_user_exists(error) ->
+    false.
+
 -spec is_user_exists(User :: jid:user(),
                      Server :: jid:server()) -> boolean().
 is_user_exists(<<"">>, _) ->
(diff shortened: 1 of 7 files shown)


mongoose-im:
[7963.1](https://travis-ci.org/esl/MongooseIM/jobs/655285393) / Erlang 22.0 / small_tests / 300eb1d0e877b75a20b3044cdc4308fe0090b21d
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2648/7963/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2648/7963/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-02-26_10.20.19/lib.mongooseim.logs/run.2020-02-26_10.20.20/suite.log.html)

src/auth/ejabberd_auth.erl
```diff
@@ -463,6 +464,14 @@ do_get_passterm_with_authmodule(LUser, LServer) ->
 
 %% @doc Returns true if the user exists in the DB or if an anonymous user is
 %% logged under the given name
+-spec is_user_exists(JID :: jid:jid()) -> boolean().
+is_user_exists(#jid{luser = <<>>}) ->
```
michalwski: Since this is a new function can we call it `does_user_exist`? I think it is more correct.

mongoose-im:
[8009.1](https://travis-ci.org/esl/MongooseIM/jobs/661187027) / Erlang 22.0 / small_tests / fdcf8d1e47def51a36208fe5de422f8f8c167f04
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2648/8009/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2648/8009/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-03-11_17.37.08/lib.mongooseim.logs/run.2020-03-11_17.37.09/suite.log.html)

PR #2649, opened 2020-02-26
Hooks cleanup

gustawlippa:
This PR follows the discussion from #2641 . It makes it mandatory to pass an accumulator to `mongoose_hooks` calls. It also adds `global` as an option for `Server` values for consistency with `ejabberd_hooks` specs.


gustawlippa:
>  There are such hooks, but we probably didn't touch them yet.

Only `ejabberd_c2s:is_ip_blacklisted` is a hook that would run with `global` option but it does not expect a Server as an argument at all, so true, there is no place to check this for now in specs. I removed this commit

PR #2650, opened 2020-02-26
Perf/ejabberd auth redundant stringprepping

NelsonVides:
Kind of a continuation over #2648, this time about `ejabberd_auth` redundant stringprepping. There's at the bottom of such file, commented out, possible exports for the old deprecated functions, if we need them (?)

### DIFF
--- a/big_tests/tests/accounts_SUITE.erl
+++ b/big_tests/tests/accounts_SUITE.erl
@@ -155,7 +155,7 @@ admin_notify(Config) ->
     [Username2, _Server2, _Pass2] = escalus_users:get_usp(Config, UserSpec2),
     [AdminU, AdminS, AdminP] = escalus_users:get_usp(Config, AdminSpec),
 
-    rpc(mim(), ejabberd_auth, try_register, [AdminU, AdminS, AdminP]),
+    rpc(mim(), ejabberd_auth, try_register, [make_jid(AdminU, AdminS), AdminP]),
     escalus:story(Config, [{admin, 1}], fun(Admin) ->
         escalus:create_users(Config, escalus:get_users([Name1, Name2])),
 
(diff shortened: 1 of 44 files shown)


big_tests/tests/accounts_SUITE.erl
```diff
@@ -386,4 +386,5 @@ watcher(Watcher) ->
 domain() ->
     ct:get_config({hosts, mim, domain}).
 
-			
+make_jid(U, S) ->
+    rpc(mim(), jid, make, [U, S, <<>>]).
```
michalwski: I think we have a helper function for this in `mongoose_helper` module in big tests.

michalwski: Moving the `make_jid/2` to the `mongoose_helper` may be handy in other places.

src/ejabberd_users.erl
```diff
@@ -173,7 +173,7 @@ code_change(_OldVsn, State, _Extra) ->
 
 -spec does_stored_user_exist(jid:luser(), jid:lserver()) -> boolean().
 does_stored_user_exist(LUser, LServer) ->
-    ejabberd_auth:is_user_exists(LUser, LServer)
+    ejabberd_auth:is_user_exists(jid:make(LUser, LServer, <<>>))
```
michalwski:
```suggestion
    ejabberd_auth:is_user_exists(jid:make_noprep(LUser, LServer, <<>>))
```

As far as I know both `LUser` and `LServer` are already stringpreped.

src/mod_disco.erl
```diff
@@ -385,10 +385,10 @@ get_sm_items(empty, From, To, _Node, _Lang) ->
 
 
 -spec is_presence_subscribed(jid:jid(), jid:jid()) -> boolean().
-is_presence_subscribed(#jid{luser=User, lserver=Server} = From,
-                       #jid{luser=LUser, lserver=LServer} = To) ->
+is_presence_subscribed(#jid{luser=User, lserver=Server} = _From,
+                       #jid{luser=LUser, lserver=LServer} = _To) ->
```
michalwski:
```suggestion
is_presence_subscribed(#jid{luser = User, lserver = Server} = _From,
                       #jid{luser = LUser, lserver = LServer} = _To) ->
```

PR #2651, opened 2020-02-27
Make s2s use hooks wrapper

gustawlippa:
I've moved the hook execution to the new wrapper module. However, all hooks except for `s2s_stream_features` are not handled as far as I can see. I removed `s2s_loop_debug` because it seems to be a counterpart of `c2s_loop_debug`, but I don't know whether the other ones can be removed. I am also not sure about naming conventions in s2s (especially `myname` and `server` kept in the state of `ejabberd_s2s_out`).

### DIFF
--- a/src/ejabberd_s2s.erl
+++ b/src/ejabberd_s2s.erl
@@ -276,10 +276,9 @@ do_route(From, To, Acc, Packet) ->
                                                   Attrs),
             NewPacket = Packet#xmlel{attrs = NewAttrs},
             #jid{lserver = MyServer} = From,
-            Acc1 = ejabberd_hooks:run_fold(s2s_send_packet,
-                                           MyServer,
-                                           Acc,
-                                           [From, To, Packet]),
+            Acc1 = mongoose_hooks:s2s_send_packet(MyServer,
+                                                  Acc,
+                                                  From, To, Packet),
             send_element(Pid, Acc1, NewPacket),
             done;
         {aborted, _Reason} ->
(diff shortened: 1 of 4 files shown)


src/ejabberd_s2s.erl
```diff
@@ -370,7 +369,7 @@ choose_pid(From, Pids) ->
                 [] -> Pids;
                 Ps -> Ps
             end,
-    % Use sticky connections based on the JID of the sender (whithout
+    % Use sticky connections based on the JID of the sender (without
```
NelsonVides: Good catch. I wouldn't mind, if we're already touching this one, if we put the parenthesis in the line below, like
```erlang
    % Use sticky connections based on the JID of the sender
    % (without the resource to ensure that a muc room always uses the same connection)
```

src/ejabberd_s2s_in.erl
```diff
@@ -202,21 +202,16 @@ wait_for_stream({xmlstreamstart, _Name, Attrs}, StateData) ->
                     send_element(StateData,
                                  #xmlel{name = <<"stream:features">>,
                                         children = SASL ++ StartTLS ++
-                                                   ejabberd_hooks:run_fold(
-                                                     s2s_stream_features,
-                                                     Server,
-                                                     [], [Server])}),
+                                                           mongoose_hooks:s2s_stream_features(Server,
+                                                                                              [])}),
```
NelsonVides: Is this indentation ok? 🤔

PR #2656, opened 2020-03-04
Rework mod_disco to use mongoose_hooks

gustawlippa:
This PR reworks `mod_disco` to use new `mongoose_hooks` module.
I tried not to change the logic of the hooks, although sometimes it seemed a good fit for refactoring (like passing `empty` atom instead of an empty list).
The `disco_sm_items` hook call is not covered by tests at all.

### DIFF
--- a/src/mongoose_hooks.erl
+++ b/src/mongoose_hooks.erl
@@ -44,7 +44,8 @@
          xmpp_bounce_message/2,
          xmpp_send_element/3]).
 
--export([roster_get_jid_info/4]).
+-export([roster_get/3,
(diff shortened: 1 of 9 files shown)


src/mongoose_hooks.erl
```diff
@@ -495,6 +504,15 @@ xmpp_send_element(Server, Acc, El) ->
 
 %% Roster related hooks
 
+%%% @doc The `roster_get' hook is called to extract user's roster.
```
NelsonVides:
```suggestion
%%% @doc The `roster_get' hook is called to extract a user's roster.
```

PR #2658, opened 2020-03-05
Change mod_amp to use mongoose_hooks

gustawlippa:
This PR changes `mod_amp` module to use new `mongoose_hooks` module.
My comments for possible future refactoring of hooks:
* `amp_notify_action_triggered` and `amp_error_action_triggered` seem to not be handled anywhere
* `amp_check_condition`, `amp_verify_support` and `amp_check_packet` seem to only be handled in AMP modules. Also `amp_check_packet` call contains a weird comment which I do not fully understand,
* `amp_determine_strategy` has specs that suggest that `To` argument may be `undefined` but only one handler reacts to this correctly (`determine_strategy` in `amp_strategy`) and handlers`determine_amp_strategy` in `mod_offline` and `mod_mam` may fail because of that. I do not know whether it is possible for `To` to really be undefined so I did not know whether to change specs in this case.
* `xmpp_stanza_dropped` seem to be only used for `mongoose_metrics` but also it seems to be skipped when creating metrics, so in the end not really used at all. Perhaps I am missing something here.

### DIFF
--- a/src/amp_strategy.erl
+++ b/src/amp_strategy.erl
@@ -26,7 +26,7 @@ determine_strategy(_, _, To, _, Event) ->
                   'expire-at' = undefined}.
 
 %% @doc This strategy will never be matched by any amp_rules.
-%% Use it as a seed parameter to ejaberd_hooks:run_fold
+%% Use it as a seed parameter to `mongoose_hooks'.
 -spec null_strategy() -> amp_strategy().
 null_strategy() ->
     #amp_strategy{deliver = undefined,
--- a/src/mod_amp.erl
+++ b/src/mod_amp.erl
@@ -88,7 +88,7 @@ check_packet(Acc, #jid{lserver = Host} = From, Event) ->
             % this hook replaces original element with something modified by amp
             % which is a hack, but since we have accumulator here we have a chance
             % to fix implementation
-            ejabberd_hooks:run_fold(amp_check_packet, Host, Acc, [From, Event]);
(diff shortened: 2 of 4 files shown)


chrzaszcz:
>     * `amp_notify_action_triggered` and `amp_error_action_triggered` seem to not be handled anywhere
This is correct.

>     * `amp_check_condition`, `amp_verify_support` and `amp_check_packet` seem to only be handled in AMP modules. Also `amp_check_packet` call contains a weird comment which I do not fully understand,
I agree, there are too many hooks there, this is an example of an overly complicated solution that has never fully grown because there was no need for it. Regarding the comment, the intention was that maybe we could avoid modifying the content of the stanza in this AMP hook and store the temporary value in the accumulator instead.

>     * `amp_determine_strategy` has specs that suggest that `To` argument may be `undefined` but only one handler reacts to this correctly (`determine_strategy` in `amp_strategy`) and handlers`determine_amp_strategy` in `mod_offline` and `mod_mam` may fail because of that. I do not know whether it is possible for `To` to really be undefined so I did not know whether to change specs in this case.
If `To` is undefined, I don't think we could end up calling `mod_offline` or `mod_mam` hooks - there is no recipient.
 
>     * `xmpp_stanza_dropped` seem to be only used for `mongoose_metrics` but also it seems to be skipped when creating metrics, so in the end not really used at all. Perhaps I am missing something here.
I think that the corresponding `xmppStanzaDropped` metric is created - I am not sure where it is missing.

PR #2666, opened 2020-03-06
Change roster related hook executions to use mongoose_hooks

gustawlippa:
I believe `roster_push`  should be called from `mod_roster:push_item/5` not `/4` . It is used for metrics and now it seems it is executed only when roster versioning is not enabled, but I didn't change the logic for now.

### DIFF
--- a/src/mongoose_hooks.erl
+++ b/src/mongoose_hooks.erl
@@ -26,10 +26,6 @@
          privacy_updated_list/4,
          push_notifications/4,
          resend_offline_messages_hook/3,
-         roster_get_subscription_lists/3,
-         roster_get_versioning_feature/2,
-         roster_in_subscription/7,
-         roster_out_subscription/5,
          session_opening_allowed_for_user/3,
          set_presence_hook/5,
          sm_broadcast/6,
@@ -46,9 +42,16 @@
          xmpp_send_element/3,
          xmpp_stanza_dropped/5]).
 
--export([roster_get/3,
+-export([roster_get/4,
+         roster_get_jid_info/4,
+         roster_get_subscription_lists/3,
+         roster_get_versioning_feature/2,
          roster_groups/2,
-         roster_get_jid_info/4]).
+         roster_in_subscription/7,
+         roster_out_subscription/5,
+         roster_process_item/2,
+         roster_push/4,
+         roster_set/5]).
 
 -export([is_muc_room_owner/4,
          can_access_identity/4,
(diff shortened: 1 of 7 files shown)


src/mongoose_hooks.erl
```diff
@@ -592,6 +551,80 @@ roster_get_jid_info(HookServer, InitialValue, LUser, RemBareJID) ->
...
+%%% @doc The `roster_process_item' hook is called when a user's roster is set.
+-spec roster_process_item(Server, Item) -> Result when
+    Server :: jid:lserver(),
+    Item :: mod_roster:roster(),
+    Result :: mod_roster:roster().
+roster_process_item(Server, Item) ->
+    ejabberd_hooks:run_fold(roster_process_item, Server, Item, [Server]).
+
+%%% @doc The `roster_push' hook is called when a roster item is being pushed and roster versioning is not enabled.
+-spec roster_push(Server, Acc, From, Item) -> Result when
```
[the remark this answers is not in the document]
gustawlippa: For now I changed the name to LServer

src/mongoose_hooks.erl
```diff
@@ -582,15 +541,93 @@ roster_groups(Server, Acc) ->
...
+    Reason :: any(),
+    Result :: mongoose_acc:t().
+roster_in_subscription(LServer, Acc, User, Server, From, Type, Reason) ->
+    ejabberd_hooks:run_fold(
+        roster_in_subscription,
+        LServer,
+        Acc,
+        [User, Server, From, Type, Reason]).
+
+%%% @doc The `roster_out_subscription' hook is a user sends out subsciption.
```
chrzaszcz: Minor: missing 'called to determine if'

src/mongoose_hooks.erl
```diff
@@ -582,15 +541,93 @@ roster_groups(Server, Acc) ->
...
+%%% @doc The `roster_get_versioning_feature' hook is called to determine if roster versioning is enabled.
+-spec roster_get_versioning_feature(Server, Acc) -> Result when
+    Server :: jid:server(),
+    Acc :: [exml:element()],
+    Result :: [exml:element()].
+roster_get_versioning_feature(Server, Acc) ->
+    ejabberd_hooks:run_fold(roster_get_versioning_feature,
+                            Server, Acc, [Server]).
+
+%%% @doc The `roster_in_subscription' hook is a subscription presence is routed to a user.
```
chrzaszcz: Minor: missing 'called to determine if'

src/mongoose_hooks.erl
gustawlippa: Oops, missed it. Is the sentence itself correct?

PR #2669, opened 2020-03-09
Change `ejabberd_*` modules to use mongoose_hooks

gustawlippa:
This PR changes `ejabberd_*` modules to use `mongoose_hooks`. I did not add many doc comments because I wasn't sure about the meaning of the hooks, this should be added later.
The `ejabberd_ctl_process` has handlers added dynamically in `ejabberd_ctl:register_commands` but this part of code is not covered by tests, so I don't know whether it is used.

### DIFF
--- a/src/mongoose_hooks.erl
+++ b/src/mongoose_hooks.erl
@@ -14,10 +14,14 @@
          c2s_unauthenticated_iq/4,
          c2s_update_presence/2,
          check_bl_c2s/2,
+         ejabberd_ctl_process/2,
          failed_to_store_message/4,
          forbidden_session_hook/3,
+         host_config_update/4,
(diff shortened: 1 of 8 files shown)


src/mongoose_hooks.erl
```diff
@@ -425,6 +491,14 @@ sm_remove_connection_hook(LServer, Acc, SID, JID, Info, Reason) ->
 unacknowledged_message(Server, Acc, JID) ->
     ejabberd_hooks:run_fold(unacknowledged_message, Server, Acc, [JID]).
 
+%%% @doc The `unregister_subhost' hook is called when a component is unregistered for ejabberd_router.
```
chrzaszcz: Minor: 'unregistered from'

PR #2673, opened 2020-03-10
Rework remaining hooks

gustawlippa:
This PR addresses changes remaining modules to use the new `mongoose_hooks` module.


src/auth/ejabberd_auth_anonymous.erl
```diff
@@ -183,12 +183,15 @@ unregister_connection(Acc, SID, #jid{luser = LUser, lserver = LServer}, _, _) ->
...
 
 
-%% @doc Launch the hook to purge user data only for anonymous users
+%% @doc Launch the hook to purge user data only for anonymous users.
 -spec purge_hook(boolean(), jid:luser(), jid:lserver()) -> 'ok'.
 purge_hook(false, _LUser, _LServer) ->
     ok;
 purge_hook(true, LUser, LServer) ->
-    ejabberd_hooks:run(anonymous_purge_hook, LServer, [LUser, LServer]).
+    Acc = mongoose_acc:new(#{ location => ?LOCATION,
```
michalwski: Good catch! I noticed that this hooks in often handled by the same handlers as for `remove_user`.

gustawlippa: Yes, exactly. I believe `mod_roster:remove_user/3` would not expect anything other than a `mongoose_acc` there, it even has a helper function `mod_roster:remove_user/2` which creates an accumulator for testing purposes. Sadly, execution of the `anonymous_purge_hook` is not covered by tests.

src/mongoose_hooks.erl
```diff
@@ -433,6 +494,24 @@ push_notifications(Server, Acc, NotificationForms, Options) ->
 register_subhost(Acc, LDomain, IsHidden) ->
     ejabberd_hooks:run_fold(register_subhost, Acc, [LDomain, IsHidden]).
 
+%%% @doc The `register_user' hook is called when a user is successfully registered in a authentication backend.
```
chrzaszcz: Minor: **an** authentication backend

PR #2677, opened 2020-03-18
Refactoring of nested case statements

janciesla8818:
Code refactoring for the cyrsasl_scram module to improve code quality and reduce amount of nested case statements.

### DIFF
--- a/src/sasl/cyrsasl_scram.erl
+++ b/src/sasl/cyrsasl_scram.erl
@@ -58,130 +58,163 @@ mech_new(_Host, Creds) ->
     {ok, #state{step = 2, creds = Creds}}.
 
 mech_step(#state{step = 2} = State, ClientIn) ->
-    case re:split(ClientIn, <<",">>, [{return, binary}]) of
-        [_CBind, _AuthorizationIdentity, _UserNameAttribute, _ClientNonceAttribute, ExtensionAttribute | _]
-          when ExtensionAttribute /= [] ->
-            {error, <<"protocol-error-extension-not-supported">>};
-        [CBind, _AuthorizationIdentity, UserNameAttribute, ClientNonceAttribute | _]
-          when (CBind == <<"y">>) or (CBind == <<"n">>) ->
-            case parse_attribute(UserNameAttribute) of
-                {error, Reason} -> {error, Reason};
-                {_, EscapedUserName} ->
-                    case unescape_username(EscapedUserName) of
-                        error -> {error, <<"protocol-error-bad-username">>};
-                        UserName ->
-                            case parse_attribute(ClientNonceAttribute) of
-                                {$r, ClientNonce} ->
-                                    Creds = State#state.creds,
-                                    LServer = mongoose_credentials:lserver(Creds),
-                                    case ejabberd_auth:get_passterm_with_authmodule(UserName, LServer) of
-                                        {false, _} -> {error, <<"not-authorized">>, UserName};
-                                        {Ret, AuthModule} ->
-                                            {StoredKey, ServerKey, Salt, IterationCount} =
-                                            if is_tuple(Ret) -> Ret;
-                                               true ->
-                                                   TempSalt =
-                                                   crypto:strong_rand_bytes(?SALT_LENGTH),
-                                                   SaltedPassword =
-                                                   mongoose_scram:salted_password(Ret,
-                                                                                  TempSalt,
-                                                                                  mongoose_scram:iterations()),
-                                                   {mongoose_scram:stored_key(mongoose_scram:client_key(SaltedPassword)),
-                                                    mongoose_scram:server_key(SaltedPassword),
-                                                    TempSalt,
-                                                    mongoose_scram:iterations()}
-                                            end,
-                                            {NStart, _} = binary:match(ClientIn, <<"n=">>),
-                                            ClientFirstMessageBare = binary:part(ClientIn,
-                                                                                 {NStart,
-                                                                                  byte_size(ClientIn)-NStart}),
-                                            ServerNonce =
-                                            jlib:encode_base64(crypto:strong_rand_bytes(?NONCE_LENGTH)),
-                                            ServerFirstMessage =
-                                            iolist_to_binary(
-                                              [<<"r=">>,
-                                               ClientNonce,
-                                               ServerNonce,
-                                               <<",s=">>,
-                                               jlib:encode_base64(Salt),
-                                               <<",i=">>,
-                                               integer_to_list(IterationCount)]),
-                                            {continue, ServerFirstMessage,
-                                             State#state{step = 4, stored_key = StoredKey,
-                                                         server_key = ServerKey,
-                                                         auth_message =
-                                                         <<ClientFirstMessageBare/binary,
-                                                           ",", ServerFirstMessage/binary>>,
-                                                         client_nonce = ClientNonce,
-                                                         server_nonce = ServerNonce,
-                                                         username = UserName,
-                                                         auth_module = AuthModule}}
-                                    end;
-                                _Else -> {error, <<"not-supported">>}
-                            end
-                    end
+    try
+        % Check received attributes
+        ClientInList = re:split(ClientIn, <<",">>, [{return, binary}]),
+        ok = check_extension_attribute(ClientInList),
+        [CBind, _AuthIdentity, UserNameAtt, ClientNonceAtt | _] = ClientInList,
+        ok = check_message_headers(CBind),
+        EscapedUserName = check_username(parse_attribute(UserNameAtt)),
+        UserName = get_username(EscapedUserName),
+        ClientNonce = get_client_nonce(parse_attribute(ClientNonceAtt)),
+        % Authenticate and prepare response message
+        Creds = State#state.creds,
+        LServer = mongoose_credentials:lserver(Creds),
+        {AuthModule, {StoredKey, ServerKey,
+                      Salt, IterationCount}} = get_scram_attributes(UserName, LServer),
+        {NStart, _} = binary:match(ClientIn, <<"n=">>),
+        ClientFirstMessageBare = binary:part(ClientIn,
+                                            {NStart, byte_size(ClientIn) - NStart}),
+        ServerNonce = jlib:encode_base64(crypto:strong_rand_bytes(?NONCE_LENGTH)),
+        ServerFirstMessage = create_server_first_message(ClientNonce, ServerNonce,
+                                                         Salt, IterationCount),
+
+        {continue, ServerFirstMessage,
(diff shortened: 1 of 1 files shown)


mongoose-im:
[8024.1](https://travis-ci.org/esl/MongooseIM/jobs/663897384) / Erlang 22.0 / small_tests / 26e7764bcde93950d2709bad51263447e5402c45
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2677/8024/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2677/8024/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-03-18_11.36.45/lib.mongooseim.logs/run.2020-03-18_11.36.45/suite.log.html)

src/sasl/cyrsasl_scram.erl
```diff
@@ -58,164 +58,196 @@ mech_new(_Host, Creds) ->
...
-                                    end;
-                                _Else -> {error, <<"not-supported">>}
-                            end
-                    end
+    try
+        % Check received attributes
+        ClientInList = re:split(ClientIn, <<",">>, [{return, binary}]),
+        ok = check_extension_attribute(ClientInList),
+        [CBind, _AuthIdentity, UserNameAtt, ClientNonceAtt | _] = ClientInList,
+        ok = check_message_headers(CBind),
```
NelsonVides: For example, these three lines could be unified as
```erlang
        {UserNameAtt, ClientNonceAtt} = check_extension_attribute(ClientInList),
```
if `check_extension_attribute/1` was implemented as
```erlang
check_extension_attribute([_,_,_,_, Extension | _]) when Extension /= [] ->
    erlang:error(<<"protocol-error-extension-not-supported">>);
check_extension_attribute([CBind | _])
  when (CBind =/= <<"y">>) andalso (CBind =/= <<"n">>) ->
    erlang:error(<<"bad-protocol">>);
check_extension_attribute([_CBind, _AuthIdentity, UserNameAtt, ClientNonceAtt | _]) ->
    {UserNameAtt, ClientNonceAtt}.
```
But I'm being very opinionated here, just giving ideas 😄

src/sasl/cyrsasl_scram.erl
```diff
@@ -58,164 +58,196 @@ mech_new(_Host, Creds) ->
...
-                       true -> {error, <<"bad-channel-binding">>}
-                    end;
-                _Else -> {error, <<"bad-protocol">>}
+    try
+        % Checked received arrtibutes
+        ClientInList = re:split(ClientIn, <<",">>),
+        [GS2ChannelBindingAtt,
+         NonceAtt,
+         ClientProofAtt] = check_attributes(ClientInList),
+        ok = check_channel_binding_attribute(GS2ChannelBindingAtt),
```
NelsonVides: Similar thing here than before 🤔

src/sasl/cyrsasl_scram.erl
```diff
@@ -58,164 +58,196 @@ mech_new(_Host, Creds) ->
...
     end;
-parse_attribute(<<Char, Rest/binary>>, PName) ->
-    case Char == $= of
-        true ->
-            {PName, Rest};
-        _ ->
-            {error, <<"bad-format first char not a letter">>}
-    end.
-
+parse_attribute(<<Char, Rest/binary>>, PName) when Char == $= ->
```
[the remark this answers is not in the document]
janciesla8818: sure

mongoose-im:
[8041.1](https://travis-ci.org/esl/MongooseIM/jobs/666026768) / Erlang 22.0 / small_tests / 74557f9391d0db1288059f025a280d569c52e7e0
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2677/8041/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2677/8041/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-03-23_19.33.50/lib.mongooseim.logs/run.2020-03-23_19.33.51/suite.log.html)

src/sasl/cyrsasl_scram.erl
```diff
@@ -58,164 +58,194 @@ mech_new(_Host, Creds) ->
...
-                                                           ",", ServerFirstMessage/binary>>,
-                                                         client_nonce = ClientNonce,
-                                                         server_nonce = ServerNonce,
-                                                         username = UserName,
-                                                         auth_module = AuthModule}}
-                                    end;
-                                _Else -> {error, <<"not-supported">>}
-                            end
-                    end
+    ClientInList = re:split(ClientIn, <<",">>, [{return, binary}]),
```
NelsonVides:
```suggestion
    ClientInList = binary:split(ClientIn, <<",">>, [global]),
```

So, I think we don't need a regex for something so simple, a binary split will do with much much better performance:
```erlang
10> Cs = [ crypto:strong_rand_bytes(100) || _ <- lists:seq(1,1000) ].
13> lists:all(fun(X) -> X == true end, [ re:split(C, <<",">>, [{return, binary}]) == binary:split(C, <<",">>, [global]) || C <- Cs ]).
true
14> lists:all(fun(X) -> X == true end, [ re:split(C, <<",">>, [{return, binary}]) == binary:split(C, <<",">>, [global]) || C <- Cs ]).
true
15> timer:tc(fun() ->   [re:split(C, <<",">>, [{return, binary}]) || C <- Cs], ok   end).
{8322,ok}
16> timer:tc(fun() ->   [binary:split(C, <<",">>, [global]) || C <- Cs], ok   end).
{3688,ok}
```
Or even in the case of a single element:
```erlang
21> C = crypto:strong_rand_bytes(100).
22> timer:tc(fun() ->   binary:split(C, <<",">>, [global]), ok   end).
{19,ok}
23> timer:tc(fun() ->   re:split(C, <<",">>, [{return, binary}]), ok   end).
{25,ok}
```

I believe this to work, as `mech_step` is called only from `cyrsasl:server_start/3`, which is called from `ejabberd_c2s` by passing something that we can be sure is a binary and nothing else, as it simply is the output of `base64:mime_decode/1`.

Would be cool to make sure nevertheless that I'm passing the correct options to the `binary:split/3` function, but that could be as easy as just throwing it to travis 😛

src/sasl/cyrsasl_scram.erl
```diff
@@ -58,164 +58,194 @@ mech_new(_Host, Creds) ->
...
-                                           end;
-                                       _Else -> {error, <<"bad-protocol">>}
-                                   end;
-                               {$r, _} -> {error, <<"bad-nonce">>};
-                               _Else -> {error, <<"bad-protocol">>}
-                           end;
-                       true -> {error, <<"bad-channel-binding">>}
-                    end;
-                _Else -> {error, <<"bad-protocol">>}
+    ClientInList = re:split(ClientIn, <<",">>),
```
NelsonVides: Same applies here I believe:
```suggestion
    ClientInList = binary:split(ClientIn, <<",">>, [global]),
```

src/sasl/cyrsasl_scram.erl
```diff
@@ -58,164 +58,194 @@ mech_new(_Host, Creds) ->
...
     end.
 
-parse_attribute(Attribute) ->
-    AttributeLen = byte_size(Attribute),
-    case AttributeLen >= 3 of
-        true ->
-            parse_attribute(Attribute, 0);
-        _ ->
-            {error, <<"bad-format attribute too short">>}
+parse_attributes(ClientInList, Steps) ->
```
NelsonVides: This one would be cool if it can have some specification, more important than all the others, considering the magic this one is doing 🤔 
The specification could look something like
```erlang
-spec parse_attributes( ... , fun((...) -> ...)) -> ok | {error, term()}.
```
The only problem is that the functions being called have different input types, which makes it not possible to specify a `TList`. Is it possible to unify all input types for the handlers? Otherwise just
```erlang
 -spec parse_attributes(any(), fun()) -> ok | {error, term()}.
```
🤷‍♂

src/sasl/cyrsasl_scram.erl
```diff
@@ -58,164 +58,194 @@ mech_new(_Host, Creds) ->
...
+                ok ->
+                    ServerSignature =
+                        mongoose_scram:server_signature(State#state.server_key,
+                                                        AuthMessage),
+                    R = [{username, State#state.username}, {sasl_success_response,
+                         <<"v=", (jlib:encode_base64(ServerSignature))/binary>>},
+                         {auth_module, State#state.username}],
+                    {ok, mongoose_credentials:extend(State#state.creds, R)};
+                {error, Reason} ->
+                    {error, Reason}
```
NelsonVides: Why don't we extract this part to another function as well? That's already two nested cases, not too much but still, not flat 🤔 
Similar idea in the step 2, what do you think?

janciesla8818: I wanted to separate validation of client input and authentication verification. The problem with using single fold (and avoiding the second case statement) would be that we'd need to carry all the arguments though all the functions. I think that with what we have now is a balance between nesting and code readability. I wanted to leave as much core functionality in mech_step as possible. Reading mech_step alone we could have some basic understanding to what is being done without going though all the functions.

src/sasl/cyrsasl_scram.erl
janciesla8818: I changed to the binary:split with the global option and it works locally.

mongoose-im:
[8045.1](https://travis-ci.org/esl/MongooseIM/jobs/666338746) / Erlang 22.0 / small_tests / ff90f4b5927db91a4e3cf570c3294d62cd1ee537
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2677/8045/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2677/8045/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-03-24_13.44.52/lib.mongooseim.logs/run.2020-03-24_13.44.53/suite.log.html)

mongoose-im:
[8046.1](https://travis-ci.org/esl/MongooseIM/jobs/666352777) / Erlang 22.0 / small_tests / 334a94d6700052fc7b23bc867aef84830c0b6a82
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2677/8046/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2677/8046/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-03-24_14.20.16/lib.mongooseim.logs/run.2020-03-24_14.20.16/suite.log.html)

PR #2689, opened 2020-03-26
Fix mod_ping tests so that they pass at the first go

bartekgorny:
There was some tricky issue: another test, ran just a little while before, did not respond to pings, waited a few seconds, then killed the connection and checked metrics. It was all fine, but the test waited for ping_req_timeout + 1, which incidentally is equal to ping_interval. Server sent one more request, which incremented metrics a few seconds later, which happened be exactly when another test was run, thus breaking that one. The second run of the other tests was ok.

### DIFF
--- a/big_tests/tests/mod_ping_SUITE.erl
+++ b/big_tests/tests/mod_ping_SUITE.erl
@@ -223,7 +223,7 @@ server_ping_pang(ConfigIn) ->
         fun(Alice) ->
                 wait_for_ping_req(Alice),
                 %% do not resp to ping req
-                ct:sleep(timer:seconds(ping_req_timeout() + 1)),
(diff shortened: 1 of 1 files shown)


mongoose-im:
[8059.1](https://travis-ci.org/esl/MongooseIM/jobs/667248918) / Erlang 22.0 / small_tests / 3a7559c30b998445441fd2f012e1d64a7a6e7043
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2689/8059/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2689/8059/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-03-26_13.19.01/lib.mongooseim.logs/run.2020-03-26_13.19.02/suite.log.html)

PR #2703, opened 2020-04-03
Extend scram sha rdbms

janciesla8818:
Adding multi scram support for rdbms backend.

### DIFF
--- a/src/auth/ejabberd_auth_rdbms.erl
+++ b/src/auth/ejabberd_auth_rdbms.erl
@@ -286,8 +286,13 @@ get_password(LUser, LServer) ->
             Password; %%Plain password
         {selected, [{_Password, PassDetails}]} ->
             case mongoose_scram:deserialize(PassDetails) of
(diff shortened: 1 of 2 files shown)


src/auth/ejabberd_auth_rdbms.erl
```diff
@@ -287,8 +287,13 @@ get_password(LUser, LServer) ->
             Password; %%Plain password
         {selected, [{_Password, PassDetails}]} ->
             case mongoose_scram:deserialize(PassDetails) of
-                {ok, Scram} ->
-                    mongoose_scram:scram_to_tuple(Scram);
+                {ok, Scram} when is_record(Scram, scram) ->
```
michalwski, line 290: How about we return the new map from `mongoose_scram:deserialize`?

janciesla8818: This was possible only after doing rebase with the base branch as deserialize was done there after this PR was created. Now it's fixed.

PR #2709, opened 2020-04-09
Transport consistency

NelsonVides:
`ejabberd_c2s` is calling `starttls` in two different ways: in one, it
calls it through the `mongoose_transport` behaviour, which in turns calls
`starttls` on the module that was given to the first as a parameter, and
on the second it calls it directly over such module, which seems like an
inconsistent way to call such function.

All the modules we have that implement mongoose_transport actually
implement two `starttls` callbacks, and even in some, like in
`mod_websockets`, we mention in the comments that some callbacks are just
to make them compatible with `ejabberd_sockets`, but the `starttls/3`
callback is not required in the behaviour definition. So that's a first
code cleaning.

### DIFF
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -381,8 +381,8 @@ determine_features(SockMod, #state{tls = TLS, tls_enabled = TLSEnabled,
     case can_use_tls(SockMod, TLS, TLSEnabled) of
         true ->
             case TLSRequired of
-                true -> [starttls(required)];
-                _    -> [starttls(optional)] ++ OtherFeatures
+                true -> [starttls_stanza(required)];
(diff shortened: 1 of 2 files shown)


mongoose-im:
[8130.1](https://travis-ci.org/esl/MongooseIM/jobs/673042533) / Erlang 22.0 / small_tests / 659b2ac79e6da0c27d4b50c4d0efa8f60c2a6663
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2709/8130/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2709/8130/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-04-09_15.36.15/lib.mongooseim.logs/run.2020-04-09_15.36.16/suite.log.html)

PR #2724, opened 2020-04-29
privacy blocking iq should not stop stanzas sent by my server to me

bartekgorny:
Previously, blocking all incoming iqs also cut off communication with the server, also in legitimate cases - like receiving an iq result from my own request. This is not required by XEP-0199, and it broke some features, like mod_ping. There was also a race condition - when a user sets a privacy lists, his c2s process routes two messages - iq result and a broadcast, which also sets the list in the same process. Whether the result came back or not, depended on which of these two messages was handled first.

This PR skips privacy checks for iq stanzas sent to a user by his own server account or by the server itself.

### DIFF
--- a/src/mongoose_privacy.erl
+++ b/src/mongoose_privacy.erl
@@ -61,15 +61,33 @@ privacy_check_packet(Acc0, Server, User, PrivacyList, From, To, Dir) ->
                        _ ->
                            {Acc0, mongoose_acc:stanza_name(Acc0), mongoose_acc:stanza_type(Acc0)}
                    end,
-    % check if it is there, if not then set default and run a hook
-    Key = {cached_check, Server, User, From, To, Name, Type, Dir},
-    case mongoose_acc:get(privacy, Key, undefined, Acc) of
-        undefined ->
-            Acc1 = mongoose_hooks:privacy_check_packet(Server, Acc, User, PrivacyList,
-                                                       {From, To, Name, Type}, Dir),
-            Res = mongoose_acc:get(hook, result, Acc1),
-            {mongoose_acc:set(privacy, Key, Res, Acc1), Res};
-        Res ->
-            {Acc, Res}
+    case basic_check(Name, From, To, PrivacyList) of
+        allow ->
+            {Acc, allow};
+        check ->
+            % check if it is there, if not then set default and run a hook
+            Key = {cached_check, Server, User, From, To, Name, Type, Dir},
+            case mongoose_acc:get(privacy, Key, undefined, Acc) of
+                undefined ->
+                    Acc1 = mongoose_hooks:privacy_check_packet(Server, Acc, User, PrivacyList,
+                                                               {From, To, Name, Type}, Dir),
+                    Res = mongoose_acc:get(hook, result, Acc1),
+                    {mongoose_acc:set(privacy, Key, Res, Acc1), Res};
+                Res ->
+                    {Acc, Res}
+            end
     end.
 
+%% allow iq result send in response to my own request
+basic_check(<<"iq">>,
+            #jid{luser = U, lserver = S} = _From,
+            #jid{luser = U, lserver = S} = _To,
+            _) -> allow;
+%% allow iqs sent to me by my server (privacy is meant to block other users)
+basic_check(<<"iq">>,
+            #jid{luser = <<>>, lserver = S} = _From,
+            #jid{lserver = S} = _To,
+            _) -> allow;
+%% do not bother checking if privacy list is empty
+basic_check(_, _, _, #privacy{lists = []}) -> allow;
+basic_check(_, _, _, _) -> check.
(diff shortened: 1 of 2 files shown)


mongoose-im:
[8205.1](https://travis-ci.org/esl/MongooseIM/jobs/680976050) / Erlang 22.0 / small_tests / 544818e2e1a4073193648ea77ed2f6906ee9656e
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2724/8205/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2724/8205/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-04-29_12.14.01/lib.mongooseim.logs/run.2020-04-29_12.14.02/suite.log.html)

src/mongoose_privacy.erl
```diff
@@ -61,15 +61,33 @@ privacy_check_packet(Acc0, Server, User, PrivacyList, From, To, Dir) ->
...
+            #jid{luser = U, lserver = S} = _From,
+            #jid{luser = U, lserver = S} = _To,
+            _) -> allow;
+%% allow iqs sent to me by my server (privacy is meant to block other users)
+basic_check(<<"iq">>,
+            #jid{luser = <<>>, lserver = S} = _From,
+            #jid{lserver = S} = _To,
+            _) -> allow;
+%% do not bother checking if privacy list is empty
+basic_check(_, _, _, #privacy{lists = []}) -> allow;
```
NelsonVides, line 92: I think this doesn't get a `#privacy{}` record, but rather a `#userlists{}` one. It's on the specs of the caller, and also codecov shows this line as never hit, with is a sign 🤔

bartekgorny: true

NelsonVides: Also, `#userlists{lists = []}` is already ignored in the handler in `mod_privacy:check_packet/`, in case you missed it. I'm not saying it's bad to duplicate logic, but maaybe it's worth giving it a thought 🤔

bartekgorny: indeed it is, but some modules use `mongoose_privacy:check_packet` directly, so this is a more appropriate place to make this check.

bartekgorny: true, again. I'd put it here, though, if only for more visibility.

chrzaszcz:
> Previously, blocking all incoming iqs also cut off communication with the server, also in legitimate cases - like receiving an iq result from my own request. This is not required by XEP-0199, and it broke some features, like mod_ping. There was also a race condition - when a user sets a privacy lists, his c2s process routes two messages - iq result and a broadcast, which also sets the list in the same process. Whether the result came back or not, depended on which of these two messages was handled first.

XEP-0199 is about Ping, the one for privacy lists is XEP-0016, which describes the IQ blocking feature is described in the following way:
**Server-side privacy lists enable a user to block incoming IQ stanzas from other entities based on the entity's JID, roster group, or subscription status (or globally).**

As an entity is a user OR any service, I don't see a reason to exclude any IQ stanzas from the blocking feature. I think that if we want to change this, we would need to get an opinion on the  [XSF mailing list](standards@xmpp.org) first.

Regarding the race condition, could we add a test for it first and make sure this test fails randomly to prove that it really exists?

bartekgorny:
> XEP-0199 is about Ping, the one for privacy lists is XEP-0016, 

Of course, my mistake.

> As an entity is a user OR any service, I don't see a reason to exclude any IQ stanzas from the blocking feature. I think that if we want to change this, we would need to get an opinion on the [XSF mailing list](standards@xmpp.org) first.

Yup, this is exactly why I expected a discussion. The XEP doesn't specifically state which IQs should be blocked and which shouldn't, so there is a need for interpretation. The purpose of privacy lists is to avoid unwanted communication, like malicious users or servers. Blocking IQs from my own server in effect breaks some other xeps (like mod_ping), but also RFC 6120 (which says that the server should reply to an IQ with a result or an error) and RFC 6121 (roster push). Even more, it breaks the same XEP, which specifically states that if you edit a privacy list you should receive a iq result.

Having said that, I'm not opposed to consulting XSF if there is a need for it.

> Regarding the race condition, could we add a test for it first and make sure this test fails randomly to prove that it really exists?

Not really, because the messages always arrived in the same order, so technically speaking it was not a race condition since one horse was winning it all the time. However, a minor change in code layout was sufficient to reverse the order, so the race must have been pretty tight.

chrzaszcz:
Actually I think that this functionality needs to be discussed first with the team before implementation. There are some new rules introduced, that were not here before: which IQs to allow and which ones to block. I am not against such changes, but I think they need to be discussed first.

Neustradamus:
@bartekgorny, @NelsonVides, @chrzaszcz: Have you progressed on this PR?

bartekgorny:
@chrzaszcz @NelsonVides Can we get back to this one? To me it seems still valid, though if you guys decide otherwise we can close this PR.

chrzaszcz:
Closing because of inactivity. The PR got outdated. We can update and reopen (or open a new PR) if the feature is needed. Currently I don't see demand for it.

PR #2726, opened 2020-04-30
Message retraction: discovery

chrzaszcz:
Add the message retraction feature to service discovery
- For the main XMPP service, register it when enabled for mod_mam
- For the MUC service, register it when enabled for mod_mam_muc

Also:
- Make the MUC test actually check MAM features
  (muc service check is left as it can be useful to ensure MUC works)
- Remove stanza printout from the MAM test as we have stanza_log
- Check for both supported MAM namespaces

Notes:
- When message retraction is disabled in MAM, retraction messages are still stored although they do not cause any change to the original messages. The motivation is to allow the clients to handle the retraction message themselves, when they are back online. In this case there seems to be no reason for advertising of message retraction by the server.

### DIFF
--- a/src/mam/mod_mam_utils.erl
+++ b/src/mam/mod_mam_utils.erl
@@ -47,7 +47,8 @@
          decode_optimizations/1,
          form_borders_decode/1,
          form_decode_optimizations/1,
-         is_mam_result_message/1]).
+         is_mam_result_message/1,
+         features/2]).
 
 %% Forms
 -export([
@@ -126,6 +127,7 @@
 
 -include("mod_mam.hrl").
 -include("mongoose_rsm.hrl").
+-include("mongoose_ns.hrl").
 
 -define(MAYBE_BIN(X), (is_binary(X) orelse (X) =:= undefined)).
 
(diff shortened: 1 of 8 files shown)


src/mam/mod_mam_utils.erl
```diff
@@ -677,6 +679,11 @@ is_mam_namespace(?NS_MAM_04) -> true;
 is_mam_namespace(?NS_MAM_06) -> true;
 is_mam_namespace(_)          -> false.
 
+features(Module, Host) ->
+    [?NS_MAM_04, ?NS_MAM_06] ++ case has_message_retraction(Module, Host) of
+                                    true -> [?NS_RETRACT];
+                                    false -> []
+                                end.
```
michalwski, line 686: I'm not fun (actually against) mixing results from `case` clauses with other operations. Can this block of code be changed to sth like the following?
```suggestion
    MAMFeatures = [?NS_MAM_04, ?NS_MAM_06],
    case has_message_retraction(Module, Host) of
        true -> [?NS_RETRACT | MAMFeatures];
        false -> MAMFeatures
    end.
```

chrzaszcz: Yes. The only question is: can `?NS_RETRACT` stay at the end? I think it's good if the list starts from the MAM features.

michalwski: Sure, it can stay at the end.

PR #2735, opened 2020-05-08
initialise session state with a hook call, so that e.g. handlers can …

bartekgorny:
…set their initial state


NelsonVides:
Closing this one and the subsequent two PRs because most of this work was already done in the latest `mongoose_c2s` rework. Only wonder if anything else is on these PRs that wasn't done in such rework, if so please open an issue or let me know to keep note of it 🙂

PR #2736, opened 2020-05-08
Refactor mod privacy as handler

bartekgorny:
Includes:
- removing the sm broadcasting (remote hook call used instead)
- moving nearly all privacy and blocking related code to dedicated modules (with just two lines left, because privacy iq handler is a bit unusual)
- minor changes to privacy tests (in a few cases iq result is now received)


NelsonVides:
Closing this one and the subsequent two PRs because most of this work was already done in the latest `mongoose_c2s` rework. Only wonder if anything else is on these PRs that wasn't done in such rework, if so please open an issue or let me know to keep note of it 🙂

PR #2738, opened 2020-05-12
Restore previous ordering of roster-related stanzas

bartekgorny:
This PR introduces a new API function of ejabberd_c2s, called `send_to_local_user`. It is meant to be used by modules which want to send a stanza to the user of the current c2s process. Instead of using `ejabberd_router:route`, they can now send a stanza immediately, which helps ensure proper ordering of stanzas. 

Tests are reverted to what they were before mod_roster refactoring, and they are supposed to pass.

### DIFF
--- a/src/admin_extra/service_admin_extra_roster.erl
+++ b/src/admin_extra/service_admin_extra_roster.erl
@@ -54,6 +54,10 @@
                                    _Group}.
 -type subs() :: atom() | binary().
 
+-type push_action() :: remove
+                       | none
+                       | {add, Nick :: binary(), Subs :: subs(),
+   Group :: binary() | string()}.
 -type delete_action() :: {'delete', Subs :: [atom()], Asks :: [atom()],
                                     [jid:user()], Contacts :: [binary()]}.
 -type list_action() :: {'list', Subs :: [atom()], Asks :: [atom()],
@@ -175,7 +179,8 @@ add_rosteritem(LocalUser, LocalServer, User, Server, Nick, Group, Subs) ->
     end.
 
 %% @doc returns result of mnesia or rdbms transaction
--spec subscribe(LocalUser :: jid:user(),
+%% add to roster directly, without generating a roster push (in case we do bulk inserts)
+-spec subscribe_silent(LocalUser :: jid:user(),
                 LocalServer :: jid:server(),
                 User :: jid:user(),
                 Server :: jid:server(),
@@ -183,11 +188,25 @@ add_rosteritem(LocalUser, LocalServer, User, Server, Nick, Group, Subs) ->
                 Group :: binary() | string(),
                 Subs :: subs(),
                 _Xattrs :: [jlib:binary_pair()]) -> any().
+subscribe_silent(LU, LS, User, Server, Nick, Group, SubscriptionS, _Xattrs) ->
+    ItemEl = build_roster_item(User, Server, {add, Nick, SubscriptionS, Group}),
+    QueryEl = #xmlel{ name = <<"query">>,
+                      attrs = [{<<"xmlns">>, <<"jabber:iq:roster">>}],
+                      children = [ItemEl]},
+    mod_roster:set_items(LU, LS, QueryEl).
+
+-spec subscribe(LocalUser :: jid:user(),
+                LocalServer :: jid:server(),
+                User :: jid:user(),
+                Server :: jid:server(),
+                Nick :: binary(),
+                Group :: binary() | string(),
+                Subs :: subs(),
+                _Xattrs :: [jlib:binary_pair()]) -> ok | error.
 subscribe(LU, LS, User, Server, Nick, Group, SubscriptionS, _Xattrs) ->
     Owner = jid:make(LU, LS, <<>>),
     PeerBin = jid:to_binary(jid:make(User, Server, <<>>)),
     mod_roster:set_roster_entry(Owner, PeerBin, Nick, [Group], binary_to_existing_atom(SubscriptionS, utf8)).
-
 
 -spec delete_rosteritem(LocalUser :: jid:user(),
                         LocalServer :: jid:server(),
@@ -304,7 +323,7 @@ subscribe_roster({Name, Server, Group, Nick}, [{Name, Server, _, _} | Roster]) -
     subscribe_roster({Name, Server, Group, Nick}, Roster);
 %% Subscribe Name2 to Name1
 subscribe_roster({Name1, Server1, Group1, Nick1}, [{Name2, Server2, Group2, Nick2} | Roster]) ->
-    subscribe(Name1, Server1, Name2, Server2, Nick2, Group2, <<"both">>, []),
+    subscribe_silent(Name1, Server1, Name2, Server2, Nick2, Group2, <<"both">>, []),
     subscribe_roster({Name1, Server1, Group1, Nick1}, Roster).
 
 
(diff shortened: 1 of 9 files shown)


mongoose-im:
[8233.1](https://travis-ci.org/esl/MongooseIM/jobs/686068702) / Erlang 22.0 / small_tests / 28a5bcf6308c8531c6b7a7f564d83724d424ce43
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2738/8233/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2738/8233/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-05-12_11.35.56/lib.mongooseim.logs/run.2020-05-12_11.35.56/suite.log.html)

erszcz:
> send_to_local_user. It is meant to be used by modules which want to send a stanza to the user of the current c2s process. Instead of using ejabberd_router:route, they can now send a stanza immediately, which helps ensure proper ordering of stanzas.

This description is great. Maybe we should put it as a `@doc` just above `send_to_local_user` definition in the `.erl` file?

src/admin_extra/service_admin_extra_roster.erl
```diff
@@ -175,20 +179,35 @@ add_rosteritem(LocalUser, LocalServer, User, Server, Nick, Group, Subs) ->
     end.
 
 %% @doc returns result of mnesia or rdbms transaction
--spec subscribe(LocalUser :: jid:user(),
+%% add to roster directly, without generating a roster push (in case we do bulk inserts)
+-spec subscribe_silent(LocalUser :: jid:user(),
```
michalwski, line 183: Looks like this is something new we didn't have before. What's the reason behind this new function and where exactly is it used?

src/mod_roster.erl
```diff
@@ -886,7 +886,7 @@ send_presence_type(From, To, Type) ->
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-%% used for testing only - should be removed
+%% used by ejabberdctl
```
michalwski, line 889: As far as I noticed, the script is now named `mongooseimctl`, unless you mean the suite `ejabberdctl_SUITE`?

src/admin_extra/service_admin_extra_roster.erl
bartekgorny: `mongooseimctl` (more specifically `service_admin_extra_roster`) adds roster entries in two ways: either one by one or in bulk. When it adds one by one tests expect roster push to be sent; for bulk inserts push should not be sent. Previously, `service_admin_extra_roster` used `mod_roster:set_items` for all operations, and for single inserts it sent iq push itself. Since now it is `mod_roster` that does the pushing, service_admin_extra_roster has to call the right api (`mod_roster:set_roster_entry` if it wants the push to be sent, or `mod_roster:set_items` if it wants it silent. Hence the split into `subscribe` and `subscribe_silent`.

src/mod_roster.erl
bartekgorny: True. It is somewhat confusing because all test suites still call it `ejabberdctl`.

bartekgorny:
> This makes me think that probably the new way of delivering stanzas ignores or breaks stream management. 

Looks like this, indeed.

src/admin_extra/service_admin_extra_roster.erl
michalwski: All right, thanks for the explanation!

mongoose-im:
[8241.1](https://travis-ci.org/esl/MongooseIM/jobs/686530306) / Erlang 22.0 / small_tests / e0759a476ebd817849276b29dce4b30324dfd9ce
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2738/8241/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2738/8241/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-05-13_11.52.15/lib.mongooseim.logs/run.2020-05-13_11.52.16/suite.log.html)

mongoose-im:
[8242.1](https://travis-ci.org/esl/MongooseIM/jobs/686538744) / Erlang 22.0 / small_tests / 1d281dc29b485b370927255d23a31c6508adb4d6
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2738/8242/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2738/8242/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-05-13_12.17.13/lib.mongooseim.logs/run.2020-05-13_12.17.14/suite.log.html)

src/ejabberd_c2s.erl
```diff
@@ -1182,6 +1184,13 @@ handle_incoming_message({run_remote_hook, HandlerName, Args}, StateName, StateDa
                        [HandlerName, Args, HandlerState, E]),
             fsm_next_state(StateName,
                            StateData);
+        {update_buffers, NewHandlerState, NewStateData} ->
+            % if a handler sent some stanzas directly it has to pass updated state like this
+            % so that we update stream management buffers
```
michalwski, line 1189: Stream management is not the only thing to consider here. How about the `csi` buffers when the client using [XEP-0352: CSI](https://xmpp.org/extensions/xep-0352.html) puts the session into inactive?

michalwski: Also, our tests are not ideal, many things are not tested with stream management enabled, not to mention `csi`. I wonder if there is a way to enforce the correct return value indicating the need for updating some parts of the state in case the handler uses the `ejabberd_c2s:send_to_local_user`.

chrzaszcz: Maybe we need to relax the separation here and just allow the handlers to access and modify the full c2s state? The calls to `ejabberd_c2s_state:get_handler_state(HandlerName, StateData)` and `ejabberd_c2s_state:set_handler_state(HandlerName, NewHandlerState, StateData1)` could happen in the handlers. Some of them may not even need their state AFAIR. I think this would be still better than trying to enumerate all possibilities in the case clauses here.

bartekgorny: I considered returning the whole state, but the design decision was to not do it... But it seems safer indeed, and shouldn't hurt.

michalwski: Right, when making the decision we didn't have in mind delivering to a local user, now considering that I'm in favour of relaxing the separation. Only this will introduce more changes to the entire refactoring at this point, right?

mongoose-im:
[8244.1](https://travis-ci.org/esl/MongooseIM/jobs/686594525) / Erlang 22.0 / small_tests / 057015d1c9e3ea0280d74ef860868303d5347625
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2738/8244/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2738/8244/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-05-13_14.22.31/lib.mongooseim.logs/run.2020-05-13_14.22.32/suite.log.html)

NelsonVides:
Closing this one as it is not relevant anymore after all the lastest `mongoose_c2s` rework, thanks for all the effort 🙂

PR #2753, opened 2020-05-25
Improve parallelism of AMP tests

chrzaszcz:
AMP test suites check a matrix of conditions, e.g.
- Actions like `notify`, `error`, `drop`
- Delivery conditions like `direct`, `stored`, `none` and their combinations

This increases the number of test cases. To avoid coding every intersection of these conditions, there is a need to generate some of them. Currently this is achieved by nesting test groups, but this prevents them from running in parallel. Long group names need to be generated, what makes the test setup less readable as well.

The proposed solution is to have a new common test hook (see https://erlang.org/doc/man/ct_hooks.html) to set up a list of Config's for each test case. This would speed up the tests significantly by running them in parallel, avoid nesting groups too much and eliminate the need for hardcoding or generating long group names.

This approach could be used for more test suites.

Other changes:
- Fix module setup - it was redundantly performed for nested groups, storing the module config multiple times and then restoring it multiple times.
- Remove some unused code
- Fix double "fresh" prefixes for Alice in some tests.
- Wait for Alice to disconnect before checking for an incoming message in Bob's archive - after increasing test parallelism this bug caused random failures, as Bob got the message while Alice was still online.

### DIFF
--- a/big_tests/tests/amp_big_SUITE.erl
+++ b/big_tests/tests/amp_big_SUITE.erl
@@ -19,56 +19,63 @@
 suite() ->
     require_rpc_nodes([mim]) ++ escalus:suite().
 
-all() -> [{group, Group} || Group <- enabled_group_names()].
-
-enabled_group_names() ->
-    [basic, offline] ++
-    case mongoose_helper:is_rdbms_enabled(domain()) of
-        true -> [mam];
-        false -> []
-    end.
+all() ->
+    [{group, G} || G <- main_group_names(), is_enabled(G)].
 
 groups() ->
-    Gs = ([{basic, [parallel], [{group, G} || G <- subgroup_names()] ++ basic_test_cases()},
-           {mam, [], [{group, mam_success},
-                      {group, mam_failure}]},
-           {mam_success, [], [{group, G} || G <- subgroup_names()]},
-           {mam_failure, [], [{group, G} || G <- subgroup_names()]},
-           {offline, [], [{group, offline_success},
-                          {group, offline_failure}]},
-           {offline_success, [], [{group, G} || G <- subgroup_names()]},
-           {offline_failure, [], [{group, G} || G <- subgroup_names()]}
-          ] ++
-          [{G, [parallel, shuffle], notify_deliver_test_cases()}
-           || G <- notify_deliver_group_names()] ++
-          [{G, [parallel, shuffle], error_deliver_test_cases()}
-           || G <- error_deliver_group_names()] ++
-          [{G, [parallel, shuffle], drop_deliver_test_cases()}
-           || G <- drop_deliver_group_names()]),
-    ct_helper:repeat_all_until_all_ok(Gs).
+    group_spec(main_group_names()).
 
+is_enabled(mam) -> mongoose_helper:is_rdbms_enabled(domain());
+is_enabled(_) -> true.
 
-subgroup_names() -> notify_deliver_group_names() ++
-                    error_deliver_group_names() ++
-                    drop_deliver_group_names().
+%% Group definitions
 
-notify_deliver_group_names() ->
-    [notify_deliver_none,
-     notify_deliver_direct,
-     notify_deliver_stored,
-     notify_deliver_none_direct_stored].
+main_group_names() ->
+    [basic, mam, offline].
 
-error_deliver_group_names() ->
-    [error_deliver_none,
-     error_deliver_direct,
-     error_deliver_stored,
-     error_deliver_none_direct_stored].
+subgroups(mam) -> [mam_success, mam_failure];
+subgroups(offline) -> [offline_success, offline_failure];
+subgroups(_) -> [].
 
-drop_deliver_group_names() ->
-    [drop_deliver_none,
-     drop_deliver_direct,
-     drop_deliver_stored,
-     drop_deliver_none_direct_stored].
+group_spec(Groups) when is_list(Groups) ->
+    lists:flatmap(fun group_spec/1, Groups);
+group_spec(Group) ->
+    case subgroups(Group) of
+        [] -> [{Group, [parallel], test_cases(Group)}];
+        SubGroups -> [{Group, [{group, SubG} || SubG <- SubGroups]} | group_spec(SubGroups)]
+    end.
+
+test_cases(Group) ->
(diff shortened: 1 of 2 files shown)


mongoose-im:
[8295.1](https://travis-ci.org/esl/MongooseIM/jobs/690833934) / Erlang 22.0 / small_tests / 40542a78a86cf37000782c1b4101e3262e48efb1
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2753/8295/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2753/8295/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-05-25_07.11.43/lib.mongooseim.logs/run.2020-05-25_07.11.44/suite.log.html)

mongoose-im:
[8296.1](https://travis-ci.org/esl/MongooseIM/jobs/690891773) / Erlang 22.0 / small_tests / 44a637bafad0451d214ddb557c89e31b3e5c5064
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2753/8296/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2753/8296/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-05-25_10.53.07/lib.mongooseim.logs/run.2020-05-25_10.53.08/suite.log.html)

mongoose-im:
[8297.1](https://travis-ci.org/esl/MongooseIM/jobs/690895144) / Erlang 22.0 / small_tests / 94dbb1b5294f976e48de5e41452ed395dc4e2f3a
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2753/8297/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2753/8297/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-05-25_11.24.48/lib.mongooseim.logs/run.2020-05-25_11.24.48/suite.log.html)

chrzaszcz:
Results on CI:
- Before: `amp_big_SUITE` took 51 s to execute
- After: it took 16.8 s

big_tests/tests/amp_big_SUITE.erl
```diff
@@ -85,39 +92,48 @@ basic_test_cases() ->
...
      drop_deliver_to_offline_user_test,
      drop_deliver_to_stranger_test].
 
-init_per_suite(C) ->
+%% Setup and teardown
+
+init_per_suite(Config) ->
     rpc(mim(), ejabberd_config, add_local_option, [{{s2s_host, <<"not a jid">>}, domain()}, deny]),
-    escalus:init_per_suite(C).
+    ConfigWithHooks = [{ct_hooks, [{multiple_config_cth, fun tests_with_config/1}]} | Config],
```
michalwski, line 116: How does it work with the `ct_hooks` defined in `default.spec` file? Are they still respected even if here we specify one element list?

big_tests/tests/amp_big_SUITE.erl
```diff
@@ -647,6 +668,23 @@ last_rule_applies_test(Config) ->
...
+        0 ->
+            ok;
+        _ when Timeout > 0 ->
+            ct:log("Waiting ~p ms for MongooseIM to close the sessions for ~s@~s", [Timeout, U, S]),
+            ct:sleep(1000),
+            wait_until_no_session(U, S, Timeout - 1000);
+        _ ->
+            ct:fail("MongooseIM failed to close the session for ~s@~s")
+    end.
+
```
michalwski, line 687: There is `mongoose_helper:wait_until` function which helps to write such kind of waiting tests. Do you think it could be used here?

michalwski:
The only disadvantage I can see with this new approach is that we lose some information about details of the test case which previously could be inferred from the group name. Do you think the `pre_test_case` or another hook could add some information to the `comment` column in the summary? Would it make sense?

chrzaszcz:
> The only disadvantage I can see with this new approach is that we lose some information about details of the test case which previously could be inferred from the group name. Do you think the `pre_test_case` or another hook could add some information to the `comment` column in the summary? Would it make sense?

In the generated HTML test log you can see this in green:

```
*** User 2020-05-26 11:11:55.721 ***
Extra test config: [{rules,[{deliver,direct,notify}]}]
```

It works for me, do you think it is enough?

big_tests/tests/amp_big_SUITE.erl
chrzaszcz: https://erlang.org/doc/apps/common_test/ct_hooks_chapter.html#overriding-cths

"By default, each installation of a CTH causes a new instance of it to be activated."

I even traced the tests to make sure it works :-)

big_tests/tests/amp_big_SUITE.erl
chrzaszcz: This code is indeed ported from a very old commit, I will update this.

big_tests/tests/amp_big_SUITE.erl
michalwski: All right, thanks for the explanation!

michalwski:
> In the generated HTML test log you can see this in green:
>```
>*** User 2020-05-26 11:11:55.721 ***
>Extra test config: [{rules,[{deliver,direct,notify}]}]
>```
>It works for me, do you think it is enough?

Yeah, I saw that. I think it's enough, adding this or similar info to the comment column in the table summarising all tests may be more difficult to do.

mongoose-im:
[8300.1](https://travis-ci.org/esl/MongooseIM/jobs/691249668) / Erlang 22.0 / small_tests / 5d07ca88bef053fceddbeb2b7992a2af2ac71dd3
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2753/8300/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2753/8300/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-05-26_10.22.36/lib.mongooseim.logs/run.2020-05-26_10.22.37/suite.log.html)

PR #1195, opened 2017-02-23
This header must be included in all valid CORS responses

Zatvobor:
This PR addresses #1045 

> Access-Control-Allow-Origin (required) - This header must be included in all valid CORS responses; omitting the header will cause the CORS request to fail. (HTML5 Rocks)

PR #2772, opened 2020-06-03
Allow extra HTTP headers in BOSH and WebSockets.

varnerac:
This is a WIP PR

Proposed changes include:
* Allowing extra headers for BOSH and WebSockets

This isn't ready yet. I just wanted to test the waters and see if you had an appetite for a patch like this. We want the ability to add CORS headers and security headers like HSTS to the BOSH and WebSockets endpoints.

### DIFF
--- a/rel/files/mongooseim.cfg
+++ b/rel/files/mongooseim.cfg
@@ -163,10 +163,16 @@
         {transport_options, [{max_connections, 1024}]},
         {{{https_config}}}
         {modules, [
-            {"_", "/http-bind", mod_bosh},
+            {"_", "/http-bind", mod_bosh, [
+            %% Uncomment to enable HTTP custom headers. Header names must be
+            %% lowercase.
+            % {headers,[{"access-control-request-origin", "*"}]}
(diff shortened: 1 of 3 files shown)


michalwski:
Hi @varnerac, thanks for the patch. There are appetite and need for the CORS headers, there is even another PR #1195 where the headers are always added in case of BOSH connection. I think making them configurable would be a more flexible solution.

When it comes to your changes, they don't integrate well with cowboy yet. All the tests connecting users over BOSH or WS failed. You can try it locally by running the following commands:
```bash
tools/test-runner.sh --skip-small-tests --skip-cover --preset internal_mnesia -- ws
```
or
```bash
tools/test-runner.sh --skip-small-tests --skip-cover --preset internal_mnesia -- bosh
```

More on the `test-runner.sh` the script you can find in the help provided by the tool and https://mongooseim.readthedocs.io/en/latest/developers-guide/Testing-MongooseIM/

varnerac:
I'll tackle this in another PR

PR #2785, opened 2020-07-07
Logger/lager to logger

NelsonVides:
This PR basically attempts to introduce a working logger configuration to MIM, without yet fixing tests and other helper functionality that depend on the old lager.

Work goes to a feature branch called `otp_logger`

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -1,4 +1,34 @@
 [
+ {kernel, [
+  {logger_level, warning},
+  {logger, [
+    %% Console logger
+    {handler, default, logger_std_h, #{}},
+    %% Disk logger for errors
+    {handler, disk_log, logger_disk_log_h,
+       #{config => #{
+           file => "{{mongooseim_log_dir}}/mongooseim.log",
+           type => wrap,
+           max_no_files => 5,
+           max_no_bytes => 2097152,
+           sync_mode_qlen => 2000, % If sync_mode_qlen is set to the same value as drop_mode_qlen,
+           drop_mode_qlen => 2000, % synchronous mode is disabled. That is, the handler always runs
+           flush_qlen => 5000,     % in asynchronous mode, unless dropping or flushing is invoked.
+           burst_limit_enable => true, % Default
+           burst_limit_max_count => 20, % Default
+           burst_limit_window_time => 500, % Default
+           overload_kill_enable => true,
+           overload_kill_qlen => 20000, % Default
+           overload_kill_mem_size => 3000000, % Default
+           overload_kill_restart_after => 5000 % Default
+         },
+         formatter => {logger_formatter, #{
+           depth => 12,
+           chars_limit => 1024
+         }}
+        }
+    }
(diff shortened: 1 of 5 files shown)


rel/files/app.config
```diff
@@ -1,4 +1,34 @@
...
+    {handler, default, logger_std_h, #{}},
+    %% Disk logger for errors
+    {handler, disk_log, logger_disk_log_h,
+       #{config => #{
+           file => "{{mongooseim_log_dir}}/ejabberd.log",
+           type => wrap,
+           max_no_files => 5,
+           max_no_bytes => 2097152,
+           sync_mode_qlen => 2000,
+           drop_mode_qlen => 2000,
```
michalwski, line 15: How exactly this work? Seeing them set to the same value I can think that log messages will be dropped the moment logging switches to the sync mode.

rel/files/app.config
```diff
@@ -1,4 +1,34 @@
 [
+ {kernel, [
+  {logger_level, warning},
+  {logger, [
+    %% Console logger
+    {handler, default, logger_std_h, #{}},
+    %% Disk logger for errors
+    {handler, disk_log, logger_disk_log_h,
+       #{config => #{
+           file => "{{mongooseim_log_dir}}/ejabberd.log",
```
michalwski, line 10: One more thing, can we finally rename the file to `mongooseim.log` ? :)

rel/files/app.config
michalwski: Ok, I just learned that from your doc:
> If sync_mode_qlen is set to the same value as drop_mode_qlen, synchronous mode is disabled. That is, the handler always runs in asynchronous mode, unless dropping or flushing is invoked.

I think this info could be added as a comment to the logger configuration.

rel/files/app.config
NelsonVides: Absolutely! 😃 
There were just some tests depending on the name, but that's fixable anyway 😛

rel/files/app.config
```diff
@@ -1,4 +1,34 @@
...
+           sync_mode_qlen => 2000, % If sync_mode_qlen is set to the same value as drop_mode_qlen,
+           drop_mode_qlen => 2000, % synchronous mode is disabled. That is, the handler always runs
+           flush_qlen => 5000,     % in asynchronous mode, unless dropping or flushing is invoked.
+           burst_limit_enable => true, % Default
+           burst_limit_max_count => 20, % Default
+           burst_limit_window_time => 500, % Default
+           overload_kill_enable => true,
+           overload_kill_qlen => 20000, % Default
+           overload_kill_mem_size => 3000000, % Default
+           overload_kill_restart_after => 5000 % Default
```
chrzaszcz, line 23: Why do we need to explicitly set the default values?

NelsonVides: I thought so for readability, so that we know the exact expected behaviour of the logging system. I could also just leave a link to the documentation now that I think about it 🤔

PR #2786, opened 2020-07-08
TOML config parser

chrzaszcz:
The goal of this PR is to provide a first POC implemenation of the TOML config parser.

**Done:**
- Simplest code that handles the typical options. All options from the `pgsql_mnesia` preset are supported.
- Both config formats supported at once, TOML takes precendence over the `cfg` file. The paths need to be the same and the configurability is limited on purpose - there is no environment variable for the TOML config format as we mignht end up supporting only one format soon.
- Handle of the happy path when the config file is configured properly.

**To do:**
- Reduce the amount of copy+paste and replacing the ad-hoc structure of the config processing with a more structured approach.
- Refactor the code for improved readability.
- Add or drop the support for config reload (most likely the former).
- Add specs.
- Template the TOML config file for testing and for the release.
- Test the TOML config file on CI.
- Validate the options and handle errors in a unified way.
- If both formats need to be supported, split `mongooseim_config_parser` into two modules: API and the backend which is interchangeable with the TOML parser.
- Use `binary_to_existing_atom` instead of `binary_to_atom`. This is not done yet as some atoms appearing in the config files are not existing yet at the time of parsing.

**Note for the reviewers:**
- Start with `mongooseim.toml` - it's the most important file right now as the format is taking the final shape at the moment.
- If possible, do not mention the things already listed as to do unless there is some important matter.

**How to test it?**
It was tested by uncommenting the `copy` operation for `mongooseim.toml` in `rebar.config`. The tests were performed with:

`./tools/test-runner.sh --one-node --db pgsql redis --skip-small-tests --preset pgsql_mnesia --`

Failing tests - all of them were expected:
- service_mongoose_system_metrics_SUITE - 1 test, requires mim3
- conf_reload_SUITE - 12 tests, reload not supported
- connect_SUITE - 72 tests, reload not supported
- component_SUITE - 21 tests, reload not supported
- login_SUITE - 14 tests, reload not supported
- metrics_api_SUITE - 3 tests, requires mim2
- rest_client_SUITE - 3 tests, requires mim3
- sasl_external_SUITE - 34 tests, reload not supported


rel/files/mongooseim.toml
```diff
@@ -0,0 +1,362 @@
...
+[shaper.mam_global_shaper]
+  max_rate = 1000
+
+[acl.local]
+  user_regexp = ""
+
+[access]
+  max_user_sessions = [
+    {acl = "all", value = 10}
+  ]
```
kmakiela, line 229: Why is this an array and not a table like the rest of the config? It looks like we deal with key-value pairs here as well

src/config/mongoose_config_parser_toml.erl
```diff
@@ -0,0 +1,354 @@
+-module(mongoose_config_parser_toml).
```
kmakiela, line 1: Should all the options be parsed in a single file? It seems like it could become monstrous in the future

src/config/mongoose_config_parser_toml.erl
```diff
@@ -0,0 +1,354 @@
...
+listener_opt(<<"c2s">>, <<"xml_socket">>, V) -> [{xml_socket, V}];
+listener_opt(<<"c2s">>, <<"zlib">>, V) -> [{zlib, V}];
+listener_opt(<<"c2s">>, <<"verify_peer">>, V) -> [{verify_peer, V}];
+listener_opt(<<"c2s">>, <<"hibernate_after">>, V) -> [{hibernate_after, V}];
+listener_opt(<<"c2s">>, <<"starttls">>, true) -> [starttls];
+listener_opt(<<"c2s">>, <<"starttls">>, false) -> [];
+listener_opt(<<"c2s">>, <<"starttls_required">>, true) -> [starttls_required];
+listener_opt(<<"c2s">>, <<"starttls_required">>, false) -> [];
+listener_opt(<<"c2s">>, <<"tls">>, true) -> [tls];
+listener_opt(<<"c2s">>, <<"tls">>, false) -> [];
```
kmakiela, line 183: Maybe group the options that may be `false` to one clause since it returns and empty list anyway?

src/ejabberd_config.erl
```diff
@@ -315,8 +315,14 @@ handle_table_does_not_exist_error(Table) ->
 %%--------------------------------------------------------------------
 -spec parse_file(file:name()) -> state().
 parse_file(ConfigFile) ->
-    Terms = get_plain_terms_file(ConfigFile),
-    mongoose_config_parser:parse_terms(Terms).
+    TOMLFile = filename:rootname(ConfigFile) ++ ".toml",
```
kmakiela, line 318: Could the filename be already given with the `.toml` extension?

rel/files/mongooseim.toml
chrzaszcz: These are lists of rules and the order of rules matters, so they cannot be KV pairs. The alternative syntax would be:

```
[[access.max_user_offline_messages]]
  acl = admin
  value = 5000


[[access.max_user_offline_messages]]
  acl = all
  value = 100
```

For me it is less readable.

src/config/mongoose_config_parser_toml.erl
chrzaszcz: I fully agree and they are currently in one place to simplify the first implementation. Apart from splitting it is quite likely that a callback mechanism will be used.

src/config/mongoose_config_parser_toml.erl
chrzaszcz: I wanted to be explicit here. I think that it would be best to have one line per option. This will be refactored for sure.

src/ejabberd_config.erl
chrzaszcz: I agree that this i a bit clunky. I just wanted it to be simple as we are still not sure if we want to support both formats or only one of them. As soon as we know that, this code will be changed.

kmakiela:
> @kmakiela do you think my answers are enough for now or do you feel that the changes need to be done before the merge to the feature branch?

They are enough, thanks for the clarification - I don't feel the need to implement any of these changes now mostly because as you said, they are a part of the refactoring process that is yet to come

src/config/mongoose_config_parser_toml.erl
```diff
@@ -0,0 +1,354 @@
...
+    {users_can_see_hidden_services, V};
+module_opt(<<"mod_register">>, <<"welcome_message">>, V) ->
+    {welcome_message, b2l(V)};
+module_opt(<<"mod_register">>, <<"ip_access">>, V) ->
+    Rules = [{Addr, b2a(P)} || #{<<"address">> := Addr, <<"policy">> := P} <- V],
+    {ip_access, Rules};
+module_opt(<<"mod_register">>, <<"access">>, V) ->
+    {access, b2a(V)};
+module_opt(<<"mod_vcard">>, <<"host">>, V) ->
+    {host, b2l(V)};
```
michalwski, line 73: I think this kind of parsing could be delegated to the respective modules so they control their specific config options. Probably in a separate PR.

rel/files/mongooseim.toml
```diff
@@ -0,0 +1,362 @@
...
+    "anonymous.localhost",
+    "localhost.bis"
+  ]
+  registration_timeout = "infinity"
+  language = "en"
+  all_metrics_are_global = false
+  sm_backend = "mnesia"
+  max_fsm_queue = 1000
+
+[[listen.cowboy]]
```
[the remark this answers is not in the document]
chrzaszcz: I could try it now and see how it looks like. I like it as it hides the dependency on `cowboy` and we will be free to change the library to something else if we need to.

rel/files/mongooseim.toml
```diff
@@ -0,0 +1,362 @@
...
+  transport.max_connections = 1024
+
+  [[listen.cowboy.modules.mongoose_api]]
+    host = "localhost"
+    path = "/api"
+    handlers = ["mongoose_api_metrics", "mongoose_api_users"]
+
+[[listen.c2s]]
+  port = 5222
+  certfile = "priv/ssl/fake_server.pem"
```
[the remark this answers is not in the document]
chrzaszcz: Let me look at it now and I will see. If it's easy, I could do it now.

rel/files/mongooseim.toml
```diff
@@ -0,0 +1,362 @@
...
+
+[s2s]
+  use_starttls = "optional"
+  certfile = "priv/ssl/fake_server.pem"
+  default_policy = "allow"
+  outgoing_port = 5299
+
+  [[s2s.address]]
+    host = "fed1"
+    ip_address = "127.0.0.1"
```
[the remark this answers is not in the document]
chrzaszcz: I can reorder the sections now, this is a good test for the equivalence check anyway ;-)

src/config/mongoose_config_parser_toml.erl
chrzaszcz: Yes I think it needs to be well thought and probably aligned with the work around the config API you are starting now.

rel/files/mongooseim.toml
```diff
@@ -0,0 +1,362 @@
+[general]
+  loglevel = 3
+  hosts = [
+    "localhost",
+    "anonymous.localhost",
+    "localhost.bis"
+  ]
+  registration_timeout = "infinity"
+  language = "en"
+  all_metrics_are_global = false
```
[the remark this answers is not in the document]
chrzaszcz: > In this review, I focused mostly on the TOML file. Below there are some more general comments:
> 
>     * If possible, I'd like to see `modules` and `services` section moved up (modules before services would be perfect). They change more often than `shapers` and `acls`.
I will experiment with the order for a moment and see how it looks like.

>     * Below there are config options which, in the following changes, could be removed from the TOML config file and exchanged with some kind of defaults (if not present in the config):
>       
>       * `registration_timeout`
>       * `language`
>       * `all_metrics_are_global`
>       * `max_fsm_queue`
I think that introducing the new defaults is a must and should be done in a new task.

> When it comes to the code parsing and translating config option to the intermediate format it could be refactored in a way that some operations are delegated to modules responsible for certain functionality.
Yes, I fully agree with that.

> Also, I think in many cases the translation from binary to list is redundant, because later the list is often converted to binaries again. This would require changing the intermediate format, which I'm not sure we are going to do in the nearest future.
Again, we think the same :-) it just became more obvious now how we convert things there and back. I left it as it is on purpose to limit the changes in one go. We definitely need to do that.

> All in all, thanks for the initial implementation. I think we can merge this PR and consider if my comments are worth addressing in the future.
I will try to address some of them quickly as some of them seem very easy to do.

rel/files/mongooseim.toml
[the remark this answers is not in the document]
chrzaszcz: I will do it later as the format for s2s and c2s-related TLS options is different, so for now it makes sense that the syntax is also different.

test/config_parser_SUITE_data/mongooseim-pgsql.cfg
```diff
@@ -0,0 +1,939 @@
+%%%
```
NelsonVides, line 1: Does this file need all the comments explaining things? Nothing important, but it might just reduce repo size a few bytes for duplicated comments coming from the original config file 🤔

chrzaszcz: Right now it's just a copy-paste of the original config file and I didn't want to change it to much, making it more difficult to compare with the original file. That's also how the rest of the tests for `ejabberd_config` work.

PR #2798, opened 2020-07-22
TOML config templating

chrzaszcz:
Use the mustache templates for the TOML config file - make it work the same as for the '.cfg' file.

Notes:
- Failures for big tests are expected on CI except `pgsql_mnesia`, which should pass
- There is a fix for getting hosts by auth method. This bug was revealed by the TOML config which provided a list of methods (which should be valid) instead of an atom with one method.
- The templating is a bit inconsistent: sometimes we template only the value and sometimes one or more key-value pairs. It could be more organised, but this might be done as a separate story - not to change too much in one go. This is consistent with how it worked with 'cfg'.
- The templated config file contains a lot of empty lines, maybe this could be improved, but this falls outside of the scope of this story.
- TOML config does **not** support 'required files' - this should be done in a separate story as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,7 @@ priv/mibs/EJABBERD-MIB.bin
 ebin/ejabberd.app
 rel/configure.vars.config
 rel/vars.config
+rel/vars-toml.config
 compile_commands.json
 # Compilation/test ephemera
 *.d
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ clean:
 	-rm -rf _build
 	-rm rel/configure.vars.config
 	-rm rel/vars.config
+	-rm rel/var-toml.config
 
 # REBAR_CT_EXTRA_ARGS comes from a test runner
 ct:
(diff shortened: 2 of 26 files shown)


chrzaszcz:
> In `rebar.config`, line 99, there's a comment that says "%% Don't forget to update tools/test_runner/apply_templates.erl, when changing templates", how does that one apply here?
I didn't notice this comment, but I updated `apply_templates` anyway.

> Also, there are some log messages about unhandled config changes, for example in https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/2798/24663/pgsql_mnesia.22.1.8-1/big/ct_run.test@default-5a02f52e-ac71-46a0-8f78-5dd1a2e3d6b1.2020-07-22_12.20.31/mongooseim@localhost.log.html#L781, why is so?

It's the same as on the master branch:

https://circleci-mim-results.s3.eu-central-1.amazonaws.com/branch/master/24523/pgsql_mnesia.22.1.8-1/big/ct_run.test@default-c6290ab2-1453-40ee-8564-7422c9b154d8.2020-07-14_13.22.26/mongooseim@localhost.log.html#L775

These are some holes in the config reload logic, it would be good to patch them later.

PR #2800, opened 2020-07-23
Update RE2 repo link

licaon-kter:
...as it currently redirects there, but I'm not sure for how long this is done.

### DIFF
--- a/rebar.lock
+++ b/rebar.lock
@@ -133,7 +133,7 @@
  {<<"rabbit_common">>,{pkg,<<"rabbit_common">>,<<"3.8.4">>},1},
  {<<"ranch">>,{pkg,<<"ranch">>,<<"1.7.1">>},1},
  {<<"re2">>,
-  {git,"https://github.com/tuncer/re2.git",
+  {git,"https://github.com/dukesoferl/re2.git",
        {ref,"2471e06ed7e548ca1af522e95b16867135afdfdc"}},
   1},
  {<<"recon">>,{pkg,<<"recon">>,<<"2.5.1">>},0},


NelsonVides:
Hi @licaon-kter,

Thanks a lot for your input here. I'm only afraid that this is not the way to update dependencies in rebar3. The lock file is generated from the config file. The thing about this dependency is that it is not MongooseIM's dependency, but a dep of a dep.

If you run `rebar3 tree`, you'll get an ascii tree in your shell listing all the hierarchy of dependencies. There we can see that `re2` is a dependency of `cqerl`, so it is `cqerl` the project that should be updated, and then MIM can be updated to use the new version of `cqerl`.

In case you're not familiar with it, you can read about dependencies in rebar3 here: https://www.rebar3.org/docs/dependencies Do you want to experiment with this update?

I'm closing this PR, as it doesn't propose a correct solution, but thank you for raising awareness, and we hope to see more PRs from you in the future! 🙂

licaon-kter:
Indeed I'm not yet familiar with the new rebar3 workflow, hope you can have a go at it.

licaon-kter:
@NelsonVides I'm hitting this again, you are using a local fork of `cqerl` that's 2 years old, so we need to update what exactly? Why do you use a fork in the first place?

NelsonVides:
> @NelsonVides I'm hitting this again, you are using a local fork of `cqerl` that's 2 years old, so we need to update what exactly? Why do you use a fork in the first place?

Hi @licaon-kter, sorry for not answering yet. I'm moving to the open issue and keeping a single thread there 👍

PR #2802, opened 2020-07-24
TOML config split

chrzaszcz:
**Functional changes**
Choose config format (cfg or TOML) by looking at file extension. Previously even though the configured path ended with 'cfg', the '.cfg' file extension was stripped and if there was a '.toml' file with the same root name, it was used instead. This was counter-intuitive and now the config file name is configured explicitly - it is `mongooseim.toml` by default.
There is no implicit fallback to `mongooseim.cfg` anymore.
Having one explicitly configured config path is less error prone.

Right now the way to change it would be to edit the `mongooseim` script. We could make it simpler if we want to support both formats. The goal for now is to keep only TOML after a possible transition period (not decided yet).

**Refactoring**
- Move the parser of the `cfg` format to `mongoose_config_parser_cfg`
- Extract term parsing to `mongoose_config_terms`
- Extract reload helpers to `mongoose_config_reload`

### DIFF
--- a/src/config/mongoose_config_terms.erl
+++ b/src/config/mongoose_config_terms.erl
@@ -1,0 +1,145 @@
+%%% @doc Parsing of the 'cfg' file to Erlang terms
+-module(mongoose_config_terms).
+
+-export([get_plain_terms_file/1]).
+
+-include("mongoose.hrl").
+-include("ejabberd_config.hrl").
+
+%% @doc Read an ejabberd configuration file and return the terms.
+%% Input is an absolute or relative path to an ejabberd config file.
+%% Returns a list of plain terms,
+%% in which the options 'include_config_file' were parsed
+%% and the terms in those files were included.
+-spec get_plain_terms_file(string()) -> [term()].
+get_plain_terms_file(File1) ->
+    File = mongoose_config_utils:get_absolute_path(File1),
+    case file:consult(File) of
+        {ok, Terms} ->
+            include_config_files(Terms);
+        {error, {LineNumber, erl_parse, _ParseMessage} = Reason} ->
+            ExitText = describe_config_problem(File, Reason, LineNumber),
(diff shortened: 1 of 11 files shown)


src/config/mongoose_config_terms.erl
```diff
@@ -0,0 +1,145 @@
+%%% @doc Parsing of the 'cfg' file to Erlang terms
+-module(mongoose_config_terms).
+
+-export([get_plain_terms_file/1]).
+
+-include("mongoose.hrl").
+-include("ejabberd_config.hrl").
+
+%% @doc Read an ejabberd configuration file and return the terms.
```
janciesla8818, line 9: Is it still reading ejabberd config?

src/config/mongoose_config_terms.erl
```diff
@@ -0,0 +1,145 @@
...
+                  end,
+                  Terms),
+    [?WARNING_MSG("event=ignore_disallowed_option option=~p", [NA])
+     || NA <- NAs],
+    As.
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%%% Errors reading the config file
+
+-type config_problem() :: atom() | {integer(), atom() | tuple(), _}. % spec me better
```
janciesla8818, line 123: I know this is how it was in the `ejabberd_config.erl` but what do you think of moving this type to the top together with the other types.

chrzaszcz:
> Good stuff! I only have two minor comments.

Thanks for the review @janciesla8818.

While I agree that `ejabberd_config_terms.erl` could be refactored, I am not convinced that it needs to be done here.

- For me it seems a bit out of scope of this story.
- This is legacy code that will be dropped - probably not in the next release, but in the one after it.

I'd rather merge the code as it is right now.

PR #2807, opened 2020-07-31
Reorganise http related test_suites

NelsonVides:
Except for the place where the http pool passes a map instead of a list, this is the state of the big_tests that #2716 should be able to pass. Once this is merged to master, the http2 PR should rebase onto it and pass these tests successfully.

Note that the problem for gun was about pools being initialised before there was an available listener: in the modified test suites, the pools where being initialised earlier in the `init_per_x` hierarchy, hence gun's connection wasn't available.

Note that there's one important test, that in my fork of http2 with these tests gun is not passing, and that is the `mod_http_listener` `simple_message_no_listener` test, where there's precisely no listener. This is failing because the request timeout is of 5 seconds, so the client's process is blocked for 5 long seconds, while escalus is waiting for a stream_end stanza with also a timeout of 5 seconds, but which started way earlier than the pool's timeout.

PS: notice also that these tests are waaay faster than the originals 😄

### DIFF
--- a/big_tests/tests/mod_http_notification_SUITE.erl
+++ b/big_tests/tests/mod_http_notification_SUITE.erl
@@ -49,11 +49,10 @@ groups() ->
     ct_helper:repeat_all_until_all_ok(G).
 
 init_per_suite(Config0) ->
-    Config1 = escalus:init_per_suite(Config0),
-    escalus:create_users(Config1, escalus:get_users([alice, bob])).
+    escalus:init_per_suite(Config0).
 
 end_per_suite(Config) ->
-    escalus:delete_users(Config, escalus:get_users([alice, bob])),
+    escalus_fresh:clean(),
     escalus:end_per_suite(Config).
 
 init_per_group(no_prefix, Config) ->
@@ -80,36 +79,30 @@ end_per_testcase(CaseName, Config) ->
 %%%===================================================================
 %%% offline tests
 %%%===================================================================
-proper_http_message_encode_decode(Config0) ->
-    Config = escalus_fresh:create_users(Config0, [{alice, 1}, {bob, 1}]),
-    {ok, Alice} = escalus_client:start(Config, alice, <<"res1">>),
-    escalus:send(Alice, escalus_stanza:presence(<<"available">>)),
-    {ok, Bob} = escalus_client:start(Config, bob, <<"res1">>),
-    escalus:send(Bob, escalus_stanza:presence(<<"available">>)),
+proper_http_message_encode_decode(Config) ->
+    escalus:fresh_story(Config, [{alice, 1}, {bob, 1}],
+        fun(Alice, Bob) ->
+            Sender = jid:nameprep(escalus_client:username(Alice)),
+            Receiver = jid:nameprep(escalus_client:username(Bob)),
+            Server = jid:nodeprep(escalus_users:get_host(Config, alice)),
+            Message = <<"Hi Test!&escape=Hello">>,
 
-    Sender = jid:nameprep(escalus_users:get_username(Config, alice)),
-    Server = jid:nodeprep(escalus_users:get_host(Config, alice)),
-    Receiver = jid:nameprep(escalus_users:get_username(Config, bob)),
-    Message = <<"Hi Test!&escape=Hello">>,
+            Stanza = escalus_stanza:chat_to(Bob, Message),
+            escalus:send(Alice, Stanza),
+            escalus:wait_for_stanza(Bob),
 
-    BobJid = escalus_users:get_jid(Config, bob),
-    Stanza = escalus_stanza:chat_to(BobJid, Message),
-    escalus:send(Alice, Stanza),
-    escalus:wait_for_stanzas(Bob, 2),
+            Body = get_http_request(),
 
-    escalus_client:stop(Config, Alice),
-    escalus_client:stop(Config, Bob),
+            ExtractedAndDecoded = rpc(mim(), cow_qs, parse_qs, [Body]),
+            ExpectedList = [{<<"author">>,<<Sender/binary>>},
+                            {<<"server">>,<<Server/binary>>},
+                            {<<"receiver">>,<<Receiver/binary>>},
+                            {<<"message">>,<<Message/binary>>}],
+            SortedExtractedAndDecoded = lists:sort(ExtractedAndDecoded),
+            SortedExpectedList = lists:sort(ExpectedList),
+            ?assertEqual(SortedExpectedList, SortedExtractedAndDecoded)
+        end).
 
-    Body = get_http_request(),
-
-    ExtractedAndDecoded = rpc(mim(), cow_qs, parse_qs, [Body]),
-    ExpectedList = [{<<"author">>,<<Sender/binary>>},
-                    {<<"server">>,<<Server/binary>>},
-                    {<<"receiver">>,<<Receiver/binary>>},
-                    {<<"message">>,<<Message/binary>>}],
-    SortedExtractedAndDecoded = lists:sort(ExtractedAndDecoded),
-    SortedExpectedList = lists:sort(ExpectedList),
-    ?assertEqual(SortedExpectedList, SortedExtractedAndDecoded).
 
 simple_message(Config) ->
     %% we expect one notification message
--- a/big_tests/tests/mod_http_notification_SUITE.erl
+++ b/big_tests/tests/mod_http_notification_SUITE.erl
@@ -14,6 +14,8 @@
 -include_lib("common_test/include/ct.hrl").
 -include_lib("eunit/include/eunit.hrl").
 
+-define(ETS_TABLE, mod_http_notification).
+
 -import(distributed_helper, [mim/0,
                              require_rpc_nodes/1,
                              rpc/4]).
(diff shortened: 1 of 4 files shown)


mongoose-im:
[8399.1](https://travis-ci.org/esl/MongooseIM/jobs/713845373) / Erlang 22.0 / small_tests / 2cab25f0edb0a330f85bd707290107a48e2b7415
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2807/8399/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2807/8399/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-07-31_22.13.33/lib.mongooseim.logs/run.2020-07-31_22.13.33/suite.log.html)

big_tests/tests/mod_http_notification_SUITE.erl
```diff
@@ -151,44 +125,94 @@ simple_message_no_listener(Config) ->
...
+    ejabberd_node_utils:call_fun(mongoose_wpool, start_configured_pools,
+                                 [[{http, global, http_pool, PoolOpts, HTTPOpts}]]).
+
+stop_pool() ->
+    ejabberd_node_utils:call_fun(mongoose_wpool, stop, [http, global, http_pool]).
+
+set_modules(Config0, Opts) ->
+    Config = dynamic_modules:save_modules(host(), Config0),
+    ModOpts = [{worker_timeout, 500}, {host, http_notifications_host()}] ++ Opts,
+    dynamic_modules:ensure_modules(host(), [{mod_http_notification, ModOpts}]),
```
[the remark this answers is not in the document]
NelsonVides: At the moment, the opts given to `mod_http_notification` are directly passed to `mod_event_pusher` in https://github.com/esl/MongooseIM/blob/f82a1aae6b9fb0c030e87fd30b6b8e39d6d07c0e/src/mod_http_notification.erl#L10-L11, as far as I understand this will still work the same way 🤔

big_tests/tests/mod_http_notification_SUITE.erl
```diff
@@ -151,44 +125,94 @@ simple_message_no_listener(Config) ->
...
-get_prefix(mod_http_notification_tests_with_prefix) ->
+get_prefix(with_prefix) ->
     "/prefix";
 get_prefix(Config) ->
     GroupName = proplists:get_value(name, proplists:get_value(tc_group_properties, Config)),
     get_prefix(GroupName).
+
+start_pool() ->
+    HTTPOpts = [{server, http_notifications_host()}],
+    PoolOpts = [{strategy, available_worker}, {workers, 1}],
```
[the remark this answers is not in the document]
NelsonVides: I thought so because tests are run in sequence, so there's no need for so many listeners. But perhaps it was mostly because log messages from gun crashing were multiplied by ten 😅  Will fix anyway 🙂

big_tests/tests/mod_http_notification_SUITE.erl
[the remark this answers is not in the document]
gustawlippa: Yes, and I may be mistaken, but I don't see a place where these options would be read. For example, in the backend:
https://github.com/esl/MongooseIM/blob/f82a1aae6b9fb0c030e87fd30b6b8e39d6d07c0e/src/event_pusher/mod_event_pusher_http.erl#L52
This made me think that these options are discarded. But it is not really in the scope of this PR and I'm really not sure, just an observation.

big_tests/tests/mod_http_notification_SUITE.erl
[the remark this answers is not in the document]
gustawlippa: I would go for more because it may show some problems, like "Why are there so many crash logs?! We should clean it up!" and because it is probably the way the pool will be ran, even if the actual number of requests will be really low.

big_tests/tests/mod_http_notification_SUITE.erl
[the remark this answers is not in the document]
NelsonVides: Yeah, this one is very tricky. Opts are stored globally on an ets table on mod_event_pusher: https://github.com/esl/MongooseIM/blob/f82a1aae6b9fb0c030e87fd30b6b8e39d6d07c0e/src/event_pusher/mod_event_pusher.erl#L62-L65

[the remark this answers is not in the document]
NelsonVides: And then it gets weirder and weirder, but yeah, that's out of the scope of this PR maybe 😅

mongoose-im:
[8401.1](https://travis-ci.org/esl/MongooseIM/jobs/714449489) / Erlang 22.0 / small_tests / 0a70d75fa9cd4b88578e8b4ba19959860c884649
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2807/8401/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2807/8401/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-08-03_12.23.13/lib.mongooseim.logs/run.2020-08-03_12.23.13/suite.log.html)

mongoose-im:
[8402.1](https://travis-ci.org/esl/MongooseIM/jobs/714449917) / Erlang 22.0 / small_tests / 4133461a23156ec1db0b4bf24e9b6a18036824a7
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2807/8402/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2807/8402/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-08-03_12.55.22/lib.mongooseim.logs/run.2020-08-03_12.55.23/suite.log.html)

big_tests/tests/mod_http_notification_SUITE.erl
```diff
@@ -66,68 +56,57 @@ end_per_suite(Config) ->
...
-    Pid ! {got_http_request, Body},
-    Req2.
-
 %%%===================================================================
 %%% offline tests
 %%%===================================================================
-proper_http_message_encode_decode(Config) ->
-    Receiver= <<"bob">>,
+proper_http_message_encode_decode(Config0) ->
+    Config = escalus_fresh:create_users(Config0, [{alice, 1}, {bob, 1}]),
```
chrzaszcz, line 84: Why is this changed? What prevents us from using `escalus_story`?

big_tests/tests/mod_http_notification_SUITE.erl
```diff
@@ -14,6 +14,8 @@
 -include_lib("common_test/include/ct.hrl").
 -include_lib("eunit/include/eunit.hrl").
 
+-define(ETS_TABLE, mod_http_notification).
```
chrzaszcz, line 17: What is the motivation for using this ETS table instead of sending the message with the result?

NelsonVides: It follows the example from `service_mongoose_system_metrics_SUITE`. The problem of passing the pid is that the listener cannot then be started until the `init_per_testcase`, where we actually know the pid of the process that will run the test. And in that case, the pool can't either. So tests still can't be parallelised (unless we start pools for different hosts and tests register their clients with different hosts as well?), and it makes the whole listener-pool initialisation and termination run too often.

But actually, this was the intention in the other test suites. In the case of this one, it all still runs on the init_per_testcase, that is true 🤔 So perhaps here it is like this just for consistency? I could revert this specific change to how it was before, using pids and messages 🤷‍♂️

big_tests/tests/mod_http_notification_SUITE.erl
NelsonVides: Users are fresh now, containing the test name, and for some mysterious reason I'm not aware of, `escalus_story` was taking several seconds to run, while as it is now takes less than 0.1s

big_tests/tests/mod_http_notification_SUITE.erl
chrzaszcz: Thanks for the explanation, I don't think it needs to be reverted.

big_tests/tests/mod_http_notification_SUITE.erl
chrzaszcz: Was it also the case with `fresh_story`?

NelsonVides: Don't remember trying with fresh_story, but I probably chose not to because in `do_simple_message` we were disconnecting users independently. I'll try out quickly with fresh_story nevertheless, might make it cleaner 🤔

mongoose-im:
[8403.1](https://travis-ci.org/esl/MongooseIM/jobs/714781274) / Erlang 22.0 / small_tests / 12fe404bedcd57e194a47a633565d4ddbc56c6a6
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2807/8403/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2807/8403/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-08-04_11.44.36/lib.mongooseim.logs/run.2020-08-04_11.44.37/suite.log.html)

PR #2814, opened 2020-08-11
TOML config parser refactoring

chrzaszcz:
1. Parse TOML structures with two functions: `parse_map` and `parse_list`.
2. Reorder functions according to the order of sections in the TOML config file.
3. Add function specs for input types (TOML sections, keys, values) and output types (config records and options).
4. Annotate each function with the TOML path it is processing.

Planned next steps:
1. Renaming the functions for increased consistency.
2. Research on converting the annotations to code - more details coming soon.

### DIFF
--- a/src/config/mongoose_config_parser_toml.erl
+++ b/src/config/mongoose_config_parser_toml.erl
@@ -8,7 +8,18 @@
 -include("mongoose.hrl").
 -include("ejabberd_config.hrl").
 
+%% Used to create per-host config when the list of hosts is not known yet
 -define(HOST_F(Expr), [fun(Host) -> Expr end]).
+
+%% Input: TOML parsed by tomerl
+-type toml_key() :: binary().
+-type toml_value() :: tomerl:value().
+-type toml_section() :: tomerl:section().
+
+%% Output: list of config records, containing key-value pairs
+-type option() :: term(). % any part of a config value, which can be a complex term
+-type config() :: #config{} | #local_config{} | acl:acl().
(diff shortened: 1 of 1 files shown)


src/config/mongoose_config_parser_toml.erl
```diff
@@ -19,186 +30,127 @@ parse_file(FileName) ->
...
     {Port, IPT, _, _, Proto, OptsClean} =
         ejabberd_listener:parse_listener_portip(PortIP, Opts),
-    {{Port, IPT, Proto}, listener_module(Type), OptsClean}.
+    [{{Port, IPT, Proto}, listener_module(Type), OptsClean}].
 
+-spec listener_portip(toml_section()) -> option().
 listener_portip(#{<<"port">> := Port, <<"ip_address">> := Addr}) -> {Port, b2l(Addr)};
 listener_portip(#{<<"port">> := Port}) -> Port.
 
+-spec listener_module(toml_key()) -> option().
```
kmakiela, line 134: Maybe we could include path here as well

chrzaszcz: The idea was that when a path is the same as for the previous function, it is not specified again to avoid repetition.

In this case it would be still `listen.*[]` as we are still parsing the node representing the module - the argument is the key represented by the wildcard `*`. Actually this function could also be placed one level higher with `listen.*` as it only needs the type.

I think that I could fix this by explicitly stating the rule of no repetition in a comment. Would this s be ok?

kmakiela: Yes, that is precisely what I had in mind

kmakiela:
Approved and merged

PR #2815, opened 2020-08-13
Toml config logger fix

chrzaszcz:
Update TOML config after merging from master.

Most of the changes are necessary after the `loglevel` format changed.

It is advised to use only textual log levels not to cause confusion when the system silently changes its loglevel from the default 3 (warning) to error. This is already done in TOML.

Regarding the cleanup of the 'cfg' file, the following was done:
- Remove comments
- Remove empty lines
- Fix indentation (automatic, no manual verification)

This would help with future maintenance of this file and creating similar files for other parser tests.


src/config/mongoose_config_parser_toml.erl
```diff
@@ -130,7 +130,7 @@ auth_option_placement(<<"sasl_mechanisms">>) -> outer;
 auth_option_placement(_) -> inner.
 
 process_general(<<"loglevel">>, V) ->
-    #local_config{key = loglevel, value = V};
+    #local_config{key = loglevel, value = b2a(V)};
```
NelsonVides, line 133: I understand here that we're accepting only the binary then, right? While I totally agree with that and prefer the keywords rather than their number counterparts, I think this will need to be explicitly stated later on for the release. This is just to raise awareness and make some final decision later and document it properly, you don't need to take any action here.

chrzaszcz: Yes.

chrzaszcz: The `4 = error` trap caught me off guard and I am sure it would catch others as well. So we definitely need to make everything clear in the docs.

PR #2818, opened 2020-08-16
and the implement the xep 0277

daixfnwpu:
This PR addresses #

Proposed changes include:

### DIFF
--- a/apps/ejabberd/src/mod_microblog.erl
+++ b/apps/ejabberd/src/mod_microblog.erl
@@ -1,0 +1,4311 @@
+%%% ====================================================================
+%%% ``The contents of this file are subject to the Erlang Public License,
+%%% Version 1.1, (the "License"); you may not use this file except in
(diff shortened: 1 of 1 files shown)


NelsonVides:
Hi @daixfnwpu,

I'm afraid that the PR having no description fails to answer a lot of questions that are necessary in order to consider the change. Why is this xep needed, what is it used for, why would you and we need it? How was it implemented, how is it tested? For us, accepting a proposal means that we'd need to maintain it later, so we'd need to be sure we want it and the code is readily maintainable.

Then, I'm afraid that this way of porting code from ejabberd doesn't work. The whole building chain is different, the file organisation is different (check umbrella apps), the internal architecture is different. So this definitely doesn't work in our project, it would need to be adapted. And is not tested. So we definitely need to reject this.

Thank you for your curiosity into the project, wish you a good learning path 🙂

PR #2824, opened 2020-08-18
convert pubsub modules for structured logging

leszke:
`pubsub` can into structured logging now

### DIFF
--- a/src/pubsub/mod_pubsub.erl
+++ b/src/pubsub/mod_pubsub.erl
@@ -280,7 +280,7 @@ get_personal_data(Acc, #jid{ luser = LUser, lserver = LServer }) ->
         -> {'ok', state()}.
 
 init([ServerHost, Opts]) ->
-    ?DEBUG("pubsub init ~p ~p", [ServerHost, Opts]),
+    ?LOG_DEBUG(#{what => pubsub_init, server => ServerHost, opts => Opts}),
     Host = gen_mod:get_opt_subhost(ServerHost, Opts, default_host()),
 
     init_backend(ServerHost, Opts),
@@ -421,13 +421,13 @@ init_plugins(Host, ServerHost, Opts) ->
     TreePlugin = tree(Host, gen_mod:get_opt(nodetree, Opts,
                                             fun(A) when is_binary(A) -> A end,
                                             ?STDTREE)),
-    ?DEBUG("** tree plugin is ~p", [TreePlugin]),
+    ?LOG_DEBUG(#{what => pubsub_tree_plugin, tree_plugin => TreePlugin}),
     gen_pubsub_nodetree:init(TreePlugin, Host, ServerHost, Opts),
     Plugins = gen_mod:get_opt(plugins, Opts,
                               fun(A) when is_list(A) -> A end, [?STDNODE]),
     PepMapping = gen_mod:get_opt(pep_mapping, Opts,
                                  fun(A) when is_list(A) -> A end, []),
-    ?DEBUG("** PEP Mapping : ~p~n", [PepMapping]),
+    ?LOG_DEBUG(#{what => pubsub_pep_mapping, pep_mapping => PepMapping}),
     PluginsOK = lists:foldl(pa:bind(fun init_plugin/5, Host, ServerHost, Opts), [], Plugins),
     {lists:reverse(PluginsOK), TreePlugin, PepMapping}.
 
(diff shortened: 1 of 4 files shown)


src/pubsub/mod_pubsub.erl
```diff
@@ -280,7 +280,7 @@ get_personal_data(Acc, #jid{ luser = LUser, lserver = LServer }) ->
         -> {'ok', state()}.
 
 init([ServerHost, Opts]) ->
-    ?DEBUG("pubsub init ~p ~p", [ServerHost, Opts]),
+    ?LOG_DEBUG(#{what => pubsub_init, server_host => ServerHost, opts => Opts}),
```
arcusfelis, line 283: We use `server` field, not `server_host` (check field description in my PR) :)

src/pubsub/mod_pubsub.erl
```diff
@@ -421,32 +421,32 @@ init_plugins(Host, ServerHost, Opts) ->
...
     {lists:reverse(PluginsOK), TreePlugin, PepMapping}.
 
 init_plugin(Host, ServerHost, Opts, Name, Acc) ->
     Plugin = plugin(Host, Name),
     case catch apply(Plugin, init, [Host, ServerHost, Opts]) of
         {'EXIT', _Error} ->
-            ?ERROR_MSG("event=pubsub_plugin_init_failed,plugin=~p,host=~s,pubsub_host=~s,opts=~p",
-                       [Plugin, ServerHost, Host, Opts]),
+            ?LOG_ERROR(#{what => pubsub_plugin_init_failed, plugin => Plugin,
+                host => ServerHost, pubsub_host => Host, opts => Opts}),
```
arcusfelis, line 439: So, `server => ServerHost, sub_host => Host`. And the same format everywhere :) (it's also described in my fields docs)

src/pubsub/mod_pubsub.erl
```diff
@@ -881,9 +881,9 @@ remove_user_per_plugin_safe(LUser, LServer, Plugin) ->
         plugin_call(Plugin, remove_user, [LUser, LServer])
     catch
         Class:Reason:StackTrace ->
-            ?WARNING_MSG("event=cannot_delete_pubsub_user,"
-                         "luser=~s,lserver=~s,class=~p,reason=~p,stacktrace=~p",
-                         [LUser, LServer, Class, Reason, StackTrace])
+            ?LOG_WARNING(#{what => pubsub_cannot_delete_user, luser => LUser,
```
arcusfelis, line 884: Just `user => LUser, server => LServer`

src/pubsub/mod_pubsub.erl
```diff
@@ -1617,14 +1620,14 @@ adhoc_get_pending_parse_options(Host, #xmlel{name = <<"x">>} = XEl) ->
...
-    ?INFO_MSG("Bad XForm: ~p", [XData]),
+    ?LOG_INFO(#{what => pubsub_bad_xform, exml_packet => XData}),
     {error, mongoose_xmpp_errors:bad_request()}.
 
 %% @doc <p>Send a subscription approval form to Owner for all pending
 %% subscriptions on Host and Node.</p>
 send_pending_auth_events(Request, Host, Node, Owner) ->
-    ?DEBUG("Sending pending auth events for ~s on ~s:~s",
-           [jid:to_binary(Owner), Host, Node]),
+    ?LOG_DEBUG(#{what => pubsub_sending_pending_auth_events,
```
arcusfelis, line 1629: I think we would use `node` field as erlang node for each event (something that logger would emit). The same with pid, module, function, line, timestamp, when fields. So, basically, if you pass node now, it would be overwritten later. You also use function below, so it should be something action_function.


So, pubsub nodes should use `pubsub_node` field instead.

src/pubsub/mod_pubsub.erl
```diff
@@ -4538,12 +4548,12 @@ timestamp() ->
 make_error_reply(#iq{ sub_el = SubEl } = IQ, #xmlel{} = ErrorEl) ->
     IQ#iq{type = error, sub_el = [ErrorEl, SubEl]};
 make_error_reply(#iq{ sub_el = SubEl } = IQ, Error) ->
-    ?ERROR_MSG("event=pubsub_crash,details=~p", [Error]),
+    ?LOG_ERROR(#{what => pubsub_crash, details => Error}),
```
arcusfelis, line 4551: `reason` field is more preferred than details.

src/pubsub/mod_pubsub_cache_rdbms.erl
```diff
@@ -83,8 +83,7 @@ convert_rdbms_response({selected, [SelectedItem]}) ->
 convert_rdbms_response({updated, _}) ->
     ok;
 convert_rdbms_response(Response) ->
-    ?ERROR_MSG("RDBMS cache failed with: ~p", [Response]),
-    {error, pubsub_rdbms_cache_failed}.
+    ?LOG_ERROR(#{what => pubsub_rdbms_cache_failed, response => Response}).
```
arcusfelis, line 86: And reason field here too, instead of response.

src/pubsub/mod_pubsub_db_rdbms.erl
```diff
@@ -375,7 +375,7 @@ get_parentnodes_tree(Key, Node) ->
 find_nodes_with_parents(_, [], _, Acc) ->
     Acc;
 find_nodes_with_parents(_, _, 100, Acc) ->
-    ?WARNING_MSG("event=max_depth_reached, nodes=~p", [Acc]),
+    ?LOG_WARNING(#{what => pubsub_max_depth_reached, nodes => Acc}),
```
arcusfelis, line 378: `pubsub_nodes`, if you use `pubsub_node` as suggested above.

src/pubsub/mod_pubsub.erl
arcusfelis: Also, `pubsub_cannot_delete_user` could be converted to `pubsub_delete_user_failed` to be more consistent with the rest of codebase.

src/pubsub/mod_pubsub.erl
```diff
@@ -4307,17 +4316,17 @@ maybe_default_node(PluginModule, Function, Args) ->
...
                BaseModule ->
                    maybe_default_node(BaseModule, Function, Args)
            end
     end.
 
 node_action(Host, Type, Function, Args) ->
-    ?DEBUG("event=node_action,pubsub_host=~p,node_type=~p,function=~p,args=~p",
-           [Host, Type, Function, Args]),
+    ?LOG_DEBUG(#{what => pubsub_node_action, sub_host => Host,
+        node_type => Type, function => Function, args => Args}),
```
arcusfelis, line 4329: action_function

arcusfelis:
merged

PR #2837, opened 2020-08-21
TOML modules parsing

kmakiela:
This PR addresses [MIM-1112](https://erlangsolutions.atlassian.net/secure/RapidBoard.jspa?rapidView=3&projectKey=MIM&modal=detail&selectedIssue=MIM-1112).
* Added parsing for all the modules and their options
* Added equivalence test for module options between TOML and CFG

### DIFF
--- a/src/config/mongoose_config_parser_toml.erl
+++ b/src/config/mongoose_config_parser_toml.erl
@@ -923,6 +923,7 @@ mod_event_pusher_backend_push_opts([<<"virtual_pubsub_hosts">> |_] = Path, V) ->
     VPH = parse_list(Path, V),
     [{virtual_pubsub_hosts, VPH}].
 
+-spec mod_event_pusher_backend_http_opts(path(), toml_value()) -> [option()].
 mod_event_pusher_backend_http_opts([<<"pool_name">>|_], V) ->
     [{pool_name, b2a(V)}];
 mod_event_pusher_backend_http_opts([<<"path">>|_], V) ->
@@ -1110,6 +1111,7 @@ mod_muc_default_room_affiliations(_, #{<<"user">> := User, <<"server">> := Serve
     <<"resource">> := Resource, <<"affiliation">> := Aff}) ->
     [{{User, Server, Resource}, b2a(Aff)}].
 
+-spec mod_muc_log_top_link(path(), toml_value()) -> [option()].
 mod_muc_log_top_link([<<"target">>|_], V) ->
     [b2l(V)];
 mod_muc_log_top_link([<<"text">>|_], V) ->
--- a/src/config/mongoose_config_parser_toml.erl
+++ b/src/config/mongoose_config_parser_toml.erl
@@ -613,7 +613,7 @@ module_opt([<<"sdp_origin">>, <<"mod_jingle_sip">>|_], V) ->
     [{sdp_origin, b2l(V)}];
 module_opt([<<"ram_key_size">>, <<"mod_keystore">>|_], V) ->
     [{ram_key_size, V}];
-module_opt([<<"key">>, <<"mod_keystore">>|_] = Path, V) ->
+module_opt([<<"keys">>, <<"mod_keystore">>|_] = Path, V) ->
     Keys = parse_list(Path, V),
     [{keys, Keys}];
 module_opt([<<"archive_chat_markers">>, <<"mod_mam_meta">>|_], V) ->
@@ -646,18 +646,18 @@ module_opt([<<"history_size">>, <<"mod_muc">>|_], V) ->
     [{history_size, V}];
 module_opt([<<"room_shaper">>, <<"mod_muc">>|_], V) ->
     [{room_shaper, b2a(V)}];
-module_opt([<<"max_room_id">>, <<"mod_muc">>|_], V) when is_integer(V) ->
-    [{max_room_id, V}];
+module_opt([<<"max_room_id">>, <<"mod_muc">>|_], <<"infinity">>) ->
+    [{max_room_id, infinite}];
 module_opt([<<"max_room_id">>, <<"mod_muc">>|_], V) ->
-    [{max_room_id, b2a(V)}];
-module_opt([<<"max_room_name">>, <<"mod_muc">>|_], V) when is_integer(V) ->
-    [{max_room_name, V}];
+    [{max_room_id, V}];
+module_opt([<<"max_room_name">>, <<"mod_muc">>|_], <<"infinity">>) ->
+    [{max_room_name, infinite}];
 module_opt([<<"max_room_name">>, <<"mod_muc">>|_], V) ->
-    [{max_room_name, b2a(V)}];
-module_opt([<<"max_room_desc">>, <<"mod_muc">>|_], V) when is_integer(V) ->
-    [{max_room_desc, V}];
+    [{max_room_name, V}];
+module_opt([<<"max_room_desc">>, <<"mod_muc">>|_], <<"infinity">>) ->
+    [{max_room_desc, infinite}];
 module_opt([<<"max_room_desc">>, <<"mod_muc">>|_], V) ->
-    [{max_room_desc, b2a(V)}];
+    [{max_room_desc, V}];
 module_opt([<<"min_message_interval">>, <<"mod_muc">>|_], V) ->
     [{min_message_interval, V}];
 module_opt([<<"min_presence_interval">>, <<"mod_muc">>|_], V) ->
@@ -700,9 +700,9 @@ module_opt([<<"dirname">>, <<"mod_muc_log">>|_], V) ->
 module_opt([<<"file_format">>, <<"mod_muc_log">>|_], V) ->
     [{file_format, b2a(V)}];
 module_opt([<<"css_file">>, <<"mod_muc_log">>|_], <<"false">>) ->
-    [{css_file, false}];
+    [{cssfile, false}];
 module_opt([<<"css_file">>, <<"mod_muc_log">>|_], V) ->
-    [{css_file, V}];
+    [{cssfile, V}];
 module_opt([<<"timezone">>, <<"mod_muc_log">>|_], V) ->
     [{timezone, b2a(V)}];
 module_opt([<<"top_link">>, <<"mod_muc_log">>|_] = Path, V) ->
@@ -786,8 +786,8 @@ module_opt([<<"access">>, <<"mod_register">>|_], V) ->
     [{access, b2a(V)}];
 module_opt([<<"welcome_message">>, <<"mod_register">>|_], V) ->
     [{welcome_message, {b2l(V)}}];
-module_opt([<<"registration_watchers">>, <<"mod_register">>|_], V) ->
-    [{registration_watchers, V}];
+module_opt([<<"registration_watchers">>, <<"mod_register">>|_] = Path, V) ->
+    [{registration_watchers, parse_list(Path, V)}];
 module_opt([<<"password_strength">>, <<"mod_register">>|_], V) ->
     [{password_strength, V}];
 module_opt([<<"ip_access">>, <<"mod_register">>|_] = Path, V) ->
@@ -830,8 +830,14 @@ module_opt([<<"ldap_gfilter">>, <<"mod_shared_roster_ldap">>|_], V) ->
     [{ldap_gfilter, b2l(V)}];
 module_opt([<<"ldap_ufilter">>, <<"mod_shared_roster_ldap">>|_], V) ->
     [{ldap_ufilter, b2l(V)}];
+module_opt([<<"buffer_max">>, <<"mod_stream_management">>|_], <<"infinity">>) ->
+    [{buffer_max, infinity}];
+module_opt([<<"buffer_max">>, <<"mod_stream_management">>|_], <<"no_buffer">>) ->
+    [{buffer_max, no_buffer}];
 module_opt([<<"buffer_max">>, <<"mod_stream_management">>|_], V) ->
     [{buffer_max, V}];
+module_opt([<<"ack_freq">>, <<"mod_stream_management">>|_], <<"never">>) ->
+    [{ack_freq, never}];
 module_opt([<<"ack_freq">>, <<"mod_stream_management">>|_], V) ->
     [{ack_freq, V}];
 module_opt([<<"resume_timeout">>, <<"mod_stream_management">>|_], V) ->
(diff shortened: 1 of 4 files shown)


test/config_parser_SUITE.erl
```diff
@@ -160,6 +174,79 @@ compare_unordered_lists(L1, L2, F) ->
...
+compare_ordered_lists([{backends, Backends}], [{backends, Backends2}], _) ->
+    Sns = proplists:get_value(sns, Backends, []),
+    Sns2 = proplists:get_value(sns, Backends2, []),
+    Push = proplists:get_value(push, Backends, []),
+    Push2 = proplists:get_value(push, Backends2, []),
+    Http = proplists:get_value(http, Backends, []),
+    Http2 = proplists:get_value(http, Backends2, []),
+    compare_unordered_lists(Sns, Sns2),
+    compare_unordered_lists(Push, Push2),
+    compare_unordered_lists(Http, Http2);
```
gustawlippa, line 186: For me something like this looks cleaner, but perhaps there is an even nicer way of expressing this. This could be reused below.
```suggestion
Keys = [push, http, sns],
       [compare_unordered_lists(proplists:get_value(Key, Backends, []),
                                proplists:get_value(Key, Backends2, [])) 
       || Key <- Keys]
```

chrzaszcz (commented):
That's a huge amount of options!
First round of review: checked the general approach and style.

test/config_parser_SUITE.erl
```diff
@@ -160,6 +174,79 @@ compare_unordered_lists(L1, L2, F) ->
...
+    compare_unordered_lists(Group, Group2),
+    compare_ordered_lists(T1, T2, F);
+compare_ordered_lists([{mod_keystore, Opts}|T1], [{mod_keystore, Opts2}|T2], F) ->
+    Keys1 = proplists:get_value(keys, Opts, []),
+    Keys2 = proplists:get_value(keys, Opts2, []),
+    compare_unordered_lists(Keys1, Keys2),
+    compare_unordered_lists(lists:keydelete(keys, 1, Opts), lists:keydelete(keys, 1, Opts2)),
+    compare_ordered_lists(T1, T2, F);
+compare_ordered_lists([{Name, Opts} = H1|T1], [{Name, Opts2} = H2|T2], F) ->
+    Names = [
```
NelsonVides, line 205: This makes the function clause too long, this list of statically-chosen keys could be moved to a helper function returning that list, I find it more readable 🤔

kmakiela: Is there something in erlang like `module attribute` in elixir?

src/config/mongoose_config_parser_toml.erl
```diff
@@ -612,10 +1283,12 @@ process_host_section(Path, Content) ->
 %%       outgoing_pools.ldap.connection.tls.*,
 %%       outgoing_pools.riak.connection.tls.*,
 %%       outgoing_pools.cassandra.connection.tls.*
+%%       modules.mod_global_distrib.connections.tls.*
 -spec client_tls_option(path(), toml_value()) -> [option()].
 client_tls_option([<<"verify_peer">>|_], V) -> [{verify, verify_peer(V)}];
 client_tls_option([<<"certfile">>|_], V) -> [{certfile, b2l(V)}];
 client_tls_option([<<"cacertfile">>|_], V) -> [{cacertfile, b2l(V)}];
+client_tls_option([<<"cafile">>|_], V) -> [{cafile, b2l(V)}];
```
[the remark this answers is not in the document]
kmakiela: I took it from [`mod_global_distrib` docs](https://mongooseim.readthedocs.io/en/latest/modules/mod_global_distrib/).
Should it simply be renamed to `cacertfile`?

test/config_parser_SUITE.erl
chrzaszcz: I agree see that this piece of code is somewhat difficult to grasp and extend according to my initial intentions (which are apparently not clear), I think I need to propose a rework here as it is me who started it all.

src/config/mongoose_config_parser_toml.erl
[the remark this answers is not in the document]
chrzaszcz: Oh I see, it's `fasttls`. I agree that it is not clear, but `client_tls_option` is for the Erlang TLS implementation. I added a handler for the `fasttls` ones in my new PR: https://github.com/esl/MongooseIM/pull/2841/commits/e6adf60ce0b1c45140aa05f6f969c57eb237b8c9#diff-a96144358bbe793a05c0358187017af2R658

chrzaszcz:
Thanks @kmakiela for drastically improving the tests, that's what I had in mind. I added a small change to make the `compare_ordered_lists` function completely generic. All exceptions are handled in the provided `F` functions now, which ensures that no bug slips accidentally under the umbrella of these special cases.

src/config/mongoose_config_parser_toml.erl
```diff
@@ -471,30 +472,700 @@ process_module([Mod|_] = Path, Opts) ->
...
+module_opt([<<"api_version">>, <<"mod_push_service_mongoosepush">>|_], V) ->
+    [{api_version, b2l(V)}];
+module_opt([<<"max_http_connections">>, <<"mod_push_service_mongoosepush">>|_], V) ->
+    [{max_http_connections, V}];
+module_opt([<<"access">>, <<"mod_register">>|_], V) ->
+    [{access, b2a(V)}];
 module_opt([<<"welcome_message">>, <<"mod_register">>|_], V) ->
     [{welcome_message, {b2l(V)}}];
+module_opt([<<"registration_watchers">>, <<"mod_register">>|_], V) ->
+    [{registration_watchers, V}];
```
chrzaszcz, line 790: This seems to be a list of binaries, I thing we should call `parse_list` here.

src/config/mongoose_config_parser_toml.erl
```diff
@@ -471,30 +472,700 @@ process_module([Mod|_] = Path, Opts) ->
...
+module_opt([<<"ldap_rfilter">>, <<"mod_shared_roster_ldap">>|_], V) ->
+    [{ldap_rfilter, b2l(V)}];
+module_opt([<<"ldap_gfilter">>, <<"mod_shared_roster_ldap">>|_], V) ->
+    [{ldap_gfilter, b2l(V)}];
+module_opt([<<"ldap_ufilter">>, <<"mod_shared_roster_ldap">>|_], V) ->
+    [{ldap_ufilter, b2l(V)}];
+module_opt([<<"buffer_max">>, <<"mod_stream_management">>|_], V) ->
+    [{buffer_max, V}];
+module_opt([<<"ack_freq">>, <<"mod_stream_management">>|_], V) ->
+    [{ack_freq, V}];
```
chrzaszcz, line 836: There are atoms like `infinity` or `no_buffer` accepted in the code. I am not sure if we need to support them if they are undocumented. Could you double check it?

src/config/mongoose_config_parser_toml.erl
```diff
@@ -471,30 +472,700 @@ process_module([Mod|_] = Path, Opts) ->
...
     [{host, b2l(V)}];
-module_opt([<<"ldap_base">>, <<"mod_vcard">>|_], V) ->
+module_opt([<<"search">>, <<"mod_vcard">>|_], V) ->
+    [{search, V}];
+module_opt([<<"matches">>, <<"mod_vcard">>|_], <<"infinity">>) ->
+    [{matches, infinity}];
+module_opt([<<"matches">>, <<"mod_vcard">>|_], V) ->
+    [{matches, V}];
+module_opt([<<"allow_return_all">>, <<"mod_vcard">>|_], V) ->
+    [{allow_return_all, V}];
```
chrzaszcz, line 851: I can't find it in the code at all, as well as `search_all_hosts` and `ldap_uids`... they might have been removed, maybe when replacing LDAP implementation with `outgoing_pools`?

src/config/mongoose_config_parser_toml.erl
```diff
@@ -471,30 +472,700 @@ process_module([Mod|_] = Path, Opts) ->
...
+module_opt([<<"defaults_bucket_type">>|_], V) ->
+    [{defaults_bucket_type, V}];
+module_opt([<<"names_bucket_type">>|_], V) ->
+    [{names_bucket_type, V}];
+module_opt([<<"version_bucket_type">>|_], V) ->
+    [{version_bucket_type, V}];
+module_opt([<<"bucket_type">>|_], V) ->
+    [{bucket_type, V}];
+module_opt([<<"search_index">>|_], V) ->
+    [{search_index, V}].
```
chrzaszcz, line 896: Are they used outside `mod_privacy_riak`? Is it possible to limit the path somehow?

test/config_parser_SUITE.erl
```diff
@@ -56,10 +55,24 @@ s2s(Config) ->
     TOML_path = ejabberd_helper:data(Config, "s2s_only.toml"),
     State2 = mongoose_config_parser_toml:parse_file(TOML_path),
     Opts2 = mongoose_config_parser:state_to_opts(State2),
-        
+
     compare_unordered_lists(lists:filter(fun filter_config/1, Opts1), Opts2,
                             fun handle_config_option/2).
 
+modules(Config) ->
```
chrzaszcz, line 124: I think we could extract the common parts of these tests to a helper function, as they are getting repetitive. I think we could separate the equivalence tests to a different suite as well. However, this can be done in a separate PR.

chrzaszcz (commented):
I reviewed the options one by one, looking at the code and added a few comments.

src/config/mongoose_config_parser_toml.erl
```diff
@@ -471,30 +472,700 @@ process_module([Mod|_] = Path, Opts) ->
...
+module_opt([<<"max_room_id">>, <<"mod_muc">>|_], V) when is_integer(V) ->
+    [{max_room_id, V}];
+module_opt([<<"max_room_id">>, <<"mod_muc">>|_], V) ->
+    [{max_room_id, b2a(V)}];
+module_opt([<<"max_room_name">>, <<"mod_muc">>|_], V) when is_integer(V) ->
+    [{max_room_name, V}];
+module_opt([<<"max_room_name">>, <<"mod_muc">>|_], V) ->
+    [{max_room_name, b2a(V)}];
+module_opt([<<"max_room_desc">>, <<"mod_muc">>|_], V) when is_integer(V) ->
+    [{max_room_desc, V}];
```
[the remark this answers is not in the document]
kmakiela: the docs stated that it is an atom and wasn't sure if that meant `infinity` only

src/config/mongoose_config_parser_toml.erl
kmakiela: The handler would just pass the value in that case so I though it was not necessary, will change that

NelsonVides (commented):
I have no comments about the code, it's huge but very consistent and predictable, looks awesome :)

src/config/mongoose_config_parser_toml.erl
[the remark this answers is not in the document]
chrzaszcz: We could keep it as it is and leave it to the validator, it's `infinite` for these three options for sure, I checked it in the code of `mod_muc`

src/config/mongoose_config_parser_toml.erl
kmakiela: You are right, also can't find it in places other than tests and docs, will remove them

src/config/mongoose_config_parser_toml.erl
kmakiela: They are used quite often when `riak` can be chosen as a backend module. I can restrict it by requiring `riak` as a step, i.e. 
```
[modules.mod_offline]
  access_max_user_messages = "max_user_offline_messages"
  backend = "riak"
  bucket_type = "offline"
```
would become 
```
[modules.mod_offline]
  access_max_user_messages = "max_user_offline_messages"
  backend = "riak"
  riak.bucket_type = "offline"
```

src/config/mongoose_config_parser_toml.erl
```diff
@@ -518,30 +519,707 @@ process_module([Mod|_] = Path, Opts) ->
...
+mod_http_upload_s3([<<"add_acl">>|_], V) ->
+    [{add_acl, V}];
+mod_http_upload_s3([<<"region">>|_], V) ->
+    [{region, b2l(V)}];
+mod_http_upload_s3([<<"access_key_id">>|_], V) ->
+    [{access_key_id, b2l(V)}];
+mod_http_upload_s3([<<"secret_access_key">>|_], V) ->
+    [{secret_access_key, b2l(V)}].
+
+-spec mod_global_distrib_connections(path(), toml_section()) -> [option()].
```
chrzaszcz, line 1024: Here it should be `toml_value()` - it is the supertype of `toml_section()`, not otherwise.

src/config/mongoose_config_parser_toml.erl
```diff
@@ -518,30 +519,707 @@ process_module([Mod|_] = Path, Opts) ->
...
+module_opt([<<"max_room_id">>, <<"mod_muc">>|_], <<"infinity">>) ->
+    [{max_room_id, infinity}];
+module_opt([<<"max_room_id">>, <<"mod_muc">>|_], V) ->
+    [{max_room_id, V}];
+module_opt([<<"max_room_name">>, <<"mod_muc">>|_], <<"infinity">>) ->
+    [{max_room_name, infinity}];
+module_opt([<<"max_room_name">>, <<"mod_muc">>|_], V) ->
+    [{max_room_name, V}];
+module_opt([<<"max_room_desc">>, <<"mod_muc">>|_], <<"infinity">>) ->
+    [{max_room_desc, infinity}];
```
chrzaszcz, line 658: It's `infinite` not `infinity`, as I stated in the comments before :-) Please change the parsed atom.

src/config/mongoose_config_parser_toml.erl
```diff
@@ -518,30 +519,707 @@ process_module([Mod|_] = Path, Opts) ->
...
+module_opt([<<"ldap_ufilter">>, <<"mod_shared_roster_ldap">>|_], V) ->
+    [{ldap_ufilter, b2l(V)}];
+module_opt([<<"buffer_max">>, <<"mod_stream_management">>|_], <<"infinity">>) ->
+    [{buffer_max, infinity}];
+module_opt([<<"buffer_max">>, <<"mod_stream_management">>|_], <<"no_buffer">>) ->
+    [{buffer_max, no_buffer}];
+module_opt([<<"buffer_max">>, <<"mod_stream_management">>|_], V) ->
+    [{buffer_max, V}];
+module_opt([<<"ack_freq">>, <<"mod_stream_management">>|_], <<"never">>) ->
+    [{buffer_max, never}];
```
chrzaszcz, line 840: Change it to `ack_freq`

chrzaszcz (commented):
Just a few remaining issues, very minor but let's fix them and it will be great :-)

src/config/mongoose_config_parser_toml.erl
kmakiela: Right, it is a bit misleading having these 2 types of "infinity" though

chrzaszcz: That's why I think it's best to change only the resulting atom and maybe get rid of this second infinity later.

chrzaszcz (approved):
Great!

PR #2839, opened 2020-08-24
Base16

NelsonVides:
Old dependency is not maintained. And code is copyrighted to ESL anyway. I forked it within esl's github and modify the implementation to use lookup tables, which is the fastest algorithm and one that uses constant memory. This way we also get rid of some old code and move it out to dependency responsibilities.

Big tests now explicitly adds `base16` dependency because it is explicitly used across code, and it works only because it is a dependency of escalus. Dependencies should be explicitly declared and not depend on dependencies of dependencies.

This is also good for OTP23 compilation, the old code won't compile with HiPE so its flags would need to be removed anyway. Also, my improvements in the dependency are even faster than this old code anyway 🔥


mongoose-im:
[8472.1](https://travis-ci.org/esl/MongooseIM/jobs/720626127) / Erlang 22.0 / small_tests / da8d87c6e977ae80344fe55c000476ac850a1602
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2839/8472/small_tests.22.0) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2839/8472/small_tests.22.0/small/ct_run.mongooseim@localhost.2020-08-24_12.30.58/lib.mongooseim.logs/run.2020-08-24_12.30.58/suite.log.html)

erszcz:
Great move! Let's wait for CI to see if we're fine with this lib.

arcusfelis (approved):

Issue #2770, opened 2020-06-02
Typo in num_acceptors


varnerac:
MongooseIM version: master
Installed from: source
Erlang/OTP version: any

This typo likely causes an issue when explicitly setting the number of acceptors in Cowboy.

https://github.com/esl/MongooseIM/blob/master/src/ejabberd_cowboy.erl#L151

NelsonVides:
Fixed twice in the linked PRs above.

PR #2841, opened 2020-08-26
TOML config options validation and testing: listeners

chrzaszcz:
Validation of the `listen` section, along with tests.

When testing it became necessary to amend and polish some listener options.
This is the reason for the first commit.

For details, see the commit messages.

### DIFF
--- a/src/ejabberd_cowboy.erl
+++ b/src/ejabberd_cowboy.erl
@@ -148,7 +148,7 @@ do_start_cowboy(Ref, Opts) ->
     SSLOpts = gen_mod:get_opt(ssl, Opts, undefined),
     NumAcceptors = gen_mod:get_opt(num_acceptors, Opts, 100),
     TransportOpts0 = gen_mod:get_opt(transport_options, Opts, #{}),
-    TransportOpts = TransportOpts0#{num_acceptros => NumAcceptors},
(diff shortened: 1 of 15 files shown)


src/ejabberd_cowboy.erl
```diff
@@ -148,7 +148,7 @@ do_start_cowboy(Ref, Opts) ->
     SSLOpts = gen_mod:get_opt(ssl, Opts, undefined),
     NumAcceptors = gen_mod:get_opt(num_acceptors, Opts, 100),
     TransportOpts0 = gen_mod:get_opt(transport_options, Opts, #{}),
-    TransportOpts = TransportOpts0#{num_acceptros => NumAcceptors},
+    TransportOpts = TransportOpts0#{num_acceptors => NumAcceptors},
```
gustawlippa, line 151: Great catch! Not related to the whole PR, but it fixes #2770, we should resolve this issue after merging.

chrzaszcz: I put it in a different commit to separate it from the rest.

kmakiela (approved):
Thanks for updating the parser to use `handlers` along the way and splitting `tls` and `fast_tls` options

NelsonVides (approved):

PR #2692, opened 2020-03-30
Add mongooseimctl bootstrap command

arcusfelis:
Proposed changes include:
* ejabberdctl command and docs

PR #2846, opened 2020-08-28
Change mongooseim_lock_dir specified for big tests so it's not /var/lock

### LINKED ISSUE #2832: Big tests write to /var/lock/ctl and fail if they can't
MongooseIM version: 3.7.1
Installed from: source
Erlang/OTP version: 22

Running big tests with command line 
```
./tools/test-runner.sh --skip-small-tests --db "mnesia minio redis" --preset internal_mnesia
```
Multiple tests fail with the following message:
```
﻿OutputStr: mkdir: cannot create directory ‘/var/lock/ctl’: Permission denied
```
One example of the 80 tests that fail this way is `conf_reload_SUITE:xmpp_domain_local_reload:domain_should_change`

Running the tests as root is a workaround, but is undesirable for obvious reasons.

The cause appears to be this line in [mim1.vars.config](https://github.com/esl/MongooseIM/blob/master/rel/mim1.vars.config#L97), [mim2.vars.config](https://github.com/esl/MongooseIM/blob/master/rel/mim2.vars.config#L67) etc.:
```
{mongooseim_lock_dir, "$EJABBERD_DIR/var/lock"}.
```
The environment variable `EJABBERD_DIR` is not set, so the lock dir ends up being `/var/lock`.

Changing this line to the following removes the issue:
```
{mongooseim_lock_dir, "$RUNNER_BASE_DIR/var/lock"}.
```
Is there a reason why it was `EJABBERD_DIR` in the first place, while the other paths in the file are `RUNNER_BASE_DIR`?

balgillo:
Closes #2832

In `mim1.vars.config`, `mim2.vars.config` etc., the `mongooseim_lock_dir` setting is configured as follows:
```
{mongooseim_lock_dir, "$EJABBERD_DIR/var/lock"}.
```
But nothing sets `EJABBERD_DIR` during the build process, so the big tests end up writing the lock file in `/var/lock`, which fails if the big tests are running as a user that doesn't have permission to write to that directory.

I've changed it to:
```
{mongooseim_lock_dir, "$RUNNER_BASE_DIR/var/lock"}.
```
Please let me know if I've missed something and there is some reason that it should be different, or some process that would set `EJABBERD_DIR` for the build.

### DIFF
--- a/rel/fed1.vars.config
+++ b/rel/fed1.vars.config
@@ -37,5 +37,5 @@
 {mongooseim_log_dir, "log"}.
 {mongooseim_mdb_dir, "$RUNNER_BASE_DIR/Mnesia.$NODE"}.
 {mongooseim_mdb_dir_toggle, "%"}.
-{mongooseim_lock_dir, "$EJABBERD_DIR/var/lock"}.
+{mongooseim_lock_dir, "$RUNNER_BASE_DIR/var/lock"}.
 {mongooseim_nodetool_etc_dir, "etc"}.
--- a/rel/fed1.vars.config
+++ b/rel/fed1.vars.config
@@ -37,5 +37,5 @@
 {mongooseim_log_dir, "log"}.
 {mongooseim_mdb_dir, "$RUNNER_BASE_DIR/Mnesia.$NODE"}.
 {mongooseim_mdb_dir_toggle, "%"}.
-{mongooseim_lock_dir, "$RUNNER_BASE_DIR/var/lock"}.
+{mongooseim_lock_dir, "$MIM_DIR/var/lock"}.
 {mongooseim_nodetool_etc_dir, "etc"}.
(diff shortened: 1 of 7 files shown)


arcusfelis (approved):
Looks fine

NelsonVides (commented):
Hi @balgillo, thanks again for your PR, it's very nice to see your contributions! 😄 

I've just realised that this is a leftover from https://github.com/esl/MongooseIM/pull/2692, where the variable `EJABBERD_DIR` was renamed to `MIM_DIR`, and we lost this usage. In the end it is all expanded to the correct value as you've set it, but it is probably better to keep it consistent and simply rename the variable, from `EJABBERD_DIR` to `MIM_DIR`.

balgillo:
@NelsonVides thanks for the review!  I've changed it to MIM_DIR and that is also working for me.  I've pushed the changes to this pull request.

NelsonVides (approved):
Excellent fix! Thanks a lot! 😄

PR #2847, opened 2020-08-28
Added support for all outgoing_pools options

kmakiela:
This PR addresses [MIM-1110](https://erlangsolutions.atlassian.net/secure/RapidBoard.jspa?rapidView=3&projectKey=MIM&modal=detail&selectedIssue=MIM-1110)

* Added support for new connection types (http, rabbit)
* Added support for missing options for already existing connection types

Note:
There are still pool options that are not parsed. We decided to only add them when needed because most of them are never used anywhere. Full list of these options is [here](https://github.com/inaka/worker_pool#starting-a-pool)

### DIFF
--- a/src/config/mongoose_config_parser_toml.erl
+++ b/src/config/mongoose_config_parser_toml.erl
@@ -410,7 +410,14 @@ pool_option([<<"call_timeout">>|_], V) -> [{call_timeout, V}].
 %% path: outgoing_pools.*.connection
 -spec connection_options(path(), toml_section()) -> [option()].
 connection_options([_, _, <<"rdbms">>|_] = Path, Options) ->
-    [{server, rdbms_server(Path, Options)}];
+    Interval = parse_kv(Path, <<"keepalive_interval">>, Options, undefined),
+    Server = rdbms_server(Path, Options),
+    case Interval of
+        undefined ->
+            [{server, Server}];
+        V ->
+            [{server, Server}, {keepalive_interval, V}]
+        end;
 connection_options([_, _, <<"riak">>|_] = Path, Options = #{<<"username">> := UserName,
                                                             <<"password">> := Password}) ->
     M = maps:without([<<"username">>, <<"password">>], Options),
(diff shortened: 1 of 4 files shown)


src/config/mongoose_config_parser_toml.erl
```diff
@@ -410,7 +410,13 @@ pool_option([<<"call_timeout">>|_], V) -> [{call_timeout, V}].
 %% path: outgoing_pools.*.connection
 -spec connection_options(path(), toml_section()) -> [option()].
 connection_options([_, _, <<"rdbms">>|_] = Path, Options) ->
-    [{server, rdbms_server(Path, Options)}];
+    {Server, Interval} = rdbms_server(Path, Options),
+    case Interval of
```
NelsonVides, line 414: I'm having the feeling that this complicates the rbdms_server function, why don't we just extract the interval in place, like
```suggestion
    Server = rdbms_server(Path, Options),
    Interval = maps:get(<<"keepalive_interval">>, M, undefined),
    case Interval of
```

src/config/mongoose_config_parser_toml.erl
```diff
@@ -410,7 +410,13 @@ pool_option([<<"call_timeout">>|_], V) -> [{call_timeout, V}].
...
+        undefined ->
+            [{server, Server}];
+        V ->
+            [{server, Server}, {keepalive_interval, V}]
+        end;
```
NelsonVides, line 419: And when it comes to this case of, we could regardless if its value, pass it along. If the default value is `undefined`, we can pass it in place and skip this case-of. @chrzaszcz what do you think it's the cleanest?

NelsonVides (commented):
Just have some comments about one function, otherwise code looks neat to me 👌

src/config/mongoose_config_parser_toml.erl
kmakiela: From the configuration standpoint it would make no difference, but from the equivalence standpoint it would be different thab `cfg` config without such value and because we test it by comparison between `cfg` and `toml`I did it this way

chrzaszcz: For me it is fine as it is now. However, calling `parse_kv` for `keepalive_interval` would make it much easier to plug in a validator for the interval.

src/config/mongoose_config_parser_toml.erl
```diff
@@ -452,46 +459,83 @@ db_tls(Path, #{<<"driver">> := <<"pgsql">>, <<"tls">> := Opts}) ->
...
 
-%% path: outgoing_pools.redis.connection.*
+%% path: outgoing_pools.http.*.connection.*
+-spec http_option(path(), toml_value()) -> [option()].
+http_option([<<"host">>|_], V) -> [{server, b2l(V)}];
+http_option([<<"path_prefix">>|_], V) -> [{path_prefix, b2l(V)}];
+http_option([<<"request_timeout">>|_], V) -> [{request_timeout, V}];
+http_option([<<"http_opts">>|_] = Path, V) ->
+    Opts = parse_section(Path, V),
+     [{http_opts, maps:from_list(Opts)}].
```
chrzaszcz, line 469: Minor: no reason to indent this. Are you using any automatic indentation helper in your editor?

test/config_parser_SUITE_data/outgoing_pools.toml
```diff
@@ -0,0 +1,77 @@
...
+    http_opts.retry = 1
+    http_opts.retry_timeout = 1000
+
+[outgoing_pools.riak.default]
+    scope = "global"
+    workers = 20 
+    strategy = "next_worker"
+    connection.address = "127.0.0.1"
+    connection.port = 8087
+    connection.credentials =  {user = "username", password = "pass"}
```
chrzaszcz, line 46: Very minor: extra space

chrzaszcz:
Only minor remarks from me, it looks good!

NelsonVides (approved):
👍 😄

chrzaszcz (approved):

PR #2852, opened 2020-08-31
TOML - Add support for last batch of options

kmakiela:
This PR addresses [MIM-1146](https://erlangsolutions.atlassian.net/secure/RapidBoard.jspa?rapidView=3&projectKey=MIM&modal=detail&selectedIssue=MIM-1146)

* [x] LDAP options
* [x] Listeners' options
* [x] Validation and tests for LDAP options

### DIFF
--- a/src/config/mongoose_config_parser_toml.erl
+++ b/src/config/mongoose_config_parser_toml.erl
@@ -196,15 +196,15 @@ http_listener_opt([<<"handlers">>|_] = Path, Handlers) ->
     [{modules, parse_section(Path, Handlers)}];
 http_listener_opt([<<"mongoose_api_admin">>|_] = Path, V) ->
     Opts = parse_section(Path, V),
-    Host = proplists:get_value(host, Opts),
-    P = proplists:get_value(path, Opts),
-    User = proplists:get_value(username, Opts),
-    Password = proplists:get_value(password, Opts),
+    {_, Host} = proplists:lookup(host, Opts),
+    {_, P} = proplists:lookup(path, Opts),
+    {_, User} = proplists:lookup(username, Opts),
+    {_, Password} = proplists:lookup(password, Opts),
     [{Host, P, mongoose_api_admin, [{auth, {User, Password}}]}];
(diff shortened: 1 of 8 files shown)


chrzaszcz:
Let me add the validation and let's ask @NelsonVides for review afterwards.

src/config/mongoose_config_parser_toml.erl
```diff
@@ -194,8 +194,38 @@ http_listener_opt([<<"protocol">>|_] = Path, Opts) ->
     [{protocol_options, parse_section(Path, Opts)}];
 http_listener_opt([<<"handlers">>|_] = Path, Handlers) ->
     [{modules, parse_section(Path, Handlers)}];
+http_listener_opt([<<"mongoose_api_admin">>|_] = Path, V) ->
+    Opts = parse_section(Path, V),
+    Host = proplists:get_value(host, Opts),
+    P = proplists:get_value(path, Opts),
+    User = proplists:get_value(username, Opts),
+    Password = proplists:get_value(password, Opts),
```
chrzaszcz, line 202: Quick remark: `proplists:get_value` can return `undefined` and we most likely don't want it. Could you amend all the `proplists:get_value` calls? I'd replcae them with `{_, Value} = proplists:lookup(Key, Opts)`

NelsonVides: @chrzaszcz if we do `{_, Value} = proplists:lookup(Key, Opts)` and the key wasn't present, `lookup/2` would return `none`, which would fail with a pattern-match. Is that what we want?

kmakiela: I can change that, but in this case we get `none` if the key does not exist instead of `undefined` - is it any better?

chrzaszcz: @NelsonVides we want to fail, this is precisely what we want. There is no point in cosntructing a tuple when a mandatory element is not present. This is a broken config and we want to fail fast.

NelsonVides (approved):
Excellent, thanks a lot for this work! 😄

chrzaszcz:
It's still not finished... validation in progress.

PR #2853, opened 2020-09-02
TOML: Validate and test rdbms pool config

chrzaszcz:
- Pool scope: make 'global' the default one
- Server options: rework them for more fine-grained validation

### DIFF
--- a/src/config/mongoose_config_parser_toml.erl
+++ b/src/config/mongoose_config_parser_toml.erl
@@ -397,13 +397,13 @@ is_inner_auth_opt(_) -> true.
 process_pool([Tag, Type|_] = Path, M) ->
     Scope = pool_scope(M),
     Options = parse_section(Path, maps:without([<<"scope">>, <<"host">>, <<"connection">>], M)),
-    ConnectionOptions = connection_options([<<"connection">> | Path],
-                                           maps:get(<<"connection">>, M, #{})),
+    ConnectionOptions = parse_kv(Path, <<"connection">>, M, #{}),
     [{b2a(Type), Scope, b2a(Tag), Options, ConnectionOptions}].
 
 -spec pool_scope(toml_section()) -> option().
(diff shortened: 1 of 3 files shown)


src/config/mongoose_config_parser_toml.erl
```diff
@@ -1518,24 +1529,40 @@ handler([_, <<"sasl_mechanisms">>, <<"auth">>]) -> fun sasl_mechanism/2;
...
+handler([_, <<"servers">>, <<"connection">>, _,
+         <<"cassandra">>, <<"outgoing_pools">>]) -> fun cassandra_server/2;
+handler([_, <<"connection">>, _,
+         <<"elastic">>, <<"outgoing_pools">>]) -> fun elastic_option/2;
+handler([_, <<"connection">>, _,
+         <<"rabbit">>, <<"outgoing_pools">>]) -> fun rabbit_option/2;
+handler([_, <<"tls">>, _, _, _, <<"outgoing_pools">>]) -> fun tls_option/2;
+handler([_, <<"versions">>, <<"tls">>, _, _, _, <<"outgoing_pools">>]) ->
     fun(_, Val) -> [b2a(Val)] end;
-handler([_, <<"ciphers">>, <<"tls">>, <<"connection">>, _, _, <<"outgoing_pools">>]) ->
```
kmakiela, line 1538: Can `tls` options be someplace different than in the `connection` section?

kmakiela (approved):
Looks good to me, thanks for making the old parts of `mongoose_config_parser_toml` consistent with the approach we adopted later on.

src/config/mongoose_config_parser_toml.erl
chrzaszcz: Actually the only issue is that now we can have either `<<"connection">>` or `{connection, DBType}` there as the DB type is stored in the path to be able to have different handlers for different DB types. I thought that pattern-matching for both was a bit verbose and that there is not much risk with parsing something else as it is at the end of the list for `outgoing_pools`

PR #2856, opened 2020-09-03
Remove redundant loglevel setting

gustawlippa:
It is really redundant as the correct key in config for OTP's logger's logging level is `logger_level` and not `loglevel`. 
This line was effectively discarding config and always setting the loglevel to `warning`.
The default loglevel for OTP's logger is `notice` and it will be set even if no such key exists in the config.

### DIFF
--- a/src/ejabberd_app.erl
+++ b/src/ejabberd_app.erl
@@ -54,7 +54,6 @@ start(normal, _Args) ->
     mongoose_service:start(),
     gen_mod:start(),
     ejabberd_config:start(),
-    mongoose_logs:set_global_loglevel(ejabberd_config:get_local_option_or_default(loglevel, warning)),
     connect_nodes(),
     mongoose_deprecations:start(),
     {ok, _} = Sup = ejabberd_sup:start_link(),


mongoose-im:
[8521.1](https://travis-ci.org/esl/MongooseIM/jobs/723749417) / Erlang 23.0.3 / small_tests / e0c55977019b1f62ffe35a8a6462b78c327e8ae0
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2856/8521/small_tests.23.0.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2856/8521/small_tests.23.0.3/small/ct_run.mongooseim@localhost.2020-09-03_11.18.48/lib.mongooseim.logs/run.2020-09-03_11.18.48/suite.log.html)

gustawlippa:
Oh, I think you are absolutely correct, `loglevel` is set in `mongooseim.cfg` and this takes precedence. In the future of course it will be set in the TOML config.
My mistake comes from looking at the `app.config` only and seeing the level always being set to `warning` (which was correct because that was what was set in the `mongooseim.cfg`). Well, this PR makes no sense then and I will close it. Thank you very much for the review and pointing it out!

PR #2857, opened 2020-09-04
Reject publish to non-existent pubsub node with nodeid-required

### LINKED ISSUE #2779: Incorrect response remote-server-not-found in pubsub_s2s tests publish_test and publish_without_node_attr_test?
MongooseIM version: 3.7.0
Installed from: source
Erlang/OTP version: 22

Describe the issue.

I have downloaded MongooseIM 3.7.0 source and am trying to run the big_tests.  All are passing except `pubsub_s2s_SUITE`.  I've looked at the public testing results for `pubsub_s2s_SUITE`; they are different from what I am getting, but they do not look right either.

I followed the chain of published test result links [Public CI](https://travis-ci.org/github/esl/MongooseIM) -> [riak/mnesia results](https://travis-ci.org/github/esl/MongooseIM/jobs/701544750) -> [Uploaded test results](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=branch/master/8347/riak_mnesia.22.0)

The results there show that in both [`publish_test`](http://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/branch/master/8347/riak_mnesia.22.0/big/ct_run.test@travis-job-69d20e69-77a8-4c80-bfc9-f01354d0aaf7.2020-06-24_07.07.36/big_tests.tests.pubsub_s2s_SUITE.logs/run.2020-06-24_07.19.12/pubsub_s2s_suite.publish_test.7971.html) and [`publish_without_node_test`](http://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/branch/master/8347/riak_mnesia.22.0/big/ct_run.test@travis-job-69d20e69-77a8-4c80-bfc9-f01354d0aaf7.2020-06-24_07.07.36/big_tests.tests.pubsub_s2s_SUITE.logs/run.2020-06-24_07.19.12/pubsub_s2s_suite.publish_without_node_attr_test.9266.html), the server responds with a `remote-server-not-found` response:

```
	<error code='404' type='cancel'>
		<remote-server-not-found xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/>
	</error>
```

This matches the expectation of [`type='cancel'`](https://github.com/esl/MongooseIM/blob/master/big_tests/tests/pubsub_s2s_SUITE.erl#L112) in the test suite, but `remote-server-not-found` does not make sense to me in this context.  It also doesn't seem to match any of the `cancel` errors specified in [XEP-0060](https://xmpp.org/extensions/xep-0060.html).

Is this the correct response?  If so, can anyone explain the code path that leads to this response?

P.S. In my local tests, instead of 404, I get:
- 403 forbidden for `publish_test`
- a success response with an auto-created node for `publish_without_node_test` (created by the "instant-nodes" feature?)

The former seems correct by XEP-0060 (but doesn't match the test suite expectation), the latter seems wrong by [XEP-0060](https://xmpp.org/extensions/xep-0060.html#publisher-publish-request):

> The <publish/> element MUST possess a 'node' attribute, specifying the NodeID of the node.

But, it's hard for me to debug my local tests because I don't understand the behaviour of MongooseIM against the public tests.

balgillo:
Closes #2779 

Change the response to a publish stanza that lacks a `node` attribute.  Currently, a node with a random name will be auto-created.  But [XEP-0060](https://xmpp.org/extensions/xep-0060.html#publisher-publish-request) states:
> The element MUST possess a 'node' attribute, specifying the NodeID of the node.

XEP-0060 is not clear on what error should be returned if the node attribute is missing, so I've implemented the same response that it specifies for [retract](https://xmpp.org/extensions/xep-0060.html#publisher-delete-error-nodeid) and [configure](https://xmpp.org/extensions/xep-0060.html#owner-configure-error-nodeid) stanzas that lack NodeIDs:
```
  <error type='modify'>
    <bad-request xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/>
    <nodeid-required xmlns='http://jabber.org/protocol/pubsub#errors'/>
  </error>
```
The unit tests in pubsub_s2s_SUITE need corresponding changes.  Currently in the public test results for [publish_test](http://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/branch/master/8347/riak_mnesia.22.0/big/ct_run.test@travis-job-69d20e69-77a8-4c80-bfc9-f01354d0aaf7.2020-06-24_07.07.36/big_tests.tests.pubsub_s2s_SUITE.logs/run.2020-06-24_07.19.12/pubsub_s2s_suite.publish_test.7971.html) and the [publish_without_node_test](http://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/branch/master/8347/riak_mnesia.22.0/big/ct_run.test@travis-job-69d20e69-77a8-4c80-bfc9-f01354d0aaf7.2020-06-24_07.07.36/big_tests.tests.pubsub_s2s_SUITE.logs/run.2020-06-24_07.19.12/pubsub_s2s_suite.publish_without_node_attr_test.9266.html), the server is responding with a remote-server-not-found response:
```
	<error code='404' type='cancel'>
		<remote-server-not-found xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/>
	</error>
```
In my local testing, I was getting that result intermittently.  The cause seems to be DNS resolution of the host `pubsub.localhost`.  I've configured the big_test fed1 instance to bypass DNS for `pubsub.localhost` (it was already configured to bypass DNS for `localhost`, but that configuration did not affect subdomains).  While making this change, I also configured it to bypass DNS for subdomain `muc.localhost` as well, because I saw intermittent failures of `muc_SUITE:register_over_s2s` for DNS-related reasons; while this doesn't seem to be a problem in the public test results, it seems like a good change anyway.

Once the DNS was bypassed, I get the following responses in the big tests:

`pubsub_s2s_SUITE:publish_test`
```
	<error code='403' type='auth'>
		<forbidden xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/>
	</error>
```

`pubsub_s2s_SUITE:publish_without_node_attr_test`
```
	<error code='400' type='modify'>
		<bad-request xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/>
		<nodeid-required xmlns='http://jabber.org/protocol/pubsub#errors'/>
	</error>
```

These responses seem correct (certainly more correct than `remote-server-not-found`) so I have updated the big test expectations to match.

### DIFF
--- a/rel/fed1.vars.config
+++ b/rel/fed1.vars.config
@@ -14,7 +14,7 @@
 {hosts, "[ \"fed1\" ]"}.
 {sm_backend, "{mnesia, []}"}.
 {auth_method, "internal"}.
-{s2s_addr, "{ {s2s_addr, \"localhost\"}, {127,0,0,1} }. { {s2s_addr, \"localhost.bis\"}, {127,0,0,1} }."}.
+{s2s_addr, "{ {s2s_addr, \"localhost\"}, {127,0,0,1} }. { {s2s_addr, \"pubsub.localhost\"}, {127,0,0,1} }. { {s2s_addr, \"muc.localhost\"}, {127,0,0,1} }. { {s2s_addr, \"localhost.bis\"}, {127,0,0,1} }."}.
 {s2s_default_policy, allow}.
 {highload_vm_args, ""}.
 {ejabberd_service, ""}.
--- a/src/pubsub/mod_pubsub.erl
+++ b/src/pubsub/mod_pubsub.erl
@@ -1389,6 +1389,8 @@ iq_pubsub_set_create(Host, Node, From,
             create_node(Host, ServerHost, Node, From, Type, Access, Config)
     end.
 
+iq_pubsub_set_publish(_Host, <<>>, _From, _ExtraArgs) ->
+    {error, extended_error(mongoose_xmpp_errors:bad_request(), <<"nodeid-required">>)};
 iq_pubsub_set_publish(Host, Node, From, #{server_host := ServerHost, access := Access,
                                           action_el := ActionEl, query_el := QueryEl}) ->
     case xml:remove_cdata(ActionEl#xmlel.children) of
--- a/big_tests/tests/pubsub_s2s_SUITE.erl
+++ b/big_tests/tests/pubsub_s2s_SUITE.erl
@@ -109,7 +109,7 @@ publish_test(Config) ->
               Node = pubsub_tools:pubsub_node(),
               pubsub_tools:create_node(Alice, Node, []),
               pubsub_tools:publish(Alice, <<"item1">>, Node, []),
-              pubsub_tools:publish(Alice2, <<"item2">>, Node, [{expected_error_type, <<"cancel">>}]),
+              pubsub_tools:publish(Alice2, <<"item2">>, Node, [{expected_error_type, <<"auth">>}]),
               pubsub_tools:delete_node(Alice, Node, [])
       end).
 
@@ -121,7 +121,7 @@ publish_without_node_attr_test(Config) ->
               Node = pubsub_tools:pubsub_node(),
               pubsub_tools:create_node(Alice, Node, []),
               pubsub_tools:publish(Alice, <<"item1">>, Node, []),
-              pubsub_tools:publish_without_node_attr(Alice2, <<"item2">>, Node, [{expected_error_type, <<"cancel">>}]),
+              pubsub_tools:publish_without_node_attr(Alice2, <<"item2">>, Node, [{expected_error_type, <<"modify">>}]),
               pubsub_tools:delete_node(Alice, Node, [])
       end).
 


rel/fed1.vars.config
```diff
@@ -14,7 +14,7 @@
 {hosts, "[ \"fed1\" ]"}.
 {sm_backend, "{mnesia, []}"}.
 {auth_method, "internal"}.
-{s2s_addr, "{ {s2s_addr, \"localhost\"}, {127,0,0,1} }. { {s2s_addr, \"localhost.bis\"}, {127,0,0,1} }."}.
+{s2s_addr, "{ {s2s_addr, \"localhost\"}, {127,0,0,1} }. { {s2s_addr, \"pubsub.localhost\"}, {127,0,0,1} }. { {s2s_addr, \"muc.localhost\"}, {127,0,0,1} }. { {s2s_addr, \"localhost.bis\"}, {127,0,0,1} }."}.
```
DenysGonchar, line 17: doesn't `{s2s_addr, "{ {s2s_addr, \"localhost\"}, {127,0,0,1} }.` cover sub-domains (`pubsub.localhost` & `muc.localhost`)?

DenysGonchar:
> Change the response to a publish stanza that lacks a `node` attribute. Currently, a node with a random name will be auto-created. But [XEP-0060](https://xmpp.org/extensions/xep-0060.html#publisher-publish-request) states:
> 
> > The element MUST possess a 'node' attribute, specifying the NodeID of the node.

taking into into account the changes in `publish_without_node_attr_test` tescase I don't see the describbed behaviour.
according to the original test there is no creation of the node with a random name, but `{expected_error_type, <<"cancel">>}` error is expected.  please clarify that.

rel/fed1.vars.config
balgillo: I believe not, for two reasons:
1. The `ejabberd_s2s_out` code looks up the full XMPP domain name in the `s2s_addr` list via `get_predefined_addresses`, which doesn't have any logic related to subdomains.  https://github.com/esl/MongooseIM/blob/master/src/ejabberd_s2s_out.erl#L1198
2. Changing this setting resolved `remote-server-not-found` in the `pubsub_s2s_SUITE` results.

balgillo:
> taking into into account the changes in publish_without_node_attr_test tescase I don't see the describbed behaviour.
> according to the original test there is no creation of the node with a random name, but {expected_error_type, <<"cancel">>} error is expected.  please clarify that.

A node with a random name is created unless `remote-server-not-found` occurs.  In your travis test environment it seems that `remote-server-not-found` always occurs (i.e. resolution of `pubsub.localhost` fails).

If you run the `pubsub_s2s_SUITE` with only the change to `fed1.vars.config` to bypass resolution of `pubsub.localhost` in place, then you will see the dynamic creation of a randomly-named pubsub node.

rel/fed1.vars.config
DenysGonchar: It seems that adding `{route_subdomains, s2s}.` option solves this problem. see [documentation](https://mongooseim.readthedocs.io/en/latest/Advanced-configuration/#served-hostnames) for more information.

DenysGonchar (changes requested):
please use `route_subdomains` option instead of explicit s2s subdomains configuration.

UPDATE: my fault, I misinterpreted the option.

rel/fed1.vars.config
balgillo: I don't see how it will help.  The problem is fed1 (which serves XMPP domain `fed1`) trying to contact mim1 (which serves XMPP domains `localhost` and `pubsub.localhost`).

Configuring `route_subdomains` on fed1 would change how fed1 routes messages to `subdomain.fed1`, and configuring `route_subdomains` on mim1 would change how mim1 routes messages to `subdmain.localhost`, but the problem is how fed1 routes messages to `subdomain.localhost`, which would not be affected as fed1 does not serve XMPP domain `localhost`.

DenysGonchar: Yes, you're right. I misinterpreted it. just had a quick conversation within the team. All subdomains must be explicitly configured using DNS or s2s_addr option. That is the only right and possible way.

DenysGonchar (approved):

DenysGonchar:
Thanks for catching and fixing that.

src/pubsub/mod_pubsub.erl
```diff
@@ -1391,18 +1391,23 @@ iq_pubsub_set_create(Host, Node, From,
...
-        [#xmlel{name = <<"item">>, attrs = ItemAttrs, children = Payload}] ->
-            ItemId = xml:get_attr_s(<<"id">>, ItemAttrs),
-            PublishOptions = exml_query:path(QueryEl,
-                                             [{element, <<"publish-options">>},
-                                              {element, <<"x">>}]),
-            publish_item(Host, ServerHost, Node, From, ItemId,
-                         Payload, Access, PublishOptions);
-        [] ->
-            {error, extended_error(mongoose_xmpp_errors:bad_request(), <<"item-required">>)};
+    case Node of
```
DenysGonchar, line 1394: that would be nice to have a separate `iq_pubsub_set_publish/4` function clause matching <<"">> as Node argument instead of this extra `case`

balgillo:
Thank you @DenysGonchar for reviewing.  I've made the suggested improvement to `mod_pubsub.erl` and rebased to keep this MR at one commit.  Hopefully good to merge now but let me know if there are any other issues.

NelsonVides (approved):
Great great catch, thank you very much for this fix 🎉 😄

PR #2861, opened 2020-09-07
Validate the 'shaper', 'acl' and 'access' sections

chrzaszcz:
Parse the 'acl' section in a more structured way,
matching exact rules from the 'acl' module.

### DIFF
--- a/src/config/mongoose_config_parser_toml.erl
+++ b/src/config/mongoose_config_parser_toml.erl
@@ -1342,7 +1342,7 @@ mod_vcard_ldap_uids(_, #{<<"attr">> := Attr}) ->
     [b2a(Attr)].
 
 -spec mod_vcard_ldap_vcard_map(path(), toml_section()) -> [option()].
-mod_vcard_ldap_vcard_map(_, #{<<"vcard_field">> := VF, <<"ldap_pattern">> := LP, 
+mod_vcard_ldap_vcard_map(_, #{<<"vcard_field">> := VF, <<"ldap_pattern">> := LP,
     <<"ldap_field">> := LF}) ->
     [{VF, LP, [LF]}].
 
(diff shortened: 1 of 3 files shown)


src/config/mongoose_config_parser_toml.erl
```diff
@@ -1365,21 +1365,44 @@ process_acl([ACLName, _|Path], Content) ->
...
+    {AclName, AclKeys} = find_acl(M, lists:sort(maps:keys(M)), acl_keys()),
+    list_to_tuple([AclName | lists:map(fun(K) -> maps:get(K, M) end, AclKeys)]).
+
+find_acl(M, SortedMapKeys, [{AclName, AclKeys}|Rest]) ->
+    case lists:sort(AclKeys) of
+        SortedMapKeys -> {AclName, AclKeys};
+        _ -> find_acl(M, SortedMapKeys, Rest)
+    end.
+
+acl_keys() ->
```
gustawlippa, line 1380: In `acl` module there is also `resource`, isn't it missing?

chrzaszcz: Good catch!

src/config/mongoose_config_parser_toml.erl
```diff
@@ -1365,21 +1365,44 @@ process_acl([ACLName, _|Path], Content) ->
...
-            list_to_tuple([b2a(Key) | Values]);
-        [{Key, Value}] when is_binary(Value) ->
-            {b2a(Key), Value}
-    end;
-acl_data(Value) when is_binary(Value) -> b2a(Value).
+acl_data(<<"all">>) -> all;
+acl_data(<<"none">>) -> none;
+acl_data(M) ->
+    {AclName, AclKeys} = find_acl(M, lists:sort(maps:keys(M)), acl_keys()),
+    list_to_tuple([AclName | lists:map(fun(K) -> maps:get(K, M) end, AclKeys)]).
```
gustawlippa, line 1372: If I understand this line correctly, I would find something like this more readable, but it's a matter of taste.
```suggestion
    list_to_tuple([AclName | maps:to_list(maps:with(AclKeys, M))).
```

chrzaszcz: Not quite the same, your version changes the ordering of the keys.

PR #2862, opened 2020-09-08
TOML: test and validate s2s options

chrzaszcz:
Validate and test the `s2s` options.

Some fixes and improvements to this section:
- Make subsections out of the `dns` and `outgoing` option groups
- Clean up the parsing logic, use the `handler` where possible
- Decouple the internal representation of the IP versions (address families) and the connection timeout to improve code structure.

See the individual commits for more details.

Note: the `domain_certfiles` options should be moved out of the s2s section, preferably to `host_config`, as it also affects c2s listeners. This is not urgent, but it would be good to create a story for it.

### DIFF
--- a/src/config/mongoose_config_parser_toml.erl
+++ b/src/config/mongoose_config_parser_toml.erl
@@ -110,10 +110,8 @@ process_general([<<"loglevel">>|_], V) ->
     [#local_config{key = loglevel, value = b2a(V)}];
 process_general([<<"hosts">>|_] = Path, Hosts) ->
     [#config{key = hosts, value = parse_list(Path, Hosts)}];
-process_general([<<"registration_timeout">>|_], <<"infinity">>) ->
-    [#local_config{key = registration_timeout, value = infinity}];
 process_general([<<"registration_timeout">>|_], V) ->
-    [#local_config{key = registration_timeout, value = V}];
+    [#local_config{key = registration_timeout, value = int_or_infinity(V)}];
 process_general([<<"language">>|_], V) ->
     [#config{key = language, value = V}];
 process_general([<<"all_metrics_are_global">>|_], V) ->
@@ -289,8 +287,7 @@ s2s_tls_option([Opt|_] = Path, Val) when Opt =:= <<"cacertfile">>;
 %% path: listen.http[].transport.*
 -spec cowboy_transport_opt(path(), toml_value()) -> [option()].
 cowboy_transport_opt([<<"num_acceptors">>|_], N) -> [{num_acceptors, N}];
-cowboy_transport_opt([<<"max_connections">>|_], <<"infinity">>) -> [{max_connections, infinity}];
-cowboy_transport_opt([<<"max_connections">>|_], N) -> [{max_connections, N}].
+cowboy_transport_opt([<<"max_connections">>|_], N) -> [{max_connections, int_or_infinity(N)}].
 
 %% path: listen.http[].protocol.*
 -spec cowboy_protocol_opt(path(), toml_value()) -> [option()].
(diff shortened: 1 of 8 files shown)


src/config/mongoose_config_parser_toml.erl
```diff
@@ -91,14 +91,6 @@ process_section([<<"services">>] = Path, Content) ->
 process_section([<<"modules">>] = Path, Content) ->
     Mods = parse_section(Path, Content),
     ?HOST_F([#local_config{key = {modules, Host}, value = Mods}]);
-process_section([<<"s2s">>] = Path, Content) ->
```
NelsonVides, line 94: At first sight, I can see that `process_section` has clauses for many other sections, but not now for `s2s`, isn't that a bit suspicious? 😕

NelsonVides (commented):
x
I think that the sections
```erlang
module_opt([<<"max_room_id">>, <<"mod_muc">>|_], <<"infinity">>) ->
    [{max_room_id, infinite}];
module_opt([<<"max_room_id">>, <<"mod_muc">>|_], V) ->
    [{max_room_id, V}];
module_opt([<<"max_room_name">>, <<"mod_muc">>|_], <<"infinity">>) ->
    [{max_room_name, infinite}];
module_opt([<<"max_room_name">>, <<"mod_muc">>|_], V) ->
    [{max_room_name, V}];
module_opt([<<"max_room_desc">>, <<"mod_muc">>|_], <<"infinity">>) ->
    [{max_room_desc, infinite}];
module_opt([<<"max_room_desc">>, <<"mod_muc">>|_], V) ->
    [{max_room_desc, V}];
```
can also take the `int_or_infinity` trick, right? Also some other about 

Also note that the `int_or_infitiny` function is already doing some validation, as it is enforcing the value to be an integer and not anything else. Is that intended?

Splitting the s2s config, excellent idea.

I've just noticed though that our docs about s2s are quite poor, so I have little to check from 😕

src/config/mongoose_config_parser_toml.erl
chrzaszcz: This is because the `s2s` section is now handled by `parse_section`, just like `shaper`, `acl` or `access`. These clauses are more like exceptions than the norm and they should disappear in the future when we get rid of the nested tuples in the internal configuration format.

chrzaszcz:
> I think that the sections
> 
> ```erlang
> module_opt([<<"max_room_id">>, <<"mod_muc">>|_], <<"infinity">>) ->
>     [{max_room_id, infinite}];
> module_opt([<<"max_room_id">>, <<"mod_muc">>|_], V) ->
>     [{max_room_id, V}];
> module_opt([<<"max_room_name">>, <<"mod_muc">>|_], <<"infinity">>) ->
>     [{max_room_name, infinite}];
> module_opt([<<"max_room_name">>, <<"mod_muc">>|_], V) ->
>     [{max_room_name, V}];
> module_opt([<<"max_room_desc">>, <<"mod_muc">>|_], <<"infinity">>) ->
>     [{max_room_desc, infinite}];
> module_opt([<<"max_room_desc">>, <<"mod_muc">>|_], V) ->
>     [{max_room_desc, V}];
> ```
> 
> can also take the `int_or_infinity` trick, right?
They cannot as the resulting atom is different. When we fix that, we can use the helper.
> Also some other about
> 
> Also note that the `int_or_infitiny` function is already doing some validation, as it is enforcing the value to be an integer and not anything else. Is that intended?

The rule I followed here was to have the code as simple as possible. I think that `int_or_atom` would be more confusing and would have less meaning. I wouldn't complicate code to separate the things that are so connected - actually conversion from binary to int is validation as well...

> 
> Splitting the s2s config, excellent idea.
> 
> I've just noticed though that our docs about s2s are quite poor, so I have little to check from 😕
Can't comment on that as there is no example provided and I don't think this PR should do anything about it.

NelsonVides (approved):
Right, I confused `infinity` with `infinite` 😅
Good to see how it is progressing and what is next, thanks for this PR!

PR #2864, opened 2020-09-09
TOML: validate host config

chrzaszcz:
Test host config: each allowed option is now validated and tested for both cases: per-host and globally.

Handling of host config was also reworked - see the individual commits for details.
The main motivation was to avoid having two separate clauses in the parser for handling the global and per-host option.

### DIFF
--- a/src/config/mongoose_config_validator_toml.erl
+++ b/src/config/mongoose_config_validator_toml.erl
@@ -46,20 +46,24 @@ validate([item, <<"override">>, <<"general">>],
 validate([<<"override">>, <<"general">>],
          Items) ->
     validate_unique_items(Items);
-validate([<<"pgsql_users_number_estimate">>, <<"general">>],
+validate([<<"pgsql_users_number_estimate">>, <<"general">>|Path],
          [#local_config{value = Value}]) ->
+    validate_root_or_host_config(Path),
     validate_boolean(Value);
-validate([<<"route_subdomains">>, <<"general">>],
+validate([<<"route_subdomains">>, <<"general">>|Path],
          [#local_config{value = Value}]) ->
+    validate_root_or_host_config(Path),
     validate_enum(Value, [s2s]);
 validate([item, <<"routing_modules">>, <<"general">>],
          [Value]) ->
     validate_module(Value);
-validate([<<"replaced_wait_timeout">>, <<"general">>],
+validate([<<"replaced_wait_timeout">>, <<"general">>|Path],
          [#local_config{value = Value}]) ->
+    validate_root_or_host_config(Path),
     validate_positive_integer(Value);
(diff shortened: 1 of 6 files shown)


test/config_parser_SUITE.erl
```diff
@@ -956,6 +971,24 @@ rdbms_opts() ->
...
+eq_host_or_global(ResultF, Config) ->
+    ?eq(ResultF(global), parse(Config)), % check for the 'global' host
+    ?eq(ResultF(?HOST), parse_host_config(Config)). % check for a single host
+
+err_host_config(Config) ->
+    ?err(parse(Config)),
+    ?err(parse_host_config(Config)).
+
+parse_host_config(Config) ->
+    parse(#{<<"host_config">> => [Config#{<<"host">> => ?HOST}]}).
```
NelsonVides, line 990: 👌

src/config/mongoose_config_validator_toml.erl
```diff
@@ -323,3 +345,6 @@ validate_password_format(Value) -> validate_enum(Value, [scram, plain]).
 
 validate_pool_scope(Value) when is_binary(Value) -> validate_non_empty_binary(Value);
 validate_pool_scope(Value) -> validate_enum(Value, [host, global]).
+
+validate_root_or_host_config([]) -> ok;
+validate_root_or_host_config([{host, _}, <<"host_config">>]) -> ok.
```
NelsonVides, line 350: I understand that this extra check should apply only to those paths which allow to be set either global and per-host, right? In that case, why doesn't it apply to other `s2s` paths like `[<<"ip_address">>, item, <<"address">>, <<"s2s">>]`? I'm probably missing something about `s2s` configuration, I'm sure 🤔

NelsonVides (commented):
Excellent, very readable! Just a couple of questions, most likely out of curiosity 🙂

src/config/mongoose_config_validator_toml.erl
chrzaszcz: Yes, this is correct, it only applies to the options that are constructed with the `?HOST` macro, creating a function which returns the config for a host. And there are only three s2s options like this: `default_policy`, `shared`, `max_retry_delay`. The `address` one you mentioned contains `host`, but it is not a per-host setting. That `host` is just a regular hostname of an external server.

arcusfelis: does it mean I need to add `validate_root_or_host_config` for **every** module option validator?

chrzaszcz: @arcusfelis for modules I'd recommend:

```validate([_, <<"modules">>|Path],
         [Value]) ->
    validate_root_or_host_config(Path);```

And then you don't need it for **every** module option validator.

kmakiela (approved):

PR #2865, opened 2020-09-10
fix mongoose_api_admin and mongoose_api_client opts

kmakiela:
This PR fixes the way we parsed `mongoose_api_client` and `mongoose_api_admin` options. They are moved to `listen.http.handlers` subsection

### DIFF
--- a/test/config_parser_SUITE_data/mongooseim-pgsql.cfg
+++ b/test/config_parser_SUITE_data/mongooseim-pgsql.cfg
@@ -25,10 +25,11 @@
                                          {timeout, infinity},
                                          {ping_rate, none},
                                          {max_stanza_size, 100}
-                                                                         ]}
-                                      ]},
-                            {"localhost", "/api", mongoose_api_admin, [{auth, {<<"ala">>, <<"makotaipsa">>}}]},
-                            {"localhost", "/api/contacts/{:jid}", mongoose_api_client_contacts, []}
+                                                                         ]},
+                                       {"localhost", "/api", mongoose_api_admin, [{auth, {<<"ala">>, <<"makotaipsa">>}}]},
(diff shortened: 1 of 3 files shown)


test/config_parser_SUITE_data/mongooseim-pgsql.cfg
```diff
@@ -25,10 +25,11 @@
                                          {timeout, infinity},
                                          {ping_rate, none},
                                          {max_stanza_size, 100}
-                                                                         ]}
-                                      ]},
-                            {"localhost", "/api", mongoose_api_admin, [{auth, {<<"ala">>, <<"makotaipsa">>}}]},
-                            {"localhost", "/api/contacts/{:jid}", mongoose_api_client_contacts, []}
```
NelsonVides, line 31: Why did `mongoose_api_client_contacts` change to `mongoose_api_client`?

NelsonVides (commented):
Excellent, much shorter and readable now! Just have one question :)

test/config_parser_SUITE_data/mongooseim-pgsql.cfg
kmakiela: I believe it was a mistake taken from [this doc](https://mongooseim.readthedocs.io/en/latest/advanced-configuration/Listener-modules/), should have been `mongoose_api_client` in the first place

NelsonVides: But then why the `path` part still contains `contacts`? 😕

chrzaszcz: Ok, I am documenting this right now, let me sort out the docs with this path as well.

chrzaszcz (approved):
The necessary fixes are OK. Thanks @kmakiela

PR #2867, opened 2020-09-11
Km/toml val out

kmakiela:
This PR addresses [MIM-1151](https://erlangsolutions.atlassian.net/secure/RapidBoard.jspa?rapidView=3&projectKey=MIM&modal=detail&selectedIssue=MIM-1151)

* Added tests and validation for every type of connection pool
* A few minor fixes in the parser

### DIFF
--- a/src/config/mongoose_config_validator_toml.erl
+++ b/src/config/mongoose_config_validator_toml.erl
@@ -248,12 +248,12 @@ validate([<<"path_prefix">>, _Conn, _Tag, <<"http">>, <<"outgoing_pools">>],
 validate([<<"request_timeout">>, _Conn, _Tag, <<"http">>, <<"outgoing_pools">>],
          [{request_timeout, Value}]) ->
     validate_non_negative_integer(Value);
-validate([<<"http_opts">>, _Conn, _Tag, <<"http">>, <<"outgoing_pools">>],
-         [{http_opts, Value}]) ->
-    validate_map(Value),
-    [validate_enum(Key, [retry, retry_timeout]) || Key <- maps:keys(Value)],
-    validate_non_negative_integer(maps:get(retry, Value, 1)),
-    validate_positive_integer(maps:get(retry_timeout, Value, 1));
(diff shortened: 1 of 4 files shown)


kmakiela:
2 comments:
1. In some cases (i.e. most `riak` options) I couldn't find if empty string are allowed. I decided not to accept empty values unless explicitly stated in the docs or I could find an example suggesting otherwise
2. In `riak` and `cassandra` option there is no `tls` validation. I went for the same approach as with `rdbms.tls` section meaning no explicit validation of `tls` options

NelsonVides (approved):
I think this is fantastic :)

I saw that `ssl`/`tls`_`opts` are not verified for Riak nor LDAP for example, I guess this is expected, last thing I remember is that we're not passing opts to internal libraries, right?

src/config/mongoose_config_validator_toml.erl
```diff
@@ -224,21 +224,112 @@ validate([<<"strategy">>, _Tag, _Type, <<"outgoing_pools">>],
...
+    validate_non_empty_string(Value);
+validate([<<"request_timeout">>, _Conn, _Tag, <<"http">>, <<"outgoing_pools">>],
+         [{request_timeout, Value}]) ->
+    validate_non_negative_integer(Value);
+validate([<<"http_opts">>, _Conn, _Tag, <<"http">>, <<"outgoing_pools">>],
+         [{http_opts, Value}]) ->
+    validate_map(Value),
+    [validate_enum(Key, [retry, retry_timeout]) || Key <- maps:keys(Value)],
+    validate_non_negative_integer(maps:get(retry, Value, 1)),
+    validate_positive_integer(maps:get(retry_timeout, Value, 1));
```
chrzaszcz, line 256: Split this, it's better to validate the value at the leaf, so that the error report will be correct. Also: I think we agreed at one point to flatten `http_opts` to avoid the `request_timeout` vs `http_opts.retry_timeout` inconsistency.

chrzaszcz (approved):

PR #2868, opened 2020-09-14
Validate modules

arcusfelis:
Check the rework instead https://github.com/esl/MongooseIM/pull/2875

This PR addresses "modules validation for TOML" task.

Proposed changes include:
* Validators for module options
* Tests (happy path/failing path) for each option
* It's good idea to start review from the approach description https://github.com/esl/MongooseIM/pull/2868/files#diff-530ac33fc6ed2655a0a48ed0b934112eR433


src/config/mongoose_config_validator_toml.erl
```diff
@@ -294,6 +295,126 @@ validate([<<"max_retry_delay">>, <<"s2s">>|Path],
...
+    validate_non_empty_atom(Value);
+validate([<<"registration_watchers">>,<<"mod_register">>,<<"modules">>],
+         [{registration_watchers, Value}]) ->
+    validate_list_of_jids(Value);
+validate([<<"password_strength">>,<<"mod_register">>,<<"modules">>],
+         [{password_strength, Value}]) ->
+    validate_non_negative_integer(Value);
+
+validate(Option, Value) ->
+    ?LOG_ERROR(#{ what => validate_unknown, option => Option, value => Value}),
```
chrzaszcz, line 416: Currently validation is supplementary, so this is not an error.

test/config_parser_SUITE.erl
```diff
@@ -932,7 +945,284 @@ s2s_max_retry_delay(_Config) ->
                   #{<<"s2s">> => #{<<"max_retry_delay">> => 120}}),
     err_host_config(#{<<"s2s">> => #{<<"max_retry_delay">> => 0}}).
 
-%% Helpers for 'listen' tests
+modopts(Mod, Opts) ->
+    [#local_config{key = {modules, ?HOST}, value = [{Mod, Opts}]}].
```
chrzaszcz, line 949: Move the helpers below.

test/config_parser_SUITE.erl
```diff
@@ -932,7 +945,284 @@ s2s_max_retry_delay(_Config) ->
...
+
+%% modules
+
+%% ---------------------------------------------------------------------------
```
chrzaszcz, line 953: I am not a fan of this kind of ascii art, for me it only clutters the code. What is more, this formatting is inconsistent with the rest of this file. I'd avoid these decorations if possible. I'd like to hear a second opinion here though.

test/config_parser_SUITE.erl
```diff
@@ -1006,6 +1296,8 @@ handle_config_option(#local_config{key = K1, value = V1},
 handle_config_option(Opt1, Opt2) ->
     ?eq(Opt1, Opt2).
 
+compare_values(_, X, X) -> %% I am a simple man
+    ok;
```
chrzaszcz, line 1300: Already handled by the last clause, remove.

test/config_parser_SUITE.erl
```diff
@@ -12,6 +12,9 @@
 
 -define(HOST, <<"myhost">>).
 
+-define(eqf(Expected, Actual), ?assertEqual(Expected, apply(hd(Actual), [?HOST]))).
+-define(errf(Expr), ?assertError(_, apply(hd(Expr), [?HOST]))).
```
chrzaszcz, line 16: These are very tricky, especially if `Actual` contains more elements, all but the head would get unnoticed... maybe you could use `eq_host_config` and `err_host_config` which also check the per-host case? See the latest `toml-config` revision.

chrzaszcz (commented):
Nice to see the progress! I have some comments.
My current bottom line is that maybe in some cases such detailed validation is not necessary at this point... especially if something is hard to do and we could deliver the whole functionality much quicker without it (e.g. the netmask validation).

src/config/mongoose_config_parser_toml.erl
```diff
@@ -864,15 +864,17 @@ module_opt([<<"max_http_connections">>, <<"mod_push_service_mongoosepush">>|_],
...
     [{registration_watchers, parse_list(Path, V)}];
 module_opt([<<"password_strength">>, <<"mod_register">>|_], V) ->
     [{password_strength, V}];
 module_opt([<<"ip_access">>, <<"mod_register">>|_] = Path, V) ->
     Rules = parse_list(Path, V),
     [{ip_access, Rules}];
+module_opt([<<"welcome_message">>, <<"mod_register">>|_], V) ->
+    Subject = maps:get(<<"subject">>, V, <<>>),
+    Body = maps:get(<<"body">>, V, <<>>),
+    [{welcome_message, {binary_to_list(Subject), binary_to_list(Body)}}];
```
kmakiela, line 877: I think we should call `parse_section` here for the `subject` and `body` opts

test/config_parser_SUITE.erl
kmakiela: I second that, the consistency being the most important reason

src/config/mongoose_config_validator_toml.erl
arcusfelis: that line for me, ignore

test/config_parser_SUITE.erl
arcusfelis: no, it fails for some option.

src/config/mongoose_config_parser_toml.erl
arcusfelis: ok, done

arcusfelis:
Moved to https://github.com/esl/MongooseIM/pull/2875

PR #2869, opened 2020-09-14
TOML documentation: listen

chrzaszcz:
Looking at the [original section](https://mongooseim.readthedocs.io/en/latest/advanced-configuration/Listener-modules/) I found some options needing changes.

They are listed in the first commit message.

The last commit documents the new TOML config section.


kmakiela (commented):
I have a comment regarding `tls` options - some of them are duplicated in this file already and could be even more redundant going forward - maybe we could create a separate section for the `tls` and `fast_tls` modules?

chrzaszcz:
> I have a comment regarding `tls` options - some of them are duplicated in this file already and could be even more redundant going forward - maybe we could create a separate section for the `tls` and `fast_tls` opts?

That was my initial idea, but when I did it that way, there were following issues:
- relying on low-level facts like the choice of `tls` and `fast_tls`
- too many redirections between sections
- the common parts are quite irregular and if I think that it's better to explicitly duplicate the identical facts

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+* **Example:** `tls.cacertfile = "ca.pem"`
+
+Path to the X509 PEM file with a CA chain that will be used to verify clients. Won't have any effect if `verify_peer` is not enabled.
+
+#### `listen.c2s.tls.dhfile`
+* **Syntax:** string, path in the file system
+* **Default:** not set
+* **Example:** `tls.dhfile = "dh.pem"`
+
+Path to the Diffie Hellman parameter file.
```
kmakiela, line 200: It's not immediately obvious where these files should be put, maybe mention absolute and relative possibilities here?

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+* **Syntax:** string with the OpenSSL cipher suite specification
+* **Default:** `"TLSv1.2:TLSv1.3"`
+* **Example:** `tls.ciphers = "ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384"`
+
+Cipher suites to use with StartTLS or TLS. Please refer to the [OpenSSL documentation](http://www.openssl.org/docs/man1.0.2/apps/ciphers.html) for the cipher string format.
+
+#### `listen.c2s.tls.ciphers` - for `just_tls`
+* **Syntax:** array of tables with the following keys: `cipher`, `key_exchange`, `mac`, `prf` and string values.
+* **Default:** not set, all supported cipher suites are accepted
+* **Example:** `tls.ciphers = "[{cipher = "aes_25_gcm", key_exchange = "any", max = "aead", "prf = sha384"}]"`
```
kmakiela, line 212:
```suggestion
* **Example:** `tls.ciphers = "[{cipher = "aes_25_gcm", key_exchange = "any", mac = "aead", "prf = sha384"}]"`
```

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+* **Syntax:** array of strings, paths in the file system
+* **Default:** not set
+* **Example:** `tls.crl_files = ["certs.crl"]`
+
+Specifies the paths to Certificate Revocation Lists.
+
+#### `listen.c2s.tls.protocol_options` - for `fast_tls`
+* **Syntax:** array of strings
+* **Default:** `["no_sslv2", "no_sslv3", "no_tlsv1", "no_tlsv1_1"]`
+* **Example:** `tls.protocol_options = ["no_tlsv1", "no_tlsv1_1"]`
```
kmakiela, line 233: the `just_tls` and `fast_tls` options are interweaven here with each other - maybe it could be better to group them by library. Right now the `cipher` option makes sense but overall we have a structure like:
```
fast_tls_1
just_tls_1
just_tls_2
just_tls_3
fast_tls_2
```
And I think we could make it a bit more organized

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+  zlib = 4096
+  access = "c2s"
+  shaper = "c2s_shaper"
+  max_stanza_size = 65536
+```
+
+* One at port 5222, which accepts a plain TCP connection and allows to use StartTLS for upgrading it to an encrypted one. The files containing the certificate and the DH parameter are also provided.
+* One at port 5223, which accepts only encrypted TLS connections - this is the legacy method as StartTLS is preferred.
+
+Both listeners use ZLIB and the `c2s` and `c2s_shaper` rules for access management and traffic shaping, respectively.
```
kmakiela, line 263:
```suggestion
Both listeners use ZLIB and the `c2s` and `c2s_shaper` rules for access management and traffic shaping respectively.
```

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+Recommended port number: 5280 for BOSH/WS.
+
+There are following options for each of the HTTP listeners:
+
+#### `listen.http.handlers`
+* **Syntax:** each handler is specified in a subsection starting with `[[listen.http.handlers.type]]` where `type` is one of the allowed handler types, handling different connection types, e.g.
+
+    * `mod_bosh` - for [BOSH](https://xmpp.org/extensions/xep-0124.html) connections,
+    * `mod_websockets` - for [WebSocket](https://tools.ietf.org/html/rfc6455) connections,
+    * `mongoose_api_*`, `mongoose_client_api_*`, ... - for REST API.
```
kmakiela, line 419: If we don't list all the handlers here I'd like to see a ref added to be able to check them quickly

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+### Handler types: Metrics API (obsolete) - `mongoose_api`
+
+REST API for accessing the internal MongooseIM metrics.
+Please refer to the [REST interface to metrics](../rest-api/Metrics-backend.md) page for more information.
+The following option is required:
+
+#### `listen.http.handlers.mongoose_api.modules`
+* **Syntax:** array of strings - Erlang modules
+* **Default:** not set, this is a mandatory option for this handler
+* **Example:** `modules = [mongoose_api_metrics]`
```
kmakiela, line 537: If I understand correctly, this option is called `handlers` and not `modules` in our parser. Also, it should be:
```suggestion
* **Example:** `modules = ["mongoose_api_metrics"]`
```

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+* **Syntax:** string, path in the file system
+* **Default:** not set
+* **Example:** `tls.dhfile = "dh.pem"`
+
+Path to the Diffie Hellman parameter file.
+
+#### `listen.http.tls.ciphers`
+* **Syntax:** array of tables with the following keys: `cipher`, `key_exchange`, `mac`, `prf` and string values.
+* **Default:** not set, all supported cipher suites are accepted
+* **Example:** `tls.ciphers = "[{cipher = "aes_25_gcm", key_exchange = "any", max = "aead", "prf = sha384"}]"`
```
kmakiela, line 607:
```suggestion
* **Example:** `tls.ciphers = "[{cipher = "aes_25_gcm", key_exchange = "any", mac = "aead", "prf = sha384"}]"`
```

kmakiela (commented):
Thanks for the PR, looks good, left some comments in specicic sections

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+#### `listen.*.port`
+* **Syntax:** integer, port number
+* **Default:** no default, this option is mandatory.
+* **Example:** `port = 5222`
+
+The port number to which the listening socket is bound.
+
+#### `listen.*.ip_address`
+* **Syntax:** string with the IP address
+* **Default:** any IP address is allowed by default (0.0.0.0 for IPv4).
```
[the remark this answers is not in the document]
chrzaszcz: It's not that simple: 
```get_ip_tuple(no_ip_option, inet) ->
    {0, 0, 0, 0};
get_ip_tuple(no_ip_option, inet6) ->
    {0, 0, 0, 0, 0, 0, 0, 0};```

So the default depends on `ip_version`

File: doc/advanced-configuration/listen.md [hunk 9]
```diff
@@ -0,0 +1,738 @@
...
+#### `listen.*.proto`
+* **Syntax:** string, `"udp"` or `"tcp"`
+* **Default:** `"tcp"`
+* **Example:** `proto = "udp"`
+
+The protocol, which is TCP by default. There is no reason to change this for XMPP or HTTP listeners.
+
+#### `listen.*.ip_version`
+* **Syntax:** integer, `4` or `6`
+* **Default:** `4` unless the `ip_address` is an IPv6 one - then it is `6`
```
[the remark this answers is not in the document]
chrzaszcz: Not that simple again. It's `4` if `ip_address` is an IPv4 one, `6` if `ip_address` is an IPv6 one or `4` if the address is unspecified.

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+
+#### `listen.*.max_stanza_size`
+* **Syntax:** positive integer
+* **Default:** not set, unlimited size
+* **Example:** `max_stanza_size = 10_000`
+
+Maximum allowed incoming stanza size in bytes.
+**Warning:** this limit is checked **after** the input data parsing, so it does not apply to the input data size itself.
+
+#### `listen.*.num_acceptors`
```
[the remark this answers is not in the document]
chrzaszcz: Yes and this remark is kind of meaningless as all of them are TCP-based at the moment.

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+Handles XMPP client-to-server (C2S) connections.
+The recommended port number for a C2S listener is 5222.
+The following options are supported for each C2S listener:
+
+#### `listen.c2s.access`
+* **Syntax:** string, access rule name
+* **Default:** `"all"`
+* **Example:** `access = "c2s"`
+
+The access rule that determines who is allowed to connect. By default the rule is `all`, what means that anyone can connect. The access rule referenced here needs to be defined in the `access` configuration section.
```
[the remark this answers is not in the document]
chrzaszcz: Right, we can grep the section names later, I wanted to avoid `TODO`'s in the docs

doc/advanced-configuration/listen.md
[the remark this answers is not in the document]
kmakiela: In this case I'd state that the default is dependant on `ip_version` and is either `0.0.0.0` or `0.0.0.0.0.0.0.0`

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+Enforces verification of a client certificate. Requires a valid `cacertfile`.
+
+#### `listen.c2s.tls.module`
+* **Syntax:** string, one of `"just_tls"`, `"fast_tls"`
+* **Default:** `"fast_tls"`
+* **Example:** `tls.module = "just_tls"`
+
+By default the TLS library used for C2S connections is `fast_tls`, which uses OpenSSL-based NIFs. It is possible to change it to `just_tls` - Erlang TLS implementation provided by OTP. Some TLS-related options described here have different formats for these two libraries.
+
+#### `listen.c2s.tls.verify_mode` - for `just_tls`
```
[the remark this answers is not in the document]
chrzaszcz: I can add the `only` here, but I am not sure if it adds any value... if on type is listed explicitly, it makes sense that it does not work with the others.

doc/advanced-configuration/listen.md
chrzaszcz: I am not sure if this is the place... the old docs did not say so and I wanted to avoid redundancy and repeating the same over and over again.

doc/advanced-configuration/listen.md
chrzaszcz: Honestly, I juggled them over and over again, trying to find the best order. I could reorder them once more... I just wanted to follow the order in example docs, the parser, the Erlang docs etc...

doc/advanced-configuration/listen.md
chrzaszcz: The point is that the user does not need to know them all, not supposed to change them one by one. That's the point when I wanted to avoid going too deeply into the low level hardcoded modules.

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+An example would be [`mod_disco`](../modules/mod_disco.md), which may be configured to filter out hidden components from disco results, so they won't be discoverable by clients.
+A reason to do so could be reduced traffic - systems with many components could return very long disco responses.
+Also, some deployments would like to avoid revealing some services; not because it is a security threat (this method does not prevent clients from communicating with hidden components), but rather because they are not meant to interact with clients directly (e.g. helper components for other components).
+
+#### `listen.service.conflict_behaviour`
+* **Syntax:** string, one of: `"disconnect"`, `"kick_old"`
+* **Default:** `"disconnect"`
+* **Example:** `conflict_behaviour = "kick_old"`
+
+By default, when a component tries to connect and a registration conflict occures, the connection is dropped with the following error:
```
NelsonVides, line 372:
```suggestion
By default, when a component tries to connect and a registration conflict occurs, the connection is dropped with the following error:
```

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+```
+
+It makes implementing the reconnection logic difficult, because the old connection would not allow any other connections.
+By setting this option to `kick_old`, we drop any old connections registered at the same host before accepting new ones.
+
+### Custom extension to the protocol
+
+In order to register a component for all virtual hosts served by the server (see `hosts` in the `general` section), the component must add the attribute `is_subdomain="true"` to the opening stream element.
+This maybe helpful if someone wants to have a single instance of a component serving multiple virtual hosts.
+The `is_subdomain` attribute is optional and the default behaviour is as described in the XEP.
```
NelsonVides, line 388: Which XEP is that?

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+  password = "secret"
+```
+
+## HTTP-based services: `[[listen.http]]`
+
+Manages all HTTP-based services, such as BOSH (HTTP long-polling), WebSocket and REST.
+It uses the [Cowboy](https://ninenines.eu/docs/en/cowboy/2.6/manual) web server.
+Recommended port number: 5280 for BOSH/WS.
+
+There are following options for each of the HTTP listeners:
```
NelsonVides, line 412:
```suggestion
There are the following options for each of the HTTP listeners:
```

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+
+#### `listen.http.handlers`
+* **Syntax:** each handler is specified in a subsection starting with `[[listen.http.handlers.type]]` where `type` is one of the allowed handler types, handling different connection types, e.g.
+
+    * `mod_bosh` - for [BOSH](https://xmpp.org/extensions/xep-0124.html) connections,
+    * `mod_websockets` - for [WebSocket](https://tools.ietf.org/html/rfc6455) connections,
+    * `mongoose_api_*`, `mongoose_client_api_*`, ... - for REST API.
+
+    These types are described below in more detail.
+    The double-bracket syntax is used because there can be multiple handlers of a given type, so technically for each type there is a TOML array of tables (subsections).
```
NelsonVides, line 422: As before

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+Overrides the default TCP backlog value.
+
+#### `listen.*.proxy_protocol`
+* **Syntax:** boolean
+* **Default:** `false`
+* **Example:** `proxy_protocol = true`
+
+When set to `true`, [Proxy Protocol](https://www.haproxy.com/blog/haproxy/proxy-protocol/) is enabled and each connecting client has to provide a proxy header. Use only with a proxy (or a load balancer) to allow it to provide the connection details (including the source IP address) of the original client. Versions 1 and 2 of the protocol are supported.
+
+This option specifies the types of incoming connections (e.g. XMPP, HTTP), for which MongooseIM is listening on the specified port numbers. By default there are no listeners defined, so MongooseIM would not accept any connection.
```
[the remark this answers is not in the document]
chrzaszcz: Good catch!

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,738 @@
...
+  shaper = "s2s_shaper"
+  max_stanza_size = 131072
+  tls.dhfile = "dh_server.pem"
+```
+
+##  XMPP Components: `[[listen.service]]`
+
+Interface for external services acting as XMPP components ([XEP-0114: Jabber Component Protocol](http://xmpp.org/extensions/xep-0114.html)), enabling communication between MongooseIM and external services over the XMPP network. The recommended port number for a component listener is 8888.
+
+According to ([XEP-0114: Jabber Component Protocol](http://xmpp.org/extensions/xep-0114.html)) component's hostname should be given in the <stream:stream> element.
```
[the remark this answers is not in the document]
chrzaszcz: ...and I already eliminated a lot of such mistakes from the copy+pasted fragments ;-)

doc/advanced-configuration/listen.md
```diff
@@ -0,0 +1,736 @@
...
+* **Syntax:** positive integer or the string `"infinity"`
+* **Default:** `1024`
+* **Example:** `transport.max_connections = "infinity"`
+
+Maximum number of open connections. The default value of 1024 is set by the [Ranch](https://ninenines.eu/docs/en/ranch/1.7/guide/) library.
+
+### TLS (HTTPS) options
+
+By default the HTTP listener does not use TLS.
+To use TLS (HTTPS), you need to add a TOML table (subsection) called `tls` to the config file with the `certfile` and `keyfile` options that specify the location of the certificate and private key files, respectively.
```
kmakiela, line 558:
```suggestion
To use TLS (HTTPS), you need to add a TOML table (subsection) called `tls` to the config file with the `certfile` and `keyfile` options that specify the location of the certificate and private key files respectively.
```

kmakiela (approved):

PR #2873, opened 2020-09-17
Fix a typo "num_acceptros"

balgillo:
Fix a small typo that would stop the `num_acceptors` configuration outside `transport_options` from being applied.


balgillo:
@NelsonVides could you take a quick look at this small PR?

NelsonVides (approved):
That's excellent. This would actually fix this long-forgotten issue #2770. I was holding it off because we've already fixed so in a different feature branch we're working on at the moment. But I just checked for future git conflicts, can easily merge this beforehand 👍

PR #2875, opened 2020-09-21
Validate modules using matching

arcusfelis:
This PR addresses matching.

Proposed changes include:
* matching
* fixed invalid welcome_message format in mod_register

Validation - review progress
- [ ] services
- [x] mod_adhoc
- [ ] mod_mam
- [ ] mod_amp
- [ ] mod_muc
- [ ] mod_auth_token
- [ ] mod_muc_commands
- [ ] mod_blocking
- [ ] mod_muc_light
- [ ] mod_bosh
- [ ] mod_muc_light_commands
- [ ] mod_caps
- [ ] mod_muc_log
- [ ] mod_carboncopy
- [ ] mod_offline
- [ ] mod_commands
- [ ] mod_offline_stub
- [ ] mod_csi
- [ ] mod_ping
- [ ] mod_disco
- [ ] mod_privacy
- [ ] mod_event_pusher
- [ ] mod_private
- [ ] mod_event_pusher_http
- [ ] mod_pubsub
- [ ] mod_event_pusher_push
- [ ] mod_push_service_mongoossh
- [ ] mod_event_pusher_rabbit
- [ ] mod_register
- [ ] mod_event_pusher_sns
- [ ] mod_revproxy
- [ ] mod_global_distrib
- [ ] mod_roster
- [ ] mod_shared_roster_ldap
- [ ] mod_http_upload
- [ ] mod_sic
- [ ] mod_inbox
- [ ] mod_stream_management
- [ ] mod_jingle_sip
- [ ] mod_time
- [ ] mod_keystore
- [ ] mod_vcard
- [ ] mod_last
- [ ] mod_version


src/config/mongoose_config_parser_toml.erl
```diff
@@ -620,7 +621,8 @@ service_opt([<<"tracking_id">>, <<"service_mongoose_system_metrics">>|_],  V) ->
 %% path: (host_config[].)modules.*
 -spec process_module(path(), toml_section()) -> [option()].
 process_module([Mod|_] = Path, Opts) ->
-    [{b2a(Mod), parse_section(Path, Opts)}].
+    %% Sort option keys to ensure options could be matched in tests
+    [{b2a(Mod), lists:sort(parse_section(Path, Opts))}].
```
chrzaszcz, line 625: I had the same issue in tests now, ended up reusing the equivalence test code:

```
eq_host_config(Result, Config) ->
    [F] = parse(Config), % check for all hosts
    compare_config(Result, F(?HOST)),
    compare_config(Result, parse_host_config(Config)). % Check for a single host

compare_config(C1, C2) ->
    compare_unordered_lists(C1, C2, fun handle_config_option/2).
```

The problem with the version in the reviewed code is that we bend the code to the tests, not otherwise.
If there was another reason for the sorting, I would do it like this - consistently:
But... I am still not convinced about it.

```
parse_section(Path, M) ->
    lists:flatmap(fun({K, V}) ->
                          Key = key(K, Path, V),
                          handle([Key|Path], V)
                  end, lists:sort(maps:to_list(M))).
```

arcusfelis: We do not preserve original key order in TOML.
The options are already sorted for 95% of cases, it's how maps work, apparently (the only case that is not sorted is the riak key for modules).
We could increase usability in tests by ensuring that they are always sorted.
It also makes analyzing cases with large number of keys easier (i.e. mod_muc module).

NelsonVides (commented):
Just some comments on things to remember for documentation.
Also, have a look at the branch https://github.com/esl/MongooseIM/tree/review_validate_modules where I put some code using a previous `int_or_infinity` pattern, could be good to reuse that code.

test/config_parser_SUITE_data/modules.cfg
```diff
@@ -19,7 +19,7 @@
   {mod_csi, [{buffer_max, 40}]},
   {mod_disco, [
       {iqdisc, one_queue},
-      {extra_domains, ["some_domain", "another_domain"]},
+      {extra_domains, [<<"some_domain">>, <<"another_domain">>]},
```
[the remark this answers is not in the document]
arcusfelis: it's note for docs

doc/modules/mod_muc.md
```diff
@@ -17,9 +17,9 @@ Also `mod_muc_log` is a logging submodule.
 * `history_size` (non-negative integer, default: 20): Room message history to be kept in RAM.
  After node restart, the history is lost.
 * `room_shaper` (atom, default: `none`): Limits per-room data throughput with traffic shaper.
-* `max_room_id` (atom or positive integer, default: `infinite`): Maximum room username length (in JID).
-* `max_room_name` (atom or positive integer, default: `infinite`): Maximum room name length.
-* `max_room_desc` (atom or positive integer, default: `infinite`): Maximum room description length.
+* `max_room_id` (atom or positive integer, default: `infinity`): Maximum room username length (in JID).
+* `max_room_name` (atom or positive integer, default: `infinity`): Maximum room name length.
+* `max_room_desc` (atom or positive integer, default: `infinity`): Maximum room description length.
```
chrzaszcz, line 22: The problem is that such docs will be replaced with TOML and the new doc will just say "use the `cfg` one form the previous version". But there it was called `infinite`. This makes it necessary to add the migration info for all such changes and this was the reason I tried to minimize them even though it is tempting to just rush with them.

I don't think this change is bad (I also considered it), but it needs to be documented.

src/config/mongoose_config_parser_toml.erl
chrzaszcz: I just think that the results are proplists and should be tested as such... I wanted to prevent anyone from doing things like `[{opt1, Val1}, {opt2, Val2}] = parse_section(...)` :-)

arcusfelis: Yeah, I am treating it as orddicts, but only for modules.
The reason is that I am trying to use simplest macros for testing, if a module does not have any tested options.

I could revert the code, but we probably would loose location information in the test errors (which would make fixing tests harder).

I haven't found a consistent way for testing though. I would like to have:
- HOF support - i.e. wrap subtests and execute all subtests in each test. If something fails - report ALL errors, not just first one. Similar to eunit underscore macros. Useful for big tests.
- keep location information, so we can jump to code.
- keep the original code snippet in tests.

Also, I haven't any time to think about it and refactor though.



Though xdd:

```erlang
#{opt1 := Val1 := opt2, Val2} = maps:from_list(parse_section(...))
```

doc/modules/mod_muc.md
arcusfelis: I've checked with code. `infinite` and `infinity` are both fine from mod_muc point of view. It just needs to be an atom to pass as an infinite case.
So, both would work for an old config format (unless we are changing code).

It could be a good time to uniform that now.

arcusfelis: We could add to the migration guide information that `infinite` should not be used though.

chrzaszcz: Ok it's fine then.

src/config/mongoose_config_parser_toml.erl
chrzaszcz: Ok, if getting the helper at the top of the stacktrace in the command-line CT report is the problem, you can refactor it according the `parse_section` example above - i would be ok for me.
I still prefer the topmost version though - I am using the HTTP test reports which are not lacking any information - actually the extra frame from the stracktrace is even slightly more informative for me... 

So: your choice, please just do it consistently - not only for modules.

src/config/mongoose_config_parser_toml.erl
```diff
@@ -1522,9 +1597,27 @@ parse_list(Path, L) ->
...
+
+validate(Path, Option) ->
+    try
+        mongoose_config_validator_toml:validate(Path, Option)
+    catch Class:Error:Stacktrace
+         when not is_map(Error); not is_map_key(path, Error) ->
+              erlang:raise(Class, #{what => toml_validate_failed,
+                                    path => Path,
+                                    reason => Error}, Stacktrace)
+    end.
```
chrzaszcz, line 1620: IMO this is against the "don't care about errors yet" statement in the JIRA story.

However, I would not vote against a change just because it's out of scope.

There are following problems here:
1. Code complexity - it increase significantly
2. Nesting and repetition - I would rather see both try...catch clauses at least at the same level in code so that one does not have to fall under the other one.
3. Overly generous catching: `_:_:_` is kind of a last resort. Here it should always be the `error` class and we could start from there - not to accidentally catch the `exit` class or the `throw` class.

Anyway, this **will** be rewritten soon, there is a story for user-friendly errors which needs to change this exact piece of code.

Bottom line: at least fix points 2 and 3 - however I do not see the necessity of the whole change here to be honest, even after seeing the examples provided by @NelsonVides  :-)

src/config/mongoose_config_validator_toml.erl
```diff
@@ -429,8 +1551,18 @@ validate_non_negative_integer(Value) when is_integer(Value), Value >= 0 -> ok.
 validate_non_negative_integer_or_infinity(Value) when is_integer(Value), Value >= 0 -> ok;
 validate_non_negative_integer_or_infinity(infinity) -> ok.
 
+validate_positive_integer_or_infinity(Value) when is_integer(Value), Value > 0 -> ok;
+validate_positive_integer_or_infinity(infinity) -> ok.
+
 validate_enum(Value, Values) ->
-    true = lists:member(Value, Values).
+    case lists:member(Value, Values) of
```
chrzaszcz, line 1558: I'm seeing these error wrappers popping at some places - is that random (as a test fails and you see a message like 'badmatch', you change this) or is there a motivation for doing it here and not for the other validator functions? I think we should keep it consistent. Any motivation might be good, could you just share it?

src/config/mongoose_config_validator_toml.erl
```diff
@@ -452,4 +1584,225 @@ validate_pool_scope(Value) -> validate_enum(Value, [host, global]).
...
+                    validation_results => Results})
+    end.
+
+safe_call_validator(F, Value) ->
+    try
+        F(Value),
+        ok
+    catch Class:Reason:Stacktrace ->
+              #{class => Class, reason => Reason, stacktrace => Stacktrace}
+    end.
```
chrzaszcz, line 1699: This catch is overly generous.

test/config_parser_SUITE.erl
```diff
@@ -1395,3 +2983,116 @@ test_equivalence_between_files(Config, File1, File2) ->
...
+            [ct:pal("Info: ~p~nResult: ~p~n", [Info, Res]) || {Res, Info} <- Failed],
+            ct:fail(#{what => run_multi_failed, failed_cases => length(Failed)})
+    end.
+
+run_case(F) ->
+    try
+        F(), ok
+    catch Class:Reason:Stacktrace ->
+        {Class, Reason, Stacktrace}
+    end.
```
chrzaszcz, line 3066: Quite a bit of complexity here :-(
I guess I would not ask for any rewrite here (change the granularity, use the actual CT test cases with, when needed, CTH), that would be cruel, let's keep it.

chrzaszcz (commented):
Thanks for the huge bit of code and the gazillion of tests.
I left some comments related to the code style - I revieved mostly the design and architectural aspect of the code. I would like to see the answers to my comments and, in some pieces, improvements would be welcome. I did not ask for any substantial changes because of the size of the changeset.

I strongly recommend smaller PR's in the future.

test/config_parser_SUITE.erl
```diff
@@ -12,6 +12,14 @@
...
 -define(HOST, <<"myhost">>).
 
+-define(add_loc(X), {X, #{line => ?LINE}}).
+
+-define(eqf(Expected, Actual), eq_host_config(Expected, Actual)).
+-define(_eqf(Expected, Actual), ?add_loc(fun() -> ?eqf(Expected, Actual) end)).
+-define(errf(Config), 
+        begin ?err(parse_with_host(Config)), ?err(parse_host_config(Config)) end).
+-define(_errf(Config), ?add_loc(fun() -> ?errf(Config) end)).
+
```
chrzaszcz, line 22: I think this requires some explanation and from my experience it is too easy to confuse `eqf` with `_eqf` and have a test that does nothing (seen even in MongooseIM code). For me it does not bring much value, but this might be my subjective point of view.

src/config/mongoose_config_parser_toml.erl
arcusfelis: We could reuse code by introducing step function, but it would be in stacktraces:

```erlang
%%% config_parser_SUITE ==> services.service_admin_extra: FAILED
%%% config_parser_SUITE ==> {#{path => [<<"submods">>,<<"service_admin_extra">>,<<"services">>],
   reason => function_clause,what => toml_parse_failed},
 [{lists,flatmap,[#Fun<mongoose_config_parser_toml.22.109831236>,1],[]},
  {mongoose_config_parser_toml,service_opt,2,
                               [{file,"/Users/mikhailuvarov/erlang/esl/MongooseIM/src/config/mongoose_config_parser_toml.erl"},
                                {line,608}]},
  {mongoose_config_parser_toml,try_step,4,
                               [{file,"/Users/mikhailuvarov/erlang/esl/MongooseIM/src/config/mongoose_config_parser_toml.erl"},
                                {line,1607}]},
  {mongoose_config_parser_toml,handle,2,
                               [{file,"/Users/mikhailuvarov/erlang/esl/MongooseIM/src/config/mongoose_config_parser_toml.erl"},
                                {line,1600}]},
  {lists,flatmap,2,[{file,"lists.erl"},{line,1250}]},
  {mongoose_config_parser_toml,process_service,2,
                               [{file,"/Users/mikhailuvarov/erlang/esl/MongooseIM/src/config/mongoose_config_parser_toml.erl"},
                                {line,603}]},
  {mongoose_config_parser_toml,try_step,4,
                               [{file,"/Users/mikhailuvarov/erlang/esl/MongooseIM/src/config/mongoose_config_parser_toml.erl"},
                                {line,1607}]},
  {mongoose_config_parser_toml,handle,2,
                               [{file,"/Users/mikhailuvarov/erlang/esl/MongooseIM/src/config/mongoose_config_parser_toml.erl"},
                                {line,1600}]}]}
```

src/config/mongoose_config_validator_toml.erl
arcusfelis: It's the only (ok, there is validate_backend) function that takes an extra argument, which is not visible in the error.

test/config_parser_SUITE.erl
```diff
@@ -15,9 +15,12 @@
 -define(add_loc(X), {X, #{line => ?LINE}}).
 
 -define(eqf(Expected, Actual), eq_host_config(Expected, Actual)).
--define(_eqf(Expected, Actual), ?add_loc(fun() -> ?eqf(Expected, Actual) end)).
 -define(errf(Config), 
         begin ?err(parse_with_host(Config)), ?err(parse_host_config(Config)) end).
+
+%% Constructs HOF to pass into run_multi/1 function
```
NelsonVides, line 21: What is a "HOF"? 🙄

arcusfelis: higher order function

src/config/mongoose_config_validator_toml.erl
chrzaszcz: That's a reason, so let's keep it :-)

chrzaszcz (approved):
I think it's good enough in terms of code quality and consistency now.

test/config_parser_SUITE.erl
```diff
@@ -1195,7 +1244,1557 @@ s2s_max_retry_delay(_Config) ->
...
+mod_caps(_Config) ->
+    run_multi(mod_caps_cases()).
+
+mod_caps_cases() ->
+    T = fun(K, V) -> #{<<"modules">> => #{<<"mod_caps">> => #{K => V}}} end,
+    M = fun(K, V) -> modopts(mod_caps, [{K, V}]) end,
+    [?_eqf(M(cache_size, 10), T(<<"cache_size">>, 10)),
+     ?_eqf(M(cache_life_time, 10), T(<<"cache_life_time">>, 10)),
+     ?_errf(T(<<"cache_size">>, -1)),
+     ?_errf(T(<<"cache_size">>, <<"infinity">>)),
```
leszke, line 1313: I think it shouldn't be erroneous case, as per this definition:
```
validate([<<"cache_size">>, <<"mod_caps">>, <<"modules">>|_], [{cache_size, V}]) ->
  validate_non_negative_integer_or_infinity(V);
```

src/config/mongoose_config_validator_toml.erl
```diff
@@ -397,16 +397,1133 @@ validate([<<"max_retry_delay">>, <<"s2s">>|Path],
...
+    validate_boolean(V);
+validate([<<"last_item_cache">>, <<"mod_pubsub">>, <<"modules">>|_],
+         [{last_item_cache, V}]) ->
+    validate_enum(V, [mnesia,rdbms,false]);
+validate([<<"max_items_node">>, <<"mod_pubsub">>, <<"modules">>|_],
+         [{max_items_node, V}]) ->
+    validate_non_negative_integer(V);
+validate([<<"max_subscriptions_node">>, <<"mod_pubsub">>, <<"modules">>|_],
+         [{max_subscriptions_node, V}]) ->
+    validate_non_negative_integer(V);
```
leszke, line 1337: According to `mod_pubsub.md` we don't cover default value here.
```
default: `undefined` - no limitation
```

leszke: The same will apply to `mod_websockets.ping_rate` field and its default `none` value.

src/config/mongoose_config_validator_toml.erl
```diff
@@ -452,4 +1584,225 @@ validate_pool_scope(Value) -> validate_enum(Value, [host, global]).
...
+    validate_non_empty_binary(Value),
+    validate_backend(nodetree, b2a(Value)).
+
+validate_pubsub_plugin(Value) ->
+    validate_non_empty_binary(Value),
+    validate_backend(node, b2a(Value)).
+
+validate_pubsub_pep_mapping({Namespace, Id}) ->
+    validate_non_empty_string(Namespace),
+    validate_non_empty_string(Id).
```
leszke, line 1767: Maybe I am nitpicking now, but wouldn't `Node` variable name be more convenient, as it would match corresponding key in struct? See `config_parser_SUITE.erl:2298,2390,2391`, as direct links to the lines of files that "are not rendered by default" doesn't make github come up with their rendering 🙄

src/config/mongoose_config_validator_toml.erl
arcusfelis: oh, nice catch, there is no validator for ping_rate.

Though it would really good to refactor undefined to infinity someday.

I think there is really easy way to set option to undefined in TOML - don't specify it :trollface:

test/config_parser_SUITE.erl
arcusfelis: good catch, validator should be `validate_non_negative_integer(V)`

leszke: 👌 corrected in docs I'm currently working on

src/config/mongoose_config_validator_toml.erl
```diff
@@ -397,16 +397,1133 @@ validate([<<"max_retry_delay">>, <<"s2s">>|Path],
...
+    validate_boolean(V);
+validate([<<"subject">>, <<"default_room">>, <<"mod_muc">>, <<"modules">>|_],
+         [{subject, V}]) ->
+    validate_string(V);
+validate([<<"subject_author">>, <<"default_room">>, <<"mod_muc">>, <<"modules">>|_],
+         [{subject_author, V}]) ->
+    validate_string(V);
+validate([<<"title">>, <<"default_room">>, <<"mod_muc">>, <<"modules">>|_],
+         [{title, V}]) ->
+    validate_string(V);
```
gustawlippa, line 1447: I think the option `hibernate_timeout` is missing in validation?
```suggestion
    validate_string(V);
validate([<<"hibernate_timeout">>, <<"mod_muc">>, <<"modules">> | _],
         [{hibernate_timeout, V}]) ->
    validate_non_negative_integer_or_infinity(V);
```
It is described as an option for `mod_muc`.

NelsonVides (approved):

PR #2872, opened 2020-09-16
Revert http2

NelsonVides:
This PR reverts https://github.com/esl/MongooseIM/pull/2716/, with all the pain of my heart after how hard it was to work on that one. It came apparent that this HTTP2 worker introduced errors and lots of ugly logging messages, and ever since we've had CI more and more unstable. So it is probably not worth introducing this risk within the upcoming release.

The PR doesn't strictly removes all that was introduced in the aforementioned PR, but rather only that which introduced the gun worker, there were superficial changes like code-style or typos in docs that are good to keep and doesn't introduce any functional changes. Such things are given in a separate commit.

We'll get back to HTTP2 in the future, surely 😄 

PS: Note that the progress in TOML might need to the updated to parse-validate-document the options as they're given to fusco and not to gun. Changes should be minimal, mainly as referred in this change: https://github.com/esl/MongooseIM/pull/2867/commits/dc1eba223bdced4f34559fea30d5eb7a54fa5e9e

PS2: small tests just passed 10k times in my laptop. `big_tests/tests/muc_http_api_SUITE.erl`, which seemed to have failed many times in CCI for mysql or riak, just passed 1k times in my laptop.

PR #2878, opened 2020-09-22
Change http_opts config from gun to fusco

NelsonVides:
Follow-up on https://github.com/esl/MongooseIM/pull/2872/ in order to remove configuration related to gun, as we've reverted back to fusco. Custom fusco opts are not allowed.


kmakiela:
Does that mean that we never pass any options to `fusco`?

NelsonVides:
> Does that mean that we never pass any options to `fusco`?

Precisely. Documentation said simply to check fusco's docs if we want to pass something here, and we weren't testing for it, so it's probably better to hide away library options, it's an implementation detail.

chrzaszcz (approved):
Nice :-)

kmakiela (approved):

PR #2882, opened 2020-09-25
Replace ejabberd:server() with jid:server(), also lserver, [l]user and [l]resource

balgillo:
I spotted a few places where types that used to be in `ejabberd` but are now only in `jid` were still being referred to as members of the `ejabberd` module.  This means dialyzer is not checking type safety for these functions.  I've done a find/replace on the code.

Once resolved, dialyzer still found no type safety issues.


NelsonVides (approved):
Amazing, I'm only surprised dialyzer never complained about this before 🤔 
Thanks for the contribution! 😄

PR #2886, opened 2020-09-28
TOML docs: host_config

chrzaszcz:
Document the `host_config` section.

Fix the 'scope' definition in the docs as well.

### DIFF
--- a/doc/Advanced-configuration.md
+++ b/doc/Advanced-configuration.md
@@ -1,12 +1,10 @@
 For advanced configuration use the following files:
 
-* `mongooseim.cfg` for pure MongooseIM settings,
+* `mongooseim.toml` for pure MongooseIM settings,
 
 * `vm.args` to affect the Erlang VM behaviour (performance tuning, node name),
 
 * `app.config` to change low-level logging parameters and settings of other Erlang applications.
-
-Since you've gotten this far, we assume you're already familiar with Erlang syntax.
 
 # mongooseim.toml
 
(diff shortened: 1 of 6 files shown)


doc/advanced-configuration/host_config.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+## `host_config.modules`
+
+This section completely overrides the top-level [`modules`](modules.md) section. All options are allowed.
+
+#### Example
+
+The modules enabled for `domain2.com` will be `mod_disco` and `mod_stream_management`.
+If we wanted to enable `mod_roster`, it would need to be repeated in `host_config`.
```
NelsonVides, line 85: Hmmm... that's interesting. Wouldn't we expect enabling mod_roster globally to be inherited by all the hosts, unless explicitly stated otherwise? Not sure which approach is better, so I'm looking for some thoughts on this 🤔

NelsonVides (commented):
I just have one question in the comments below. Also, the ordering of the inherited ACL rules was confusing to read if reading fast, but I think that's inherent to the complexity of the acl rules themselves, it's perfectly well explained as it is now, thanks 😄

doc/Advanced-configuration.md
```diff
@@ -1,13 +1,11 @@
 For advanced configuration use the following files:
 
-* `mongooseim.cfg` for pure MongooseIM settings,
+* `mongooseim.toml` for pure MongooseIM settings,
 
 * `vm.args` to affect the Erlang VM behaviour (performance tuning, node name),
 
 * `app.config` to change low-level logging parameters and settings of other Erlang applications.
 
-Since you've gotten this far, we assume you're already familiar with Erlang syntax.
```
kmakiela, line 9: [*]

kmakiela (approved):
Thanks for the detailed examples, they make it much easier to understand

doc/advanced-configuration/host_config.md
chrzaszcz: That's, how it works now. One tuple `{modules, ...}` = all or nothing, no way to inherit module by module.

NelsonVides (approved):
Thanks for the rewritten commit message, it is more helpful now. As I told you, this makes the code unfortunately a bit harder to understand, so we should keep in mind figuring out clearer ways to do it in the future. For now this PR is perfect 👍

PR #2903, opened 2020-10-07
TOML: user-friendly errors

chrzaszcz:
Make the TOML errors more readable by the user.

When loading the configuration fails:
- Log all the TOML-related errors, make sure the 'what' and 'text' attributes in the log message contain the proper description of the error
- Exit with a concise error message

For more details, see the individual commit messages.

**Note:** the 'tomerl' library is quite limited:
- it is not possible to print out the faulty lines in the config
- some error messages are unhandled in 'tomerl:format_error' 
- neither this nor any other Erlang library allows formatting/printing TOML

We might consider forking and extending this library.

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -13,7 +13,8 @@
            {format_packet_filter, {fun mongoose_log_filter:format_packet_filter/2, no_state}},
            {format_acc_filter, {fun mongoose_log_filter:format_acc_filter/2, no_state}},
            {format_c2s_state_filter, {fun mongoose_log_filter:format_c2s_state_filter/2, no_state}},
-           {format_stacktrace_filter, {fun mongoose_log_filter:format_stacktrace_filter/2, no_state}}
+           {format_stacktrace_filter, {fun mongoose_log_filter:format_stacktrace_filter/2, no_state}},
+           {format_term_filter, {fun mongoose_log_filter:format_term_filter/2, [toml_value]}}
         ]},
 
     %% Shell log handler:
--- a/src/config/mongoose_config_parser_toml.erl
+++ b/src/config/mongoose_config_parser_toml.erl
@@ -6,7 +6,8 @@
 -export([parse_file/1]).
 
 -ifdef(TEST).
--export([parse/1]).
+-export([parse/1,
+         extract_errors/1]).
 -endif.
 
 -include("mongoose.hrl").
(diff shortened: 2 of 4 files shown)


mongoose-im:
[8624.1](https://travis-ci.org/esl/MongooseIM/jobs/733656249) / Erlang 23.0.3 / small_tests / 6eb61f71a99778e804cefd6ee0e287166d2f1a34
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2903/8624/small_tests.23.0.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2903/8624/small_tests.23.0.3/small/ct_run.mongooseim@localhost.2020-10-07_13.19.32/lib.mongooseim.logs/run.2020-10-07_13.19.33/suite.log.html)

rel/files/app.config
```diff
@@ -13,7 +13,8 @@
            {format_packet_filter, {fun mongoose_log_filter:format_packet_filter/2, no_state}},
            {format_acc_filter, {fun mongoose_log_filter:format_acc_filter/2, no_state}},
            {format_c2s_state_filter, {fun mongoose_log_filter:format_c2s_state_filter/2, no_state}},
-           {format_stacktrace_filter, {fun mongoose_log_filter:format_stacktrace_filter/2, no_state}}
+           {format_stacktrace_filter, {fun mongoose_log_filter:format_stacktrace_filter/2, no_state}},
+           {format_toml_filter, {fun mongoose_log_filter:format_toml_filter/2, no_state}}
```
arcusfelis, line 17: Good.
An alternative could be, because it could be not the first field with that issue:

```
{format_term_filter, {fun mongoose_log_filter:format_term_filter/2, [toml_value]}}
```

arcusfelis (approved):
good to see some progress here

rel/files/app.config
chrzaszcz: Why not, I will do that, I am only concerned about `app.config` becoming the new `mongooseim.cfg` with all the lower level options. This might be the way to go though as long as the settings here are not expected to be changed by non-Erlang users like devops.

mongoose-im:
[8626.1](https://travis-ci.org/esl/MongooseIM/jobs/733893499) / Erlang 23.0.3 / small_tests / 4667ab0f65aad3f1bc88cdb2dac25449ef23f1ae
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/2903/8626/small_tests.23.0.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/2903/8626/small_tests.23.0.3/small/ct_run.mongooseim@localhost.2020-10-08_07.21.03/lib.mongooseim.logs/run.2020-10-08_07.21.03/suite.log.html)

chrzaszcz:
@arcusfelis could you re-review the minor fixes I made?

rel/files/app.config
arcusfelis: Yep, I was thinking about asking **MongooseIM** to create handlers after the start and keep app.config minimal.
And after reading logger docs carefully, it looks like one of use cases OTP team had in mind.


But if we want to configure logging, a very good place would be to put options into mongooseim.toml, but it would limit logging usage until configuration parser has finished its work.

Alternative I see is to put `[logging]` block into toml and parse it first at start to get loglevel and filenames for logs, init logging subsystem dynamically using https://erlang.org/doc/man/logger.html#add_handler-3 and continue parsing all other options (which are more likely to contain user-made mistakes, that would need reporting).

While I like that approach, it could be hard to do with the monolithic parser as we have now, so it would be interesting to discuss :)

PR #2919, opened 2020-10-23
Remove the 'cfg' config format

chrzaszcz:
This PR removes the 'cfg' configuration format from code and tests.

Key points:
- Test suites were updated to either use TOML or to directly check the expected parsed config options.
- Tests for 'mod_revproxy' and 'mod_cowboy' checked functionality that is NOT supported in TOML. As much as possible was left intact here, removing only the reload tests that required parsing the 'cfg' files. However, we need to resolve this quickly and either drop (most likely) or add this in TOML (less likely). This will be addressed in a new story.

See individual commits for more detailed description of each step.


arcusfelis (approved):
looks good

PR #2931, opened 2020-11-06
TOML declarative parsing: 'general' section

chrzaszcz:
First section parsed with the declarative specification

See commit messages for more details.

Key points:
- type specs will be added when the spec format stops changing
- new fields like 'mandatory_keys' might be added with the next sections
- the support for imperative and declarative formats is temporary - when everything is declarative, the code will be simplified
- `export_all` is also temporary


arcusfelis (approved):
looks good

janciesla8818 (approved):
Thanks for the PR. Good work with the initial work for the declarative approach, looks really promising. 

Just one question, as we changed the `general.mongooseim_access_commands.commands` so it no longer accepts the string `all` could we update the docs accordingly or create another task to do so? I just want to make sure we don't forget about this.

janciesla8818:
💪 thanks for the additional changes.

PR #2943, opened 2020-11-17
TOML declarative config: 'auth' section

chrzaszcz:
This PR adds the declarative spec for the `auth` config section.

See the commit messages for more information.


janciesla8818 (approved):
Thanks for the changes with the `auth` section, looks really good. The more I review the more I like the declarative approach.

PR #2948, opened 2020-11-18
TOML improved config templates

chrzaszcz:
The main change is to introduce [Mustache sections](http://mustache.github.io/mustache.5.html) to remove whitespaces occuring in the templated config files, including the default `prod` config file.

**Example.** The block below has the following semantics: if `var` is defined and not `false`, render an empty line followed by its contents.

```
  {{#var}}     

  {{{var}}}
  {{/var}}
```

There is one caveat for `bbmustache` - strings are lists and would be treated as Mustache lists, so they need to be binarized in tests (rebar3 does this already when processing the overlays, but tests call bbmustache separately). Alternatively, all template variables could be converted to binaries, but I think it could make them a bit less readable.

### DIFF
--- a/big_tests/run_common_test.erl
+++ b/big_tests/run_common_test.erl
@@ -294,7 +294,7 @@ enable_preset_on_node(Node, PresetVars, HostVarsFilePrefix) ->
     ok.
 
 template_config(Template, Vars) ->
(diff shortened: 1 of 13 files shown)


arcusfelis (approved):
looks good.

PR #2962, opened 2020-12-02
TOML declarative spec for services

chrzaszcz:
New callback for services:
  config_spec/0


janciesla8818 (approved):
Thanks for the changes. All looks good.

PR #2965, opened 2020-12-03
TOML declarative modules (1/4)

chrzaszcz:
Modules covered in this PR:
- [x] mod_adhoc
- [x] mod_auth_token
- [x] mod_bosh
- [x] mod_caps
- [x] mod_carboncopy
- [x] mod_csi

To be continued in the next PR:
- mod_disco
- mod_event_pusher

Checklist for each module:
- Write the new spec in the module as a callback function: `config_spec/0`. Some cases need reworking the options, e.g. when the same value can be a primitive (e.g. string or atom) or a nested section/list
- Add the module to `mongoose_config_spec:all_modules/0` and `HAS_NO_SPEC`
- Double check with the docs: `https://esl.github.io/MongooseDocs` that the options and their format, validator etc are correct
- Remove `module_opt` and `handler` clauses from `mongoose_config_parser_toml` as well as module-specific functions for nested options
- Remove `validate` clauses from `mongoose_config_validator_toml`
- Make sure this passes locally: `./rebar3 ct --suite=config_parser_SUITE` before pushing anything, also check `./rebar3 dialyzer`
- Check test coverage after pushing, there might be some options not covered by tests, you can do it by looking at the test cases as well
- Regarding tests, it's good to simplify them and fix indentation/maximum line length of 100 character, see the change for `mod_adhoc` or `mod_auth_token`.
- Sanity check: grep for the module name in `src/config`

### DIFF
--- a/src/mod_bosh.erl
+++ b/src/mod_bosh.erl
@@ -21,7 +21,8 @@
 
 %% gen_mod callbacks
 -export([start/2,
-         stop/1]).
+         stop/1,
+         config_spec/0]).
 
 %% cowboy_loop_handler callbacks
 -export([init/2,
@@ -40,6 +41,7 @@
 -include("jlib.hrl").
 -include_lib("exml/include/exml_stream.hrl").
 -include("mod_bosh.hrl").
+-include("ejabberd_config.hrl").
 
 -define(DEFAULT_MAX_AGE, 1728000).  %% 20 days in seconds
 -define(DEFAULT_INACTIVITY, 30).  %% seconds
@@ -146,6 +148,21 @@ start(_Host, Opts) ->
 
 stop(_Host) ->
     ok.
+
+-spec config_spec() -> mongoose_config_spec:config_section().
+config_spec() ->
+    #section{
+       items = #{<<"inactivity">> => #option{type = int_or_infinity,
+                                             validate = non_negative},
+                 <<"max_wait">> => #option{type = int_or_infinity,
+                                           validate = non_negative},
+                 <<"server_acks">> => #option{type = boolean},
+                 <<"max_pause">> => #option{type = integer,
+                                            validate = positive,
+                                            format = {kv, maxpause}}
+                }
+      }.
+
 %%--------------------------------------------------------------------
 %% Hooks handlers
 %%--------------------------------------------------------------------
--- a/test/config_parser_SUITE.erl
+++ b/test/config_parser_SUITE.erl
@@ -1394,14 +1394,11 @@ mod_carboncopy(_Config) ->
     check_iqdisc(mod_carboncopy).
 
 mod_csi(_Config) ->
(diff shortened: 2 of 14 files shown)


janciesla8818 (approved):
Thanks for the changes. The code looks good 💪

test/config_parser_SUITE.erl
```diff
@@ -1337,82 +1336,69 @@ s2s_max_retry_delay(_Config) ->
...
-
-mod_caps_cases() ->
     T = fun(K, V) -> #{<<"modules">> => #{<<"mod_caps">> => #{K => V}}} end,
     M = fun(K, V) -> modopts(mod_caps, [{K, V}]) end,
-    [?_eqf(M(cache_size, 10), T(<<"cache_size">>, 10)),
-     ?_eqf(M(cache_life_time, 10), T(<<"cache_life_time">>, 10)),
-     ?_errf(T(<<"cache_size">>, -1)),
-     ?_errf(T(<<"cache_size">>, <<"infinity">>)),
-     ?_errf(T(<<"cache_life_time">>, -1)),
-     ?_errf(T(<<"cache_life_time">>, <<"cache_life_time">>))].
```
leszke, line 1402: Just one (or two) question from my side:
- you changed `cache_life_time` to `infinity` because the previous one wouldn't make its way to this place in code due to preceding validators, so it hadn't any sense here, correct?
- you changed - in my impression still - correctly erroneous value of `-1` to `0` cause zero is now the closest one value to the boundary value defined per `positive integer`, as `-1` was to `non-negative integer`, correct? I mean `-1` would still fail these tests as well and I'm trying to find the reasoning behind this

leszke (approved):

src/mod_bosh.erl
```diff
@@ -146,6 +148,21 @@ start(_Host, Opts) ->
...
+config_spec() ->
+    #section{
+       items = #{<<"inactivity">> => #option{type = int_or_infinity,
+                                             validate = non_negative},
+                 <<"max_wait">> => #option{type = int_or_infinity,
+                                           validate = non_negative},
+                 <<"server_acks">> => #option{type = boolean},
+                 <<"max_pause">> => #option{type = integer,
+                                            validate = positive,
+                                            format = {kv, maxpause}}
```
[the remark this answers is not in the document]
chrzaszcz: I think we could just grep for `{kv, ...}` and fix all of them.

test/config_parser_SUITE.erl
chrzaszcz: The reason was indeed to bring the values closer to the boundary, e.g. 0 for the `positive` validator.
We could test both (-1 and 0), but I was trying not to have too much verbosity in tests. There are no strict rules here as long as one successful and one failing case is tested for each option. It's good to have some variety as well, e.g. not to test the same value for all options to have more chance of catching bugs.

PR #2966, opened 2020-12-03
gh-pages deployment with material theme

NelsonVides:
So this PR does several things. The point is to migrate to gh-pages, as it allows us to use any theme we want, which enables us to move forward with the idea of improving the look&feel of the mongooseim documentation.

This documentation is generated using [`mike`](https://github.com/jimporter/mike), which, using [mkdocs](https://www.mkdocs.org/) and [material](https://squidfunk.github.io/mkdocs-material/), enables us to build several versions, tagging them and pushing them.

How mkdocs generate these sites? Compiling all the markdown files and the mkdocs.yml into a huge bunch of static html and css files. mike injects some minor javascript then to allow to change versions. These lots of autogenerated files are then pushed to a github repo that github allows to serve as static pages. But the amount of autogenerated stuff is too big to keep in our main repo, so these things are pushed to a new repo we've made, https://github.com/esl/MongooseDocs.

More things that this PR does: migrating all links to readthedocs to our new pages, improving some http-to-https links (my firefox complained about that), improving some minor indentation and whitespace in doc files, and, importantly, adding building docs to the CircleCI pipeline. All branches will be built, but only tags and master will then be pushed.


leszke (approved):
Good effort of moving our shiny new docs to gh-pages! Now we all can be proud of its new look ❤️

PR #2969, opened 2020-12-04
Ensure envvar is in the environment for the next job

NelsonVides:
This follows exactly what was done for mongoose-push, hopefully should work here as well.


leszke (approved):
🔥

PR #2970, opened 2020-12-04
TOML declarative modules 1/4  - continued

chrzaszcz:
Modules covered in this PR:
- [x] mod_disco
- [x] mod_event_pusher

Checklist for each module:
- Write the new spec in the module as a callback function: `config_spec/0`. Some cases need reworking the options, e.g. when the same value can be a primitive (e.g. string or atom) or a nested section/list
- Add the module to `mongoose_config_spec:all_modules/0` and `HAS_NO_SPEC`
- Double check with the docs: `https://esl.github.io/MongooseDocs` that the options and their format, validator etc are correct
- Remove `module_opt` and `handler` clauses from `mongoose_config_parser_toml` as well as module-specific functions for nested options
- Remove `validate` clauses from `mongoose_config_validator_toml`
- Make sure this passes locally: `./rebar3 ct --suite=config_parser_SUITE` before pushing anything, also check `./rebar3 dialyzer`
- Check test coverage after pushing, there might be some options not covered by tests, you can do it by looking at the test cases as well
- Regarding tests, it's good to simplify them and fix indentation/maximum line length of 100 character, see the change for `mod_adhoc` or `mod_auth_token`.
- Sanity check: grep for the module name in `src/config`


janciesla8818 (approved):
Thanks for the changes and the additional work of paying the technical debt from the recent transition. Much appreciated!

PR #2977, opened 2020-12-10
Add prepared queries for the remaining parts of MAM - user & prefs

arcusfelis:
This PR addresses "Add prepared queries for the remaining parts of MAM"

Proposed changes include:
* mod_mam_rdbms_user
* mod_mam_rdbms_prefs

### DIFF
--- a/src/mam/mod_mam_rdbms_prefs.erl
+++ b/src/mam/mod_mam_rdbms_prefs.erl
@@ -192,12 +192,14 @@ set_prefs(_Result, Host, UserID, _ArcJID, DefaultMode, AlwaysJIDs, NeverJIDs) ->
 
 set_prefs1(Host, UserID, DefaultMode, AlwaysJIDs, NeverJIDs) ->
     Rows = prefs_to_rows(UserID, DefaultMode, AlwaysJIDs, NeverJIDs),
-    run_transaction_or_retry_on_abort(Host, fun() ->
+    %% MySQL sometimes aborts transaction with reason:
+    %% "Deadlock found when trying to get lock; try restarting transaction"
+    mongoose_rdbms:transaction_with_delayed_retry(Host, fun() ->
             {updated, _} =
                 mongoose_rdbms:execute(Host, mam_prefs_delete, [UserID]),
             [mongoose_rdbms:execute(Host, mam_prefs_insert, Row) || Row <- Rows],
             ok
-        end, UserID, 5),
+        end, #{user_id => UserID, retries => 5, delay => 100}),
     ok.
 
 -spec get_prefs(mod_mam:preference(), _Host :: jid:server(),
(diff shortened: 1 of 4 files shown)


arcusfelis:
https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/2977/40312/elasticsearch_and_cassandra_mnesia.23.0.3-1/big/ct_run.test@default-2b0f9a12-81b8-46de-9bd3-9e4b40a77f88.2020-12-14_13.29.07/big_tests.tests.mam_SUITE.logs/run.2020-12-14_13.34.14/mam_suite.init_per_group.5426.html


Badarg inserting new prepared queries.

- Why do we enable rdbms module on ES and cassandra? Maybe the ETS table does not exist.
- Do we properly handle a case when a module is restarted? (two statements could be inserted in parallel too).

src/mam/mod_mam_rdbms_prefs.erl
```diff
@@ -160,157 +190,85 @@ set_prefs(_Result, Host, UserID, _ArcJID, DefaultMode, AlwaysJIDs, NeverJIDs) ->
...
+    Result = mongoose_rdbms:sql_transaction(Host, F),
+    case Result of
+        {atomic, _} ->
+            Result;
+        {aborted, Reason} when Retries > 0 ->
+            ?LOG_WARNING(#{what => mam_transaction_aborted,
+                           text => <<"Transaction aborted. Restart">>,
+                           user_id => UserID, reason => Reason, retries => Retries}),
+            timer:sleep(100),
+            run_transaction_or_retry_on_abort(Host, F, UserID, Retries-1);
```
chrzaszcz, line 248: Minor, style: add spaces to `Retries - 1`

src/mam/mod_mam_rdbms_prefs.erl
```diff
@@ -160,157 +190,85 @@ set_prefs(_Result, Host, UserID, _ArcJID, DefaultMode, AlwaysJIDs, NeverJIDs) ->
...
 %% Helpers
 
--spec encode_behaviour(always | never | roster) -> string().
-encode_behaviour(roster) -> "R";
-encode_behaviour(always) -> "A";
-encode_behaviour(never)  -> "N".
+%% Possible error with mysql
+%% Reason "Deadlock found when trying to get lock; try restarting transaction"
+%% triggered mod_mam_utils:error_on_sql_error
+run_transaction_or_retry_on_abort(Host, F, UserID, Retries) ->
```
chrzaszcz, line 238: 1. Why is this a problem only for `prefs`? It seems like it should be generic and occur somewhere else as well - and should be handled in `mongoose_rdbms`.
2. Does this deadlock happen during regular tests? If so, I think the problem would escalate during load tests? Maybe our transaction logic/isolation level is broken? I would investigate this as maybe it's just broken tests, e.g. setting prefs in parallel for the same user.
3. How did you test it?

chrzaszcz (commented):
A few questions from me...

src/mam/mod_mam_rdbms_prefs.erl
arcusfelis: 1. yep, there is generic version in mongoose_rdbms, but it does not has delay (because it's inside a worker, we don't want to wait in the DB worker).
2. regular. Waaaay too often during tests, rarely IRL.
3. Big tests sometimes fail without it (and actually it prints sometimes this warning in log).

arcusfelis: we probably can use just a builtin

chrzaszcz (approved):
Great!

PR #2985, opened 2020-12-18
Prepared queries for mod_roster

arcusfelis:
This PR addresses "prepared queries for mod_roster".

### DIFF
--- a/priv/mysql.sql
+++ b/priv/mysql.sql
@@ -58,7 +58,7 @@ CREATE INDEX i_last_seconds ON last(seconds);
 
 CREATE TABLE rosterusers (
     username varchar(250) NOT NULL,
-    jid varchar(250) NOT NULL,
+    jid varchar(250) NOT NULL, -- must be a parsable jid
     nick text NOT NULL,
     subscription character(1) NOT NULL,
     ask character(1) NOT NULL,
--- a/src/mod_roster_rdbms.erl
+++ b/src/mod_roster_rdbms.erl
@@ -56,9 +56,6 @@ prepare_queries(Host) ->
            " FROM rosterusers WHERE username=? AND jid=?">>),
     mongoose_rdbms:prepare(roster_group_get, rostergroups, [username],
         <<"SELECT jid, grp FROM rostergroups WHERE username=?">>),
-    mongoose_rdbms:prepare(roster_sub_get_by_jid, rosterusers, [username, jid],
-        <<"SELECT subscription FROM rosterusers "
-          "WHERE username=? AND jid=?">>),
     mongoose_rdbms:prepare(roster_group_get_by_jid, rostergroups, [username, jid],
         <<"SELECT grp FROM rostergroups "
           "WHERE username=? AND jid=?">>),
@@ -155,21 +152,19 @@ decode_roster_entry_rows(LUser, LServer, LJID, [Row]) ->
     case Rec of
         %% Bad JID in database:
         error ->
-            ?LOG_ERROR(#{what => roster_parse_failed, row => io_lib:format("~0p", [Row])}),
+            ?LOG_ERROR(#{what => roster_parse_failed, row => format_term(Row)}),
             #roster{usj = USJ, us = US, jid = LJID};
         _ ->
             Rec#roster{usj = USJ, us = US, jid = LJID}
     end.
 
+%% full means we should query for groups too
 get_roster_entry(LUser, LServer, LJID, full) ->
     case get_roster_entry(LUser, LServer, LJID) of
         does_not_exist -> does_not_exist;
-        Rentry ->
-            case read_subscription_and_groups(LUser, LServer, LJID) of
-                error -> error;
-                {Subscription, Groups} ->
-                    Rentry#roster{subscription = Subscription, groups = Groups}
-            end
+        Rec ->
+            Groups = get_groups_by_jid(LUser, LServer, LJID),
+            Rec#roster{groups = Groups}
     end.
 
 get_roster_entry_t(LUser, LServer, LJID, full) ->
@@ -219,36 +214,16 @@ del_roster_t(LUser, LServer, LJID) ->
     mongoose_rdbms:execute(LServer, roster_delete_by_jid, [LUser, BinJID]),
     mongoose_rdbms:execute(LServer, roster_group_delete_by_jid, [LUser, BinJID]).
 
-read_subscription_and_groups(LUser, LServer, LJID) ->
+get_groups_by_jid(LUser, LServer, LJID) ->
     BinJID = jid:to_binary(LJID),
-    SubResult = mongoose_rdbms:execute(LServer, roster_sub_get_by_jid, [LUser, BinJID]),
-    case SubResult of
-        {selected, [{ExtSubscription}]} ->
-            GroupResult = mongoose_rdbms:execute(LServer, roster_group_get_by_jid, [LUser, BinJID]),
-            Subscription = decode_subscription(ExtSubscription),
-            case GroupResult of
-                {selected, GroupRows} ->
-                    Groups = [JGrp || {JGrp} <- GroupRows],
-                    {Subscription, Groups};
-                _ ->
-                    ?LOG_ERROR(#{what => read_subscription_and_groups_failed,
-                                 reason => GroupResult, user => LUser,
-                                 host => LServer}),
-                    error
-            end;
-        E ->
-           ?LOG_ERROR(#{what => read_subscription_and_groups_failed,
-                        reason => E, user => LUser, host => LServer}),
-            error
-    end.
+    {selected, Rows} = mongoose_rdbms:execute_successfully(
+                         LServer, roster_group_get_by_jid, [LUser, BinJID]),
+    [Group || {Group} <- Rows].
 
 %%==============================================================================
 %% Helper functions
 %%==============================================================================
 
-%% PgSQL returns CHAR as an integer
-%% MySQL returns CHAR as a string
-decode_subscription(<<X>>) -> decode_subscription(X);
 decode_subscription($B) -> both;
 decode_subscription($T) -> to;
 decode_subscription($F) -> from;
@@ -259,7 +234,6 @@ encode_subscription(to)   -> <<"T">>;
 encode_subscription(from) -> <<"F">>;
 encode_subscription(none) -> <<"N">>.
 
-decode_ask(<<X>>) -> decode_ask(X);
 decode_ask($S) -> subscribe;
 decode_ask($U) -> unsubscribe;
 decode_ask($B) -> both;
@@ -298,8 +272,8 @@ raw_to_record(LServer,
             error;
         JID ->
             LJID = jid:to_lower(JID),
-            Subscription = decode_subscription(ExtSubscription),
-            Ask = decode_ask(ExtAsk),
+            Subscription = decode_subscription(mongoose_rdbms:character_to_integer(ExtSubscription)),
+            Ask = decode_ask(mongoose_rdbms:character_to_integer(ExtAsk)),
             #roster{usj = {User, LServer, LJID},
                     us = {User, LServer}, jid = LJID, name = Nick,
                     subscription = Subscription, ask = Ask,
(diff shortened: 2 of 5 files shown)


src/mod_roster_rdbms.erl
```diff
@@ -275,40 +246,86 @@ read_subscription_and_groups(LUser, LServer, LJID, GSFunc, GRFunc) ->
...
+    [LUser, BinJID, Nick, ExtSubscription, ExtAsk, AskMessage,
+     <<"N">>, <<>>, <<"item">>].
+
+groups_to_rows(#roster{us = {LUser, _LServer}, jid = JID, groups = Groups}) ->
+    BinJID = jid:to_binary(jid:to_lower(JID)),
+    lists:foldl(fun (<<>>, Acc) -> Acc;
+                    (Group, Acc) -> [[LUser, BinJID, Group] | Acc]
+                end, [], Groups).
+
+raw_to_record(LServer,
```
chrzaszcz, line 292: we have `record_to_row` but `raw_to_record`. This is asking for typos. Please make it consistent.

src/mod_roster_rdbms.erl
```diff
@@ -145,22 +173,13 @@ get_roster_entry(LUser, LServer, LJID, full) ->
...
-        does_not_exist -> does_not_exist;
-        Rentry ->
-        case read_subscription_and_groups_t(LUser, LServer, LJID) of
-            error -> error;
-            {Subscription, Groups} ->
-                Rentry#roster{subscription = Subscription, groups = Groups}
-        end
-    end.
-
+    get_roster_entry(LUser, LServer, LJID, full).
```
chrzaszcz, line 176: My question about the transaction context remains valid here. I am starting to think it deserves an explanation in the commit message - the motivation and reasoning behind removing these `_t` versions.

src/mod_roster_rdbms.erl
```diff
@@ -275,40 +246,86 @@ read_subscription_and_groups(LUser, LServer, LJID, GSFunc, GRFunc) ->
...
+    lists:foldl(fun (<<>>, Acc) -> Acc;
+                    (Group, Acc) -> [[LUser, BinJID, Group] | Acc]
+                end, [], Groups).
+
+raw_to_record(LServer,
+              {User, SJID, Nick, ExtSubscription, ExtAsk, AskMessage,
+               _Server, _ExtSubscribe, _Type} = I) ->
+    case jid:from_binary(SJID) of
+        error ->
+            ?LOG_ERROR(#{what => roster_parse_failed, row => io_lib:format("~0p", [I])}),
```
chrzaszcz, line 297: Did you check manually that the log message is correct?

chrzaszcz, line 297: Did you check manually that the log message is correct?

src/mod_roster_rdbms.erl
```diff
@@ -275,40 +246,86 @@ read_subscription_and_groups(LUser, LServer, LJID, GSFunc, GRFunc) ->
...
-    SJID = mongoose_rdbms:escape_string(jid:to_binary(jid:to_lower(JID))),
-    lists:foldl(fun (<<"">>, Acc) -> Acc;
-                    (Group, Acc) ->
-                        G = mongoose_rdbms:escape_string(Group),
-                        [[Username, SJID, G] | Acc]
-                end,
-                [], Groups).
+%% PgSQL returns CHAR as an integer
+%% MySQL returns CHAR as a string
+decode_subscription(<<X>>) -> decode_subscription(X);
```
chrzaszcz, line 251: Do not accept binaries and integers in the same function. Previous version was much better.

src/mod_roster_rdbms.erl
```diff
@@ -275,40 +246,86 @@ read_subscription_and_groups(LUser, LServer, LJID, GSFunc, GRFunc) ->
...
+decode_subscription($T) -> to;
+decode_subscription($F) -> from;
+decode_subscription($N) -> none.
+
+encode_subscription(both) -> <<"B">>;
+encode_subscription(to)   -> <<"T">>;
+encode_subscription(from) -> <<"F">>;
+encode_subscription(none) -> <<"N">>.
+
+decode_ask(<<X>>) -> decode_ask(X);
```
chrzaszcz, line 262: Do not accept binaries and integers in the same function.

src/mod_roster_rdbms.erl
```diff
@@ -275,40 +246,86 @@ read_subscription_and_groups(LUser, LServer, LJID, GSFunc, GRFunc) ->
...
+        R ->
+            BinJID = jid:to_binary(R#roster.jid),
+            [R#roster{groups = dict_find_list(BinJID, GroupsDict)}]
+    end.
+
+dict_find_list(K, Dict) ->
+    case dict:find(K, Dict) of
+        {ok, Values} -> Values;
+        error -> []
+    end.
```
chrzaszcz, line 331: I get the idea of splitting the function into smaller chunks, but this one looks a bit awkward on its own.
If we are trying to be generic here, I think a `dict_find(Key, Dict, Default)` would be best. Otherwise it's difficult to understand the `_list` suffix as the only thing it does is to set the default to an empty list.

chrzaszcz (commented):
A few comments from me.

src/mod_roster_rdbms.erl
```diff
@@ -36,101 +37,128 @@
...
-            rdbms_queries:set_roster_version(Username, EVer);
-        _ ->
-            rdbms_queries:sql_transaction(LServer,
-                                          fun () ->
-                                                  rdbms_queries:set_roster_version(Username,
-                                                                                   EVer)
-                                          end)
-    end.
+write_roster_version(LUser, LServer, _InTransaction, Ver) ->
+    version_upsert(LServer, LUser, Ver).
```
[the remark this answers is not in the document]
arcusfelis: Actually, the old code uses:

```erlang
update_t(Table, Fields, Vals, Where) ->
    UPairs = lists:zipwith(fun(A, B) -> [A, "=", mongoose_rdbms:use_escaped(B)] end,
                           Fields, Vals),
    case mongoose_rdbms:sql_query_t(
           [<<"update ">>, Table, <<" set ">>,
            join(UPairs, ", "),
            <<" where ">>, Where, ";"]) of
        {updated, 1} ->
            ok;
        _ ->
            mongoose_rdbms:sql_query_t(
              [<<"insert into ">>, Table, "(", join(Fields, ", "),
               <<") values (">>, join_escaped(Vals), ");"])
    end.
 ```
 
 And, mod_vcard rework shows, that execution time of upsert and `update_t` is the same ;) 
 I've microbenchmarked it when I've introduced prepare queries.
 
The new upsert should behave the same way as `update_t` logically, just in one query.

File: src/mod_roster_rdbms.erl [hunk 8]
```diff
@@ -36,101 +37,128 @@
...
-        true ->
-            rdbms_queries:set_roster_version(Username, EVer);
-        _ ->
-            rdbms_queries:sql_transaction(LServer,
-                                          fun () ->
-                                                  rdbms_queries:set_roster_version(Username,
-                                                                                   EVer)
-                                          end)
-    end.
+write_roster_version(LUser, LServer, _InTransaction, Ver) ->
```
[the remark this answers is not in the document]
arcusfelis: `mongoose_rdbms` has logic like:

```
case get(TransactionState) of
       undefined -> ...do query to a random worker, use LServer variable...;
       State -> ...oh, we already in transaction, we would use the current worker and ignore LServer variable...
end
```

So, `sql_query/2` behaves like `sql_query_t/1`, if it is executed inside a transaction.
The reason our current code looks like this, is that long time ago there were two separate functions which had common logic. We refactored it, but never had time to refactor it more :)

src/mod_roster_rdbms.erl
arcusfelis: DB returns binaries or integers though...

priv/mysql.sql
```diff
@@ -58,7 +58,7 @@ CREATE INDEX i_last_seconds ON last(seconds);
 
 CREATE TABLE rosterusers (
     username varchar(250) NOT NULL,
-    jid varchar(250) NOT NULL,
```
chrzaszcz, line 61: What does this comment mean? How does introducing prepared queries affect the schema?

src/rdbms/mongoose_rdbms.erl
```diff
@@ -477,6 +479,12 @@ result_to_integer(Bin) when is_binary(Bin) ->
 selected_to_integer({selected, [{BInt}]}) ->
     result_to_integer(BInt).
 
+%% Converts a value from a CHAR field to integer
+%% PgSQL returns CHAR as an integer
+%% %% MySQL returns CHAR as a string
```
chrzaszcz, line 484: Did you mean "binary"?

src/mod_roster_rdbms.erl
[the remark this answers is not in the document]
chrzaszcz: I would put a comment about it in the commit message - it's not obvious where these changes are coming from.

priv/mysql.sql
arcusfelis: Basically, it's a contract on the data format with DB administrators.
For some reason, mod_roster assumes that there could be an invalid jids in DB, which never happens.
Unless, there is another app modifying the data.

So, it's a constraint that it's not just `NOT NULL`, but the data should be verified before inserting in the DB.
The current behaviour was to just ignore such data. Which is such a corner case actually and resulted in a lot of non-covered code.

src/mod_roster_rdbms.erl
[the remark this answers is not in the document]
arcusfelis: There is actually comment:
```
%% `mongoose_rdbms:execute' automatically detects if we are in a transaction or not.
```
in the code ;)

src/rdbms/mongoose_rdbms.erl
arcusfelis: I mean pgsql silently converts `CHAR(1)` into integer `UNSIGNED TINYINT` :D 
While mysql still returns it as a string data type `CHAR(size)`.

chrzaszcz: What I mean is that the comments mention the types `integer` and `string` while the types accepted below are `binary` and `integer`. I think this comment is misleading as it suggests the Erlang type is `string()`, which is not.

priv/mysql.sql
chrzaszcz: Does introducing prepared queries affect this or not?

src/mod_roster_rdbms.erl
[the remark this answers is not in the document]
chrzaszcz: Ok... it's a bit inconsistent though, e.g. no comment in line 145, what may suggest it's different there.

priv/mysql.sql
arcusfelis: Not strictly. But each time we trying to parse value from `rosterusers.jid`, we have some logic attached to handle case when it's not a valid jid.
But this logic is unttested, uncovered with tests and I have severe questions about it (basically, it silently discards rows with invalid jids in some cases, but fails in another cases).

So, I've decided to remove "error handling logic" and just enforce the rule, that jids should be parsable. 

If you don't like comments in schema, I could remove them. But I think it could be helpful for some developers, who do not understand erlang.

src/mod_roster_rdbms.erl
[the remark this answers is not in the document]
arcusfelis: I am pretty sure, it should be a common knowledge.
Look, these functions are in mongoose_rdbms:

```
-export([prepare/4,
         prepared/1,
         execute/3,
         execute_successfully/3,
         sql_query/2,
         sql_query_t/1,
         sql_transaction/2,
         transaction_with_delayed_retry/3,
         sql_dirty/2,
         to_bool/1,
         db_engine/1,
         print_state/1,
         use_escaped/1]).
```

There is no `execute_t` function :D 
`sql_query_t` is just a helper for cases, when we are too lazy to pass `Host` argument.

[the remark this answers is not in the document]
arcusfelis: added comment to the commit message.

src/rdbms/mongoose_rdbms.erl
arcusfelis: Changed this comment.

src/mod_roster_rdbms.erl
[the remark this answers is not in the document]
chrzaszcz: Ok, maybe I am too picky here, it's ok.

src/mod_roster_rdbms.erl
```diff
@@ -36,279 +37,249 @@
...
+    BinJID = jid:to_binary(jid:to_lower(JID)),
+    lists:foldl(fun (<<>>, Acc) -> Acc;
+                    (Group, Acc) -> [[LUser, BinJID, Group] | Acc]
+                end, [], Groups).
+
+%% We must not leak our external RDBMS format representation into MongooseIM
+raw_to_record(_LServer, Rec) -> Rec.
+
+%% Decode fields from `roster_fields()' into a record
+row_to_record(LServer,
```
chrzaszcz, line 241: I don't want to be picky... but this `raw` vs `row` is quite unexpected.

chrzaszcz (commented):
One minor comment...

chrzaszcz (approved):
Good, thanks for introducing the prepared queries and fixing other issues along the way.

Issue #2879, opened 2020-09-24
MUC Light chat messages not getting when member is offline


antivista:
MongooseIM version: rel-3.7
Installed from: pkg
Erlang/OTP version: R16B03-1

I am having trouble in retrieving messages regarding MUC Light rooms, using MAM, for offline users. 
The users who are online are getting the messages without any problem.
I was expecting that when a member of a room is offline and messages in the room are exchanged, when the offline member comes online, messages in the room exchanged when the member was offline would be send to it (as it works fine with one-to-one messages using mod_offline). 
Nevertheless no messages from the rooms are send to the offline member when it comes online.

Does the offline member coming online have to send necessarily a certain stanza to retrieve the "missed" messages from its rooms?
Or is there any way to have a behaviour for MUC Light messages offline equal to the one-to-one messages offline when mod_offline is used?

My current configuration - mod_muc_light:

{mod_muc_light, [
 {host, "muclight.@HOST@"},
             {backend, rdbms},
             {equal_occupants, false},
             {legacy_mode, false},
             {rooms_per_user, 50},
             {blocking, false},
             {all_can_configure, false},
             {all_can_invite, false},
             {max_occupants, 50},
             {rooms_per_page, 10},
             {rooms_in_rosters, true},
             {config_schema, [{"roomname", "The Room"}]}
 ]},

My current configuration - mod_mam:
{mod_mam_meta, [
             {backend, rdbms},
             {is_archivable_message, mod_mam_utils},
             {archive_chat_markers, false},
             {user_prefs_store, false},
             {full_text_search, false},
             {async_writer, true},
%% Cache information about users (default)
             {cache_users, true},
%% Enable archivization for private messages (default)
             {pm, false},
  {muc, [
               {host, "muclight.@HOST@"},
                {is_archivable_message, mod_mam_utils},
                {rdbms_message_format, simple},
                {async_writer, false},
                {user_prefs_store, false}
  ]},
%% Do not use a <stanza-id/> element (by default stanzaid is used)
             {no_stanzaid_element, true}
]}

Thanks in advance for the help.

DenysGonchar:
You can try to enable `store_groupchat_messages` option for `mod_offline`:
```
{mod_offline, [{store_groupchat_messages, true},
               {backend, rdbms}]}.
```
But MAM is a preferable way, we do not recommend `mod_offline` because it sends out all the messages immediately  w/o any way to control it from the client device. This affects connection time and traffic.

[...]

PR #2992, opened 2021-01-02
[skip ci] Reference about store_groupchat_messages option in mod_offline added

antivista:
This PR adds documentation for mod_offline, regarding the option store_groupchat_messages. [#2879](https://github.com/esl/MongooseIM/issues/2879)

### DIFF
--- a/doc/modules/mod_offline.md
+++ b/doc/modules/mod_offline.md
@@ -1,7 +1,7 @@
 ## Module Description
 
 This module implements an offline messages storage compliant with [XEP-0160: Best Practices for Handling Offline Messages](http://xmpp.org/extensions/xep-0160.html). 
-It stores one-to-one messages only when the recipient has no online resources. 
+It can store one-to-one and groupchat messages only when the recipient has no online resources. 
 It is not well suited for applications supporting multiple user devices, because anything saved in the DB can be retrieved only once, so the message history is not synchronised between devices. 
 Although `mod_offline` may be sufficient in some cases, it is preferable to use [mod_mam](mod_mam.md).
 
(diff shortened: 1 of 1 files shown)


DenysGonchar (changes requested):
Thanks for PR. historically there are several reasons why `store_groupchat_messages` option is not documented:

1. XEP-0160 doesn't specify this functionality,  because the classic MUC even doesn't send the messages to offline users. 
2. As this will not work as expected with the classic MUC, the option makes sense only with MUC-light.
3. XEP-0160 is not recommended way of storing offline messages (especially group-chat messages), better use message archive with [MUC storage enabled](https://esl.github.io/MongooseDocs/latest/modules/mod_mam/#enable-muc-message-archive).

So this option is rather confusing than brings any benefit, it's kept for compatibility reasons.

So let's either keep it not documented or fully document it.

antivista:
Although the use of `mod_mam` is the better approach in most of the cases, client side requires more effort (compared with XEP-0160) to implement XEP-0313, so I have thought that to document this option could be good in these scenarios:
- As a momentary alternative waiting the XEP-0313 implementation client side; 
- If it is expected the use of just one client per user and when a message is delivered (for some reason) we do not want allow the rescue of that message again, to do this with `mod_mam` should require more work because XEP-0313 does not foresee a mechanism of message deletion except for the older messages in the archive.

If these scenarios are considered of interest, I can try to fully document, referring to explicitly to MUC-light case as DenysGonchar rightly pointed out to me.

antivista (commented):
I hope that it is adequately documented now.

NelsonVides (approved):
Fine by me 👍🏽

DenysGonchar (approved):
👍

PR #3002, opened 2021-01-07
validate mod_event_pusher_push.virtual_pubsub_hosts properly

leszke:
In case of `@HOSTS@` wildcard usage, there were validation errors reported. For more details, please see [MIM-1330](https://erlangsolutions.atlassian.net/browse/MIM-1330) description.

### DIFF
--- a/src/config/mongoose_config_validator.erl
+++ b/src/config/mongoose_config_validator.erl
@@ -11,7 +11,8 @@
 -type validator() ::
         any | non_empty | non_negative | positive | module | {module, Prefix :: atom()}
       | jid | domain | domain_template | url | ip_address | ip_mask | network_address | port
-      | filename | dirname | loglevel | pool_name | shaper | access_rule | {enum, list()}.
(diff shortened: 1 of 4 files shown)


src/config/mongoose_config_validator.erl
```diff
@@ -11,7 +11,8 @@
 -type validator() ::
         any | non_empty | non_negative | positive | module | {module, Prefix :: atom()}
       | jid | domain | domain_template | url | ip_address | ip_mask | network_address | port
-      | filename | dirname | loglevel | pool_name | shaper | access_rule | {enum, list()}.
+      | filename | dirname | loglevel | pool_name | shaper | access_rule | {enum, list()}
+      | {domain_template, string()}.
```
chrzaszcz, line 15: Minor: the order here is getting more out of sync with the order of clauses in `validate/3`... but I guess it would require some shuffling to make it the same anyway. I think this is OK as it is.

chrzaszcz (approved):
The code is good, the minor remark does not need any action now.

PR #3007, opened 2021-01-14
Update migration guide with TOML changes

leszke:
Basically I went through the output of `git diff master toml-config doc/` command and wrote it down, already excluding couple of type alignments. I also added missing link to the newest migration guide's chapter.

### DIFF
--- a/doc/migrations/4.0.1_4.x.x.md
+++ b/doc/migrations/4.0.1_4.x.x.md
@@ -6,46 +6,30 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
 All major, modern client libraries and applications support the 0.3.0+ specification.
 If you experience any issues with making requests to the HTTP File Upload service, please update your client.
 
-## Retirement of old `*.cfg` format
+## Retirement of the old `*.cfg` format
 
-Since ____ release, we are no longer supporting `*.cfg` MongooseIM configuration format. For further application's usage, please use `TOML` format.
+Since release 4.x.x, we are no longer supporting the `*.cfg` MongooseIM configuration format. Please use the `TOML` format instead.
 
-## Minor changes in `TOML` config format
+## Minor changes in the `TOML` config format
 
 * [`mod_bosh.max_pause`](../../modules/mod_bosh/#modulesmod_boshmax_pause) instead of `maxpause`
 
-* [`mod_caps.cache_life_time`](../../modules/mod_caps/#modulesmod_capscache_life_time): the value should be provided milliseconds
-
 * [`mod_disco.server_info.module`](../../modules/mod_disco/#modulesmod_discoserver_info): the field is optional, no longer required
-
-* [`mod_register`](../../modules/mod_register/#modulesmod_registeriqdisctype), [`mod_vcard`](../../modules/mod_vcard/#modulesmod_vcardiqdisctype) & [`mod_version`](../../modules/mod_version/#modulesmod_versioniqdisctype): `iqdisc.type` default value was changed from `"no_queue"` to `"one_queue"`
-
-* [`mod_event_pusher_push.virtual_pubsub_hosts`](../../modules/mod_event_pusher_push/#modulesmod_event_pusher_pushvirtual_pubsub_hosts): the default value was changed to `[]`
-
-* [`mod_event_pusher_sns.presence_updates_topic`](../../modules/mod_event_pusher_sns/#modulesmod_event_pusher_snspresence_updates_topic): default not set
-
-* `mod_event_pusher_sns.muc_host`: the option was removed
 
 * [`mod_global_distrib.connections.advertised_endpoints`](../../modules/mod_global_distrib/#modulesmod_global_distribconnectionsadvertised_endpoints): default not set (`false` is no longer accepted)
 
-* `mod_global_distrib.connections.tls.enabled`: flag was removed, TLS is enabled by providing [`cacertfile`](../../modules/mod_global_distrib/#modulesmod_global_distribconnectionstlscacertfile) and [`certfile`](../../modules/mod_global_distrib/#modulesmod_global_distribconnectionstlscertfile) options
+* `mod_global_distrib.connections.tls.enabled`: the flag was removed, TLS is enabled by providing the [`cacertfile`](../../modules/mod_global_distrib/#modulesmod_global_distribconnectionstlscacertfile) and [`certfile`](../../modules/mod_global_distrib/#modulesmod_global_distribconnectionstlscertfile) options
 
-* [`mod_http_upload.s3`](../../modules/mod_http_upload/#s3-backend-options): the section is mandatory
+* [`mod_http_upload.max_file_size`](../../modules/mod_http_upload/#modulesmod_http_uploadmax_file_size): `undefined` is no longer allowed
 
-* [`mod_mam_meta.user_prefs_store`](../../modules/mod_mam/#modulesmod_mam_metauser_prefs_store): `false` value is no longer allowed
+* [`mod_mam_meta.user_prefs_store`](../../modules/mod_mam/#modulesmod_mam_metauser_prefs_store): `false` is no longer allowed
 
-* [`mod_muc_light.config_schema`](../../modules/mod_muc_light/#modulesmod_muc_lightconfig_schema): minor config fields simplification
+* [`mod_muc_light.config_schema`](../../modules/mod_muc_light/#modulesmod_muc_lightconfig_schema): the usage of `value` and `type` fields was replaced with one of the following fields: `string_value`, `integer_value` or `float_value`
 
-* `mod_muc_log`: change of default values for [`css_file`](../../modules/mod_muc_log/#modulesmod_muc_logcss_file) and [`top_link`](../../modules/mod_muc_log/#modulesmod_muc_logtop_link)
-
-* `mod_ping`: [`ping_interval`](../../modules/mod_ping/#modulesmod_pingping_interval) and [`ping_req_timeout`](../../modules/mod_ping/#modulesmod_pingping_req_timeout) explicitly said to be in seconds
-
-* [`mod_push_service_mongoosepush.api_version`](../../modules/mod_push_service_mongoosepush/#modulesmod_push_service_mongoosepushapi_version): the value is now limited to be `"v2"` or `"v3"`
+* [`mod_muc_log.css_file`](../../modules/mod_muc_log/#modulesmod_muc_logcss_file): the default value was changed from `"false"` to `not set`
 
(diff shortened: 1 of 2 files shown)


doc/migrations/4.0.1_4.x.x.md
```diff
@@ -6,3 +6,46 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
 All major, modern client libraries and applications support the 0.3.0+ specification.
 If you experience any issues with making requests to the HTTP File Upload service, please update your client.
 
+## Retirement of old `*.cfg` format
+
+Since ____ release, we are no longer supporting `*.cfg` MongooseIM configuration format. For further application's usage, please use `TOML` format.
```
leszke, line 11: I left it as a placeholder for upcoming TOMLed release version, once we clarify it.

doc/migrations/4.0.1_4.x.x.md
```diff
@@ -6,3 +6,46 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
...
+
+* `mod_event_pusher_sns.muc_host`: the option was removed
+
+* [`mod_global_distrib.connections.advertised_endpoints`](../../modules/mod_global_distrib/#modulesmod_global_distribconnectionsadvertised_endpoints): default not set (`false` is no longer accepted)
+
+* `mod_global_distrib.connections.tls.enabled`: flag was removed, TLS is enabled by providing [`cacertfile`](../../modules/mod_global_distrib/#modulesmod_global_distribconnectionstlscacertfile) and [`certfile`](../../modules/mod_global_distrib/#modulesmod_global_distribconnectionstlscertfile) options
+
+* [`mod_http_upload.s3`](../../modules/mod_http_upload/#s3-backend-options): the section is mandatory
+
+* [`mod_mam_meta.user_prefs_store`](../../modules/mod_mam/#modulesmod_mam_metauser_prefs_store): `false` value is no longer allowed
```
chrzaszcz, line 35: the value 'false' is not longer allowed
alternatively: 'false' is no longer allowed

doc/migrations/4.0.1_4.x.x.md
```diff
@@ -6,3 +6,46 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
...
+
+* [`mod_muc_light.config_schema`](../../modules/mod_muc_light/#modulesmod_muc_lightconfig_schema): minor config fields simplification
```
chrzaszcz, line 37: I would rephrase it somehow... to make it more obvious that the format has changed.

doc/migrations/4.0.1_4.x.x.md
```diff
@@ -6,3 +6,46 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
...
+
+* `mod_muc_log`: change of default values for [`css_file`](../../modules/mod_muc_log/#modulesmod_muc_logcss_file) and [`top_link`](../../modules/mod_muc_log/#modulesmod_muc_logtop_link)
```
chrzaszcz, line 39: The value for `top_link` was just documented incorrectly.

doc/migrations/4.0.1_4.x.x.md
```diff
@@ -6,3 +6,46 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
...
+
+* `mod_ping`: [`ping_interval`](../../modules/mod_ping/#modulesmod_pingping_interval) and [`ping_req_timeout`](../../modules/mod_ping/#modulesmod_pingping_req_timeout) explicitly said to be in seconds
```
chrzaszcz, line 41: I think it might be skipped here as there is no action needed - it was always seconds, so it works the same.

doc/migrations/4.0.1_4.x.x.md
```diff
@@ -6,3 +6,46 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
...
+
+* [`mod_push_service_mongoosepush.api_version`](../../modules/mod_push_service_mongoosepush/#modulesmod_push_service_mongoosepushapi_version): the value is now limited to be `"v2"` or `"v3"`
```
chrzaszcz, line 43: It was like this, now it's just better documented.

doc/migrations/4.0.1_4.x.x.md
```diff
@@ -6,3 +6,46 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
...
+
+* `mod_stream_management`: minor adjustments of [`buffer_max`](../../modules/mod_stream_management/#modulesmod_stream_managementbuffer_max) and [`ack_freq`](../../modules/mod_stream_management/#modulesmod_stream_managementack_freq) options, [`buffer`](../../modules/mod_stream_management/#modulesmod_stream_managementbuffer) and [`ack`](../../modules/mod_stream_management/#modulesmod_stream_managementack) booleans were added
+
+* [`listen.c2s.tls.ciphers`](../../advanced-configuration/listen/#listenc2stlsciphers), [`listen.http.tls.ciphers`](../../advanced-configuration/listen/#listenhttptlsciphers) and [`outgoing_pools.*.*.connection.tls.ciphers`](../../advanced-configuration/outgoing-connections/#outgoing_poolsconnectiontlsciphers): the ciphers should now be formatted as a specification string
+
+* [listen.http.handlers.mod_websockets.ping_rate](../../advanced-configuration/listen/#listenhttphandlersmod_websocketsping_rate): `none` value is no longer allowed
+
+* `outgoing_pools.generic`: the option was removed
```
chrzaszcz, line 51: This pool type was removed as it was an internal setting, not supposed to be set in the configuration file.

doc/migrations/4.0.1_4.x.x.md
```diff
@@ -6,3 +6,46 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
...
+
+* [`mod_push_service_mongoosepush.api_version`](../../modules/mod_push_service_mongoosepush/#modulesmod_push_service_mongoosepushapi_version): the value is now limited to be `"v2"` or `"v3"`
+
+* `mod_stream_management`: minor adjustments of [`buffer_max`](../../modules/mod_stream_management/#modulesmod_stream_managementbuffer_max) and [`ack_freq`](../../modules/mod_stream_management/#modulesmod_stream_managementack_freq) options, [`buffer`](../../modules/mod_stream_management/#modulesmod_stream_managementbuffer) and [`ack`](../../modules/mod_stream_management/#modulesmod_stream_managementack) booleans were added
+
+* [`listen.c2s.tls.ciphers`](../../advanced-configuration/listen/#listenc2stlsciphers), [`listen.http.tls.ciphers`](../../advanced-configuration/listen/#listenhttptlsciphers) and [`outgoing_pools.*.*.connection.tls.ciphers`](../../advanced-configuration/outgoing-connections/#outgoing_poolsconnectiontlsciphers): the ciphers should now be formatted as a specification string
+
+* [listen.http.handlers.mod_websockets.ping_rate](../../advanced-configuration/listen/#listenhttphandlersmod_websocketsping_rate): `none` value is no longer allowed
```
chrzaszcz, line 49: I would say "the value 'none'" or just "'none'", just like you say "the letter A".

chrzaszcz (commented):
I think the migration guide (contrary to the changelog) should only list the changes that require actions from the user. This is why I pointed out all the changes that actually are only doc corrections - I think they shouldn't be listed here as they require no action from the user. If they had a correct setting, it remains correct, so no there is no migration.

doc/migrations/4.0.1_4.x.x.md
```diff
@@ -6,3 +6,46 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
...
+
+Since ____ release, we are no longer supporting `*.cfg` MongooseIM configuration format. For further application's usage, please use `TOML` format.
+
+## Minor changes in `TOML` config format
+
+* [`mod_bosh.max_pause`](../../modules/mod_bosh/#modulesmod_boshmax_pause) instead of `maxpause`
+
+* [`mod_caps.cache_life_time`](../../modules/mod_caps/#modulesmod_capscache_life_time): the value should be provided milliseconds
+
+* [`mod_disco.server_info.module`](../../modules/mod_disco/#modulesmod_discoserver_info): the field is optional, no longer required
```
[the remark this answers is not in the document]
leszke: Where? [Here](https://github.com/esl/MongooseIM/blob/master/doc/modules/mod_disco.md#modulesmod_discoserver_info) it's still required.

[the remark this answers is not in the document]
chrzaszcz: Correct, I looked at the whole `server_info` by mistake.

doc/migrations/4.0.1_4.x.x.md
```diff
@@ -6,3 +6,30 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
 All major, modern client libraries and applications support the 0.3.0+ specification.
 If you experience any issues with making requests to the HTTP File Upload service, please update your client.
 
+## Retirement of the old `*.cfg` format
+
+Since release 4.x.x, we are no longer supporting `*.cfg` MongooseIM configuration format. For further application's usage, please use `TOML` format.
```
chrzaszcz, line 11: something is wrong about this sentence. I would definitely say `the cfg format` and `the TOML format` - with the definite articles. But the `futher application's usage` is also stylistically awkward, I found zero matches in Google so I am most likely correct here. Could you rephrase this, e.g. just "Please use the TOML format instead."?

doc/migrations/4.0.1_4.x.x.md
```diff
@@ -6,3 +6,30 @@ Currently, only the `urn:xmpp:http:upload:0` XMLNS is served.
...
+
+* [`mod_bosh.max_pause`](../../modules/mod_bosh/#modulesmod_boshmax_pause) instead of `maxpause`
+
+* [`mod_disco.server_info.module`](../../modules/mod_disco/#modulesmod_discoserver_info): the field is optional, no longer required
+
+* [`mod_global_distrib.connections.advertised_endpoints`](../../modules/mod_global_distrib/#modulesmod_global_distribconnectionsadvertised_endpoints): default not set (`false` is no longer accepted)
+
+* `mod_global_distrib.connections.tls.enabled`: the flag was removed, TLS is enabled by providing the [`cacertfile`](../../modules/mod_global_distrib/#modulesmod_global_distribconnectionstlscacertfile) and [`certfile`](../../modules/mod_global_distrib/#modulesmod_global_distribconnectionstlscertfile) options
+
+* [`mod_http_upload.max_file_size`](../../modules/mod_http_upload/#modulesmod_http_uploadmax_file_size): `undefined` is no longer allowed
```
chrzaszcz, line 23: Actually it seems that this was already changed (accidentally?) when introducing TOML, but I believe there were no comments about it, so let's keep it.

chrzaszcz (approved):
Looks great!

PR #3018, opened 2021-01-29
Mongoose session

NelsonVides:
Plenty of explanations and profiling and details and whatnot, in the commit messages 😁

### DIFF
--- a/src/ejabberd_sm.erl
+++ b/src/ejabberd_sm.erl
@@ -130,11 +130,6 @@
                      }.
 -type info() :: #{info_key() => any()}.
 
-%% Session representation as 4-tuple.
--type ses_tuple() :: {USR :: jid:simple_jid(),
-                      Sid :: ejabberd_sm:sid(),
-                      Prio :: priority(),
-                      Info :: info()}.
 -type backend() :: ejabberd_sm_mnesia | ejabberd_sm_redis.
 -type close_reason() :: resumed | normal | replaced.
 -type info_key() :: atom().
@@ -143,7 +138,6 @@
 -export_type([session/0,
               sid/0,
               priority/0,
-              ses_tuple/0,
               backend/0,
               close_reason/0,
               info/0
--- a/src/ejabberd_sm.erl
+++ b/src/ejabberd_sm.erl
@@ -266,7 +266,7 @@ close_session(Acc, SID, JID, Reason) ->
 store_info(JID, {Key, Value} = KV) ->
     case get_session(JID) of
         offline -> {error, offline};
-        {_SUser, SID, SPriority, SInfo} ->
+        #session{sid = SID, priority = SPriority, info = SInfo} ->
             case SID of
                 {_, Pid} when self() =:= Pid ->
                     %% It's safe to allow process update its own record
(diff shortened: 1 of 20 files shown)


mongoose-im:
[9049.1](https://travis-ci.org/esl/MongooseIM/jobs/756765213) / Erlang 23.0.3 / small_tests / 8c2297fbfe8de6703d5af97a9c6bf94d8ff44e93
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/3018/9049/small_tests.23.0.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/3018/9049/small_tests.23.0.3/small/ct_run.mongooseim@localhost.2021-01-29_22.53.53/lib.mongooseim.logs/run.2021-01-29_22.53.54/suite.log.html)

src/mod_carboncopy.erl
```diff
@@ -300,29 +300,33 @@ complete_packet(_From, #xmlel{name = <<"message">>, attrs=OrigAttrs} = Packet, r
...
+              end
+      end,
+      AllSessions).
+
+filter_priority_resources(AllSessions) ->
+    lists:filtermap(
+      fun(#session{usr = {_, _, R}, priority = P}) when is_integer(P) -> {true, {P, R}};
+         (_) -> false
+      end,
+     AllSessions).
```
chrzaszcz, line 326: Minor, not a must: here I'd rather have the inner `fun` as a separate function and call it like `lists:filtermap(fun filter_cc_enabled_resource/1, AllSessions)`. Reason: to have function names for tracing and debugging as the funs are not trivial.

src/ejabberd_sm.erl
```diff
@@ -128,7 +128,7 @@
                       priority :: priority(),
                       info     :: info()
                      }.
--type info() :: [info_item()].
+-type info() :: #{info_key() => any()}.
 
 %% Session representation as 4-tuple.
 -type ses_tuple() :: {USR :: jid:simple_jid(),
```
chrzaszcz, line 134: It looks like we are leaving an unused type behind, could you validate that nothing else is missed? I only checked this one.

chrzaszcz (commented):
Good changes! A few comments from me.

src/ejabberd_sm_redis.erl
```diff
@@ -34,32 +34,24 @@ start(_Opts) ->
...
     Keys = mongoose_redis:cmd(["KEYS", hash(<<"*">>)]),
     lists:flatmap(fun(K) ->
                           Sessions = mongoose_redis:cmd(["SMEMBERS", K]),
                           lists:map(fun(S) ->
-                                            Session = binary_to_term(S),
-                                            { Session#session.usr,
-                                              Session#session.sid,
-                                              Session#session.priority,
-                                              Session#session.info }
+                                            binary_to_term(S)
```
chrzaszcz, line 43: I wonder how performant this is (mem, CPU) compared to tuples.

NelsonVides: `binary_to_term` returns a C array of size six with six pointers to six created-on-the-stop ERL_TERMs.
Then the next instruction allocates a new C array of size four, and copies four specific fields of those pointers from the previous array. It shouldn't need to copy the contents of the pointers, as they're immutable. So technically speaking all that is happening is an extra 4*64bit memory allocation, with the CPU it takes to allocate that and later on GC both arrays. Difference is probably so so minimal it really doesn't matter, but it's still totally unnecessary 🤷🏽‍♂️ Also, records are clearer than arbitrary tuples 😐

src/mod_carboncopy.erl
NelsonVides: Sure thing, good argument 👍🏽

src/ejabberd_sm_redis.erl
chrzaszcz: Never mind, I thought for some reason that this PR changed the data stored in Redis, which is not the case.

chrzaszcz (approved):
Very good changes!

PR #3019, opened 2021-02-01
Prepared queries for mod_last

arcusfelis:
This PR addresses prepared queries for mod_last.
Proposed changes include:
* Prepared queries

### DIFF
--- a/src/mod_last_rdbms.erl
+++ b/src/mod_last_rdbms.erl
@@ -25,52 +25,66 @@
          remove_user/2]).
 
 -spec init(jid:server(), list()) -> ok.
-init(_Host, _Opts) ->
+init(Host, _Opts) ->
+    prepare_queries(Host),
     ok.
 
+%% Prepared query functions
+prepare_queries(Host) ->
+    mongoose_rdbms:prepare(last_select, last, [username],
+                           <<"SELECT seconds, state FROM last WHERE username=?">>),
+    mongoose_rdbms:prepare(last_count_active, last, [seconds],
+                           <<"SELECT COUNT(*) FROM last WHERE seconds > ?">>),
+    mongoose_rdbms:prepare(last_delete, last, [username],
+                           <<"delete from last where username=?">>),
+    rdbms_queries:prepare_upsert(Host, last_upsert, last,
+                                 [<<"username">>, <<"seconds">>, <<"state">>],
+                                 [<<"seconds">>, <<"state">>],
+                                 [<<"username">>]).
+
+execute_get_last(LServer, LUser) ->
+    mongoose_rdbms:execute_successfully(LServer, last_select, [LUser]).
+
+execute_count_active_users(LServer, Seconds) ->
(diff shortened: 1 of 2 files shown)


src/mod_last_rdbms.erl
```diff
@@ -25,52 +25,66 @@
...
 -spec get_last(jid:luser(), jid:lserver()) ->
-    {ok, non_neg_integer(), binary()} | {error, term()} | not_found.
+    {ok, non_neg_integer(), binary()} | not_found.
 get_last(LUser, LServer) ->
-    Username = mongoose_rdbms:escape_string(LUser),
-    case catch rdbms_queries:get_last(LServer, Username) of
-        {selected, []} ->
-            not_found;
-        {selected, [{STimeStamp, Status}]} ->
-            case catch mongoose_rdbms:result_to_integer(STimeStamp) of
```
chrzaszcz, line 39: Could you comment on why we are not returning the errors anymore? For me it looks like this code was too defensive and tried to protect against faulty data, which should not be present in the DB. Is my assumption correct?

src/mod_last_rdbms.erl
```diff
@@ -25,52 +25,66 @@
...
 
 -spec count_active_users(jid:lserver(), non_neg_integer()) -> non_neg_integer().
-count_active_users(LServer, TimeStamp) ->
-    TimeStampBin = integer_to_binary(TimeStamp),
-    WhereClause = <<"where seconds > ", TimeStampBin/binary >>,
-    case rdbms_queries:count_records_where(LServer, <<"last">>, WhereClause) of
-        {selected, [{Count}]} ->
-            mongoose_rdbms:result_to_integer(Count);
-        _ ->
-            0
```
chrzaszcz, line 56: This case was removed. Did you make sure it does not happen during normal operation? Is this again the code being overly defensive?

chrzaszcz (commented):
Looks good! Could you just answer my questions?

src/mod_last_rdbms.erl
arcusfelis: I prefer to use `execute_successfully` everywhere. It would report to logs and throw an error.
So, less code here (no need to match on `aborted` and `error`).

Using 0 here is just an attempt to handle the error by faking data.
I am trying to remove this kinda code from backend modules code.

I think the caller (i.e. the main module) should handle the error explicitly (or just fail, it's an option too).

src/mod_last_rdbms.erl
arcusfelis: `mongoose_rdbms:result_to_integer(STimeStamp)` would never fail, we don't allow NULLs for this field.

```sql
CREATE TABLE last (
    username varchar(250) PRIMARY KEY,
    seconds int NOT NULL,
    state text NOT NULL
) CHARACTER SET utf8mb4
```

We use `execute_successfully`, so we need to handle only two cases: `{selected, []}` and `{selected, [{Timestamp, Status}]`, which `decode_last_result` function does.

chrzaszcz (approved):

PR #3022, opened 2021-02-04
Use prepared queries for mod_offline

arcusfelis:
This PR addresses prepared queries for mod_offline.

Proposed changes include:
* Prepared
* Queries


arcusfelis:
Would create another PR

PR #3027, opened 2021-02-05
Prepared queries for mod_offline - version 2

arcusfelis:
This PR is mod_offline prepared queries on top of mu-prepared-queries-2

Proposed changes include:

* rebased
* queries

### DIFF
--- a/src/offline/mod_offline_rdbms.erl
+++ b/src/offline/mod_offline_rdbms.erl
@@ -34,6 +34,8 @@
          remove_old_messages/2,
          remove_user/2]).
 
+-import(mongoose_rdbms, [prepare/4, execute_successfully/3]).
+
 -include("mongoose.hrl").
 -include("jlib.hrl").
 -include("mod_offline.hrl").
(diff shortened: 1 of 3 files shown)


src/offline/mod_offline_rdbms.erl
```diff
@@ -34,22 +34,91 @@
...
+execute_fetch_offline_messages(LServer, LUser, ExtTimeStamp) ->
+    execute_successfully(LServer, offline_select, [LServer, LUser, ExtTimeStamp]).
+
+execute_remove_expired_offline_messages(LServer, ExtTimeStamp) ->
+    execute_successfully(LServer, offline_delete_expired, [ExtTimeStamp]).
+
+execute_remove_old_offline_messages(LServer, ExtTimeStamp) ->
+    execute_successfully(LServer, offline_delete_old, [ExtTimeStamp]).
+
+execute_remove_user(LServer, LUser) ->
```
chrzaszcz, line 96: Could it be renamed? It does not remove the user, but their offline messages.

src/offline/mod_offline_rdbms.erl
```diff
@@ -58,123 +127,84 @@ pop_messages(#jid{} = To) ->
...
-    #offline_msg{us = US,
-             timestamp = TimeStamp,
-             expire = never,
-             from = From,
-             to = To,
-             packet = Packet,
-             permanent_fields = PermanentFields}.
-
-extract_permanent_fields(null) ->
-    []; %% This is needed in transition period when upgrading to MongooseIM above 3.5.0
```
chrzaszcz, line 91: Why remove this clause? It's missing in line 206. If so, I guess we'll need to add it to the migration guide. How not to forget to do it?

src/offline/mod_offline_rdbms.erl
```diff
@@ -58,123 +127,84 @@ pop_messages(#jid{} = To) ->
...
+%% Pure helper functions
+record_to_row(LUser, LServer,
+              #offline_msg{from = From, packet = Packet, timestamp = TimeStamp,
+                           expire = Expire, permanent_fields = PermanentFields}) ->
+    ExtFrom = jid:to_binary(From),
+    ExtTimeStamp = encode_timestamp(TimeStamp),
+    ExtExpire = maybe_encode_timestamp(Expire),
+    ExtPacket = exml:to_binary(Packet),
+    ExtFields = encode_permanent_fields(PermanentFields),
+    prepare_offline_message(LUser, LServer, ExtFrom, ExtTimeStamp,ExtExpire,
```
chrzaszcz, line 180: Minor: missing comma

src/offline/mod_offline_rdbms.erl
```diff
@@ -34,22 +34,91 @@
...
+    prepare_queries(),
+    ok.
+
+prepare_queries() ->
+    prepare(offline_insert, offline_message,
+            [username, server, from_jid, timestamp, expire,
+             packet, permanent_fields],
+            <<"INSERT INTO offline_message "
+              "(username, server, from_jid, timestamp, expire,"
+              " packet, permanent_fields) "
```
chrzaszcz, line 55: Why is the order changed here? It was different before and this one is also different from the one in the SQL schema...

chrzaszcz (commented):
Nice to see more prepared queries. A few comments from me.

src/offline/mod_offline_rdbms.erl
arcusfelis: We have `remove_user` hook callback, which does not remove the user, but their offline messages...

src/offline/mod_offline_rdbms.erl
arcusfelis: there is no test for it, so if we keep that, we would need to add one.

arcusfelis: put it back too, should be outdated.

arcusfelis:
Having a strange error in escalus, need to rebase on master

PR #3028, opened 2021-02-05
Use prepared queries for MUC legacy - version 2

arcusfelis:
Proposed changes include:
* prepared queries
* rebased

### DIFF
--- a/src/mod_muc_db_rdbms.erl
+++ b/src/mod_muc_db_rdbms.erl
@@ -13,7 +13,7 @@
          unset_nick/3
         ]).
 
--define(ESC(T), mongoose_rdbms:use_escaped_string(mongoose_rdbms:escape_string(T))).
+-import(mongoose_rdbms, [prepare/4, execute_successfully/3]).
 
 %% Defines which RDBMS pool to use
 %% Parent host of the MUC service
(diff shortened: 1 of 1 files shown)


src/mod_muc_db_rdbms.erl
```diff
@@ -25,122 +25,134 @@
...
+    prepare(muc_select_aff, muc_room_aff,
+            [room_id],
+            <<"SELECT luser, lserver, resource, aff "
+              "FROM muc_room_aff WHERE room_id = ?">>),
+    prepare(muc_delete_aff, muc_room_aff, [room_id],
+            <<"DELETE FROM muc_room_aff WHERE room_id = ?">>),
+    %% Queries to muc_registered table
+    prepare(muc_select_nick_user, muc_registered,
+            [muc_host, lserver, nick],
+            <<"SELECT luser FROM muc_registered WHERE muc_host = ?"
```
DenysGonchar, line 73: does it mean that we can have one and the same nick reused by the users from the different hosts?

arcusfelis: I think it's checking for conflicts inside one subdomain.
If we don't check for muc_host, than if someone hosts two domains: `muc.cats.com` and `muc.cars.com`, than there could not be duplicates between these two. Which is something we probably don't wanna do, because users on `cats` could not know anything about `cars`.

DenysGonchar (approved):
looks ok. and according to codecov pretty covered

PR #3031, opened 2021-02-10
Better timestamp logic

NelsonVides:
Just converting some of those ugly triplets-of-~~death~~timestamp things erlang used to do. Just generate everything with smaller and faster `system_time/0,1` calls.

Just converting some of those ugly triplets-of-deathtimestamp things erlang used to do. Just generate everything with smaller and faster system_time/0,1 calls.

TL;DR: everything is extensively explained here: https://erlang.org/doc/apps/erts/time_correction.html

TL;DR;2:
* erlang:monotonic_time/0,1 for measuring time differences. It's what receive ... after and all the timer bifs are using for example. Scalable, always monotonic, non-locking, self-adjusting, etc. Better precision by using /0, that is, in native units, whatever that means for the current erlang runtime, but fine to specify a human-friendly unit if it makes code simpler.
* erlang:system_time/0,1 for timestamps that don't warp or warp gently according to the runtime's configuration.
* os:system_time/0,1 for timestamps that are going somewhere outside of this system, like sending them to a client. The OS has a better view of hardware clocks or NTP syncs. There's no control on warping though.

---

What's the point: system_time and monotonic_time are two completely different units of measure. Most operating systems have some kind of monotonic clock, either in software or by having a (not as precise as a super-expensive atomic clock) hardware device. Then, periodically, the OS syncs over the internet with the Network Time Protocol, hence reflecting the real system-time. In practice, this means than the system time can change in unpredictable ways. A POSIX day is always 86400 seconds long, but UTC days can be 86400, or 86401, or 86399: this knowledge is adjusted through NTP. So a view of the system time can in theory arbitrarily jump back and forth for adjustments, and two measures done exactly one second apart can report any positive or negative number for their difference.

What are all the time functions? Basically, the most important is erlang:monotonic_time() :: integer(), it is the "time engine" for the whole BEAM. All the receive ... after, the BIF timers, and all other timeouts, are actually measured using erlang:monotonic_time(). It is internal to the current runtime (two different runtimes will have completely different start times). It is always monotonic and unshared across threads, and it periodically corrects (unless disabled, which you should basically never do these days). It's all you need, and your best bet, for measuring time differences: the fastest to query, and to compute with, as it is just a 64bits int, and not requiring slow system calls. It is periodically adjusted by the runtime system with regards to the system time, OS time, NTP, and everything else, in very complex ways.

Then there's erlang:system_time, which basically means erlang's point of view of POSIX, with all its internal adjustments, but periodically adjusted to track the OS system time, as the OS has a better view of hardware clocks and NTP. Has correction and warp configurations to adjust that, and adjusts gently to the OS time. Good enough for measuring time diffs and setting up timestamps.

Then even more precise (but not necessarily with more resolution), there's the equivalent OS monotonic time and system time, which are system calls, and most importantly, the erlang runtime has no control on how these warp. They can arbitrarily jump back and forth, or only forth in the case of the monotonic variant, according to the OS adjustments, in very imprecise intervals. Very bad for measuring time diffs.

Corrections and time warps also nicely explained here: https://learnyousomeerlang.com/time (maybe we should add the multi-warp flag to our vm.config?)


mongoose-im:
[9085.1](https://travis-ci.org/esl/MongooseIM/jobs/758429766) / Erlang 23.0.3 / small_tests / 596621067e7827590aa698c5b1dcb455534e1cb7
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/3031/9085/small_tests.23.0.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/3031/9085/small_tests.23.0.3/small/ct_run.mongooseim@localhost.2021-02-10_21.10.43/lib.mongooseim.logs/run.2021-02-10_21.10.44/suite.log.html)

chrzaszcz (commented):
My feelings are mixed about this changeset.
Firstly, it's great to get rid of `erlang:timestamp()`.

But... after the changes I am seeing the following:
`erlang:monotonic_time(microsecond)`
`erlang:system_time(microsecond)`
`erlang:system_time()`

I would like to use only one whenever possible with clear rule when to use the monotonic variant. Otherwise it looks like a bunch of adhoc changes.

src/mam/mod_mam.erl
```diff
@@ -588,18 +588,18 @@ lookup_messages_without_policy_violation_check(Host, #{search_text := SearchText
...
-            Diff = timer:now_diff(os:timestamp(), StartT),
+            Diff = erlang:monotonic_time(microsecond) - StartT,
             mongoose_metrics:update(Host, [backends, ?MODULE, lookup], Diff),
             R
     end.
 
 -spec archive_message(jid:server(), mod_mam:archive_message_params()) -> ok | {error, timeout}.
 archive_message(Host, Params) ->
-    StartT = os:timestamp(),
+    StartT = erlang:monotonic_time(microsecond),
```
chrzaszcz, line 600: Why switch from non-monotonic to monotonic? It should be consistent with the one used when archiving the message.

NelsonVides: Because this is to measure a time difference, non-monotonic doesn't ensure that they won't be the same value and it will report zero, even if it was actually not-zero.

NelsonVides:
> My feelings are mixed about this changeset.
> Firstly, it's great to get rid of `erlang:timestamp()`.
> 
> But... after the changes I am seeing the following:
> `erlang:monotonic_time(microsecond)`
> `erlang:system_time(microsecond)`
> `erlang:system_time()`
> 
> I would like to use only one whenever possible with clear rule when to use the monotonic variant. Otherwise it looks like a bunch of adhoc changes.

Seems legit, I'll have a look and explain why each is used and if some could be unified 👍🏽

src/mam/mod_mam.erl
chrzaszcz: I understand, the natural choice here might be `erlang:timestamp`... it is still not clear for me when to use them and the current choice looks a bit random. I think some rules explaining them are needed before we change everything to make it consistent.

chrzaszcz:
> Seems legit, I'll have a look and explain why each is used and if some could be unified 👍🏽

If you already have a set of your own rules for doing the changes, share them as it is difficult to figure out the motivation of the chagnes, e.g.

- Change `os:timestamp` to `erlang:system_time(...)` in this and that situation - just a few bullet points might be enough.

src/mam/mod_mam.erl
NelsonVides: Not really:
https://erlang.org/doc/man/erlang.html#timestamp-0 says `system_time/0,1` has better performance, and in the do and don't section of https://erlang.org/doc/apps/erts/time_correction.html#new-time-api it says `timestamp()` is there only to assume backwards compatibility with the code that uses the old format. We don't ned to create a triplet of bignums when we can have a single bignum consuming less memory and avoid converting from-and-to, if in the end what is stored in the database or send to metrics are bignums-as-timestamps.

NelsonVides:
> > Seems legit, I'll have a look and explain why each is used and if some could be unified 👍🏽
> 
> If you already have a set of your own rules for doing the changes, share them as it is difficult to figure out the motivation of the changes, e.g.
> 
>     * Change `os:timestamp` to `erlang:system_time(...)` in this and that situation - just a few bullet points might be enough, like when to use monotonic and when not.
> 
> 
> So... I just want to make it clear that I wanted to say it's difficult to figure out the motivation and reasoning - not that there was no motivation or reasoning behind it.
> 
> I think it would be good to use the same reasoning for unifying all timestamp logic, as I am assuming there is much more to cover? maybe we could have a story with the motivation described in a bit more detail?

Will surely prepare all of that when I have a bit of time, it's actually an interesting exercise to write it down 🙂 👍🏽

src/shaper_srv.erl
```diff
@@ -126,7 +126,7 @@ handle_call({wait, Host, Action, FromJID, Size},
             From, State=#state{max_delay=MaxDelayMs}) ->
     Key = new_key(Host, Action, FromJID),
     Shaper = find_or_create_shaper(Key, State),
-    State1 = update_access_time(Key, erlang:timestamp(), State),
+    State1 = update_access_time(Key, erlang:system_time(), State),
```
NelsonVides, line 129: This for example is knowledge of when a shaper was accessed for the last time, and it's local knowledge to the running system, so `erlang:system_time()` is enough.

mongoose-im:
[9091.1](https://travis-ci.org/esl/MongooseIM/jobs/759254328) / Erlang 23.0.3 / small_tests / 5bca679281cc0cb417f88c430782b04cd47a7417
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/3031/9091/small_tests.23.0.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/3031/9091/small_tests.23.0.3/small/ct_run.mongooseim@localhost.2021-02-16_22.27.13/lib.mongooseim.logs/run.2021-02-16_22.27.13/suite.log.html)

NelsonVides:
> > Seems legit, I'll have a look and explain why each is used and if some could be unified 👍🏽
> 
> If you already have a set of your own rules for doing the changes, share them as it is difficult to figure out the motivation of the changes, e.g.
> 
>     * Change `os:timestamp` to `erlang:system_time(...)` in this and that situation - just a few bullet points might be enough, like when to use monotonic and when not.
> 
> 
> So... I just want to make it clear that I wanted to say it's difficult to figure out the motivation and reasoning - not that there was no motivation or reasoning behind it.
> 
> I think it would be good to use the same reasoning for unifying all timestamp logic, as I am assuming there is much more to cover? maybe we could have a story with the motivation described in a bit more detail?

@chrzaszcz Description of the PR has been updated, when merging all that big message can be added to the merge commit, so that it's part of the git history.

When it comes to a general unification of all timestamp logic, hmm... it's a bit of an investigation to find out what other parts of MIM need to improve their timestamps, I wouldn't like this PR to be blocked by that 😕

NelsonVides:
Nevertheless, maybe I'll just try to get rid of usec, there's a ticket for that in our Jira 🤔

mongoose-im:
[9092.1](https://travis-ci.org/esl/MongooseIM/jobs/759262238) / Erlang 23.0.3 / small_tests / 3b970fe8f511d2e0e6ae4fd48739180231197132
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/3031/9092/small_tests.23.0.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/3031/9092/small_tests.23.0.3/small/ct_run.mongooseim@localhost.2021-02-16_23.48.06/lib.mongooseim.logs/run.2021-02-16_23.48.06/suite.log.html)

mongoose-im:
[9095.1](https://travis-ci.org/esl/MongooseIM/jobs/759263764) / Erlang 23.0.3 / small_tests / 71fe5575ee8f118d809e4a291da65c8e089e57a0
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/3031/9095/small_tests.23.0.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/3031/9095/small_tests.23.0.3/small/ct_run.mongooseim@localhost.2021-02-17_00.05.52/lib.mongooseim.logs/run.2021-02-17_00.05.52/suite.log.html)

mongoose-im:
[9096.1](https://travis-ci.org/esl/MongooseIM/jobs/759313551) / Erlang 23.0.3 / small_tests / abdbd15439bb6df3bb7e2bbcef48a20e8ca331c9
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/3031/9096/small_tests.23.0.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/3031/9096/small_tests.23.0.3/small/ct_run.mongooseim@localhost.2021-02-17_09.54.34/lib.mongooseim.logs/run.2021-02-17_09.54.34/suite.log.html)

NelsonVides:
Damn, the PR message messed up. Now it's done, and CI is passing, and usec is out.
@chrzaszcz

src/pubsub/mod_pubsub.erl
```diff
@@ -4470,7 +4470,7 @@ string_to_ljid(JID) ->
 
 -spec uniqid() -> mod_pubsub:itemId().
 uniqid() ->
-    {T1, T2, T3} = timestamp(),
+    {T1, T2, T3} = os:timestamp(),
     iolist_to_binary(io_lib:fwrite("~.16B~.16B~.16B", [T1, T2, T3])).
```
chrzaszcz, line 4474: Maybe we could put some kind of a UUID here or just use the integer timestamp? I think that the triplet is not even guaranteed to be unique, as it is the raw OS timestamp - not monotonic, right?

chrzaszcz (commented):
Great work, just a small comment from me.

src/pubsub/mod_pubsub.erl
NelsonVides: Excellent idea, there's a tiny chance the os will warp at the same time two of these are generated and they might end up generating the same id. And this doesn't require any time knowledge, it's nothing more than a unique binary. Moving it to uuid_v4 :)

mongoose-im:
[9099.1](https://travis-ci.org/esl/MongooseIM/jobs/759644891) / Erlang 23.0.3 / small_tests / 0b8748643774dc549d63c59734de4458665a6102
Reports [root](http://esl.github.io/mongooseim-ct-reports/s3_reports.html?prefix=PR/3031/9099/small_tests.23.0.3) / [small](https://mongooseim-ct-results.s3-eu-west-1.amazonaws.com/PR/3031/9099/small_tests.23.0.3/small/ct_run.mongooseim@localhost.2021-02-19_15.28.38/lib.mongooseim.logs/run.2021-02-19_15.28.39/suite.log.html)

chrzaszcz (approved):

PR #3034, opened 2021-02-18
Insert only a diff of privacy_data

arcusfelis:
This PR addresses "Deadlock occurring in MySQL from time to time in the prepared queries branch".

Proposed changes include:
* Insert only a diff of privacy_data
* Retry privacy update transaction
* Update eodbc. Return bigints as numbers, not binaries now
* Also NUMERIC type is overkill in pgsql, it's at least 10 bytes variable encoding. New installations would benefit from a smaller data type.

### DIFF
--- a/src/mod_privacy_rdbms.erl
+++ b/src/mod_privacy_rdbms.erl
@@ -68,17 +68,26 @@ prepare_queries(Host) ->
     prepare_default_list_upsert(Host),
     %% Queries to privacy_list_data table
     mongoose_rdbms:prepare(privacy_data_get_by_id, privacy_list_data, [id],
-                           <<"SELECT t, value, action, ord, match_all, match_iq, "
+                           <<"SELECT ord, t, value, action, match_all, match_iq, "
                              "match_message, match_presence_in, match_presence_out "
                              "FROM privacy_list_data "
                              "WHERE id=? ORDER BY ord">>),
     mongoose_rdbms:prepare(delete_data_by_id, privacy_list_data, [id],
                            <<"DELETE FROM privacy_list_data WHERE id=?">>),
+    mongoose_rdbms:prepare(privacy_data_delete, privacy_list_data, [id, ord],
+                           <<"DELETE FROM privacy_list_data WHERE id=? AND ord=?">>),
+    mongoose_rdbms:prepare(privacy_data_update, privacy_list_data,
+                           [t, value, action, match_all, match_iq,
+                            match_message, match_presence_in, match_presence_out, id, ord],
+                           <<"UPDATE privacy_list_data SET "
+                             "t=?, value=?, action=?, match_all=?, match_iq=?, "
+                             "match_message=?, match_presence_in=?, match_presence_out=? "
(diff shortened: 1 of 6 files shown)


src/mod_privacy_rdbms.erl
```diff
@@ -311,18 +314,47 @@ execute_privacy_list_insert(LServer, LUser, Name) ->
...
-     || Args <- Rows],
+replace_data_rows(LServer, ID, []) when is_integer(ID) ->
+    %% Just remove the data, nothing should be inserted
+    execute_delete_data_by_id(LServer, ID);
+replace_data_rows(LServer, ID, Rows) when is_integer(ID) ->
+    {selected, OldRows} = execute_privacy_data_get_by_id(LServer, ID),
+    New = lists:sort(Rows),
+    Old = lists:sort([tuple_to_list(Row) || Row <- OldRows]),
+    Diff = diff_rows(ID, New, Old, []),
+    [mongoose_rdbms:execute_successfully(LServer, Q, Args) || {Q, Args} <- Diff],
```
chrzaszcz, line 325: Maybe rewrite it to protect against erroneous items in the list that are silently ignored by LC?

src/mod_privacy_rdbms.erl
```diff
@@ -311,18 +314,47 @@ execute_privacy_list_insert(LServer, LUser, Name) ->
...
     ok.
 
+%% We assume that there are no record duplicates with the same Order.
+%% It's checked in the main module for the New argument.
+%% It's checked by the database for the Old argument.
+diff_rows(ID, [H|New], [H|Old], Ops) ->
+    diff_rows(ID, New, Old, Ops); %% Not modified
+diff_rows(ID, [NH|NT] = New, [OH|OT] = Old, Ops) ->
```
chrzaszcz, line 333: Could you extend the variable names as they are quite cryptic? `NewH` etc would suffice.

src/mod_privacy_rdbms.erl
```diff
@@ -311,18 +314,47 @@ execute_privacy_list_insert(LServer, LUser, Name) ->
...
+    end;
+diff_rows(ID, [], [OH|OT], Ops) ->
+    OO = hd(OH),
+    diff_rows(ID, [], OT, [{privacy_data_delete, [ID, OO]}|Ops]);
+diff_rows(ID, [NH|NT], [], Ops) ->
+    diff_rows(ID, NT, [], [{privacy_data_insert, [ID|NH]}|Ops]);
+diff_rows(_ID, [], [], Ops) ->
+    Ops.
+
+
```
chrzaszcz, line 351: Minor: double empty line.

chrzaszcz (commented):
Looks good! A few small remarks from me.

Also: we need to update the migration guide when merging the prepared queries branch, right? Does this `int` change require any action?

arcusfelis:
The code works both for BIGINT, NUMBER and INT. INT is just a faster version of three.
So, I don't think we need a migration.

chrzaszcz (approved):
Good!

PR #3044, opened 2021-03-01
Fixing GH Actions builds

DenysGonchar:
This PR addresses:

- fix for GH Actions builds
- README.md update
- Travis builds disabling
- Enabling RMQ tests
- fixing elasticsearch & cassandra preset.


src/wpool/mongoose_wpool_redis.erl
```diff
@@ -30,7 +30,7 @@ wpool_spec(WpoolOptsIn, ConnOpts) ->
 
 
 makeargs(RedisOpts) ->
-    Host = proplists:get_value(host, RedisOpts, "localhost"),
+    Host = proplists:get_value(host, RedisOpts, "127.0.0.1"),
```
NelsonVides, line 33: Every backend and ci tool has different requirements for this unfortunately. Like, `mongoose_wpool_elastic` is still hardcoding `localhost` there, would be good to have them all consistent. Perhaps out of the scope of the PR though.

tools/travis-setup-rmq.sh
```diff
@@ -1,6 +1,13 @@
...
+source tools/travis-common-vars.sh
+NAME=$(db_name rmq)
+RMQ_PORT=5672
+
+docker rm -v -f $NAME || echo "Skip removing the previous container"
 docker run -d \
-       --name mongooseim-rmq \
-       -p 5672:5672 \
+       --name $NAME \
+       -p $RMQ_PORT:$RMQ_PORT \
```
NelsonVides, line 10: The inner port is fixed by the container, wouldn't make any sense to be able to change it here, this pattern is usually so that we can map to a chosen port on the host, not on the container.
```suggestion
       -p $RMQ_PORT:5672 \
```

README.md
```diff
@@ -1,11 +1,11 @@
 # MongooseIM platform
 
-[![Build Status](https://github.com/esl/MongooseIM/workflows/CI/badge.svg)](https://github.com/esl/MongooseIM/actions)
-[![CircleCI](https://circleci.com/gh/esl/MongooseIM.svg?style=shield&circle-token=53cdda11cfb95bd7a7044c509c89cb73d3fdeb5a)](https://github.com/esl/MongooseIM/)
-[![codecov](https://codecov.io/gh/esl/MongooseIM/branch/master/graph/badge.svg)](https://codecov.io/gh/esl/MongooseIM)
+[![GitHub Actions](https://github.com/esl/MongooseIM/workflows/CI/badge.svg?branch=master)](https://github.com/esl/MongooseIM/actions?query=workflow%3ACI+branch%3Amaster)
+[![CircleCI](https://circleci.com/gh/esl/MongooseIM.svg?style=shield&circle-token=53cdda11cfb95bd7a7044c509c89cb73d3fdeb5a)](https://app.circleci.com/pipelines/github/esl/MongooseIM)
```
NelsonVides, line 4: I'm actually not sure the `https://app.circleci.com/pipelines/github/esl/MongooseIM` link is public, I thought it's only visible if you log in to a github account that's under the esl organization. Might be worth checking so, just to be sure.

chrzaszcz (commented):
Good changes, I have a few comments:
- Are the RabbitMQ changes also necessary to  fix the GH build? it seems to me that you just spotted that the tests for Rabbit are not performed for GH Actions and fixed that, right?
- The commit history is a bit strange with most of the changes rolled back later. I would prefer to simplify the commit history, but it's not a must, I see that it gives a hint of the possibilities you checked.

NelsonVides (commented):
Note that, upon removing the travis.yml file, we'll also need to configure travis itself to stop the builds, in the travis UI.

Also, together with the README, please update `doc/index.md` (we need to do something about that duplication some day)

DenysGonchar:
Right, the fix itself is replacement of "localhost" with "127.0.0.1".
The history of commits contains investigations steps. So it's possible to return to the testing procedure if required.

chrzaszcz:
> Right, the fix itself is replacement of "localhost" with "127.0.0.1".
Could you then mention the RabbitMQ changes separately? We need to see it passing on CircleCI as well as GH Actions btw before merging. Are you going to create a separate PR for this?

tools/travis-setup-rmq.sh
DenysGonchar: the first port is the host port
https://docs.docker.com/engine/reference/run/#expose-incoming-ports

big_tests/test.config
```diff
@@ -269,7 +269,7 @@
       {mod_roster, "[modules.mod_roster]
   backend = \"rdbms\""}]},
     {mysql_redis,
-     [{dbs, [redis, mysql]},
+     [{dbs, [redis, mysql, rmq]},
```
chrzaszcz, line 272: I think CircleCI just starts Rabbit for all presets (see `config.yml`. Did you check what happens to it after this change? My guess is that it would be started twice and it would be better to remove it from `config.yml`, right?

DenysGonchar:
> > Right, the fix itself is replacement of "localhost" with "127.0.0.1".
> 
> Could you then mention the RabbitMQ changes separately? We need to see it passing on CircleCI as well as GH Actions btw before merging. Are you going to create a separate PR for this? For some reason no checks are performed for this PR.

Sorry, I don't really have for splitting. I'm too busy with other stuff, and it already took one of my weekend days.
Anyway, this change will not affect CircleCI builds because `DB` environment variable is set explicitly in `.circleci/config.yml`

NelsonVides:
> > > Right, the fix itself is replacement of "localhost" with "127.0.0.1".
> > 
> > 
> > Could you then mention the RabbitMQ changes separately? We need to see it passing on CircleCI as well as GH Actions btw before merging. Are you going to create a separate PR for this? For some reason no checks are performed for this PR.
> 
> Sorry, I don't really have for splitting. I'm too busy with other stuff, and it already took one of my weekend days.
> Anyway, this change will not affect CircleCI builds because `DB` environment variable is set explicitly in `.circleci/config.yml`

To be precise, I've seen test suites checking the env-var, and other test suites checking the db values in the preset. So that ain't consistent anyway 🤷🏽‍♂️

chrzaszcz:
> Sorry, I don't really have for splitting.
There's no need for splitting, just mention it in the PR description.

big_tests/test.config
DenysGonchar: it won't.  we use this list to set up DB env for GH Actions in automatic way.
CircleCI sets it explicitly for different presents..

DenysGonchar:
> > > > Right, the fix itself is replacement of "localhost" with "127.0.0.1".
> > > 
> > > 
> > > Could you then mention the RabbitMQ changes separately? We need to see it passing on CircleCI as well as GH Actions btw before merging. Are you going to create a separate PR for this? For some reason no checks are performed for this PR.
> > 
> > 
> > Sorry, I don't really have for splitting. I'm too busy with other stuff, and it already took one of my weekend days.
> > Anyway, this change will not affect CircleCI builds because `DB` environment variable is set explicitly in `.circleci/config.yml`
> 
> To be precise, I've seen test suites checking the env-var, and other test suites checking the db values in the preset. So that ain't consistent anyway 🤷🏽‍♂️

just made a quick recheck, currently none of the suites use DB env variable. 
suites must use only `test.config` data. if you know the place where DB env is checked instead then it's wrong and must be fixed.

DenysGonchar:
> Could you then mention the RabbitMQ changes separately in the PR description? We need to see it passing on CircleCI as well as GH Actions btw before merging. Are you going to create a separate PR for this? For some reason no checks are performed for this PR.

Here is GH Actions build for the last commit:
https://github.com/DenysGonchar/MongooseIM/actions/runs/610442594

Sorry I cannot trigger CircleCI for this PR, there's no such option.

chrzaszcz:
> Here is GH Actions build for the last commit:
> https://github.com/DenysGonchar/MongooseIM/actions/runs/610442594
Nice!

> Sorry I cannot trigger CircleCI for this PR, there's no such option.
We could merge it to `esl:fix-gh-actions` and then make a second PR to master. But if you're sure it does not break anything on CircleCI then we could merge it directly.

src/wpool/mongoose_wpool_redis.erl
DenysGonchar: actually there's problem with eredis library, it seems that they fixed it in the latest version. but they have just released it and it's buggy. hex package is broken and not building at all, GH release just crashes.

DenysGonchar:
> Note that, upon removing the travis.yml file, we'll also need to configure travis itself to stop the builds, in the travis UI.
> 
> Also, together with the README, please update `doc/index.md` (we need to do something about that duplication some day)

is it the same? can i just copy past the text?

DenysGonchar:
> We could merge it to esl:fix-gh-actions and then make a second PR to master. 

good idea, i changed destination branch.

NelsonVides:
> > Note that, upon removing the travis.yml file, we'll also need to configure travis itself to stop the builds, in the travis UI.
> > Also, together with the README, please update `doc/index.md` (we need to do something about that duplication some day)
> 
> is it the same? can i just copy/past the text?

I'm not sure, I think it's the same but just in case just do the same changes you did here also there, I'll check the differences and unify them in a separate effort soon.

README.md
chrzaszcz: "every ever"? I think the previous link was maybe better as it was public?

DenysGonchar: > "every ever"? I think the previous link was maybe better as it was public?

If you check commit verification statuses they all have the same links to CircleCI, the same story with our PRs.
so yes, these links are everywhere in our repo.

NelsonVides: They are in the PRs only, in github, not in the source code.

tools/travis-setup-rmq.sh
DenysGonchar: sorry i'm "blind", I misunderstood the comment and somehow skipped the suggestion. you're right

PR #3045, opened 2021-03-01
Fix gh actions

DenysGonchar:
This PR addresses:

    fix for GH Actions builds
    README.md update
    Travis builds disabling
    Enabling RMQ tests
    fixing elasticsearch & cassandra preset.


README.md
```diff
@@ -1,11 +1,11 @@
 # MongooseIM platform
 
-[![Build Status](https://github.com/esl/MongooseIM/workflows/CI/badge.svg)](https://github.com/esl/MongooseIM/actions)
-[![CircleCI](https://circleci.com/gh/esl/MongooseIM.svg?style=shield&circle-token=53cdda11cfb95bd7a7044c509c89cb73d3fdeb5a)](https://github.com/esl/MongooseIM/)
-[![codecov](https://codecov.io/gh/esl/MongooseIM/branch/master/graph/badge.svg)](https://codecov.io/gh/esl/MongooseIM)
+[![GitHub Actions](https://github.com/esl/MongooseIM/workflows/CI/badge.svg?branch=master)](https://github.com/esl/MongooseIM/actions?query=workflow%3ACI+branch%3Amaster)
+[![CircleCI](https://circleci.com/gh/esl/MongooseIM.svg?style=shield&circle-token=53cdda11cfb95bd7a7044c509c89cb73d3fdeb5a)](https://app.circleci.com/pipelines/github/esl/MongooseIM)
```
NelsonVides, line 4: If the link is not publicly available, do leave the link to the source repo instead. Don't put links nobody can access.

NelsonVides: Likewise for the index.md file and the other circleci link below.

tools/travis-setup-rmq.sh
```diff
@@ -1,6 +1,13 @@
...
+source tools/travis-common-vars.sh
+NAME=$(db_name rmq)
+RMQ_PORT=5672
+
+docker rm -v -f $NAME || echo "Skip removing the previous container"
 docker run -d \
-       --name mongooseim-rmq \
-       -p 5672:5672 \
+       --name $NAME \
+       -p $RMQ_PORT:$RMQ_PORT \
```
NelsonVides, line 10: Again, this must be:
```suggestion
       -p $RMQ_PORT:5672 \
```
I know the first port is the host port, that's exactly what I said, that's the one we want to have configurable. The container's port is fixed, we can't change that.

NelsonVides (changes requested):

README.md
DenysGonchar: If you check commit verification statuses they all have the same links to CircleCI, the same story with our PRs.
so these links are everywhere in our repo, and being honest I find it quite confusing when the link that reports verification status lead to the repo itself. I would rather want to see some message from CircleCI that I have no access to the results. At least I would know that it's CircleCI.

chrzaszcz: It's difficult to decide for me, I'm fine with either version as long as it is consistent.

NelsonVides: Ok, not insisting then here either 👍🏽

NelsonVides (approved):
Good stuff, thanks for the important fixes! 👍🏽

chrzaszcz (approved):
Good!

PR #3046, opened 2021-03-01
General docs improvements

NelsonVides:
version css and js is not necessary as materials now accepts mike as a provider and automatically fills that for us more beautifully (but I honestly have no way to test this without

The site_url was missing the `latest/` suffix and hence the sitemap wasn't exactly accurate.

And some metadata is now added, for more html tagging with references to us.

You can have a look in my fork about how would it look: https://nelsonvides.github.io/MongooseIM/latest/, and also check the sitemap generated: https://nelsonvides.github.io/MongooseIM/latest/sitemap.xml


chrzaszcz (approved):
Great!

PR #3047, opened 2021-03-02
prepared queries pubsub

leszke:
My part of work related to `mod_pubsub` and what have I done so far. Once everything is done, I think I'll squash into 1 commit for better readability.

### DIFF
--- a/src/pubsub/mod_pubsub_db_rdbms.erl
+++ b/src/pubsub/mod_pubsub_db_rdbms.erl
@@ -85,7 +85,8 @@ start() ->
     mod_pubsub_db_mnesia:start(),
     % -------------------- State building ----------------------------
     mongoose_rdbms:prepare(pubsub_get_item_rows_id, pubsub_items, [nidx],
-        <<"SELECT nidx, created_luser, created_lserver, itemid FROM pubsub_items WHERE nidx = ?">>),
+        <<"SELECT nidx, created_luser, created_lserver, itemid "
+          "FROM pubsub_items WHERE nidx = ?">>),
     mongoose_rdbms:prepare(pubsub_get_affiliation_rows_id, pubsub_affiliations, [nidx],
         <<"SELECT nidx, luser, lserver, aff FROM pubsub_affiliations WHERE nidx = ?">>),
     mongoose_rdbms:prepare(pubsub_get_subscriptions_rows_id, pubsub_subscriptions, [nidx],
@@ -100,15 +101,20 @@ start() ->
     mongoose_rdbms:prepare(pubsub_get_subscriptions_rows, pubsub_subscriptions, [luser, lserver],
         <<"SELECT nidx, luser, lserver, lresource, type, sub_id FROM pubsub_subscriptions "
           "WHERE luser = ? AND lserver = ?">>),
-    mongoose_rdbms:prepare(pubsub_get_subscriptions_rows_resource, pubsub_subscriptions, [luser, lserver, lresource],
+    mongoose_rdbms:prepare(pubsub_get_subscriptions_rows_resource, pubsub_subscriptions,
+        [luser, lserver, lresource],
         <<"SELECT nidx, luser, lserver, lresource, type, sub_id FROM pubsub_subscriptions "
           "WHERE luser = ? AND lserver = ? AND lresource = ?">>),
-    mongoose_rdbms:prepare(pubsub_get_idxs_of_own_nodes_with_pending_subs, pubsub_affiliations, [aff, luser, lserver, type],
-        <<"SELECT DISTINCT s.nidx"" FROM pubsub_affiliations AS a INNER JOIN pubsub_subscriptions s "
-          "ON a.nidx = s.nidx WHERE a.aff = ? AND a.luser = ? AND a.lserver = ? AND s.type = ?">>),
+    mongoose_rdbms:prepare(pubsub_get_idxs_of_own_nodes_with_pending_subs, pubsub_affiliations,
+        [aff, luser, lserver, type],
+        <<"SELECT DISTINCT s.nidx"" FROM pubsub_affiliations AS a "
+          "INNER JOIN pubsub_subscriptions s ON a.nidx = s.nidx "
+          "WHERE a.aff = ? AND a.luser = ? AND a.lserver = ? AND s.type = ?">>),
     %% not used anywhere
-    mongoose_rdbms:prepare(pubsub_delete_node_entity_items, pubsub_items, [nidx, created_luser, created_lserver],
-        <<"DELETE FROM pubsub_items WHERE nidx = ? AND created_luser = ? AND created_lserver = ?">>),
+    mongoose_rdbms:prepare(pubsub_delete_node_entity_items, pubsub_items,
+        [nidx, created_luser, created_lserver],
+        <<"DELETE FROM pubsub_items WHERE nidx = ? AND created_luser = ? "
+          "AND created_lserver = ?">>),
 
     % ------------------- Affiliations --------------------------------
     mongoose_rdbms:prepare(pubsub_get_affiliation, pubsub_affiliations, [nidx, luser, lserver],
@@ -121,8 +127,8 @@ start() ->
     % ------------------- Subscriptions --------------------------------
     mongoose_rdbms:prepare(pubsub_insert_subscription, pubsub_subscriptions,
         [nidx, luser, lserver, lresource, type, sub_id, options],
-        <<"INSERT INTO pubsub_subscriptions (nidx, luser, lserver, lresource, type, sub_id, options) "
-          "VALUES (?, ?, ?, ?, ?, ?, ?)">>),
+        <<"INSERT INTO pubsub_subscriptions (nidx, luser, lserver, lresource, "
+            "type, sub_id, options) VALUES (?, ?, ?, ?, ?, ?, ?)">>),
     mongoose_rdbms:prepare(pubsub_update_subscription_opts, pubsub_subscriptions,
         [options, nidx, luser, lserver, lresource, sub_id],
         <<"UPDATE pubsub_subscriptions SET options = ? WHERE nidx = ? "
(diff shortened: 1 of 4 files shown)


src/pubsub/mod_pubsub_db_rdbms.erl
```diff
@@ -86,6 +83,146 @@
...
+        <<"SELECT aff, ", PubsubNodeFieldsPrefixed/binary, " FROM pubsub_affiliations AS pa "
+          "INNER JOIN pubsub_nodes AS pn ON pa.nidx = pn.nidx WHERE luser = ? AND lserver = ?">>),
+    mongoose_rdbms:prepare(pubsub_select_subnodes_empty_node, pubsub_nodes, [key],
+        <<"SELECT ", PubsubNodeFieldsPrefixed/binary, " from pubsub_nodes as pn "
+          "LEFT JOIN pubsub_node_collections as collection ON pn.name = collection.name "
+          "WHERE p_key = ? AND collection.parent_name IS NULL">>),
+    mongoose_rdbms:prepare(pubsub_select_subnodes, pubsub_nodes, [parent_name, key],
+        <<"SELECT ", PubsubNodeFieldsPrefixed/binary, " from pubsub_nodes as pn "
+          "INNER JOIN pubsub_node_collections as collection ON pn.name = collection.name AND "
+          "collection.parent_name = ? WHERE p_key = ?">>),
```
arcusfelis, line 204: This query whole not work, because it has arguments from two tables.
I would have to add that feature.

src/pubsub/mod_pubsub_db_rdbms.erl
```diff
@@ -86,6 +83,146 @@
...
+    mongoose_rdbms:prepare(pubsub_del_parents, pubsub_node_collections, [node],
+        <<"DELETE FROM pubsub_node_collections WHERE name = ?">>),
+
+    % ------------------- GDPR --------------------------------
+    mongoose_rdbms:prepare(pubsub_get_user_items, pubsub_items, [created_luser, created_lserver],
+        <<"SELECT name, itemid, payload FROM pubsub_items INNER JOIN pubsub_nodes "
+          "ON pubsub_items.nidx = pubsub_nodes.nidx WHERE created_luser = ? "
+          "AND created_lserver = ?">>),
+    mongoose_rdbms:prepare(pubsub_select_nodes_by_owner_mysql, pubsub_nodes, [ljid],
+        <<"SELECT name, type FROM pubsub_nodes WHERE owners = convert([\"?\"], JSON);">>),
```
arcusfelis, line 216: that is something weird.

Probably should be
```
<<"SELECT name, type FROM pubsub_nodes WHERE owners = convert([?], JSON);">>
```

Also, that query is very slow and we would have to get rid of this mole (maybe it's not used or there is more efficient way to do that).

src/pubsub/mod_pubsub_db_rdbms.erl
```diff
@@ -86,6 +83,146 @@
...
+    mongoose_rdbms:prepare(pubsub_select_nodes_by_owner_pgsql, pubsub_nodes, [ljid],
+        <<"SELECT name, type"" FROM pubsub_nodes WHERE owners ::json->>0 like ? "
```
arcusfelis, line 218: `""` is not needed.

src/pubsub/mod_pubsub_db_rdbms.erl
```diff
@@ -86,6 +83,146 @@
...
+          "AND JSON_ARRAY_LENGTH(owners) = 1">>),
+    mongoose_rdbms:prepare(pubsub_select_nodes_by_owner_odbc, pubsub_nodes, [ljid],
+        <<"SELECT name, type FROM pubsub_nodes WHERE cast(owners as nvarchar(max)) = [\"?\"];">>),
+    mongoose_rdbms:prepare(pubsub_get_user_subscriptions, pubsub_subscriptions, [luser, lserver],
+        <<"SELECT name"" FROM pubsub_subscriptions INNER JOIN pubsub_nodes "
```
arcusfelis, line 223: `""` is not needed.

src/pubsub/mod_pubsub_db_rdbms.erl
```diff
@@ -632,6 +781,23 @@ decode_pubsub_node_with_aff_row(Row) ->
...
+     <<"nidx, itemid, created_luser, created_lserver, created_at, "
+       "modified_luser, modified_lserver, modified_lresource, modified_at, "
+       "publisher, payload">>.
+
+-spec pubsub_node_fields() -> binary().
+pubsub_node_fields() ->
+    <<"nidx, p_key, name, type, owners, options">>.
+
+-spec pubsub_node_fields(string()) -> binary().
+pubsub_node_fields(Prefix) ->
```
arcusfelis, line 795: There is only one prefix is used: `pn`.

So we just need:

```
pubsub_node_fields_pn() ->
    <<"pn.nidx, pn.p_key, pn.name, pn.type, pn.owners, pn.options">>.
```

src/pubsub/mod_pubsub_db_rdbms.erl
```diff
@@ -86,6 +83,146 @@
...
+          "WHERE luser = ? AND lserver = ?">>),
+    mongoose_rdbms:prepare(pubsub_get_subscriptions_rows_resource, pubsub_subscriptions,
+        [luser, lserver, lresource],
+        <<"SELECT nidx, luser, lserver, lresource, type, sub_id FROM pubsub_subscriptions "
+          "WHERE luser = ? AND lserver = ? AND lresource = ?">>),
+    mongoose_rdbms:prepare(pubsub_get_idxs_of_own_nodes_with_pending_subs, pubsub_affiliations,
+        [aff, luser, lserver, type],
+        <<"SELECT DISTINCT s.nidx"" FROM pubsub_affiliations AS a "
+          "INNER JOIN pubsub_subscriptions s ON a.nidx = s.nidx "
+          "WHERE a.aff = ? AND a.luser = ? AND a.lserver = ? AND s.type = ?">>),
```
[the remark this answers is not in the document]
leszke: for the simple path: it's hardcoded to be always `sub2int(pending) == 1`, should we really care about the impossible cases and introduce dead code?

src/pubsub/mod_pubsub_db_rdbms_sql.erl
```diff
@@ -381,13 +71,6 @@ insert_pubsub_node(Key, Name, Type, Owners, Options) ->
...
     sql_node_insert(EscKey, EscName, EscType, EscOwners, EscOptions, RDBMSType).
 
-update_pubsub_node(Nidx, Type, Owners, Options) ->
-    EscNidx = esc_int(Nidx),
-    EscType = esc_string(Type),
-    EscOwners = esc_string(Owners),
-    EscOptions = esc_string(Options),
-    sql_node_update(EscNidx, EscType, EscOwners, EscOptions).
-
 sql_node_insert(EscKey, EscName, EscType, EscOwners, EscOptions, {odbc, mssql}) ->
```
chrzaszcz, line 74: My suggestion is to just execute an insert followed by a select. Other option might be to allow this optimized INSERT+SELECT as a simple query, but I think this should be done separately, not in this PR.

src/pubsub/mod_pubsub_db_rdbms_sql.erl
```diff
@@ -420,80 +103,15 @@ common_node_insert(EscKey, EscName, EscType, EscOwners, EscOptions) ->
...
      EscOptions, ")"].
 
-sql_node_update(EscNidx, EscType, EscOwners, EscOptions) ->
-    Query = [" UPDATE pubsub_nodes SET type = ", EscType, ", "
-             " owners = ", EscOwners, ", "
-             " options = ", EscOptions,
-             " WHERE nidx = ", EscNidx, ";"],
-    {updated, _} = mongoose_rdbms:sql_query(global, Query).
-
 set_parents(Node, Parents) ->
```
chrzaszcz, line 106: Let's do one query per parent here.

src/pubsub/mod_pubsub_db_rdbms_sql.erl
```diff
@@ -420,80 +103,15 @@ common_node_insert(EscKey, EscName, EscType, EscOwners, EscOptions) ->
...
-
--spec select_nodes_in_list_with_key(Key :: binary(), Nodes :: [binary()]) -> iolist().
-select_nodes_in_list_with_key(Key, Nodes) ->
-    EscapedNames = [esc_string(Node) || Node <- Nodes],
-    NodeNames = rdbms_queries:join(EscapedNames, ","),
-    ["SELECT ", pubsub_node_fields(), " from pubsub_nodes "
-     "WHERE p_key = ", esc_string(Key),
-     " AND name IN (", NodeNames, ")"].
-
 -spec select_nodes_by_key_and_names_in_list_with_parents(Key :: binary(), Nodes :: [binary()]) -> iolist().
```
chrzaszcz, line 115: These two are quite tricky. Currently I would just do one query per each `pn.name` and concatenate the results later. We could optimize it separately if we find a need for it.

src/rdbms/mongoose_rdbms.erl
```diff
@@ -197,7 +197,7 @@ execute(Host, Name, Parameters) when is_atom(Name), is_list(Parameters) ->
     sql_call(Host, {sql_execute, Name, Parameters}).
 
 %% Same as execute/3, but would fail loudly on any error.
--spec execute_successfully(Host :: server(), Name :: atom(), Parameters :: [term()]) ->
+-spec execute_successfully(Host :: server() | ignored, Name :: atom(), Parameters :: [term()]) ->
```
chrzaszcz, line 200: This `ignored` would need to be added to many more function specs in this module. My opinion is that we can use `global` in the queries for now.

src/pubsub/mod_pubsub_db_rdbms.erl
```diff
@@ -577,35 +918,30 @@ get_entity_items(Nidx, { LU, LS, _ }) ->
...
+        {updated, _} = execute_delete_item(LS, Nidx, LU, ItemId)
+    end, ItemIds).
 
 -spec remove_all_items(Nidx :: mod_pubsub:nodeIdx()) ->
     ok.
 remove_all_items(Nidx) ->
-    SQL = mod_pubsub_db_rdbms_sql:delete_all_items(Nidx),
-    {updated, _} = mongoose_rdbms:sql_query_t(SQL),
+    {updated, _} = execute_delete_all_items(Nidx),
     ok.
```
chrzaszcz, line 928: I looked at the callers of these functions as they essentially duplicate `del_item` and `del_items` and I found that they are called one after another in `node_flat`. In `mod_pubsub_db_mnesia` they are doing separate things:

- del_item(s) remove `pubsub_item` records
- remove_(all)_items remove items from the `pubsub_state` record

However, in RDBMS they are doing the same thing, so it is done twice. One solution might be to make `del_item` and `del_items` do nothing in the DB, but this is counter-intuitive, so I think that we could keep it as it is for now and fix it in `node_flat` in a separate PR.

chrzaszcz (commented):
A lot of good work done! I added a few comments.

chrzaszcz (approved):
Good work, a lot of queries! The remaining ones are more complex and will be handlend in separate PR's.

Issue #2933, opened 2020-11-09
crash when handover_session


ThomasChiroux:
MongooseIM version: 4.0.0
Installed from: source
Erlang/OTP version: 23 

Hi,
I'm in the process of migrating from ejabberd to mongoose and while doing my tests I have a case with mongoose crashing while trying to recover session when using stream management.

Here are the logs, below my mongoose configuration and some datas about the client.
I'm in a test environment with docker images, rdbms mysql. My mongoose image is self compiled running last alpine release 3.12

# TLDR log of the error:

```
when=2020-11-09T13:34:20.917723+00:00 level=error pid=<0.7072.0> at=: unstructured_log="** State machine <0.7072.0> terminating \
** Last message in was {'$gen_sync_event',\
                           {<0.7162.0>,#Ref<0.2309600686.4056154113.172964>},\
                           resume}\
** When State == resume_session\
**      Data  == {state,\
                  {websocket,<0.7070.0>,{{172,18,0,2},58416},undefined},\
                  mod_websockets,#Ref<0.2309600686.4056154113.165547>,true,\
                  <<\"f294afdf6507b7f8\">>,\
                  {sasl_state,<<\"jabber\">>,<<\"mongoose\">>,<<>>,undefined,\
                   undefined,\
                   {mongoose_credentials,<<\"mongoose\">>,[],[]}},\
                  all,none,\
                  {false,0},\
                  false,false,false,\
                  [verify_none],\
                  verify_none,true,\
                  {jid,<<\"testjid\">>,<<\"mongoose\">>,\
                   <<\"61c2500c5ffafbbd1604-928595-344402\">>,\
                   <<\"testjid\">>,<<\"mongoose\">>,\
                   <<\"61c2500c5ffafbbd1604-928595-344402\">>},\
                  <<\"testjid\">>,<<\"mongoose\">>,\
                  <<\"61c2500c5ffafbbd1604-928595-344402\">>,\
                  {1604928595423503,<0.7072.0>},\
                  {pres_t,2},\
                  {pres_f,2},\
                  {pres_a,2},\
                  {pres_i,0},\
                  [],\
                  {xmlel,<<\"presence\">>,\
                   [{<<\"xml:lang\">>,<<\"en\">>},\
                    {<<\"xmlns\">>,<<\"jabber:client\">>},\
                    {<<\"id\">>,<<\"9714bd47-0940-424d-91d4-4e75dbc7e363\">>}],\
[...]

PR #3048, opened 2021-03-03
Fix crash during handover session

imcyee:
This PR addresses #
Fix issue crash during handover_session https://github.com/esl/MongooseIM/issues/2933.

Issue summary:
`ejabberd_c2s:handover_session` terminate with `{stop, {handover_session, From}, SD}.` which `p1_fsm_old:terminate` did not handle, so it goes to end of the case of `_` that generates an error info.

During handover_session, its last line call `p1_fsm_old` to terminate:
https://github.com/esl/MongooseIM/blob/135fec226fa3be355dcac6be5536821c78b34bb4/src/ejabberd_c2s.erl#L3183-L3192

in `p1_fsm_old`, there is no case to handle stop_reason `handover_session` hence it generates error:
https://github.com/esl/MongooseIM/blob/135fec226fa3be355dcac6be5536821c78b34bb4/src/p1_fsm_old.erl#L711-L735

Proposed changes include:
added a `{handover_session, _From}` as `Reason` to exit normally

### DIFF
--- a/src/p1_fsm_old.erl
+++ b/src/p1_fsm_old.erl
@@ -724,6 +724,8 @@ terminate(Reason, Name, Msg, Mod, StateName, StateData, Debug, Queue) ->
                     %% Priority shutdown should be considered as
                     %% shutdown by SASL
                     exit(shutdown);
+                {handover_session, _From} ->
+                    exit(normal);
                 {process_limit, _Limit} ->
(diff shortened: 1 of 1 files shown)


gustawlippa:
Thank you, it was indeed the problem, closing this one as we merged #3056 with added testcase.

PR #3050, opened 2021-03-05
Use prepared queries in ejabberd_auth_rdbms

chrzaszcz:
**This PR includes the following changes:**
- Replace all plain queries in `ejabberd_auth_rdbms.erl` with their prepared equivalents
- Refactor code as it is already changed
- Improve and test Erlang-shell-only utilities for obtaining list and number of registered users.
- New tests revealed the following issues:
  - There were bugs in the corresponding functions in `ejabberd_auth_internal`- this is fixed now.
  - Prepared statements with `LIKE` were missing escaping. This is fixed and tested now.
  - The option `pgsql_users_number_estimate` was working for Postgres, but for MySQL an estimate was used by default, causing confusion and test failures. This is unified now and a new option `auth.rdbms.users_number_estimate` supports both Postgres and MySQL.

**Out of scope:**
- Change error handling in auth backends not to fabricate results in case of DB errors. It would require significant changes in multiple auth backends as they need to have consistent behaviour.
- Use prepared `LIKE` escaping in other places in the code base. This needs to be done later.
- `LIKE` escaping seems to be incomplete for MSSQL, even for plain queries. This might be investigated in the future.
- `LIKE` escaping is (and was) broken for plain queries, as shown by the new tests, but we are getting rid of such queries.

### DIFF
--- a/big_tests/tests/accounts_SUITE.erl
+++ b/big_tests/tests/accounts_SUITE.erl
@@ -45,10 +45,13 @@ groups() ->
                                              registration_failure_timeout]},
          {change_account_details, [parallel], change_password_tests()},
          {change_account_details_store_plain, [parallel], change_password_tests()},
-         {utilities, [parallel], [list_users,
+         {utilities, [{group, user_info},
+                      {group, users_number_estimate}]},
+         {user_info, [parallel], [list_users,
                                   list_selected_users,
                                   count_users,
-                                  count_selected_users]}
+                                  count_selected_users]},
+         {users_number_estimate, [], [count_users_estimate]}
         ],
     ct_helper:repeat_all_until_all_ok(G).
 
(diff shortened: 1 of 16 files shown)


DenysGonchar:
It seems that we are missing testing of queries with `users_number_estimate`  option

chrzaszcz:
> Looks great. just a few comments from my site:
> 
>     * It seems that we are missing testing of queries with `users_number_estimate`  option.
>     * Plus a couple of more places at `src/auth/ejabberd_auth_rdbms.erl` reported as uncovered by codecov.

Yes, I skipped some places which were difficult to check and just tested them manually. This includes queries with `users_number_estimate` (they apparently just return zero for small tables, but there is no guarantee for this) as well as the whole `scram_passwords` utility which can be used only in the shell. IMO this is enough for now, what do you think?

DenysGonchar (approved):
> IMO this is enough for now, what do you think?

👍  it's ok

chrzaszcz:
I just double checked and there is an issue with the estimate query for Postgres, I will update the PR with the fix.

PR #3052, opened 2021-03-08
Service for database domain management

arcusfelis:
Proposed changes include:

- Service
- Core
- DB queries
- Docs

DONE:
- We need to write a gen_server, which removes old records from the domain_events table.
- Pagination tests - many records inserted. Maybe proper tests.
- Other DB backends, not just Postgres.
- Distributed tests - more than one node (though do we really need it? Other nodes would read events from the domain_events table in exactly the same way, as one node does).
**- Need to add tests to check what is happening when service  goes down.**


src/domain/mongoose_domain_sql.erl
```diff
@@ -0,0 +1,192 @@
...
+selected_to_int({selected, [{null}]}) -> 0;
+selected_to_int({selected, [{UpdateNum}]}) ->
+    mongoose_rdbms:result_to_integer(UpdateNum).
+
+get_db_specific_limits() ->
+    %% XXX Once merged with upstream:
+    % rdbms_queries:get_db_specific_limits_binaries().
+    {<<"LIMIT ?">>, <<>>}.
+
+add_limit_arg(Limit, Args) ->
```
arcusfelis, line 181: do we wanna rebase the base branch?

chrzaszcz:
Please describe the architecture, e.g. why two database tables are needed. One commit with 1000 lines of code and no description of the changes makes it difficult to figure out the reasoning behind all the changes.

src/domain/mongoose_domain_core.erl
```diff
@@ -0,0 +1,130 @@
...
+dump() ->
+    ets:tab2list(?TABLE).
+
+restore(Dump) ->
+    ets:delete_all_objects(?TABLE),
+    ets:insert_new(?TABLE, Dump),
+    ok.
+
+%% Global's callback
+resolve(_Name, Pid1, Pid2) -> Pid1.
```
arcusfelis, line 93: delete me

src/domain/mongoose_domain_loader.erl
```diff
@@ -0,0 +1,82 @@
+-module(mongoose_domain_loader).
+-export([load_data_from_base/2]).
+-export([check_for_updates/2]).
+
+-include("mongoose_logger.hrl").
+
+load_data_from_base(FromNum, PageSize) ->
```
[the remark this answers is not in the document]
arcusfelis: Yeah, I was thinking about that. There are some issues:
- Routing to unknown domain would become slower.
- Cache invalidation? Maybe nope. rpc could fail, updates could come from a node, which is net-splited from the main cluster.
- The cost of not applying some change is extremely high. Basically, the whole tenant would be not available.
- After the full restart we would have different performance characteristics, because the ETS table is cold.
- By the way, 10k requests at once to a single cold domain - totally possible, we would have to have logic to track loading. Which is also code, just not on the DB side ;) 
- We can still do lazy init for domains, if we need to do init for domains (like starting some gen_servers).
- Time of execution of get_host_type becomes random, which could break some time-sensitive code.

That is something similar to MUC rooms.
As long as there is less than 1M domains, it makes sense to store the whole set on MIM side.
Once we want 100M domains - we would have to do lazy loading (which is not the case for us).


My main considerations:
- to have DB to be the source of truth.
- to never miss disable/enable/remove/insert of a domain.
- To not use erlang distribution for anything critical. 
- To not use Mnesia.
- Once node starts, it is ready to work as fast as possible.
- If DB is not available on any of the nodes - we would miss the updates, but once it's back - MIM recovers easily.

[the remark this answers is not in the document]
chrzaszcz: Thanks for sharing the rationale behind your solution, this was the main goal of my comment :-)

I think that relying on periodic updates is risky as we are deliberately making the data out of sync for some time, even on the very node that performs the change (albeit for a shorter period). When routing the message to a different node, we might get different domain mapping for many stanzas which might cause a flood of errors (for some modules) until the eventual consistency is reached. I think that preventing such errors is non-trivial.

Periodic events also make bugs more random and difficult to reproduce. In all systems with multiple periodic events there was always a tricky bug caused by an unlikely ordering of such events. With more complexity there is always more opportunities for bugs to creep in.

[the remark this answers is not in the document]
chrzaszcz: >     * We can still do lazy init for domains, if we need to do init for domains (like starting some gen_servers).
I don't think any `gen_server` needs to be started per domain. The only values initialized per domain would be very simple like the mapping from a MUC subdomain to the host type. It can be either lazily stored (cached) in an ETS or just replaced with a simple bit of code that calculates it dynamically like I did for Motorola. It had no  impact on load test results.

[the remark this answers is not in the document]
arcusfelis: Each time we change a domain, we notify all nodes, that there is something new in the database.
So, it speeds up the update from the default 0-30 seconds, till the domain is updated.

But there is no requirement to block the insert_domain request, until the operation has been applied across all nodes, so I haven't added it.

But if you want to block until it's applied across all nodes, just call `service_domain_db:sync()`, it will block until everything is consumed from the DB. We use this function in our tests.

src/domain/mongoose_domain_core.erl
```diff
@@ -0,0 +1,139 @@
+%% Generally, you should not call anything from this module.
```
DenysGonchar, line 1: This module doesn't correspond to initial requirements:
* ETS must be read only for public access
* init must accept 2 parameters:
    * the list of initial (not changeable) host_type/domain pairs
    * the list of acceptable host_type
* host_type cannot be changed for the existing domain, it should be possible only to add or remove domain.
* domains can be added only for acceptable host_types

src/domain/mongoose_domain_core.erl
```diff
@@ -0,0 +1,139 @@
...
+
+is_locked(Domain) ->
+    case ets:lookup(?TABLE, Domain) of
+        [{_Domain, _HostType, _Locked = true}] ->
+            true;
+        _ ->
+            false
+    end.
+
+remove_all_unlocked() ->
```
DenysGonchar, line 60: What is the purpose of such interface? There must be per domain operations, group removal should not be possible. Otherwise, it's impossible to implement domain add/remove hooks.

src/domain/mongoose_domain_core.erl
```diff
@@ -0,0 +1,139 @@
...
+get_all_locked() ->
+    heads(ets:match(?TABLE, {'$1', '_', true})).
+
+get_domains_by_host_type(HostType) when is_binary(HostType) ->
+    heads(ets:match(?TABLE, {'$1', HostType, '_'})).
+
+heads(List) ->
+    [H || [H|_] <- List].
+
+upsert_unlocked(Domain, HostType) ->
```
DenysGonchar, line 72: There should be no support for upsert, we don't want to change host_type/domain pair ever. We want only to add or remove pairs. Nothing else.

src/domain/mongoose_domain_core.erl
```diff
@@ -0,0 +1,139 @@
...
+    case is_locked(Domain) of
+        true ->
+            %% Ignore any locked domains
+            ?LOG_ERROR(#{what => domain_locked_but_in_db, domain => Domain});
+        false ->
+            ets:insert(?TABLE, new_object(Domain, HostType, false))
+    end,
+    ok.
+
+remove_unlocked(Domain) ->
```
DenysGonchar, line 82: Remove should accept domain/host_type pair to avoid any race conditions.

src/domain/mongoose_domain_api.erl
```diff
@@ -0,0 +1,130 @@
...
+%% must be added into persistent storage and distributed across all the nodes
+%% in the cluster.
+%% Domain and HostType should be nameprepped using `jid:nameprep'
+-spec insert_domain(domain(), host_type()) ->
+    ok  | {error, duplicate} | {error, {db_error, term()}}
+    | {error, service_disabled}.
+insert_domain(Domain, HostType) ->
+    case mongoose_domain_core:is_locked(Domain) of
+        true ->
+            {error, duplicate};
```
DenysGonchar, line 38: all other places return `{error, locked}` error in this case, let's be consistent and return the same here.

src/domain/mongoose_domain_api.erl
```diff
@@ -0,0 +1,130 @@
...
+%% Domain should be nameprepped using `jid:nameprep'
+-spec get_host_type(domain()) ->
+    {ok, host_type()} | {error, not_found}.
+get_host_type(Domain) ->
+    mongoose_domain_core:get_host_type(Domain).
+
+%% Get the list of the host_types provided during initialisation
+%% This has complexity N, where N is the number of online domains.
+-spec get_all_locked() -> [domain()].
+get_all_locked() ->
```
DenysGonchar, line 123: not sure if we need this interface at all, but if we do then why not to return host_types as well?

src/domain/mongoose_domain_core.erl
arcusfelis: It is the simplest way to recover if gen_server of a DB service crashes.

src/domain/mongoose_domain_core.erl
arcusfelis: > ETS must be read only for public access

Moving logic inside a core gen_server could crash the server, which would get ETS table lost. Less logic - less ways to fail. It also would make debugging and testing harder.

src/domain/mongoose_domain_core.erl
arcusfelis: unsert is just ets:insert. The preconditions are checked by the caller (i.e. db server). Could be replaced with ets:insert_new though.

src/domain/mongoose_domain_core.erl
arcusfelis: And what is the behaviour in the case of race condition detected?
DB does not know, what what the host_type of the removed domain.

src/domain/mongoose_domain_api.erl
```diff
@@ -0,0 +1,130 @@
...
+
+-type domain() :: jid:lserver().
+-type host_type() :: jid:lserver().
+-type pair() :: {domain(), host_type()}.
+
+%% Init - on init all the “enabled” domain names from the persistent storage
+%% must be added to the core MIM component described above.
+%% Domains should be nameprepped using `jid:nameprep'
+-spec init([pair()]) -> ok | {error, term()}.
+init(Pairs) ->
```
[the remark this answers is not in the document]
arcusfelis: > Also host types are just strings. So every correct domain name is a correct host type identifier, but not every host type is a correct domain name

so, it would not make it interchangeable with any code, which does nameprepping for the `Host` argument. Or requires host-type in the specification (hook api-s, pool api-s).

src/domain/mongoose_domain_api.erl
```diff
@@ -0,0 +1,130 @@
...
+%% Removal the domain name - This function must be idempotent.
+%% domain name must be removed from the core MIM component (if required)
+%% and from the DB. this action must be distributed across
+%% all the nodes in the cluster.
+%% Returns ok, if domain not found.
+%% Domain should be nameprepped using `jid:nameprep'.
+-spec remove_domain(domain()) ->
+    ok | {error, locked} | {error, {db_error, term()}}
+    | {error, service_disabled}.
+remove_domain(Domain) ->
```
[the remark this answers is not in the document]
arcusfelis: ok, but only for the api module. DB could check for the host_type and cancel the call. Though it looks like very artificial limitation. Why would we add an extra argument for remove_domain, but not require it for enable/disable?

Core module removal has different logic. DB knows that what it wants to remove from the core is what has to be done. Basically core pulls data from the DB. So, there is no record in DB for the domain and the domain is unlocked - the entry is removed from the ETS table without any if.

src/domain/mongoose_domain_api.erl
arcusfelis: it was in the spec...

src/domain/mongoose_domain_core.erl
DenysGonchar: must be reworked, this is not acceptable. ETS must be a source of truth for other components, it cannot spontaneously lost all the domains.

src/domain/mongoose_domain_core.erl
DenysGonchar: This solution is not acceptable, there won't be too much code for gen_server and that is very easy to test.
Later on we will have to introduce hooks for adding and removing domains (executed on each and every node). Current implementation doesn't allow to do this.

src/domain/mongoose_domain_core.erl
DenysGonchar: this must not be possible, and it must be clear from the code of this exact module!

src/domain/mongoose_domain_core.erl
DenysGonchar: report error! do not remove domain!

src/domain/mongoose_domain_core.erl
arcusfelis: ok, fine. I will make it slower and with more code.

src/domain/mongoose_domain_api.erl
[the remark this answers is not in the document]
DenysGonchar: the host types automatically created for hosts declaration in config file will be interchangeable.

but the final goal is to ensure that MIM supports any string id for the host_type. that is one on the sanity checks.

src/domain/mongoose_domain_core.erl
arcusfelis: Again.
DB has information about the fact that a domain is removed. 
If a domain is removed from DB, DB has no information about the domain type. It's `null`.

DenysGonchar:
Just a general comment. there is one big acceptance criteria for this sub-sytem - it must be possible to add some domain related hooks in the future:
- hooks on domains insertion/removal in the ETS table. this might be required for domain enabling/disabling logic, these hooks should be executed on each and every node in the cluster.
- hooks on domains insertion/removal in the DB. These hooks should be executed only once on one of the nodes in the cluster:
   - DB insertion hook must be triggered before domain insertion in the ETS tables on any node in the cluster (with the only possible exception when the new node is added to the cluster right in the middle of the domain insertion process). 
   - DB removal hook must be triggered only after domain removal from the ETS tables on all the nodes in the cluster.
   - This hooks can be used for adding persistent data required for the new domain (if any) and removal of all the persistent data associated with the domain

src/domain/mongoose_domain_core.erl
DenysGonchar: ok, look. we have domain/host_type pair as input parameter, the behaviour of the function should  be like this:

- there is such domain/host_type pair and we can remove it (it's not configured on init) - we remove it and return success
- there is such domain/host_type pair and we can not remove it (it is configured on init) - we do nothing and return a corresponding error.
- there is such domain but host_type is not matching - we do nothing and return a corresponding error. (MB different errors for domains configured dynamically and on init)
- there's no such domain - we do nothing and return success

src/domain/mongoose_domain_api.erl
DenysGonchar: let's return domains' host types as well.

src/domain/mongoose_domain_api.erl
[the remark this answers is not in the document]
DenysGonchar: please add it here as well.

src/domain/mongoose_domain_core.erl
DenysGonchar: as for the hooks, we may want to do execution asynchronously to prevent blocking operations on the ets. We can design it later (if required at all), for now we just need to ensure that there's an option to implement it.

src/domain/mongoose_domain_core.erl
arcusfelis: ok, it would require rework of DB layer/schema/tests.
Once domain is removed, we would not be able to provide hosttype from db.

DenysGonchar: I'm not sure i understand how the core module is related to DB storage.

arcusfelis:
@DenysGonchar Please, resolve the outdated conversations.

Diffing db and core tables would require extra code and tests.
Storing info about which domain has been removed would require queries and schema rework. I don't really see a clear algorithm for sync anymore at this point.

src/domain/mongoose_domain_api.erl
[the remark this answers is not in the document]
DenysGonchar: After in person discussion I understand better the problem highlighted by @arcusfelis in the previous comment:
```
Event table doesn't store the kind of changes but only the fact that some domain has changed.
it's joined with the settings table (LEFT JOIN on domain field) and if settings table has no 
domain record anymore then returned host type field is "null". 
 ```
 
of course, we can check the host type from the ETS table, but it doesn't really make any sense to get the host type to just remove it.  So as agreed, we will have domain/host_type pair input parameter for `mongoose_domain_api` interface only, `mongoose_domain_core` will use only domain name for removal. anyway `mongoose_domain_core` module must be used only by the domains service modules.

@arcusfelis thanks for the good and very detailed explanation.

File: src/domain/mongoose_domain_core.erl (hunk 7 above)
Comment (DenysGonchar) [reply src/domain/mongoose_domain_core.erl:84]: After in person discussion I understand better the problem highlighted by @arcusfelis in the previous comment:
```
Event table doesn't store the kind of changes but only the fact that some domain has changed.
it's joined with the settings table (LEFT JOIN on domain field) and if settings table has no 
domain record anymore then returned host type field is "null". 
 ```
 
of course, we can check the host type from the ETS table, but it doesn't really make any sense to get the host type to just remove it.  So as agreed, we will have domain/host_type pair input parameter for `mongoose_domain_api` interface only, `mongoose_domain_core` will use only domain name for removal. anyway `mongoose_domain_core` module must be used only by the domains service modules.

@arcusfelis thanks for the good and very detailed explanation.

src/domain/mongoose_domain_core.erl
DenysGonchar: putting here some notes after today's discussion:

there are 2 different cases when domains DB service can potentially crash:
- During the MIM startup. before the initial filling of the ETS table (from DB) is done.
- after initial ETS filling.

for the first case it's acceptable to completely drop the core's ETS table, because we cannot use events DB table to perform an update of the ETS table with the latest changes.

for the second case, we must use events DB table.

implementation notes:
- core's gen_server must store the last applied event id.
- at the beginning, before the initial ETS filling, this event id is undefined.
- once initial ETS filling is done, event id is set for the first time.
- after that event id is updated with each and every change of the ETS.
- it must be possible to drop ETS data only if event id is not set yet. 
- after initialization of the event id, DB service (on restart) must use that stored event id to apply the latest changes to the ETS.
- also we may want to extend the cleaning interval and max age parameters for the events DB table to some higher value, e.g. 1 day

@arcusfelis please add some implementation note in the source code.

DenysGonchar:
> @DenysGonchar Please, resolve the outdated conversations.
> 
> Diffing db and core tables would require extra code and tests.
> Storing info about which domain has been removed would require queries and schema rework. I don't really see a clear algorithm for sync anymore at this point.

feel free to resolve anything that is not relevant any more by yourself.

src/domain/mongoose_domain_sql.erl
```diff
@@ -0,0 +1,232 @@
...
+    selected_to_int(execute_successfully(Pool, domain_events_max, [])).
+
+remove_events_older_than(Id) ->
+    transaction(fun(Pool) ->
+            MaxId = get_max_event_id(),
+            if MaxId =:= 0 ->
+                    skipped;
+               MaxId < Id ->
+                    %% Removal would erase all the events, which we don't want.
+                    %% We want to keep at least one event.
```
chrzaszcz, line 159: Why do we want to keep at least one event? At start there is no event and the server is doing fine.

src/domain/mongoose_domain_api.erl
```diff
@@ -0,0 +1,136 @@
...
+%% Change of the status must be distributed across all the nodes in the cluster.
+-spec disable_domain(domain()) ->
+    ok | {error, not_found} | {error, locked} | {error, duplicate}
+    | {error, service_disabled} | {error, unknown_host_type}.
+disable_domain(Domain) ->
+    case mongoose_domain_core:is_locked(Domain) of
+        true ->
+            {error, locked};
+        false ->
+            case service_domain_db:enabled() of
```
chrzaszcz, line 74: This check seems prone to race conditions, why not just check if the service is enabled in the config?

src/domain/service_domain_db.erl
```diff
@@ -0,0 +1,141 @@
...
+                                                        validate = positive},
+               <<"event_max_age">> => #option{type = integer,
+                                              validate = positive}
+              }}.
+
+start_link() ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+enabled() ->
+    is_pid(whereis(?MODULE)).
```
chrzaszcz, line 52: This check seems prone to race conditions, why not just check if the service is enabled in the config?

chrzaszcz (commented):
lLots of good work! I have some comments, but with such a big change it is inevitable.

src/domain/mongoose_domain_sql.erl
DenysGonchar: We need it to get the value of the automatic RDBMS counter. We assume that it's 0 if the table has no records. and on the very first start (before any event is added) it's indeed 0. but it's not 0 if anything has been ever added.

chrzaszcz: But... why? If there are no records, then you don't have anything to remove, so the cleaner will just do nothing when the timeout occurs.

DenysGonchar: how to understand then that we were disconnected for too long, and we cannot make an update using events table, e.g. we missed the events that were necessary to make an update.

arcusfelis: That is exactly right, we need one record, because there are some nodes, that keep LastEventId in the service gen_server. They would need to verify that their LastEventId is bigger than the minimum event id in the table (i.e. they are still allowed to sync from that table). If there is no records in the table, such check is impossible. If such check fails, the gen_server would be stopped (**TODO: node should be killed instead**).

src/domain/service_domain_db.erl
```diff
@@ -0,0 +1,141 @@
...
+%% Client code
+
+start(Opts) ->
+    mongoose_domain_sql:start(Opts),
+    mongoose_domain_db_cleaner:start(Opts),
+    ChildSpec =
+        {?MODULE,
+         {?MODULE, start_link, []},
+         permanent, infinity, worker, [?MODULE]},
+    supervisor:start_child(ejabberd_sup, ChildSpec),
```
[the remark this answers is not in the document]
arcusfelis: it could be.

src/domain/mongoose_domain_core.erl
```diff
@@ -0,0 +1,196 @@
+%% Generally, you should not call anything from this module.
+%% Use mongoose_domain_api module instead.
+-module(mongoose_domain_core).
+-include("mongoose_logger.hrl").
+
+-export([start/2, stop/0]).
+-export([start_link/2]).
+-export([get_host_type/1]).
+-export([is_locked/1]).
```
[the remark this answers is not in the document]
arcusfelis: Fixed means that the rest are broken :D 
Could be static though.

big_tests/tests/service_domain_db_SUITE.erl
```diff
@@ -0,0 +1,501 @@
...
+
+end_per_testcase(_TestcaseName, Config) ->
+    Config.
+
+%%--------------------------------------------------------------------
+%% Tests
+%%--------------------------------------------------------------------
+
+core_lookup_works(_) ->
+    precond(mim(), off, [<<"example.com">>, <<"type1">>], [<<"type1">>]),
```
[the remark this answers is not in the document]
arcusfelis: It's pretty hard to mentally keep which were parameters for each test.
I.e. which are the initial domains.

The only reason to put it into init_per_testcase is to be more CT friendly. Though the side effect is that we would need to check two places in code to understand each testcase. And if we want to compare two tests - we need to check four places in the code.

Though I probably can init the module once with "example.cfg". And if I need to test for non existing domain - I can use "example.missing".

src/domain/mongoose_domain_core.erl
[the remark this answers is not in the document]
chrzaszcz: I just quoted the meaning of "fixed" from a dictionary. "Locked" definitely means "can't access it". "Static" is OK for me.

src/domain/service_domain_db.erl
```diff
@@ -0,0 +1,141 @@
...
+    TRef = erlang:send_after(Interval, self(), check_for_updates),
+    State#{last_event_id => LastEventId2, check_for_updates => TRef}.
+
+maybe_cancel_timer(#{check_for_updates_tref := TRef}) ->
+    erlang:cancel_timer(TRef);
+maybe_cancel_timer(_) ->
+    ok.
+
+receive_all_check_for_updates() ->
+    receive check_for_updates -> ok after 0 -> ok end.
```
chrzaszcz, line 136: Wasn't this supposed to receive all `check_for_updates` messages? It receives just one.

src/domain/mongoose_domain_sql.erl
chrzaszcz: Ok, so there is no way to discover the issue before first insert. My guess would be to solve this by failing fast - on the first failed query to read updates. This is the time when we could ensure it loads the whole table next time.

src/domain/mongoose_domain_api.erl
```diff
@@ -0,0 +1,136 @@
...
+%% Domain should be nameprepped using `jid:nameprep'
+-spec get_host_type(domain()) ->
+    {ok, host_type()} | {error, not_found}.
+get_host_type(Domain) ->
+    mongoose_domain_core:get_host_type(Domain).
+
+%% Get the list of the host_types provided during initialisation
+%% This has complexity N, where N is the number of online domains.
+-spec get_all_locked() -> [{domain(), host_type()}].
+get_all_locked() ->
```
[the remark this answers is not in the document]
arcusfelis: was required in spec

src/domain/mongoose_domain_api.erl
```diff
@@ -0,0 +1,136 @@
...
+-type domain() :: jid:lserver().
+-type host_type() :: binary().
+-type pair() :: {domain(), host_type()}.
+
+%% Init - on init all the “enabled” domain names from the persistent storage
+%% must be added to the core MIM component described above.
+%% Domains should be nameprepped using `jid:nameprep'
+-spec init([pair()], [host_type()]) -> ok | {error, term()}.
+init(Pairs, AllowedHostTypes) ->
+    mongoose_domain_core:start(Pairs, AllowedHostTypes).
```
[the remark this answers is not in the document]
arcusfelis: So we can test it.

src/domain/mongoose_domain_api.erl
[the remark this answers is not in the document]
DenysGonchar: it's required more for troubleshooting than for a real use cases.

doc/modules/service_domain_db.md
```diff
@@ -0,0 +1,128 @@
+# MongooseIM core component
+
+Implemented by `mongoose_domain_core` module.
+
+It must be based on gen\_server & ETS table w. public read access.
```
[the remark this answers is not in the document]
arcusfelis: moved to doc/developers-guide/

big_tests/tests/service_domain_db_SUITE.erl
```diff
@@ -0,0 +1,501 @@
...
+    precond(mim2(), on, [], [<<"dbgroup">>]),
+    %% Service is disable on the second node
+    service_disabled(mim2()),
+    ok = insert_domain(mim(), <<"example.com">>, <<"dbgroup">>),
+    %% Sync is working again
+    service_enabled(mim2()),
+    {ok, <<"dbgroup">>} = get_host_type(mim2(), <<"example.com">>),
+    ok.
+
+db_reinserted_from_one_node_while_service_disabled_on_another(_) ->
```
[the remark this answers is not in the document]
arcusfelis: yea, but if there is some static data in DB tied to the host-type, it would not be migrated.
It's also the behaviour **from the requirements**, I've just added test to document it too.

big_tests/tests/service_domain_db_SUITE.erl
```diff
@@ -0,0 +1,501 @@
...
+    rpc(mim(), mongoose_domain_core, is_locked, [Domain]).
+
+%% Call sync before get_host_type, if there are some async changes expected
+sync() ->
+    rpc(mim(), service_domain_db, sync, []).
+
+precond(Node, off, FlatPairs, AllowedHostTypes) ->
+    service_disabled(Node),
+    erase_database(Node),
+    init_with(Node, unflat(FlatPairs), AllowedHostTypes);
```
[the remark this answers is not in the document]
arcusfelis: more readable.

src/domain/mongoose_domain_api.erl
arcusfelis: changed

src/domain/mongoose_domain_core.erl
```diff
@@ -0,0 +1,196 @@
...
+            true;
+        _ ->
+            false
+    end.
+
+is_host_type_allowed(HostType) ->
+    ets:member(?HOST_TYPE_TABLE, HostType).
+
+get_all_locked() ->
+    pairs(ets:match(?TABLE, {'$1', '$2', true})).
```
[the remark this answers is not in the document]
arcusfelis: was in spec.
not optimized, because I don't see any code calling it.

src/domain/mongoose_domain_db_cleaner.erl
```diff
@@ -0,0 +1,89 @@
...
+%% ---------------------------------------------------------------------------
+%% Server callbacks
+
+init(Opts) ->
+    Interval = proplists:get_value(event_cleaning_interval, Opts, default_cleaning_interval()),
+    MaxAge = proplists:get_value(event_max_age, Opts, default_max_age()),
+    ?LOG_INFO(#{what => domain_cleaner_start, cleaning_interval => Interval, max_age => MaxAge}),
+    State = #{max_age => MaxAge},
+    self() ! schedule_removal,
+    timer:send_interval(timer:seconds(Interval), schedule_removal),
```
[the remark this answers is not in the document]
arcusfelis: added

big_tests/tests/service_domain_db_SUITE.erl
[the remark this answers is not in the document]
chrzaszcz: For me it is not - harder to say what is the key and what is the value, easier to make a mistake, but keep it if this helps you.

src/domain/mongoose_domain_loader.erl
```diff
@@ -0,0 +1,82 @@
...
+    end.
+
+apply_changes([Row|Rows]) ->
+    apply_change(Row),
+    apply_changes(Rows);
+apply_changes([]) ->
+    ok.
+
+apply_change({_Id, Domain, null}) ->
+    %% Missing record in domain_settings table.
```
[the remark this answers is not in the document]
arcusfelis: we expecting it. Because we've just joined onto it.

src/domain/mongoose_domain_loader.erl
```diff
@@ -0,0 +1,82 @@
...
+    mongoose_domain_core:remove_unlocked(Domain);
+apply_change({_Id, Domain, HostType}) ->
+    %% Inserted or updated record
+    mongoose_domain_core:insert_unlocked(Domain, HostType).
+
+insert_rows_to_core([Row|Rows]) ->
+    insert_row_to_core(Row),
+    insert_rows_to_core(Rows);
+insert_rows_to_core([]) ->
+    ok.
```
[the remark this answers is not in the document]
arcusfelis: It is used to be slower.

src/domain/service_domain_db.erl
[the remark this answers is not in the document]
arcusfelis: Cleaner does nothing most of the period of time, so would never trigger intensity limit.

Fail in the initial init of the service should crash everything anyway.

The only other thing the service does is calling mongoose_domain_loader:check_for_updates, which has `try...catch all` in it, so not likely to fail.

src/domain/mongoose_domain_loader.erl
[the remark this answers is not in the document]
chrzaszcz: I think it is a book example of premature optimization. But if you are concerned about performance, do this:

`[insert_rows_to_core(Row) || Row <- Rows]`

On my machine it's the fastest one and it easier to read.

src/domain/service_domain_db.erl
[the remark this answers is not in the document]
chrzaszcz: I understand, but I really think that if the supervision tree has a clean hierarchy, it is easier to reason about it. A lat tree makes it more difficult to figure out where a process comes from. But it can stay as it is now, I am just using tools like Observer from time to time and it is really nice when supervision tree reflects the structure of the application.

src/ejabberd_app.erl
```diff
@@ -58,6 +58,8 @@ start(normal, _Args) ->
     connect_nodes(),
     mongoose_deprecations:start(),
     {ok, _} = Sup = ejabberd_sup:start_link(),
+    %% TODO provide Pairs and AllowedHostTypes arguments
+    mongoose_domain_api:init([], []),
```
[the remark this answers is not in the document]
chrzaszcz: After discussion with @DenysGonchar we agreed it will be done later.

src/domain/mongoose_domain_sql.erl
arcusfelis: There is no way to discover the issue before first insert because there is no data in DB, no inserts were ever made.
And no changes were ever applied in the cluster.

Though, if there would be a node applying events normally, and someone erased event table manually (so a corner case), such node would still work, until first event appears. After that it crashes.

src/domain/mongoose_domain_utils.erl
```diff
@@ -0,0 +1,6 @@
+-module(mongoose_domain_utils).
+-export([halt_node/1]).
+
+halt_node(ExitText) ->
```
DenysGonchar, line 4: Just a side note, no need to change anything:

we should think about one common MIM helper module for such functionality

DenysGonchar (approved):
just a side note, no need to change anything in this PR:

in the future we should split service_domain_db_SUITE.erl into 2 separate suites:
- one for the domains core module testing (MB even make it a unit test)
- another for service testing

big_tests/tests/service_domain_db_SUITE.erl
```diff
@@ -0,0 +1,515 @@
...
+db_cases() -> [
+     db_inserted_domain_is_in_db,
+     db_inserted_domain_is_in_core,
+     db_deleted_domain_from_db,
+     db_deleted_domain_fails_with_wrong_host_type,
+     db_deleted_domain_from_core,
+     db_disabled_domain_is_in_db,
+     db_disabled_domain_not_in_core,
+     db_reanabled_domain_is_in_db,
+     db_reanabled_domain_is_in_core,
```
chrzaszcz, line 37: Minor typo: `reenabled`

src/domain/mongoose_domain_api.erl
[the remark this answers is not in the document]
chrzaszcz: Agreed with @DenysGonchar that we will do it later.

src/domain/mongoose_domain_api.erl
```diff
@@ -0,0 +1,118 @@
...
+get_host_type(Domain) ->
+    mongoose_domain_core:get_host_type(Domain).
+
+%% Get the list of the host_types provided during initialisation
+%% This has complexity N, where N is the number of online domains.
+-spec get_all_static() -> [{domain(), host_type()}].
+get_all_static() ->
+    mongoose_domain_core:get_all_static().
+
+%% Get the list of the host\_types provided during initialisation
```
chrzaszcz, line 99: Minor:

```suggestion
%% Get the list of the host_types provided during initialisation
```

src/domain/mongoose_domain_loader.erl
```diff
@@ -0,0 +1,90 @@
...
+    case mongoose_domain_sql:select_from(FromId, PageSize) of
+        [] -> ok;
+        Rows ->
+            PageMaxId = row_to_id(lists:last(Rows)),
+            insert_rows_to_core(Rows),
+            load_data_from_base_loop(PageMaxId, PageSize)
+    end.
+
+check_for_updates(FromId, PageSize) ->
+    check_if_from_num_still_relevant(FromId),
```
chrzaszcz, line 31: Minor:
```suggestion
    check_if_from_id_still_relevant(FromId),
```

src/domain/mongoose_domain_loader.erl
```diff
@@ -0,0 +1,90 @@
...
+    %% Ordered by the earlist events first
```
chrzaszcz, line 32: Minor:
```suggestion
    %% Ordered by the earliest events first
```

src/domain/mongoose_domain_loader.erl
```diff
@@ -0,0 +1,90 @@
...
+            ok;
+       Min > FromId ->
+           %% Looks like this node has no DB connection for a long time.
+           %% But the event log in the DB has been truncated by some other node
+           %% meanwhile. We have to load the whole set of data from DB.
+           Text = <<"DB domain log had some updates to domains deleted, "
+                    " which we have not applied yet. Have to crash.">>,
+           ?LOG_CRITICAL(#{what => events_log_out_of_sync,
+                           text => Text, min_db => Min, from_id => FromId}),
+           mongoose_domain_utils:halt_node(Text);
```
chrzaszcz, line 65: Why not just resync the DB and continue working?

src/domain/mongoose_domain_loader.erl
```diff
@@ -0,0 +1,90 @@
...
+            ?LOG_ERROR(#{what => domain_check_for_updates_failed,
+                         class => Class, reason => Reason,
+                         stacktrace => StackTrace}),
+            FromId
+    end.
+
+%% Be aware that for this check to work, the cleaner should keep at least
+%% one record in domain_events table.
+check_if_from_num_still_relevant(0) ->
+    ok;
```
chrzaszcz, line 51: This returns `ok` but what if some records were added in the meantime (while there was no connection to the DB) and then cleanup happened? I think that we should still check that `Min = 1` or something like that to make sure no cleanup was done.

chrzaszcz (approved):
Minor comments that should be addressed in subsequent PR's.

src/domain/mongoose_domain_sql.erl
chrzaszcz: I still think we are not on the same page here, I added a new comment, closing this one.

src/domain/mongoose_domain_loader.erl
arcusfelis: We need to unload hosts and load them again. Too much code for something rarely happening.

chrzaszcz: Ok, let's keep it as it is.

PR #3053, opened 2021-03-08
Multi tenancy config

DenysGonchar:
This PR addresses #

- [x] parsing of the host_type configuration
- [x] init modules per host type 
- [x] Extension of gen_mod behaviour with extra callback to define modules capabilities. 
- [x] crashing if configured modules don’t support multi-tenancy.

see ` test/config_parser_SUITE_data/host_types.toml` file for example of the new config format.


test/config_parser_SUITE_data/host_types.toml
```diff
@@ -0,0 +1,25 @@
...
+[[host_config]]
+## host is just synonym for host_type now
+host = "some host type"
+## another syntax for reseting the modules for this host type
+[host_config.modules]
+
+[[host_config]]
+host_type = "another host type"
+## reseting the modules for this host type
+[host_config.modules.another_dummy_module]
```
chrzaszcz, line 25: Missing newline

test/config_parser_SUITE.erl
```diff
@@ -266,7 +274,26 @@ hosts(_Config) ->
...
-    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => #{}})). % hosts are mandatory
+    % either hosts or host_types must be provided
+    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => #{}})).
+
+host_types(_Config) ->
+    %% ./rebar3 ct --suite config_parser_SUITE --group general --case host_types
+    ?eq([#config{key = host_types, value = [<<"type1">>, <<"type2">>]}],
+        parse(#{<<"general">> => #{<<"host_types">> => [<<"type1">>, <<"type2">>]}})),
+    ?eq([#config{key = host_types, value = [<<"what is this?">>]}],
+        parse(#{<<"general">> => #{<<"host_types">> => [<<"what is this?">>]}})),
```
chrzaszcz, line 285: Maybe it's just me, but "what is this" looks like a wrong value, could you replace it with something like "my host type"?

Side note: I am not sure what would happen if we allow spaces here, they might be processed in the code as well-formed domain names, e.g. in places where process names are constructed from them.

test/config_parser_SUITE.erl
```diff
@@ -266,7 +274,26 @@ hosts(_Config) ->
     ?err(parse(#{<<"general">> => #{<<"hosts">> => [<<>>]}})),
     ?err(parse(#{<<"general">> => #{<<"hosts">> => []}})),
     ?err(parse(#{<<"general">> => #{<<"hosts">> => [<<"host1">>, <<"host1">>]}})),
-    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => #{}})). % hosts are mandatory
+    % either hosts or host_types must be provided
+    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => #{}})).
+
+host_types(_Config) ->
+    %% ./rebar3 ct --suite config_parser_SUITE --group general --case host_types
+    ?eq([#config{key = host_types, value = [<<"type1">>, <<"type2">>]}],
```
chrzaszcz, line 282: There is a test case that would fail in the current code:

`parse(#{<<"general">> => #{<<"host_types">> => [<<"my host type">>], <<"hosts">> => []}})`

Could you add it?

src/config/mongoose_config_parser_toml.erl
```diff
@@ -306,12 +309,22 @@ get_hosts(Config) ->
...
 
--spec build_state([jid:server()], [top_level_config()], [override()]) ->
+-spec get_host_types(config_list()) -> [jid:server()].
+get_host_types(Config) ->
+    case lists:filter(fun(#config{key = host_types}) -> true;
+                         (_) -> false
+                      end, Config) of
+        [] -> [];
+        [#config{value = HostTypes}] -> HostTypes
+    end.
```
chrzaszcz, line 319: Minor: code duplication, the previous function is very similar. Not mandatory to fix.

src/config/mongoose_config_parser_toml.erl
```diff
@@ -61,15 +61,17 @@ parse_file(FileName) ->
...
     Hosts = get_hosts(Config),
+    HostTypes = get_host_types(Config),
     {FOpts, Config1} = lists:partition(fun(Opt) -> is_function(Opt, 1) end, Config),
     {Overrides, Opts} = lists:partition(fun({override, _}) -> true;
                                            (_) -> false
                                         end, Config1),
-    HOpts = lists:flatmap(fun(F) -> lists:flatmap(F, Hosts) end, FOpts),
-    AllOpts = Opts ++ HOpts,
+    HostsOpts = lists:flatmap(fun(F) -> lists:flatmap(F, Hosts ++ HostTypes) end, FOpts),
+    HostTypesOpts = lists:flatmap(fun(F) -> lists:flatmap(F, HostTypes) end, FOpts),
```
chrzaszcz, line 70: This looks redundant to me - it just duplicates some of the options from the previous line. `dedup_state_opts` removes them later.

src/config/mongoose_config_spec.erl
```diff
@@ -1078,6 +1094,25 @@ process_host(Host) ->
...
+    AllHostTypes = get_all_hosts_and_host_types(General),
+    true = lists:sort(AllHostTypes) =:= lists:usort(AllHostTypes),
+    true = [] =/= AllHostTypes.
+
+get_all_hosts_and_host_types(General) ->
+    lists:foldl(fun
+                    ({config, hosts, Hosts}, Acc) ->
+                        Acc ++ Hosts;
+                    ({config, host_types, HostTypes}, Acc) ->
+                        Acc ++ HostTypes;
```
chrzaszcz, line 1111: I think adding an include is better than this because record format is an implementation detail which can change.

chrzaszcz (commented):
Looking good, a few comments from me.

src/config/mongoose_config_spec.erl
```diff
@@ -171,6 +183,10 @@ general() ->
                                                       process = fun ?MODULE:process_host/1},
                                       validate = unique_non_empty,
```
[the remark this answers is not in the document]
DenysGonchar: I think this doesn't make hosts mandatory. what made hosts mandatory is [this line](https://github.com/esl/MongooseIM/pull/3053/files#diff-e0e1d2948b30a346140efe9c7750bef22c779f6089a5f165e1385e05813bd374L119)

and `unique_non_empty` prohibits to declare `hosts` as an empty list, if it's declared at all.

test/config_parser_SUITE.erl
```diff
@@ -266,7 +274,26 @@ hosts(_Config) ->
...
+    ?eq([#config{key = host_types, value = [<<"what is this?">>]}],
+        parse(#{<<"general">> => #{<<"host_types">> => [<<"what is this?">>]}})),
+    ?err(parse(#{<<"general">> => #{<<"host_types">> => [<<>>]}})),
+    ?err(parse(#{<<"general">> => #{<<"host_types">> => []}})),
+    ?err(parse(#{<<"general">> => #{<<"host_types">> => [<<"type1">>, <<"type1">>]}})),
+    % either hosts and host_types cannot have the same values
+    io:format("~n!!! non-unique host types error: ~p~n",
+              [mongoose_config_parser_toml:parse(#{<<"general">> => #{
+                  <<"host_types">> => [<<"type1">>],
+                  <<"hosts">> => [<<"type1">>]}})]),
```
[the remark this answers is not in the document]
DenysGonchar: 👍 just not removed yet. will get rid of it in the next commit.

src/config/mongoose_config_spec.erl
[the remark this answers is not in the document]
chrzaszcz: OK, let me rephrase this:

This line does **not** make `hosts` mandatory. It ensures that it is not declared as `hosts = []`. This causes two issues:

1. The config with `hosts = []` is equivalent to the one without the `hosts` option. Hence there is no reason to disallow `hosts = []`.
2. As `host_types = []` is valid, I see no reason for `hosts = []` being invalid.

test/config_parser_SUITE.erl
```diff
@@ -266,7 +274,26 @@ hosts(_Config) ->
     ?err(parse(#{<<"general">> => #{<<"hosts">> => [<<>>]}})),
     ?err(parse(#{<<"general">> => #{<<"hosts">> => []}})),
     ?err(parse(#{<<"general">> => #{<<"hosts">> => [<<"host1">>, <<"host1">>]}})),
-    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => #{}})). % hosts are mandatory
+    % either hosts or host_types must be provided
+    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => #{}})).
+
+host_types(_Config) ->
+    %% ./rebar3 ct --suite config_parser_SUITE --group general --case host_types
```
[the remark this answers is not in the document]
DenysGonchar: yeah, yet another leftover. will remove it soon.

test/config_parser_SUITE.erl
DenysGonchar: we should not have this restriction for host_types. 
host_types should be never processed as domain names. so whites paces seem to be a good option for sanity check.

"what is this"  it's leftover after copy past, will change it

src/config/mongoose_config_spec.erl
DenysGonchar: You're absolutely right, I'm just getting a bit lost in this code so didn't figure out from the first look that it's actually a record

src/config/mongoose_config_parser_toml.erl
DenysGonchar: I would rather want to leave it for now as it, it simplifies the following change with modules verifications.

test/config_parser_SUITE.erl
DenysGonchar: as discussed in other thread, this won't fail any more.

src/config/mongoose_config_parser_toml.erl
chrzaszcz: I think I was not understood, see my suggestion below.

chrzaszcz, line 70:
```suggestion
    HostsOpts = lists:flatmap(fun(F) -> lists:flatmap(F, Hosts) end, FOpts),
    HostTypesOpts = lists:flatmap(fun(F) -> lists:flatmap(F, HostTypes) end, FOpts),
```

DenysGonchar: Damn, I think I removed that code at least twice, no idea how it reappeared :)

src/config/mongoose_config_parser_toml.erl
```diff
@@ -292,31 +296,76 @@ node_to_string({host, _}) -> [];
...
     lists:foldl(fun(F, StateIn) -> F(StateIn) end,
                 mongoose_config_parser:new_state(),
                 [fun(S) -> mongoose_config_parser:set_hosts(Hosts, S) end,
+                 fun(S) -> mongoose_config_parser:set_host_types(HostTypes, S) end,
                  fun(S) -> mongoose_config_parser:set_opts(Opts, S) end,
                  fun mongoose_config_parser:dedup_state_opts/1,
                  fun mongoose_config_parser:add_dep_modules/1,
                  fun(S) -> set_overrides(Overrides, S) end]).
 
+-spec check_dynamic_domains_support(mongoose_config_parser:state()) -> any().
```
DenysGonchar, line 328: not sure if `src/config/mongoose_config_parser_toml.erl` is the right place for this check, MB `src/config/mongoose_config_parser.erl` fits better.
@chrzaszcz, what's your opinion on it?

chrzaszcz: If possible, I'd rather have it in `src/config/mongoose_config_parser.erl`

src/config/mongoose_config_parser.erl
```diff
@@ -200,3 +201,58 @@ add_dep_modules_opt({local_config, {modules, Host}, Modules}) ->
...
     Other.
+
+%% local functions
+-ifdef(TEST).
+halt_with_msg(ExitMsg, Errors) ->
+    config_error(ExitMsg, Errors).
+-else.
+halt_with_msg(ExitMsg, Errors) ->
+    [?LOG_ERROR(Error) || Error <- Errors],
+    mongoose_config_utils:exit_or_halt(ExitMsg).
```
chrzaszcz, line 212: Please check it manually (if not done yet) by starting MongooseIM with some errors in the config file.

src/config/mongoose_config_parser.erl
```diff
@@ -200,3 +201,58 @@ add_dep_modules_opt({local_config, {modules, Host}, Modules}) ->
...
+    FoldFN = fun
+                 (#local_config{key = {modules, H}, value = Modules}, Acc) ->
+                     case lists:member(H, HostTypes) of
+                         false -> Acc;
+                         true ->
+                             BadModules = check_modules_for_host_type(Modules),
+                             Acc ++ invalid_modules_for_host_type(H, BadModules)
+                     end;
+                 (_, Acc) -> Acc
+             end,
```
chrzaszcz, line 233: Could you use a named function here to improve debugging?

test/config_parser_SUITE.erl
```diff
@@ -260,13 +307,34 @@ loglevel(_Config) ->
     ?err(parse_host_config(#{<<"general">> => #{<<"loglevel">> => <<"debug">>}})).
 
 hosts(_Config) ->
-    ?eq([#config{key = hosts, value = [<<"host1">>, <<"host2">>]}],
-        parse(#{<<"general">> => #{<<"hosts">> => [<<"host1">>, <<"host2">>]}})),
+    ?eq([#config{key = hosts, value = [<<"host1">>]}],
+        parse(#{<<"general">> => #{<<"hosts">> => [<<"host1">>]}})),
+    ?eq_list([#config{key = hosts, value = [<<"host1">>, <<"host2">>]},
```
chrzaszcz, line 312: Just FYI: there is a function `compare_config` that is meant for this purpose - it compares two lists of config records with using `compare_ordered_lists` and `compare_unordered_lists` when needed.

src/config/mongoose_config_spec.erl
```diff
@@ -137,11 +139,22 @@ root() ->
...
+                 %% is declared automatically. As host_config section is now
+                 %% used for changing configuration of the host_type, we don't
+                 %% need host option any more. but to stay compatible with an
+                 %% old config format we keep host option as well. now it is
+                 %% just a synonym to host_type.
                  <<"host">> => #option{type = binary,
                                        validate = non_empty,
                                        format = skip},
 
+                 <<"host_type">> => #option{type = binary,
```
chrzaszcz, line 154: **Just a thought - no need to change anything:**

There is one thought that I have to share here: if someone makes the mistake below it is difficult to tell which one is a host and which one is a host type.

```
[general]
  hosts = ["host1"]
  host_types = ["type1"]

[[host_config]]
  host = "type1"
  (...)

[[host_config]]
  host_type = "host1"
```

Maybe it would make sense to have a strict check here. it's not mandatory now of course - just a thought. If we make it mandatory some time in the future and put `{host_type, ...}` or `{host, ...}` in the path, we could avoid the whole second pass through the config done in `check_dynamic_domains_support/1`.

chrzaszcz (commented):
Looking good! Just a few minor comments.

src/config/mongoose_config_parser.erl
DenysGonchar: well, it was crashing even test node. that's the reason to change implementation conditionally.

chrzaszcz: I fully understand. What I meant was that these lines are moved so it would be good to make sure they still work. A sanity check.

src/config/mongoose_config_spec.erl
DenysGonchar: I don't want to introduce that limitation, since we declare that for each and every host there's an implicit definition of the host type of the same name as host's domain name. I see `host_config.host` as a synonym for `host_config.host_type` which is necessary only for the backward compatibility purposes.

src/config/mongoose_config_parser.erl
DenysGonchar: It must be an anonymous function because of the closure (HostTypes variable).

so we have 2 options here:
* I can either create a function accepts HostTypes as input parameter and then generates FoldN function.
* Or move current FoldN implementation inside some named function that accepts 3 parameters (HostTypes, Element, Acc) and make a call of that function from the anonymous FoldN function.

I think only the second one can improve debugging a bit, but still want to clarify this. Which one were you having in mind?

chrzaszcz: I meant the second option (replace the fun with a 3-argument function) that would also make the code structure a bit more flat.

src/config/mongoose_config_spec.erl
chrzaszcz: Yes I am fine to keep it as it is.

chrzaszcz (approved):
Looks good!

PR #3055, opened 2021-03-10
Allow to specify arguments from the multiple tables for prepared queries

arcusfelis:
This PR addresses allows to specify arguments from the multiple table. Which is needed for pubsub queries.

Proposed changes include:
* Allow to pass `table.column` as a field name into the `prepare` call
* `column` still works.

### DIFF
--- a/src/rdbms/mongoose_rdbms_odbc.erl
+++ b/src/rdbms/mongoose_rdbms_odbc.erl
@@ -169,7 +169,7 @@ tabcol_to_mapper(_ServerType, _TableDesc, {_, <<"limit">>}) ->
 tabcol_to_mapper(_ServerType, _TableDesc, {_, <<"offset">>}) ->
     fun(P) -> {sql_integer, [P]} end;
 tabcol_to_mapper(_ServerType, TableDesc, TabCol) ->
-    ODBCType = field_to_odbc_type(TabCol, TableDesc),
+    ODBCType = tabcol_to_odbc_type(TabCol, TableDesc),
     case simple_type(just_type(ODBCType)) of
         binary ->
             fun(P) -> binary_mapper(P) end;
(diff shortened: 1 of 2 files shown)


leszke (approved):

src/rdbms/mongoose_rdbms_odbc.erl
```diff
@@ -76,18 +78,20 @@ query(Connection, Query, Timeout) ->
               Fields :: [binary()], Statement :: iodata()) ->
                      {ok, {binary(), [fun((term()) -> tuple())]}}.
 prepare(Connection, Name, Table, Fields, Statement) ->
-    try prepare2(Connection, Table, Fields, Statement)
+    TabCols = fields_to_tabcol(Fields, Table),
```
chrzaszcz, line 81: Just to make sure: the only reason we don't just accept `TabCols` as an argument instead of `Fields` is that you consider the `'tab.col'` syntax better than `{tab, col}`, right?

chrzaszcz (commented):
Looks good, I have a few minor remarks.

src/rdbms/mongoose_rdbms_odbc.erl
arcusfelis: the main reason is that it's the same syntax as in SQL queries. 
We don't do "SELECT {tab, col}".

arcusfelis:
tabcol is just because "fully specified column name" is too long.

src/rdbms/mongoose_rdbms_odbc.erl
arcusfelis: Also, it's the same reason why Fields are atoms, but we convert it to binaries in the prepare call.
It just looks a bit cleaner.

PR #3056, opened 2021-03-11
Fix handover crash

gustawlippa:
This PR addresses #2933.

It contains a fix to the issue as well as a related testcase.

### DIFF
--- a/big_tests/tests/sm_SUITE.erl
+++ b/big_tests/tests/sm_SUITE.erl
@@ -988,8 +988,6 @@ resume_session_kills_old_C2S_gracefully(Config) ->
     escalus_connection:send(Alice, InitialPresence),
     Presence = escalus_connection:get_stanza(Alice, presence1),
     escalus:assert(is_presence, Presence),
-    escalus_connection:send(Alice, escalus_stanza:presence(<<"available">>)),
-    _Presence = escalus_connection:get_stanza(Alice, presence2),
     discard_vcard_update(Alice),
 
     %% Monitor the C2S process and disconnect Alice.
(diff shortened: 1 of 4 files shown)


erszcz (approved):
Looks good to me, good job!

big_tests/tests/sm_SUITE.erl
```diff
@@ -975,6 +976,45 @@ session_resumption_expects_item_not_found(Config, SMID) ->
...
+    {ok, Alice = #client{props = Props}, _} = escalus_connection:start(AliceSpec, Steps),
+    {ok, C2SPid} = get_session_pid(AliceSpec, escalus_client:resource(Alice)),
+
+    InitialPresence = setattr(escalus_stanza:presence(<<"available">>),
+                              <<"id">>, <<"presence1">>),
+    escalus_connection:send(Alice, InitialPresence),
+    Presence = escalus_connection:get_stanza(Alice, presence1),
+    escalus:assert(is_presence, Presence),
+    escalus_connection:send(Alice, escalus_stanza:presence(<<"available">>)),
+    _Presence = escalus_connection:get_stanza(Alice, presence2),
```
chrzaszcz, line 992: Could you explain the need for two presences?

chrzaszcz (commented):
Looks good, I have one small question.

big_tests/tests/sm_SUITE.erl
gustawlippa: That's a great question, I also wanted to ask it as I was looking at some other test and it didn't work with one only. I see now that forgot to change the value of `h` in `connection_steps_to_stream_resumption` from 2 to 1 and that's why I've seen failures with only one presence sent and went with 2, since it worked. However it is indeed unnecessary - it's removed in the next commit.

chrzaszcz (approved):
Great!

PR #3059, opened 2021-03-19
Use prepared queries in mod_auth_token

chrzaszcz:
The module is using prepared queries now.

Other changes:
- The insert-if-not-exists query can be performed as an upsert with empty update, so this case is added to `rdbms_queries`. It looks like a typical case that can be reused in the future.
  - For MS SQL and Postgres the following SELECT could be integrated into the query, but not for MySQL (it does not support OUTPUT or RETURNING). It would also make the upsert preparation code more complex, so this was not done.
- The code was supporting only Postgres, but prepared queries work for all databases, so it was now easy to extend the support to MySQL and MS SQL.


arcusfelis (approved):
looks good

PR #3060, opened 2021-03-22
Use prepared queries in mod_event_pusher_push

chrzaszcz:
- Introduce prepared queries
- Update the backend API to remove an unused case that would require a separate prepared query

### DIFF
--- a/big_tests/tests/push_SUITE.erl
+++ b/big_tests/tests/push_SUITE.erl
@@ -61,7 +61,8 @@ groups() ->
                                              pm_no_msg_notifications_if_user_online,
                                              pm_msg_notify_if_user_offline,
                                              pm_msg_notify_if_user_offline_with_publish_options,
-                                             pm_msg_notify_stops_after_disabling
+                                             pm_msg_notify_stops_after_disabling,
+                                             pm_msg_notify_stops_after_removal
                                             ]},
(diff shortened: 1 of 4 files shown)


big_tests/tests/push_SUITE.erl
```diff
@@ -580,6 +581,31 @@ pm_msg_notify_stops_after_disabling(Config) ->
...
+        Config, [{bob, 1}],
+        fun(Bob) ->
+            %% Enable
+            escalus:send(Bob, enable_stanza(PubsubJID, <<"NodeId">>, [])),
+            escalus:assert(is_iq_result, escalus:wait_for_stanza(Bob)),
+
+            %% Remove account - this should disable the notifications
+            escalus_connection:send(Bob, escalus_stanza:remove_account()),
+            escalus:assert(is_iq_result, escalus:wait_for_stanza(Bob)),
+            Pid = mongoose_helper:get_session_pid(Bob, distributed_helper:mim()),
```
DenysGonchar, line 596: Shouldn't we first get the session PID and only then remove the account?

DenysGonchar (changes requested):
Generally looks good, thanks for extending push_SUITE.
I just have one small comment to the new testcase

DenysGonchar (approved):
👍

PR #3061, opened 2021-03-23
Multi tenancy init

DenysGonchar:
this PR introduces the next changes:

- solves race condition at `check_for_updates/2` function
- eliminates the need to restart `mongoose_domain_core` if `service_domain_db` crashes on initial loading
- eliminates the need to halt the node if `service_domain_db` goes out of sync
- ensures proper init of the mongoose_domain_core
- introduces 2 dummy test suites for the future development.

Also now it's possible to implement "hooks" for domains adding/removing if there's such a need

### DIFF
--- a/big_tests/tests/dynamic_domains_pm_SUITE.erl
+++ b/big_tests/tests/dynamic_domains_pm_SUITE.erl
@@ -1,13 +1,4 @@
-%%%-------------------------------------------------------------------
-%%% @author denysgonchar
-%%% @copyright (C) 2021, <COMPANY>
-%%% @doc
-%%%
-%%% @end
-%%% Created : 22. Mar 2021 19:38
-%%%-------------------------------------------------------------------
 -module(dynamic_domains_pm_SUITE).
--author("denysgonchar").
 
 %% API
 -compile(export_all).
--- a/big_tests/tests/dynamic_domains_pm_SUITE.erl
+++ b/big_tests/tests/dynamic_domains_pm_SUITE.erl
@@ -1,0 +1,35 @@
+%%%-------------------------------------------------------------------
+%%% @author denysgonchar
+%%% @copyright (C) 2021, <COMPANY>
+%%% @doc
+%%%
+%%% @end
+%%% Created : 22. Mar 2021 19:38
+%%%-------------------------------------------------------------------
+-module(dynamic_domains_pm_SUITE).
+-author("denysgonchar").
+
+%% API
+-compile(export_all).
+-import(distributed_helper, [mim/0, mim2/0, require_rpc_nodes/1, rpc/4]).
+
+suite() ->
+    require_rpc_nodes([mim, mim2]).
+
+
+all() ->
+    [can_authenticate].
+
+init_per_suite(Config) ->
+    rpc(mim(), mongoose_domain_core, insert, [<<"example.com">>, <<"test type">>]),
+    rpc(mim2(), mongoose_domain_core, insert, [<<"example.com">>, <<"test type">>]),
+    Config.
+
+end_per_suite(Config) ->
+    rpc(mim(), mongoose_domain_core, delete, [<<"example.com">>]),
+    rpc(mim2(), mongoose_domain_core, delete, [<<"example.com">>]),
+    Config.
+
+can_authenticate(Config) ->
+    %% TODO: implement later
+    ok.
--- a/big_tests/tests/dynamic_domains_pm_SUITE.erl
+++ b/big_tests/tests/dynamic_domains_pm_SUITE.erl
@@ -12,8 +12,11 @@ all() ->
     [can_authenticate].
 
 init_per_suite(Config) ->
-    rpc(mim(), mongoose_domain_core, insert, [<<"example.com">>, <<"test type">>]),
-    rpc(mim2(), mongoose_domain_core, insert, [<<"example.com">>, <<"test type">>]),
+    Domain = <<"example.com">>,
+    HostType = <<"test type">>,
+    Source = dummy_source,
+    rpc(mim(), mongoose_domain_core, insert, [Domain, HostType, Source]),
+    rpc(mim2(), mongoose_domain_core, insert, [Domain, HostType, Source]),
     Config.
 
 end_per_suite(Config) ->
--- a/big_tests/tests/service_domain_db_SUITE.erl
+++ b/big_tests/tests/service_domain_db_SUITE.erl
@@ -11,14 +11,13 @@ suite() ->
 
 all() ->
     [
-     core_lookup_works,
-     core_lookup_not_found,
-     core_static_domain,
-     core_cannot_insert_static,
-     core_cannot_disable_static,
-     core_cannot_enable_static,
-     core_get_all_static,
-     core_get_domains_by_host_type,
+     api_lookup_works,
+     api_lookup_not_found,
+     api_cannot_insert_static,
+     api_cannot_disable_static,
+     api_cannot_enable_static,
+     api_get_all_static,
+     api_get_domains_by_host_type,
      {group, db}
     ].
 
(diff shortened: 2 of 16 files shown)


big_tests/tests/service_domain_db_SUITE.erl
```diff
@@ -11,14 +11,13 @@ suite() ->
...
-     core_cannot_disable_static,
-     core_cannot_enable_static,
-     core_get_all_static,
-     core_get_domains_by_host_type,
+     api_lookup_works,
+     api_lookup_not_found,
+     api_cannot_insert_static,
+     api_cannot_disable_static,
+     api_cannot_enable_static,
+     api_get_all_static,
```
arcusfelis, line 19: why to rename it?
get_all_static is function from the core module, not the api module.

src/domain/mongoose_domain_sql.erl
```diff
@@ -138,34 +143,51 @@ select_from(FromId, Limit) ->
...
+                    %% is modified externally. this is critical error!
+                    %% TODO: figure out what should we do in this case
+                    ?LOG_ERROR(#{what => select_updates_from_failed, limit => Limit,
+                                 from_id => FromId, max_id => MaxId}),
+                    []
+            end;
+        {selected, Rows} -> Rows
+    end.
 
 get_max_event_id() ->
```
arcusfelis, line 165: get_max_event_id_or_set_dummy.

`get_` something means that it does not change anything.

arcusfelis (changes requested):
added some comments

src/domain/service_domain_db.erl
```diff
@@ -78,18 +81,26 @@ sync_local() ->
 init([]) ->
     pg2:create(?GROUP),
     pg2:join(?GROUP, self()),
-    LastEventId = initial_load(mongoose_domain_core:get_last_event_id()),
-    ?LOG_INFO(#{what => domains_loaded, last_event_id => LastEventId}),
-    State = #{last_event_id => LastEventId,
-              check_for_updates_interval => 30000},
-    {ok, handle_check_for_updates(State)}.
+    gen_server:cast(self(),initial_loading),
```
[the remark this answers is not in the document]
DenysGonchar: I was thinking about unblocking initialisation of other modules/services.
Mostly like we will have to introduce some REST API check to ensure  that node is ready for adding to load balancer.

Anyway it's relatively easy to roll back that change if required later.

src/domain/service_domain_db.erl
```diff
@@ -36,6 +36,9 @@ stop() ->
     supervisor:delete_child(ejabberd_sup, ?MODULE),
     ok.
 
+reset() ->
```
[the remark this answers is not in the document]
DenysGonchar: yeah, probably `restart` name will be more self-explanatory. will rename it and add comments.

big_tests/tests/service_domain_db_SUITE.erl
DenysGonchar:
```
get_all_static(Node) ->
    rpc(Node, mongoose_domain_api, get_all_static, []).
```

src/domain/mongoose_domain_sql.erl
```diff
@@ -138,34 +143,51 @@ select_from(FromId, Limit) ->
...
+                MaxId < Id ->
                     %% Removal would erase all the events, which we don't want.
                     %% We want to keep at least one event.
-                    %% This should never happen though, unless the events table
-                    %% is modified externally (for example, after DB restored
-                    %% from the dump to the previous state).
-                    %% Skipping is not critical.
+                    %% This must never happen though, unless the events table
+                    %% is modified externally. this is critical error!
+                    %% TODO: figure out what should we do in this case
```
chrzaszcz, line 185: Actually I would just crash in this case. I don't think we need protection against someone calling `TRUNCATE TABLE` or something like that while the system is running. We can keep this code, no problem from me, we just can't try to fix all such cases.

src/domain/mongoose_domain_sql.erl
```diff
@@ -220,8 +242,8 @@ get_db_pool() ->
     hd(ejabberd_config:get_global_option(hosts)).
 
 selected_to_int({selected, [{null}]}) -> 0;
-selected_to_int({selected, [{UpdateNum}]}) ->
-    mongoose_rdbms:result_to_integer(UpdateNum).
+selected_to_int({selected, [{Num}]}) ->
+    mongoose_rdbms:result_to_integer(Num).
```
chrzaszcz, line 246: We don't need this, use `mongoose_rdbms:selected_to_integer/1` instead.

src/domain/mongoose_domain_loader.erl
```diff
@@ -27,44 +28,50 @@ load_data_from_base_loop(FromId, PageSize) ->
...
+check_if_id_is_still_relevant(FromId, Rows) ->
+    MinId = row_to_id(hd(Rows)),
+    if
+        FromId =:= MinId ->
+            tl(Rows);
+        FromId =:= MinId - 1 ->
+            %% looks like someone completely erased the events table
+            %% this should not happen, but we are still fine.
+            Rows;
+        true ->
```
chrzaszcz, line 65: Just a note: this covers both `FromId < MinId - 1` and `FromId > MinId`. I think it is ok to crash in both cases, but the error message is a bit irrelevant in the second one, which should not happen.

src/domain/mongoose_domain_loader.erl
```diff
@@ -27,44 +28,50 @@ load_data_from_base_loop(FromId, PageSize) ->
...
-    check_if_from_num_still_relevant(FromId),
-    %% Ordered by the earlist events first
+    %% Ordered by the earliest events first
     try mongoose_domain_sql:select_updates_from(FromId, PageSize) of
-        [] -> FromId;
         Rows ->
-            PageMaxId = row_to_id(lists:last(Rows)),
-            apply_changes(Rows),
-            check_for_updates(PageMaxId, PageSize)
+            case check_if_id_is_still_relevant(FromId, Rows) of
```
chrzaszcz, line 39: Why check it while paging? It's not possible to lose the connection to the DB for a long time while paging...

chrzaszcz (commented):
Good changes! Just a few comments from me.

big_tests/tests/dynamic_domains_pm_SUITE.erl
```diff
@@ -0,0 +1,35 @@
+%%%-------------------------------------------------------------------
+%%% @author denysgonchar
+%%% @copyright (C) 2021, <COMPANY>
+%%% @doc
+%%%
+%%% @end
+%%% Created : 22. Mar 2021 19:38
+%%%-------------------------------------------------------------------
```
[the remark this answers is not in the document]
DenysGonchar: sure, it's not intentionally added, just default IDE stuff.

src/domain/mongoose_domain_loader.erl
DenysGonchar: you never know when exactly another node can run table clean up

src/domain/mongoose_domain_loader.erl
DenysGonchar: Maybe it's not obvious from the code of this function, but `FromId > MinId` case is impossible because we query data (in the caller function) starting from `FromID`, see  [check_for_updates/2 ](https://github.com/esl/MongooseIM/pull/3061/files/4e5aad046d69610a4ffd351163ccbc191aee6d3f#diff-0ce59ca69aebb5afbe5caf8b74f0bca372211b36840261e527deb4864661fc7dR37) function.

src/domain/mongoose_domain_sql.erl
DenysGonchar: actually we must restart the service and make a complete reload of the table in this case, but I wrote this comment before introduction of service restart procedure.

src/domain/mongoose_domain_loader.erl
chrzaszcz: It is clear now, what if we keep `FromId < MinId - 1` instead of `true` there to make it less confusing for the reader? Just a suggestion, not a must.

src/domain/mongoose_domain_loader.erl
chrzaszcz: Ok, let's keep it then.

src/domain/mongoose_domain_sql.erl
```diff
@@ -138,34 +143,51 @@ select_from(FromId, Limit) ->
...
-    Pool = get_db_pool(),
-    selected_to_int(execute_successfully(Pool, domain_events_min, [])).
+    case execute_successfully(Pool, domain_select_events_from, Args) of
+        {selected, []} ->
+            %% get MaxID, this inserts the dummy record
+            %% in the events table if required.
+            MaxId = get_max_event_id_or_set_dummy(),
+            if
+                MaxId > FromId ->
+                    select_updates_from(FromId, Limit);
```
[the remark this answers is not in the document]
chrzaszcz: I don't want to edit comments, but I've just realized that that it might be actually possible to get **more** than just the dummy row there if case another node made an update to the table, so I edited the suggestion above.

big_tests/tests/service_domain_db_SUITE.erl
```diff
@@ -374,19 +370,20 @@ db_reinserted_from_one_node_while_service_disabled_on_another(_) ->
...
-    %% But if we delete it, it would be deleted everywhere
-    ok = delete_domain(mim(), <<"example.com">>, <<"dbgroup2">>),
+    {ok, <<"dbgroup2">>} = get_host_type(mim2(), <<"example.com">>),
+    %% check deleting
+    ok = delete_domain(mim2(), <<"example.com">>, <<"dbgroup2">>),
     sync(),
     {error, not_found} = get_host_type(mim(), <<"example.com">>),
     {error, not_found} = get_host_type(mim2(), <<"example.com">>).
 
 db_initial_load_crashes_node(_) ->
```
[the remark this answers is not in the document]
DenysGonchar: actually it's not, it's crashing on initial loading because of mocking.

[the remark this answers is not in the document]
DenysGonchar: but renaming is still required.

test/mongoose_domain_core_SUITE.erl
```diff
@@ -2,50 +2,123 @@
 
 -compile(export_all).
 
+-include_lib("eunit/include/eunit.hrl").
+
+-define(STATIC_PAIRS, [{<<"example.cfg">>, <<"type #1">>},
+                       {<<"erlang-solutions.com">>, <<"type #2">>},
+                       {<<"erlang-solutions.local">>, <<"static type">>}, %% not allowed type
+                       {<<"example.org">>, <<"type #2">>}]).
+-define(ALLOWED_TYPES, [<<"type #1">>, <<"type #2">>, <<"type #3">>]).
```
chrzaszcz, line 11: Interesting, so we are testing the cases that cannot be configured (static pairs have to be `{Host, Host}`) here. Did you want explicitly that?

chrzaszcz (commented):
I also had a quick look, it seems good!

test/mongoose_domain_core_SUITE.erl
DenysGonchar: While it always will be `{Host, Host}`, I just didn't want to have this limit to be hardcoded in the module.

PR #3063, opened 2021-03-30
Multi tenancy auth

DenysGonchar:
This PR introduces the new callback supported_features/0 for auth backends:
  * for now there is the only one optional feature - dynamic_domains.
  * currently only ejabberd_auth_dummy supports dynamic_domains feature.
  * all other auth modules must be rechecked and adopted in order to
    support this feature.
  * the following things must be done per auth module:
       - recheck that ejabberd_auth set_opts and get_opt interfaces are
         used with host type, not domain name.
       - recheck that mongoose_scram interfaces are used with host type,
         not domain name.
       - recheck that ejabberd_config get_local_option interfaces are
         used with host type, not domain name.
       - create proper tests to ensure that auth module works correctly
         with dynamic domains.

In order to reduce the number of calls to `mongoose_domain_api:get_host_type/1` HostType parameter is added to some callbacks of ejabberd_gen_auth behaviour.

### DIFF
--- a/include/ejabberd_c2s.hrl
+++ b/include/ejabberd_c2s.hrl
@@ -29,6 +29,7 @@
                 tls_options = [],
                 tls_verify            :: verify_none | verify_peer,
                 authenticated = false :: authenticated_state(),
+                host_type = <<>>      :: binary(),
                 jid                   :: jid:jid() | undefined,
                 user = <<>>           :: jid:user(),
                 server = <<>>         :: jid:server(),
--- a/src/auth/ejabberd_auth.erl
+++ b/src/auth/ejabberd_auth.erl
@@ -76,7 +76,7 @@
 %%%----------------------------------------------------------------------
 -spec start() -> 'ok'.
 start() ->
-    lists:foreach(fun start/1, ?MYHOSTS).
+    lists:foreach(fun start/1, ?ALL_HOST_TYPES).
 
 -spec start(Host :: jid:server()) -> 'ok'.
 start(Host) ->
@@ -122,9 +122,10 @@ get_opt(Host, Opt, Default) ->
 get_opt(Host, Opt) ->
     get_opt(Host, Opt, undefined).
 
--spec supports_sasl_module(jid:lserver(), cyrsasl:sasl_module()) -> boolean().
-supports_sasl_module(Server, Module) ->
-    lists:any(fun(M) -> M:supports_sasl_module(Server, Module) end, auth_modules(Server)).
+-spec supports_sasl_module(binary(), cyrsasl:sasl_module()) -> boolean().
+supports_sasl_module(HostType, Module) ->
+    lists:any(fun(M) -> M:supports_sasl_module(HostType, Module) end,
+              auth_modules_for_host_type(HostType)).
 
 -spec authorize(mongoose_credentials:t()) -> {ok, mongoose_credentials:t()}
                                            | {error, any()}.
@@ -247,11 +248,11 @@ do_try_register_if_does_not_exist(true, _, _) ->
     {error, exists};
 do_try_register_if_does_not_exist(_, JID, Password) ->
     {LUser, LServer} = jid:to_lus(JID),
-    case lists:member(LServer, ?MYHOSTS) of
-        true ->
-            timed_call(LServer, try_register,
+    case mongoose_domain_api:get_host_type(LServer) of
+        {ok, HostType} ->
+            timed_call(HostType, try_register,
                        fun do_try_register_if_does_not_exist_timed/3, [LUser, LServer, Password]);
-        false ->
+        {error, not_found} ->
             {error, not_allowed}
     end.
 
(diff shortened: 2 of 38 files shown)


include/ejabberd_c2s.hrl
```diff
@@ -29,6 +29,7 @@
                 tls_options = [],
                 tls_verify            :: verify_none | verify_peer,
                 authenticated = false :: authenticated_state(),
+                host_type = <<>>      :: binary(),
```
chrzaszcz, line 32: Is there any reason to set an empty host type?

src/auth/ejabberd_auth.erl
```diff
@@ -338,19 +343,6 @@ do_get_vh_registered_users_number(LServer, Opts) ->
             end, auth_modules(LServer))).
 
 
-%% @doc Get the password of the user.
```
chrzaszcz, line 341: Why remove this one? Was it unused? Maybe it was intended as a debug shell utility?

chrzaszcz (commented):
Lots of good work! Just a few comments from me.

Also: please take care of the failing tests.

src/auth/ejabberd_auth_anonymous.erl
```diff
@@ -39,13 +39,13 @@
...
-         set_password/3,
+         set_password/4,
          authorize/1,
-         try_register/3,
+         try_register/4,
          dirty_get_registered_users/0,
          get_vh_registered_users/1,
          get_password/2,
-         get_password/3,
+         %get_password/3,
```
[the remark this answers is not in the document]
DenysGonchar: yeah, should be removed. There is no such interface in ejabberd_gen_auth behaviour.

test/auth_http_SUITE.erl
```diff
@@ -78,18 +78,16 @@ init_per_suite(Config) ->
               Pool = {http, host, auth,
                       [{strategy, random_worker}, {call_timeout, 5000}, {workers, 20}],
                       [{path_prefix, "/auth/"}, {http_opts, []}, {server, ?AUTH_HOST}]},
-              Hosts = [?DOMAIN1, ?DOMAIN2],
```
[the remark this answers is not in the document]
DenysGonchar: well we don't use ?DOMAIN2 anywhere else.
and this extra domain tests more mongoose_wpool functionality then auth_http.
and the last thing - it does it in the init_per_suite() which is not the right place for it.

but if you think that brings any benefits, I can  revert this change

include/ejabberd_c2s.hrl
DenysGonchar: otherwise, dialyzer complains :)

test/config_parser_SUITE.erl
```diff
@@ -280,29 +280,55 @@ outgoing_pools(Config) ->
...
+                   #{reason := not_supported_module, module := test_mim_module2,
                      host_type := <<"another host type">>}]},
                  test_config_file(Config, "host_types")).
 
+host_types_unsupported_auth_methods(Config) ->
+    Modules = [test_mim_module1, test_mim_module2, ejabberd_auth_test1],
+    FN = fun() -> [dynamic_domains] end,
+    [meck:expect(M, supported_features, FN) || M <- Modules],
+    ?assertError({config_error, "Invalid host type configuration",
+                  [#{reason := not_supported_auth_method, auth_method := test2,
```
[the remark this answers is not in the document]
DenysGonchar: yeah, it's a bit tricky if we want to make processing of the returned error value. 
we cannot use `?assertError` macro in this case.
don't know, maybe it's better to rework it in the way how it's done in `host_types_missing_auth_methods_and_modules/1`

src/config/mongoose_config_parser.erl
```diff
@@ -262,3 +264,34 @@ invalid_modules_for_host_type(HostType, Modules) ->
                   what => toml_processing_failed}
             end,
     lists:map(MapFN, Modules).
+
+maybe_check_auth_methods_for_host_types(#local_config{key = {auth_method, HostOrHostType},
```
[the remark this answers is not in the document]
DenysGonchar: would you like to mark it somehow? any `TODO` or `FIXME` comment?

src/auth/ejabberd_auth.erl
DenysGonchar: it's almost identical to `get_passterm_with_authmodule/1` interface, and we were using it only in integration tests.

test/commands_SUITE.erl
```diff
@@ -91,11 +91,12 @@ init_per_testcase(_, C) ->
     meck:expect(ejabberd_config, get_global_option, fun ggo/1),
     meck:new(ejabberd_auth_dummy, [non_strict]),
     meck:expect(ejabberd_auth_dummy, get_password_s, fun(_, _) -> <<"">> end),
+    meck:new(mongoose_domain_api),
```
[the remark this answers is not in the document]
DenysGonchar: it's required for `ejabberd_auth:get_password_s/1` interface

test/config_parser_SUITE.erl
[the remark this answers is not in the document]
chrzaszcz: We could comment that this order is arbitrary.

src/config/mongoose_config_parser.erl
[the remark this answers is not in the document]
chrzaszcz: Not needed now IMO.

include/ejabberd_c2s.hrl
chrzaszcz: Hm... maybe it has a reason to complain? What is the error? If it's in tests, then ok, but are you sure that we won't get an empty binary for an online user?

DenysGonchar: it complains that `undefined` is not binary() type. and nope - we won't have it empty for an online user.
we set a valid host_type at the very early stage of session, actually authentication is impossible without it.

test/config_parser_SUITE.erl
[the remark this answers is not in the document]
DenysGonchar: Technically it's not arbitrary, this check will not result in accidental & floating failures. But it may change if internal implementation changes.

I will add a comment about it.

test/auth_http_SUITE.erl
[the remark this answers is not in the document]
DenysGonchar: ok, added some dummy domain.

include/ejabberd_c2s.hrl
arcusfelis: could be `binary() | undefined` type

DenysGonchar: Then it will complain about usage of `undefined` type everywhere where only binary() is expected.
There is no problem with the current type definition, there is no problem with the default value either.

chrzaszcz: I like what @arcusfelis suggested. Dialyzer does not complain, I checked it.

DenysGonchar: Dialyzer doesn't complain about it right now, because we don't really use that field yet. But I will start in the following PRs, I'm absolutely against this change. Dialyzer is not smart enough to understand FSM lifecycle and understand that in some state it cannot be undefined any more. 
empty binary is the same illegal value for the host type as undefined, we cannot configure it in toml.

chrzaszcz: I still think Dialyzer would not complain because of how success typing works. However, it is a minor thing for me and I can merge the changes as they are now.

chrzaszcz (approved):
Looks good to me.

arcusfelis (approved):
looks reasonable

include/ejabberd_c2s.hrl
DenysGonchar: did some extra tests, looks you're right. will change in the following PR.

PR #3064, opened 2021-03-31
Added mod_domain_isolation

arcusfelis:
This PR addresses isolation.

Proposed changes include:
* isolation

### DIFF
--- a/src/mod_domain_isolation.erl
+++ b/src/mod_domain_isolation.erl
@@ -45,14 +45,16 @@ stop(Host) ->
 -spec filter_local_packet(Value :: fpacket() | drop) -> fpacket() | drop.
 filter_local_packet(drop) ->
     drop;
-filter_local_packet({From = #jid{lserver = FromServer},
-                     To = #jid{lserver = ToServer}, Acc, Packet}) ->
+filter_local_packet({#jid{lserver = Server},
+                     #jid{lserver = Server}, _Acc, _Packet} = Arg) ->
+    Arg;
+filter_local_packet({#jid{lserver = FromServer} = From,
+                     #jid{lserver = ToServer} = To, Acc, Packet} = Arg) ->
     FromHost = domain_to_host(FromServer),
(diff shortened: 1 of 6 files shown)


src/mod_domain_isolation.erl
```diff
@@ -0,0 +1,73 @@
...
+    AllHosts = [Host|SubHosts],
+    [ejabberd_hooks:delete(filter_local_packet, H, ?MODULE, filter_local_packet, 10)
+     || H <- AllHosts],
+    ok.
+
+-spec filter_local_packet(Value :: fpacket() | drop) -> fpacket() | drop.
+filter_local_packet(drop) ->
+    drop;
+filter_local_packet({From = #jid{lserver = FromServer},
+                     To = #jid{lserver = ToServer}, Acc, Packet}) ->
```
NelsonVides, line 49: Won't it happen that if we can be sure both `FromServer` and `ToServer` are the same, including subdomains, then this is allowed? In that case we could do
```suggestion
filter_local_packet({From = #jid{lserver = SameServer},
                     To = #jid{lserver = SameServer}, Acc, Packet}) ->
    {From, To, Acc, Packet};
filter_local_packet({From = #jid{lserver = FromServer},
                     To = #jid{lserver = ToServer}, Acc, Packet}) ->
```
as a shortcut path to skip extracting domains?

src/mod_domain_isolation.erl
```diff
@@ -0,0 +1,73 @@
...
+    FromHost = domain_to_host(FromServer),
+    ToHost = domain_to_host(ToServer),
+    Pass = FromHost =:= ToHost,
+    case Pass of
```
NelsonVides, line 53:
```suggestion
    case FromHost =:= ToHost of
```
At the very least, please

NelsonVides (approved):
LGTM

arcusfelis:
replaced with #3070

PR #3066, opened 2021-03-31
Prepared queries for chatmarkers and smart_markers

leszke:


src/mod_smart_markers_rdbms.erl
```diff
@@ -21,6 +21,10 @@ init(Host, _) ->
     QueryName = smart_markers_upsert,
     rdbms_queries:prepare_upsert(Host, QueryName, smart_markers,
                                  InsertFields, UpdateFields, KeyFields),
+    mongoose_rdbms:prepare(smart_markers_select, smart_markers,
+        [to_jid, thread, timestamp],
+        <<"select from_jid, to_jid, thread, type, msg_id, timestamp from smart_markers "
+          "WHERE to_jid = ? AND thread = ? AND timestamp >= ? ;">>),
```
chrzaszcz, line 27: Semicolon is not necessary, please **check other queries** as well.

```suggestion
          "WHERE to_jid = ? AND thread = ? AND timestamp >= ?">>),
```

src/offline/mod_offline_chatmarkers_rdbms.erl
```diff
@@ -39,20 +63,13 @@ get(#jid{lserver = Host} = Jid) ->
...
 maybe_store(#jid{lserver = Host} = Jid, Thread, Room, Timestamp) ->
-    JidEscaped = escape(encode_jid(Jid)),
-    ThreadEscaped = escape(encode_thread(Thread)),
-    RoomEscaped = escape(encode_jid(Room)),
-    TSEscaped = escape(Timestamp),
-    InsertQuery = ["INSERT INTO offline_markers (jid, thread, room, timestamp) VALUES (",
-                   JidEscaped, ",", ThreadEscaped, ",", RoomEscaped, ",", TSEscaped, ");"],
-    Res = mongoose_rdbms:sql_query(Host, InsertQuery),
-    ok = check_insert_result(Res).
+    {updated, _} = execute_maybe_store(Host, encode_jid(Jid), encode_thread(Thread),
```
chrzaszcz, line 66: Minor: in case of `execute_successfully` with a `UPDATE/DELETE` query, I see no reason to pattern match the result as it can be only `{updated, _}`. So I think we could simplify the code:

```suggestion
    execute_maybe_store(Host, encode_jid(Jid), encode_thread(Thread),
```

src/offline/mod_offline_chatmarkers_rdbms.erl
```diff
@@ -39,20 +63,13 @@ get(#jid{lserver = Host} = Jid) ->
...
+                                       encode_jid(Room), Timestamp),
+    ok.
 
 -spec remove_user(Jid :: jid:jid()) -> ok.
 remove_user(#jid{lserver = Host} = Jid) ->
-    JidEscaped = escape(encode_jid(Jid)),
-    DelQuery = ["DELETE FROM offline_markers WHERE jid = ", JidEscaped, ";"],
-    {updated, _} = mongoose_rdbms:sql_query(Host, DelQuery),
+    {updated, _} = execute_delete_user(Host, encode_jid(Jid)),
```
chrzaszcz, line 72: See the comment above.

chrzaszcz (commented):
Good work! I only have a few minor comments.

chrzaszcz (approved):
Great!

PR #3068, opened 2021-04-02
Improve pipeline for mod_smart_markers

NelsonVides:
This module was doing one non-optimal check, that is, it was checking if the user has rights to access a room using the `can_access_room` hook, _before_ checking that the message being processed actually has any marker to begin with. This is non-optimal, as `can_access_room` requires another hook call and a DB query to verify affiliations, and will usually return true unless misbehaving clients, while checking whether a message has actually any marker is an in-place check with higher chances to return false.


DenysGonchar (approved):
looks good.

PR #3069, opened 2021-04-02
Don't print test summary if CT reports directory is not created

arcusfelis:
This PR addresses "We print reports for the last **successful** run. But not the last executed run".

Proposed changes include:
* Pass to summarise-ct-results directories actually created since the start of the test


leszke (approved):
Looks good, thanks for the change!

chrzaszcz (approved):
Checked manually, I can confirm that it work.

PR #3070, opened 2021-04-02
Host isolation

arcusfelis:
it's a variant of https://github.com/esl/MongooseIM/pull/3064 but to master.

Proposed changes include:
* there is no code or solution for multi-tenancy.
* the code would probably change, once we add dynamic domains.
* could be merged to master, but multi-tenancy would have to be rebased on master to continue work on this feature. Routing would also have to be working.


NelsonVides (approved):
LGTM

PR #3075, opened 2021-04-07
Multi tenancy pm

DenysGonchar:
this PR introduces:
* a new router `mongoose_router_dynamic_domains` that is responcible for a lazy adding (on the first routing attempt) of dynamic domains. removal of dynamic domains happens synchronously on `disable_domain` hook.
* a new mandatory configuration item `default_server_domain` - it must be used for reporting stream errors before user clarifies the domain name he wants to use (before user sets up the stream).

### DIFF
--- a/test/config_parser_SUITE.erl
+++ b/test/config_parser_SUITE.erl
@@ -49,6 +49,7 @@ groups() ->
      {general, [parallel], [loglevel,
                             hosts,
                             host_types,
+                            default_server_domain,
                             registration_timeout,
                             language,
                             all_metrics_are_global,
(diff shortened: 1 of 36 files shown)


test/config_parser_SUITE.erl
```diff
@@ -346,19 +347,22 @@ loglevel(_Config) ->
...
     ?err(parse(#{<<"general">> => #{<<"hosts">> => [<<"what is this?">>]}})),
     ?err(parse(#{<<"general">> => #{<<"hosts">> => [<<>>]}})),
     ?err(parse(#{<<"general">> => #{<<"hosts">> => [<<"host1">>, <<"host1">>]}})),
     % either hosts or host_types must be provided
-    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => #{}})),
-    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => #{<<"host">> => [],
-                                                                <<"host_types">> => []}})),
-    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => #{<<"host">> => []}})),
-    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => #{<<"host_types">> => []}})).
+    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => GenM})),
```
chrzaszcz, line 361: I think we lost the test case for the empty map.

```suggestion
    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => #{}})),
    ?err(mongoose_config_parser_toml:parse(#{<<"general">> => GenM})),
```

test/config_parser_SUITE.erl
```diff
@@ -2979,19 +2996,34 @@ err_host_or_global(Config) ->
...
-    %% remove hosts key only if value is equal to [?HOST]
-    lists:filter(fun(#config{key = hosts, value = [?HOST]}) -> false;
-                    (_) -> true
-                 end, Config).
+    maybe_filter_out_dummy_domain(Config, DummyDomainName).
+
+maybe_insert_dummy_domain(M, DomainName) ->
+    DummyGenM = #{<<"default_server_domain">> => DomainName,
+                  <<"hosts">> => [DomainName]},
+    OldGenM = maps:get(<<"general">> ,M, #{}),
```
chrzaszcz, line 3016:
```suggestion
    OldGenM = maps:get(<<"general">>, M, #{}),
```

test/config_parser_SUITE.erl
```diff
@@ -2979,19 +2996,34 @@ err_host_or_global(Config) ->
...
+parse(M0) ->
+    %% 'hosts' (or 'host_types') and `default_server_domain` options are mandatory.
+    %% this function does the following things:
+    %%   1) plugs that mandatory options with dummy values (if required).
+    %%   2) executes parsing.
+    %%   3) removes extra 'hosts'/'default_server_domain' config keys (but only if
+    %%      they have dummy values).
+    %% DummyDomainName value must be unique to avoid accidental config keys removal.
+    DummyDomainName = <<"dummy.domain.name">>,
+    M = maybe_insert_dummy_domain(M0, DummyDomainName),
```
chrzaszcz, line 3009: The idea of putting `?HOST` here was to create a valid config. Now the config is invalid in some tests as `dummy.domain.name` is not the same host as in `host_config` (see `eq_host_config`). So the tests now show (and expect) that an invalid config is parsed correctly... This might be confusing.

I'd rather keep a valid host name here and fix the issue later than have a wrong value set on purpose.

chrzaszcz (commented):
Looks good, I have a few specific comments. Also, as we agreed, there is an urgent need to fix the use of the `?MYNAME` macro that is now doing something else than expected in some places in the code.

test/config_parser_SUITE.erl
DenysGonchar: `eq_host_config` explicitly uses `?HOST` value in both checks.

chrzaszcz: The point is that after the change some test cases are parsing inconsistent configuration, e.g.

```
#{<<"hosts">> => [<<"dummy...">>], <<"host_config">> => [{<<"host_type">> => ?HOST, ....}]}
```

Such configuration does not make sense and I think that it should result in an error and should be checked in a test case.

chrzaszcz (approved):

PR #3077, opened 2021-04-09
Update rebar3_codecov to use the latest hex package

NelsonVides:
Very quick update to use the hex package, now that it has been updated :)


leszke (approved):
🔥

PR #3078, opened 2021-04-09
Prepared queries for inbox

chrzaszcz:
- Use prepared queries in `mod_inbox_rdbms`
- Make it possible to include update expressions like 'value = column.value + 1' in upsert queries
- Simplify some queries, e.g. property updates

Details in commit messages


src/rdbms/rdbms_queries.erl
```diff
@@ -89,35 +89,37 @@ execute_upsert(Host, Name, InsertParams, UpdateParams, UniqueKeyValues) ->
 
 %% @doc
 %% This functions prepares query for inserting a row or updating it if already exists
+%% Updates can be either fields or literal expressions like "column = tab.column + 1"
 -spec prepare_upsert(Host :: mongoose_rdbms:server(),
                      QueryName :: atom(),
                      TableName :: atom(),
                      InsertFields :: [binary()],
-                     UpdateFields :: [binary()],
```
arcusfelis, line 96: could be an extra argument `UpdateSQL`, which is empty by default.
And so we would have prepare_upsert/6 and prepare_upsert/7.

But  not thaaat critical, it's called once per statement anyway.

arcusfelis (approved):
looks good to me

src/rdbms/rdbms_queries.erl
chrzaszcz: I admit I avoided adding the extra argument. I could do it like you suggest to make it more obvious for the user.

leszke (approved):
Looks good!

PR #3080, opened 2021-04-13
Unify checks for chat markers

NelsonVides:
Currently checking if a stanza contains a chat markers is triplicated in
three different modules. Implement all of it in mod_smart_markers and
make the other two places simply call this one.


DenysGonchar (approved):
thanks for moving it in a separate helper module

PR #3086, opened 2021-04-16
Remove untested and outdated 'azuresql.sql'

chrzaszcz:
This file was outdated, unsupported and untested.
The docs were misleading, because they gave an impression that it was up to date and working.

We can bring it back when a need arises and there is a way to test it.


doc/advanced-configuration/database-backends-configuration.md
```diff
@@ -200,11 +200,11 @@ ALTER DATABASE $name_of_your_db SET READ_COMMITTED_SNAPSHOT ON
 
 The command above may take some time.
 
-Then you need to import the SQL schema from either ``mssql2012.sql`` or ``azuresql.sql`` file depending on which database you are using.
+Then you need to import the SQL schema from  ``mssql2012.sql``.
```
leszke, line 203: Minor: I think that single grave accents are enough: 
```suggestion
Then you need to import the SQL schema from  `mssql2012.sql`.
```

leszke (approved):
All is good! I noticed only one cosmetic issue, no need to take care of it.

PR #3089, opened 2021-04-18
Multi tenancy hooks

DenysGonchar:
this PR reworks `ejabberd_hooks` in the following way:
* hooks are running either globally or for a host type, not domain name, so they must be registered globally or for a host type as well.
* hooks execution is fully wrapped in `mongoose_hooks` module.
* when hook handler expect some exact constant initial Acc value, such value is hardcoded  `mongoose_hooks` module. It should not be possible to run the hook with another initial acc value.
* `mongoose_acc:t()` type is extended with `host_type` field, this allows to reduce the number of calls to `mongoose_domain_api:get_host_type/1` function.
* ejabberd_c2s modules is now running hooks for a `host_type`,  note that only hooks execution is updated. if some hook was expecting a domain name as one of the parameters, it still receives a domain name.  we may need to change it in the future, but it requires updates of the hook handlers (which are mostly in the MIM modules).
* `remove_domain` hook handler implementation for `ejaberd_auth`

Another big generic module that triggers hooks is `ejabberd_sm` (13 calls to `mongoose_hook`). But it should be reworked in a separate PR, and probably after switching back to the master branch.


test/acc_SUITE.erl
```diff
@@ -133,38 +126,92 @@ produce_iq_meta_automatically(_C) ->
...
+                    (_) -> {error, not_found}
+                end),
+    AccParams=#{location => ?LOCATION},
+    StripParams = #{element => sample_stanza()},
+    %% new with <<"host 1">> host type
+    Acc1 = mongoose_acc:new(AccParams#{lserver => <<"host1">>}),
+    ?assertEqual(<<"host 1">>, mongoose_acc:host_type(Acc1)),
+    ?assertEqual(1, meck:num_calls(mongoose_domain_api, get_host_type, [<<"host1">>])),
+    %% new with 'undefined' host type
+    Acc2 = mongoose_acc:new(AccParams#{lserver => <<"unknown.host1">>}),
```
chrzaszcz, line 193: Would it be possible to break it into several test cases? It is a bit difficult to read and understand.

test/acc_SUITE.erl
```diff
@@ -133,38 +126,92 @@ produce_iq_meta_automatically(_C) ->
...
+    ?assertEqual(undefined, mongoose_acc:host_type(Acc2)),
+    ?assertEqual(1, meck:num_calls(mongoose_domain_api, get_host_type, [<<"unknown.host1">>])),
+    %% host type changes from <<"host 1">> to 'undefined'
+    Acc3 = mongoose_acc:strip(StripParams#{lserver => <<"unknown.host1">>},Acc1),
+    ?assertEqual(undefined, mongoose_acc:host_type(Acc3)),
+    ?assertEqual(2, meck:num_calls(mongoose_domain_api, get_host_type, [<<"unknown.host1">>])),
+    %% host type changes from <<"host 1">> to <<host 2>>
+    Acc6 = mongoose_acc:strip(StripParams#{lserver => <<"host2">>},Acc1),
```
chrzaszcz, line 201: The ordering does not help either: Acc1, Acc2, Acc3, Acc6, Acc4, Acc5

test/acc_SUITE.erl
```diff
@@ -133,38 +126,92 @@ produce_iq_meta_automatically(_C) ->
...
     ?assertEqual(jid:from_binary(<<"a@localhost">>), mongoose_acc:to_jid(NAcc2)),
     ?assertEqual(Ref, mongoose_acc:ref(NAcc2)),
     ?assertEqual(997, mongoose_acc:get(ns, ppp, NAcc2)),
-    ?assertEqual([], mongoose_acc:get(ns2, NAcc2)).
-
+    ?assertEqual([], mongoose_acc:get(ns2, NAcc2)),
+    ?assertEqual(sample_stanza(), mongoose_acc:element(NAcc2)),
+    ?assertEqual(NewServer, mongoose_acc:lserver(NAcc2)),
+    ?assertEqual(NewHostType, mongoose_acc:host_type(NAcc2)),
+    NAcc3 = mongoose_acc:strip(Acc3),
```
chrzaszcz, line 167: At this point it becomes really hard to follow the test case. Please break it into smaller ones.

chrzaszcz (commented):
I appreciate the huge amount of good work, especially putting the defaults in `mongoose_hooks`. I have a few comments:
- Are you planning to use `mongoose_acc:get_host_type` inside the hooks to get rid of the `host_type` argument for each of them? For me it looks like a natural follow-up.
- Please revise the line breaks in `mongoose_hooks` to break the comments so that expressions like `a server`, `the client`, `should be`, `client jid` etc remain in one line. I think it's good practice to do so.

src/auth/ejabberd_auth.erl
```diff
@@ -503,6 +505,19 @@ get_auth_method_as_a_list(undefined) -> [];
 get_auth_method_as_a_list(AuthMethod) when is_list(AuthMethod) -> AuthMethod;
 get_auth_method_as_a_list(AuthMethod) when is_atom(AuthMethod) -> [AuthMethod].
 
+remove_domain(Acc, HostType, Domain) ->
+    Modules = auth_modules_for_host_type(HostType),
+    lists:foreach(fun(M) -> maybe_return_domain(M, HostType, Domain) end, Modules),
+    Acc.
+
+maybe_return_domain(Module, HostType, Domain) ->
```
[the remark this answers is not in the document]
DenysGonchar: oh my, yes it's a typo

src/auth/ejabberd_auth_dummy.erl
```diff
@@ -92,7 +92,7 @@ does_user_exist(_User, _Server) ->
 remove_user(_User, _Server) ->
     ok.
 
-remove_domain(_Server) -> ok.
+remove_domain(_HostType, _Server) -> ok.
```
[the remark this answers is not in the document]
DenysGonchar: it's required for testing the ejabberd_auth hook callback.

src/mongoose_hooks.erl
```diff
@@ -1232,187 +1231,179 @@ s2s_connect_hook(Name, Acc, Server) ->
...
+%%% to the rules and the event (i.e. notify the sender).
 -spec amp_check_packet(Server, Acc, From, Event) -> Result when
     Server :: jid:lserver(),
     Acc :: mongoose_acc:t(),
     From :: jid:jid(),
     Event :: mod_amp:amp_event(),
     Result :: mongoose_acc:t().
 amp_check_packet(Server, Acc, From, Event) ->
-    ejabberd_hooks:run_fold(amp_check_packet, Server, Acc, [From, Event]).
+    %%TODO: no-one is calling this hook, recheck why!
```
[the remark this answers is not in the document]
DenysGonchar: -1 TODO :)

test/acc_SUITE.erl
DenysGonchar: yeah, i've decided to move Acc6 test but forgot to change the ids.

test/acc_SUITE.erl
DenysGonchar: we can have 2 test cases, one with Acc1 and striping operations, another with Acc2

DenysGonchar:
>     * Are you planning to use `mongoose_acc:get_host_type` inside the hooks to get rid of the `host_type` argument for each of them? For me it looks like a natural follow-up.

The idea was to reduce calls to mongoose_acc:get_host_type/1 as much as possible. Ideally it should be limited to just one call per incoming message (to identify receiver's host_type, if required), we don't want it to become a bottleneck.

arcusfelis:
We could always ask for acc argument, that would have host_type field.
and if hook handlers need host_type - they could extract it from the acc.
And if handlers need a domain, they already have server in the acc.

But it's more lines of code to change :D 

And after that we could pass Acc everywhere, even into the backend functions.

test/acc_SUITE.erl
```diff
@@ -163,55 +198,57 @@ strip(_C) ->
...
     AccParams=#{location => ?LOCATION},
     StripParams = #{element => sample_stanza()},
     %% new with <<"host 1">> host type
     Acc1 = mongoose_acc:new(AccParams#{lserver => <<"host1">>}),
     ?assertEqual(<<"host 1">>, mongoose_acc:host_type(Acc1)),
     ?assertEqual(1, meck:num_calls(mongoose_domain_api, get_host_type, [<<"host1">>])),
-    %% new with 'undefined' host type
-    Acc2 = mongoose_acc:new(AccParams#{lserver => <<"unknown.host1">>}),
+    %% host type changes from <<"host 1">> to 'undefined'
+    Acc2 = mongoose_acc:strip(StripParams#{lserver => <<"unknown.host1">>},Acc1),
```
chrzaszcz, line 224: Minor: missing commas before `Acc1`, this happens a few times below.

chrzaszcz (commented):
One minor comment, please tell me whether you are correcting it or if I should merge.

chrzaszcz (approved):

PR #3091, opened 2021-04-22
Remove ?MYNAME macros

chrzaszcz:
Remove all occurences of `?MYNAME` except stream errors and the initial value in `ejabberd_c2s`. These are the only expected uses of this macro.


DenysGonchar (approved):
looks good

PR #3092, opened 2021-04-22
Add remove_domain handler for MAM

arcusfelis:
This PR addresses "Ensure efficient indexing and removal of persistent data by domain name".
MAM already have proper indexes in the mam_server_user table. We should just LEFT JOIN it. OK, we can't LEFT JOIN in DELETEs. But we can use `WHERE IN` instead.

Proposed changes include:
* remove_domain/3 handler for MAM/MUC MAM.
* nobody runs this hook in this PR. @DenysGonchar will code that part.
* a tip in schema. Because we use the same table to track MAM archive IDs for both MAM and MUC MAM.

### DIFF
--- a/big_tests/tests/domain_removal_SUITE.erl
+++ b/big_tests/tests/domain_removal_SUITE.erl
@@ -111,6 +111,4 @@ mam_muc_removal(Config0) ->
     escalus_fresh:story_with_config(Config0, [{alice, 1}], F).
 
 run_remove_domain() ->
-    Acc = #{},
-    rpc(mim(), ejabberd_hooks, run_fold,
-        [remove_domain, domain(), Acc, [domain(), domain()]]).
+    rpc(mim(), mongoose_hooks, remove_domain, [domain(), domain()]).
--- a/big_tests/tests/domain_removal_SUITE.erl
+++ b/big_tests/tests/domain_removal_SUITE.erl
@@ -1,0 +1,127 @@
+-module(domain_removal_SUITE).
+
+%% API
+-export([all/0,
+         groups/0,
+         init_per_suite/1,
+         end_per_suite/1,
+         init_per_group/2,
+         end_per_group/2,
+         init_per_testcase/2,
+         end_per_testcase/2]).
+
+-export([mam_pm_removal/1,
+         mam_muc_removal/1]).
+
+-import(mam_helper,
+        [stanza_archive_request/2,
+         wait_archive_respond/1,
+         assert_respond_size/2,
+         respond_messages/1,
+         parse_forwarded_message/1]).
+
+-import(distributed_helper, [mim/0,
+                             require_rpc_nodes/1,
+                             rpc/4]).
+
+-include("mam_helper.hrl").
+-include_lib("escalus/include/escalus.hrl").
+-include_lib("escalus/include/escalus_xmlns.hrl").
+-include_lib("common_test/include/ct.hrl").
+-include_lib("exml/include/exml_stream.hrl").
+
+all() ->
+    [{group, mam_removal}].
+
+groups() ->
+    [
+     {mam_removal, [], [mam_pm_removal,
+                        mam_muc_removal]}
+    ].
+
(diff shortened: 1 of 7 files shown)


src/mam/mod_mam_rdbms_arch.erl
```diff
@@ -102,6 +103,7 @@ start_hooks(Host, _Opts) ->
     ejabberd_hooks:add(mam_archive_size, Host, ?MODULE, archive_size, 50),
     ejabberd_hooks:add(mam_lookup_messages, Host, ?MODULE, lookup_messages, 50),
     ejabberd_hooks:add(mam_remove_archive, Host, ?MODULE, remove_archive, 50),
+    ejabberd_hooks:add(remove_domain, Host, ?MODULE, remove_domain, 50),
```
DenysGonchar, line 106: please, make a `hooks/0` function that returns a list of hooks and use `ejabbed_hooks:add/1` and `ejabberd_hooks:delete/1` functions to start/stop module

src/mam/mod_mam_muc_rdbms_arch.erl
```diff
@@ -86,6 +87,7 @@ start_hooks(Host, _Opts) ->
     ejabberd_hooks:add(mam_muc_archive_size, Host, ?MODULE, archive_size, 50),
     ejabberd_hooks:add(mam_muc_lookup_messages, Host, ?MODULE, lookup_messages, 50),
     ejabberd_hooks:add(mam_muc_remove_archive, Host, ?MODULE, remove_archive, 50),
+    ejabberd_hooks:add(remove_domain, Host, ?MODULE, remove_domain, 50),
```
DenysGonchar, line 90: please, make a hooks/0 function that returns a list of hooks and use ejabbed_hooks:add/1 and ejabberd_hooks:delete/1 functions to start/stop module

src/mam/mod_mam_muc_rdbms_arch.erl
```diff
@@ -112,6 +115,11 @@ register_prepared_queries() ->
     mongoose_rdbms:prepare(mam_muc_archive_remove, mam_muc_message, [room_id],
                            <<"DELETE FROM mam_muc_message "
                              "WHERE room_id = ?">>),
+    mongoose_rdbms:prepare(mam_muc_remove_domain, mam_muc_message, ['mam_server_user.server'],
+                           <<"DELETE FROM mam_muc_message "
+                             "WHERE room_id IN (SELECT id FROM mam_server_user where server = ?)">>),
```
DenysGonchar, line 120: how optimal is this query? what if we have over 100k records selected by `SELECT id FROM mam_server_user where server = ?`

big_tests/tests/domain_removal_SUITE.erl
```diff
@@ -0,0 +1,127 @@
...
+        mam_helper:wait_for_room_archive_size(MucHost, Room, 1),
+        run_remove_domain(),
+        mam_helper:wait_for_room_archive_size(MucHost, Room, 0),
+        ok
+        end,
+    escalus_fresh:story_with_config(Config0, [{alice, 1}], F).
+
+run_remove_domain() ->
+    Acc = #{},
+    rpc(mim(), ejabberd_hooks, run_fold,
```
DenysGonchar, line 126: `remove_domain` hook code is already merged into master. I know it was not in original scope, and it's not critical to be a part of this PR.
but please change this rpc call to `mongoose_hooks:remove_domain/2` in the following PR.

DenysGonchar (changes requested):
Thanks. The changes look good, I've added just a few minor comments

src/mam/mod_mam_muc_rdbms_arch.erl
arcusfelis: it should work.

ids would go into a temp table in this case. And use similar logic as left joins.

 It would be a pretty long transaction though.

DenysGonchar (approved):
Approved

PR #3093, opened 2021-04-22
Use map for mod_mam:message_row()

arcusfelis:
This PR addresses "what if we use map, will tests still pass?"

Proposed changes include:
* map
* Allow to override mod_mam_utils:send_message/4 callback. 
* Pass caller_jid into lookup_messages hook as a part of params (useful for MUC MAM).

As a dev, I want to extract some extra fields from MAM (or during MAM lookup).
I am currently only allowed to return message-id, source-jid and stanza packet. But sometimes I need to return more fields, so I can format them later (to build a proper forwarded message). Maps allows to attach some meta info to each message.

_Also, it's really hard to track what is what in 3 fields tuple :trollface:_

### DIFF
--- a/big_tests/tests/mam_send_message_SUITE.erl
+++ b/big_tests/tests/mam_send_message_SUITE.erl
@@ -1,0 +1,142 @@
+-module(mam_send_message_SUITE).
+
+%% API
+-export([all/0,
+         groups/0,
+         init_per_suite/1,
+         end_per_suite/1,
+         init_per_group/2,
+         end_per_group/2,
+         init_per_testcase/2,
+         end_per_testcase/2]).
+
+-export([mam_muc_send_message/1,
+         mam_pm_send_message/1]).
+
+-import(mam_helper,
+        [stanza_archive_request/2,
+         wait_archive_respond/1,
+         assert_respond_size/2,
+         respond_messages/1,
+         parse_forwarded_message/1]).
+
+-import(distributed_helper, [mim/0,
+                             require_rpc_nodes/1,
+                             rpc/4]).
+
+-include("mam_helper.hrl").
+-include_lib("escalus/include/escalus.hrl").
+-include_lib("escalus/include/escalus_xmlns.hrl").
+-include_lib("common_test/include/ct.hrl").
+-include_lib("exml/include/exml_stream.hrl").
+
+all() ->
+    [{group, send_message}].
+
+groups() ->
+    [
+     {send_message, [], [mam_pm_send_message,
+                         mam_muc_send_message]}
+    ].
+
+domain() ->
+    ct:get_config({hosts, mim, domain}).
+
+%%%===================================================================
+%%% Overall setup/teardown
+%%%===================================================================
+init_per_suite(Config) ->
+    escalus:init_per_suite(Config).
+
+end_per_suite(Config) ->
+    escalus:end_per_suite(Config).
+
+%%%===================================================================
+%%% Group specific setup/teardown
+%%%===================================================================
+init_per_group(Group, Config) ->
+    case mongoose_helper:is_rdbms_enabled(domain()) of
+        true ->
+            load_custom_module(),
+            Config2 = dynamic_modules:save_modules(domain(), Config),
+            rpc(mim(), gen_mod_deps, start_modules, [domain(), group_to_modules(Group)]),
+            Config2;
+        false ->
+            {skip, require_rdbms}
+    end.
+
+end_per_group(_Groupname, Config) ->
+    case mongoose_helper:is_rdbms_enabled(domain()) of
+        true ->
+            dynamic_modules:restore_modules(domain(), Config);
+        false ->
+            ok
+    end,
+    ok.
+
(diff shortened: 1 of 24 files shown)


big_tests/tests/mam_send_message_SUITE.erl
```diff
@@ -0,0 +1,142 @@
...
+
+end_per_testcase(TestCase, Config) ->
+    escalus:end_per_testcase(TestCase, Config).
+
+%%%===================================================================
+%%% Test Cases
+%%%===================================================================
+
+mam_pm_send_message(Config) ->
+    P = ?config(props, Config),
```
NelsonVides, line 104: It doesn't seem to me like this variable has been loaded into the config, isn't it? I think this was one of the mam helper functions, might as well be used here if it is really needed. Same in the test below.

big_tests/tests/mam_send_message_SUITE.erl
```diff
@@ -0,0 +1,142 @@
...
+        escalus:wait_for_stanza(Bob),
+        mam_helper:wait_for_archive_size(Alice, 1),
+        mam_helper:wait_for_archive_size(Bob, 1),
+        escalus:send(Alice, stanza_archive_request(P, <<"q1">>)),
+        Res = wait_archive_respond(Alice),
+        assert_respond_size(1, Res),
+        [Msg] = respond_messages(Res),
+        verify_has_some_hash(Msg)
+        end,
+    escalus_fresh:story(Config, [{alice, 1}, {bob, 1}], F).
```
NelsonVides, line 116: Also, don't we usually use `escalus:fresh_story`? No big deal, but maybe a good practice.

big_tests/tests/mam_send_message_SUITE.erl
```diff
@@ -0,0 +1,142 @@
...
+        escalus:send(Alice, escalus_stanza:groupchat_to(RoomAddr, <<"text">>)),
+        M = escalus:wait_for_stanza(Alice),
+        escalus:assert(is_message, M),
+        escalus_assert:has_no_stanzas(Alice),
+        mam_helper:wait_for_room_archive_size(MucHost, Room, 1),
+        escalus:send(Alice, escalus_stanza:to(stanza_archive_request(P, <<"q1">>), RoomAddr)),
+        [Msg] = respond_messages(assert_respond_size(1, wait_archive_respond(Alice))),
+        verify_has_some_hash(Msg)
+        end,
+    escalus_fresh:story_with_config(Config0, [{alice, 1}], F).
```
NelsonVides, line 136: Why `story_with_config`?

NelsonVides (commented):
Looks very very good, just a few comments in the tests 😄

big_tests/tests/mam_send_message_SUITE.erl
arcusfelis: Because we need the updated user list in `muc_light_helper:create_room`

big_tests/tests/mam_send_message_SUITE.erl
arcusfelis: it could be undefined.
which still works.
But I want to keep the test similar to how mam_SUITE handles it (i.e. setting this variable and using stanza_archive_request).

Ok, I will set it to 06.

NelsonVides (approved):
Totally love it. It has a theoretical impact on performance, but it's probably just theoretical. This is one of the cases where flexibility outweighs those tiny nanoseconds. And this is a very useful one to me 🙂

src/mam/mam_lookup.erl
```diff
@@ -158,7 +158,8 @@ decode_rows(MessageRows, Env) ->
     [decode_row(Row, Env) || Row <- MessageRows].
 
 %% First element of the tuple is decoded message ID
```
chrzaszcz, line 160: There is no tuple now, please amend this comment.

src/mam/mod_mam.erl
```diff
@@ -119,7 +118,7 @@
 
 -type borders()             :: #mam_borders{}.
 
--type message_row() :: {message_id(), jid:jid(), exml:element()}.
+-type message_row() :: #{id => message_id(), jid => jid:jid(), packet => exml:element()}.
```
chrzaszcz, line 121: Are they optional? Could we make the keys mandatory in this type?

chrzaszcz (commented):
Looks good. I have a few comments and Elvis has two that I agree with:

```
mod_mam.erl   460   1 warning  invalid_dynamic_call Remove the dynamic function call on line 460. Only modules that define callbacks should make dynamic calls. (lsp)
mod_mam.erl   617   1 warning  line_length Line 617 is too long: message_row_to_xml(MamNs, #{id := MessID, jid := SrcJID, packet := Packet}, QueryID, SetClientNs)  ->. (lsp)
```

NelsonVides (approved):
👍🏽

chrzaszcz (commented):
I'm a bit late, but it looks great, thanks for the extra work to address the comments!

PR #3096, opened 2021-04-23
Start only used metrics

gustawlippa:
This PR addresses the problem of starting metrics for not used modules in MIM which is a performance concern and bad user experience.
Most of the culprits are metrics in the file `mongoose_metrics_definitions`, which are custom hook metrics (they should override the automatic hook metrics). Their handlers, which implement custom logic for updating the metrics, are present in files `mongoose_metrics_hooks` and `mongoose_metrics_mam_hooks`.
The proposition is to move the place of starting the metrics and registering these custom hooks back to the modules which these metrics concern, while keeping the files with hook code as is, at least for now.

There are some problems left but would require metrics name changes or even a rework of some metrics, which could be breaking changes for the users.
I'm not sure how to approach `modPrivacy*` metrics and would appreciate feedback.

There is an issue right now that `mongoose_metrics` contains a list of hooks that don't have generic metrics and have some custom logic to calculate metrics (often weirdly named) based on them (this logic is for example in `mongoose_metrics_hooks`).
This list can lead to inconsistencies, and the custom logic may sometimes be too coupled with the modules.
My proposition would be to get rid of this list, unify the hook metrics and keep the automatic metrics for all of them (for example with a `hooks` prefix).

The custom logic to calculate more complex metrics could maybe be than moved from files like `mongoose_metrics_hooks` to the modules which evoke `ensure_metrics` or similar functions on startup, so the places which are "responsible" for the metrics. `mongoose_metrics_hooks` and similar would keep some very general metrics like the xmpp ones.
Other solution could just be to check the metrics ensured based on `mongoose_metrics_definitions` against the configured modules. 

Metrics that are run for MIM configured without any modules:

Changed by this PR:
-  [modCSIActive], changed - now will start in `mod_csi`.
-  [modCSIInactive], same.
-  [modGlobalDistribMessagesReceived], removed because it was unused, probably deprecated by `?GLOBAL_DISTRIB_MESSAGES_SENT` in `global_metrics.hrl`
-  [modGlobalDistribMessagesSent], same.
-  [modMamArchiveRemoved], changed - now will start in `mod_mam`, like the ones below.
-  [modMamArchived],
-  [modMamDropped],
-  [modMamDropped2], this name...
-  [modMamDroppedIQ],
-  [modMamFlushed],
-  [modMamForwarded],
-  [modMamLookups],
-  [modMamMultiplePurges], removed - wasn't used.
-  [modMamPrefsGets],
-  [modMamPrefsSets],
-  [modMamSinglePurges], removed - wasn't used.
-  [modMucMamArchiveRemoved], changed similarly to `mod_mam` - now will start in `mod_mam_muc`. Added `modMucMamDroppedIQ` because it was being updated and didn't exist.
-  [modMucMamArchived],
-  [modMucMamFlushed],
-  [modMucMamForwarded],
-  [modMucMamLookups],
-  [modMucMamMultiplePurges], removed - wasn't used.
-  [modMucMamPrefsGets],
-  [modMucMamPrefsSets],
-  [modMucMamSinglePurges], removed - wasn't used.
-  [sessionAuthAnonymous], removed - wasn't used.

Problematic/other comments - for now unchanged:
-  [backends,auth,authorize] - these 4 seem OK, although I don't really get the name - should the metrics be split by the different backends? Bumped in `ejabberd_auth:timed_call/4`.
-  [backends,auth,check_password],
-  [backends,auth,does_user_exist],
-  [backends,auth,try_register],
-  [local_send_to_resource_hook], generic automatic hook metric, comes from `ejabberd_local`. Some hooks and their metrics have the `_hook` suffix, other not, maybe this could be unified.
-  [mam_lookup_messages], added in `mongoose_metrics_mam_hooks` as an automatic hook, I think it was supposed to be filtered by `mongoose_metrics:filter_hook/1` and is the same as `modMamLookups` but left as is.
-  [modPresenceSubscriptions], # not sure what to do with these 2  - updated in sm, maybe the name should be changed.
-  [modPresenceUnsubscriptions],
-  [modPrivacyGets], didn't touch these for now. The difference to `mod_mam` ones is that the hook is called from `ejabberd_c2s` - can these be sent without `mod_privacy` on? If so, maybe it should stay on by default? If this is not a concern should be moved to start with `mod_privacy`. Should anyway be replaced by a generic hook metric in my opinion.
-  [modPrivacyPush],
-  [modPrivacySets],
-  [modPrivacySetsActive],
-  [modPrivacySetsDefault],
-  [modPrivacyStanzaAll],
-  [modPrivacyStanzaBlocked],
-  [modPrivacyStanzaDenied],
-  [modRegisterCount], it does not actually count the users registered through `mod_register` but just all registered users. I would just delete these (maybe add a different metric for `mod_register` itself) and keep the automatic hook metric instead of this.
-  [modRosterGets], similarly to `modPrivacy*` is run from different places in the code even if `mod_roster` is not enabled.
-  [modRosterPush],
-  [modRosterSets],
-  [modUnregisterCount], similar to `modRegisterCount`.
-  [offline_groupchat_message_hook], gets created in sm because a handler is registered.
-  [offline_message_hook], same.
-  [roster_in_subscription], same.
-  [sessionAuthFails], could be replaced by an automatic hook metric.

Good ones:
- [sessionCount], counts sessionSuccessfulLogins times hosts.
- [sessionLogouts]
- [sessionSuccessfulLogins], same as sessionCount but per Host.
- [sm_broadcast], automatic hook, counts the number of modPrivacyPushes. `modPrivacyPush` metric is sm_broadcast times sessionCounts per Host.
- [xmppErrorIq],
- [xmppErrorMessage],
- [xmppErrorPresence],
- [xmppErrorTotal],
- [xmppIqReceived],
- [xmppIqSent],
- [xmppMessageBounced],
- [xmppMessageReceived],
- [xmppMessageSent],
- [xmppPresenceReceived],
- [xmppPresenceSent],
- [xmppStanzaCount],
- [xmppStanzaDropped],
- [xmppStanzaReceived],
- [xmppStanzaSent]


src/mam/mod_mam.erl
```diff
@@ -684,5 +682,30 @@ is_archivable_message(Host, Dir, Packet) ->
...
+    lists:foreach(fun(Name) ->
+                      mongoose_metrics:ensure_metric(Host, Name, spiral)
+                  end,
+                  spirals()),
+    Hooks = mongoose_metrics_mam_hooks:get_hooks(Host),
+    ejabberd_hooks:add(Hooks).
+
+spirals() ->
+    [modMamPrefsSets,
+    modMamPrefsGets,
```
chrzaszcz, line 701: Minor: indentation

chrzaszcz (approved):
Looks good! Let's leave `mod_privacy` for now.

PR #3098, opened 2021-04-23
Prepare queries in mongoose_cluster_id

leszke:

### DIFF
--- a/src/mongoose_cluster_id.erl
+++ b/src/mongoose_cluster_id.erl
@@ -89,7 +89,7 @@ maybe_prepare_queries(rdbms) ->
 
 -spec execute_cluster_insert_new(binary()) -> mongoose_rdbms:query_result().
 execute_cluster_insert_new(ID) ->
-    mongoose_rdbms:execute(global, cluster_insert_new, [ID]).
+    mongoose_rdbms:execute_successfully(global, cluster_insert_new, [ID]).
 
 -spec make_cluster_id() -> cluster_id().
 make_cluster_id() ->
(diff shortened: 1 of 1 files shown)


arcusfelis (approved):
looks ok

src/mongoose_cluster_id.erl
```diff
@@ -96,20 +110,8 @@ store_cluster_id(ID) ->
...
-    catch
-        Class:Reason:Stacktrace ->
-            ?LOG_WARNING(#{what => cluster_id_set_failed,
-                           text => <<"Error inserting cluster ID into RDBMS">>,
-                           sql_query => SQLQuery, cluster_id => ID,
-                           class => Class, reason => Reason, stacktrace => Stacktrace}),
-            {error, {Class, Reason}}
-    end;
+    Res = execute_cluster_insert_new(ID),
+    convert_insert_response(Res, ID);
```
chrzaszcz, line 114: I think we need to keep the try...catch logic here as the code intentionally falls back to Mnesia if RDBMS is faulty to prevent a crash.

src/mongoose_cluster_id.erl
```diff
@@ -122,18 +124,9 @@ set_new_cluster_id(ID, mnesia) ->
 %% Get cluster ID
 -spec get_backend_cluster_id(mongoose_backend()) -> maybe_cluster_id().
 get_backend_cluster_id(rdbms) ->
-    SQLQuery = [<<"SELECT v FROM mongoose_cluster_id WHERE k='cluster_id'">>],
-    try mongoose_rdbms:sql_query(global, SQLQuery) of
+    case mongoose_rdbms:execute_successfully(global, cluster_select, []) of
```
chrzaszcz, line 127: Could you call `mongoose_rdbms:execute` here to prevent crashing in case of DB errors? (see the comment above).

chrzaszcz (commented):
Nice to see the prepared queries introduced here. I have a few comments.

src/mongoose_cluster_id.erl
chrzaszcz: Update: we checked that MongooseIM crashes when the DB is not available anyway, so no need for try...catch.

src/mongoose_cluster_id.erl
chrzaszcz: Like above.

src/mongoose_cluster_id.erl
```diff
@@ -77,6 +78,19 @@ init_mnesia_cache() ->
...
+maybe_prepare_queries(rdbms) ->
+    mongoose_rdbms:prepare(cluster_insert_new, mongoose_cluster_id, [v],
+        <<"INSERT INTO mongoose_cluster_id(k,v) VALUES ('cluster_id', ?)">>),
+    mongoose_rdbms:prepare(cluster_select, mongoose_cluster_id, [],
+        <<"SELECT v FROM mongoose_cluster_id WHERE k='cluster_id'">>),
+    ok.
+
+-spec execute_cluster_insert_new(binary()) -> mongoose_rdbms:query_result().
+execute_cluster_insert_new(ID) ->
+    mongoose_rdbms:execute(global, cluster_insert_new, [ID]).
```
chrzaszcz, line 92: Please just use execute_successfully in both queries for consistency.

chrzaszcz (commented):
Looks good, one minor comment from me.

chrzaszcz (approved):
Looks good!

PR #3101, opened 2021-04-26
Pass the accumulator along

NelsonVides:
There could be more important data in this, like the mam-id, that we
might prefer rather than the message's attribute id.

For MUC, well, inbox with regular muc is not the most common use case.

### DIFF
--- a/src/inbox/mod_inbox_muc.erl
+++ b/src/inbox/mod_inbox_muc.erl
@@ -93,8 +93,8 @@ direction(From, To) ->
       Packet :: packet().
 handle_outgoing_message(Host, Room, To, Packet) ->
     maybe_reset_unread_count(Host, To, Room, Packet),
-    TS = erlang:system_time(microsecond),
-    maybe_write_to_inbox(Host, To, Room, Packet, TS, fun write_to_sender_inbox/5).
+    Acc = mongoose_acc:new(#{location => ?LOCATION, lserver => Host}),
+    maybe_write_to_inbox(Host, To, Room, Packet, Acc, fun write_to_sender_inbox/5).
 
 -spec handle_incoming_message(Host, Room, To, Packet) -> term() when
       Host :: receiver_host(),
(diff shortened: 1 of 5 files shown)


arcusfelis (approved):
ok

src/inbox/mod_inbox_muc.erl
```diff
@@ -93,29 +93,31 @@ direction(From, To) ->
...
 
 -spec handle_incoming_message(Host, Room, To, Packet) -> term() when
       Host :: receiver_host(),
       Room :: room_bare_jid(),
       To :: receiver_bare_user_jid(),
       Packet :: packet().
 handle_incoming_message(Host, Room, To, Packet) ->
-    TS = erlang:system_time(microsecond),
-    maybe_write_to_inbox(Host, Room, To, Packet, TS, fun write_to_receiver_inbox/5).
+    % Acc = #{mongoose_acc => true,
```
DenysGonchar, line 105: looks like left over

NelsonVides: 🤦🏽‍♂️

DenysGonchar (approved):
looks good

src/inbox/mod_inbox_muc.erl
```diff
@@ -93,29 +93,29 @@ direction(From, To) ->
       Packet :: packet().
 handle_outgoing_message(Host, Room, To, Packet) ->
     maybe_reset_unread_count(Host, To, Room, Packet),
-    TS = erlang:system_time(microsecond),
-    maybe_write_to_inbox(Host, To, Room, Packet, TS, fun write_to_sender_inbox/5).
+    Acc = mongoose_acc:new(#{location => ?LOCATION, lserver => Host, host_type => undefined}),
```
chrzaszcz, line 96: Actually in https://github.com/esl/MongooseIM/pull/3097 I am trying to ensure the consistency between `lserver` and `host_type` in accumulators. For me this is against the principle of least surprise, hopefully we could rework this code later to accept `Acc` as one of its arguments.

NelsonVides: Makes sense, but muc code is a bit unfortunate to not really use accs, so I think it would need quite a complex muc rework, then it would all be much better 😕

PR #3102, opened 2021-04-27
Fix Swagger docs rendering

gustawlippa:
My first front-end PR. The Swagger part was obscured on the left side and could not even be scrolled to see the content. Affected places: https://esl.github.io/MongooseDocs/latest/rest-api/Client-frontend/ and https://esl.github.io/MongooseDocs/latest/rest-api/Administration-backend/. I think this should be the fix but I have no idea what I am doing.

Before:
Screenshot_2021-04-27 Administration backend - MongooseIM
After:
Screenshot_2021-04-27 Administration backend - MongooseIM(1)


arcusfelis (approved):
good

PR #3103, opened 2021-04-27
Handle remove_domain in mod_inbox

arcusfelis:
This PR addresses "tenant removal epic".

Proposed changes include:
* Reorder indexes in the inbox table
* Add a prepared query for removal
* Add host_type type

### DIFF
--- a/doc/developers-guide/Testing-MongooseIM.md
+++ b/doc/developers-guide/Testing-MongooseIM.md
@@ -108,6 +108,30 @@ The command can be used instead:
 And `mam` is used to run `mam_SUITE` suite only.
 
 
+## Debugging big tests database
+
+This command opens MySQL shell interface:
+
+```bash
+./tools/open-test-database-shell.sh mysql
+```
+
+This command opens PgSQL shell interface:
+
+```bash
+./tools/open-test-database-shell.sh pgsql
+```
+
+This command opens [MSSQL shell interface](https://docs.microsoft.com/en-us/sql/tools/sqlcmd-utility):
+
+```bash
+./tools/open-test-database-shell.sh mssql
+```
+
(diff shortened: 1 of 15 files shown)


doc/developers-guide/Testing-MongooseIM.md
```diff
@@ -108,6 +108,30 @@ The command can be used instead:
 And `mam` is used to run `mam_SUITE` suite only.
 
 
+## Debugging big tests database
```
DenysGonchar, line 111: absolutely love it.
it's needed probably twice a year but always time consuming to recall how to do it. thanks a lot

src/inbox/mod_inbox.erl
```diff
@@ -267,6 +271,12 @@ remove_user(Acc, User, Server) ->
     mod_inbox_utils:clear_inbox(User, Server),
     Acc.
 
+-spec remove_domain(mongoose_acc:t(), mongooseim:host_type(), jid:lserver()) ->
```
DenysGonchar, line 274: it's not `mongoose_acc:t()`, MB we need some special type to highlight that Acc is not used!

src/mongooseim.erl
```diff
@@ -15,6 +15,8 @@
 %%==============================================================================
 -module(mongooseim).
 
+-type host_type() :: binary().
```
DenysGonchar, line 18: 👍

DenysGonchar (approved):
Looks good, just one minor comment regarding Acc type. But it can be changed later in the following PRs

src/inbox/mod_inbox.erl
arcusfelis: ok, changed.

PR #3104, opened 2021-04-27
rework of the gen_mod module, part #1

DenysGonchar:
this PR contains these changes:
* rework of the subdomain (template) configurations.
* rework of the `virtual_pubsub_hosts` configuration (documentation should be updated in a separate PR).
* beginning of the `gen_mod` module rework (to be continued in the following PRs).
* `mongoose_domain_core:for_each_domain/2` interface (to be used in the following PR).


src/domain/mongoose_subdomain_utils.erl
```diff
@@ -0,0 +1,32 @@
...
+
+%% API
+-export([make_subdomain_pattern/1,
+         get_fqdn/2,
+         is_subdomain/3]).
+
+-type subdomain_pattern() :: {fqdn | prefix, binary()}.
+-export_type([subdomain_pattern/0]).
+
+-spec make_subdomain_pattern(SubdomainPatternConfigOpt :: binary() | string()) ->
```
arcusfelis, line 11: maybe some comment on what it does and why is needed.

arcusfelis (approved):
good

src/gen_mod.erl
```diff
@@ -84,23 +79,24 @@
...
 
 -record(ejabberd_module, {
-          module_host, % {module(), jid:server()},
+          module_host_type, % {module(), host_type()},
           opts % list()
          }).
 
 -type module_feature() :: atom().
+-type domain_name() :: jid:server().
+-type host_type() :: binary().
```
chrzaszcz, line 88: There is `mongoose:host_type()` already.

src/gen_mod.erl
```diff
@@ -110,60 +106,68 @@
 %% case of cycle (in that case soft dependency may be started after the
 %% dependent module).
 %%
-%% -callback deps(Host :: jid:server(), Opts :: proplists:list()) ->
-%%     deps_list().
+%% TODO: think about getting rid of HostType param for deps/2 interface, currently
+%% it's used only by global_distrib modules (see mod_global_distrib_utils:deps/4
+%% function).
+-callback deps(HostType :: binary(), Opts :: proplists:list()) -> deps_list().
```
chrzaszcz, line 112: Here the type is `binary()`. We shouldn't allow the types to be so random.

src/gen_mod.erl
```diff
@@ -277,27 +282,21 @@ config_spec(Module) ->
...
-    receive
-        {'DOWN', MonitorReference, _Type, _Object, _Info} ->
-            ok
-    after 5000 ->
+    case wait_for_stop(MonitorReference) of
+        ok -> ok;
+        timeout ->
             catch exit(whereis(Process), kill),
-            wait_for_stop1(MonitorReference)
+            wait_for_stop(MonitorReference),
```
chrzaszcz, line 289: This is a significant change of behaviour - before it waited up to 2*5 = 10 seconds, now it can get stuck in an infinite loop. Of course it should't if the `DOWN` message is received, but I wouldn't set up such a trap in the code. Let's have a maximum retry count here.

src/event_pusher/mod_event_pusher_push.erl
```diff
@@ -273,18 +257,10 @@ parse_form(Form) ->
...
-            store_session_info(From, {BarePubSubJID, Node, FormFields}),
-            IQ#iq{type = result, sub_el = []};
-        false ->
-            NewSubEl = [IQ#iq.sub_el, mongoose_xmpp_errors:remote_server_not_found()],
-            IQ#iq{type = error, sub_el = NewSubEl}
-    end.
+-spec enable_node(jid:jid(), jid:jid(), pubsub_node(), form(), jlib:iq()) ->
+    ok | {error, Reason :: term()}.
+enable_node(From, BarePubSubJID, Node, FormFields, IQ) ->
+    mod_event_pusher_push_backend:enable(jid:to_bare(From), BarePubSubJID, Node, FormFields).
```
chrzaszcz, line 263: Why did we have this check in the code? I am not sure if it can be just removed, this changes the functionality.
Why not check it with `mongoose_domain_api:get_host_type` (and maybe use the version for subdomains that we are going to introduce)?

test/mongoose_domain_core_SUITE.erl
```diff
@@ -138,15 +138,13 @@ can_get_outdated_domains(_) ->
     [] = mongoose_domain_core:get_all_outdated(another_dummy_src).
 
 run_for_each_domains(_) ->
-    NumOfDomains = 10,
+    NumOfDomains = 1234,
```
chrzaszcz, line 141: Please add a comment so it does not seem to be a magic random number. It took me a while to figure it out and I know why we need it to large (above the batch size), but it's not obvious.

chrzaszcz (commented):
I like the introduction of domain name patterns, good work!

One general comment: We need to keep the types consistent, they are being introduced ad-hoc now and I am not liking it very much as it's becoming more and more random. I'd rather introduce them once and fix the whole codebase in one PR. Using `binary()`, `mongoose:host_type()` and locally defined `host_type()` interchangeably makes the code less readable. I am afraid that `domain_name()` will become another randomly used type.

test/mongoose_domain_core_SUITE.erl
```diff
@@ -23,7 +23,8 @@ all() ->
      get_all_static,
      get_domains_by_host_type,
      host_type_check,
-     can_get_outdated_domains].
+     can_get_outdated_domains,
+     run_for_each_domains].
```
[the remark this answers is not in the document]
DenysGonchar: damn, i was selecting between `run for all domains` and `run for each domain`... and it looks like i overuse autocompletion if I haven't even noticed that typo.

src/mongooseim.erl
```diff
@@ -16,7 +16,8 @@
 -module(mongooseim).
 
 -type host_type() :: binary().
--export_type([host_type/0]).
+-type domain_name() :: jid:lserver().
```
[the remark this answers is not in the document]
DenysGonchar: it's hard to judge, when exactly we should introduce this type, I would say a soon as possible. I really need it for gen_mod.

src/config/mongoose_config_validator.erl
```diff
@@ -143,8 +141,12 @@ validate_binary_domain(Domain) when is_binary(Domain) ->
     #jid{luser = <<>>, lresource = <<>>} = jid:from_binary(Domain),
     validate_domain_res(binary_to_list(Domain)).
 
-validate_domain_template(Domain, Pattern) ->
-    validate_binary_domain(re:replace(Domain, Pattern, <<"example.com">>, [{return, binary}])).
+validate_subdomain_template(SubdomainTemplate) ->
+    Pattern = mongoose_subdomain_utils:make_subdomain_pattern(SubdomainTemplate),
+    Domain = mongoose_subdomain_utils:get_fqdn(Pattern, <<"example.com">>),
+    %% TODO: do we want warning printed by validate_domain_res, especially when
```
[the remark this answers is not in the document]
DenysGonchar: I'm not sure that we need DNS validation at all

src/gen_mod.erl
DenysGonchar: yes, right. i started work on this PR before Mikhail introduced `mongoose:host_type()` type, and then just forgot to change it after rebasing.

src/gen_mod.erl
DenysGonchar: It's not a recursive call, `wait_for_process/1` function just calls `wait_for_stop/1` 2 times. So the behaviour is exactly the same.

src/event_pusher/mod_event_pusher_push.erl
DenysGonchar: The check is incorrect, it's against XMPP push XEP(s). we can specify the node from another federation here, and that check will fail. I think it was accidentally ported from some private code into open-source MIM.

src/config/mongoose_config_validator.erl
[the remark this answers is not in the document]
DenysGonchar: Anyway, I would rather want to do this in another PR.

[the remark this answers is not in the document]
chrzaszcz: It should be done in a separate PR for sure. @arcusfelis, you introduced the DNS check, are you OK to remove it from config validators? For me only syntax should be checked here.

src/mongooseim.erl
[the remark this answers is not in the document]
chrzaszcz: But why not `jid:lserver()`? The domain part of a JID is an XMPP domain name, why reinvent it?

src/event_pusher/mod_event_pusher_push.erl
chrzaszcz: I don't know much about this functionality, by "against XMPP push XEP(s)" do you mean that XEP-0357 does not mention any checks on the server side in the [Enabling](https://xmpp.org/extensions/xep-0357.html#enabling) section?

Just asking: "node from another federation" - do you mean that the domain is not in the MIM routing table (yet)?

src/mongooseim.erl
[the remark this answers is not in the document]
DenysGonchar: The main reason is that an old Host concept uses `jid:lserver()` and I want to make the wrong code look wrong. I don't want to guess if some code that uses `jid:lserver()` is already checked, and it's correct. Or no one did the check yet, and it must be performed. Ideally we should get rid of the `host` term in codebase at all, and it should be replaced with either `host type` or `domain` (or `domain name`).  It significantly simplifies formal verification.

src/event_pusher/mod_event_pusher_push.erl
DenysGonchar: s2s connections are not reflected in ejabberd_router tables at all, they are managed by `ejabberd_s2s` module. but even if don't have any established s2s connection yet, it doesn't mean that we should reject such a pubsub node. because normally s2s connection is established on the first routing.

src/domain/mongoose_subdomain_utils.erl
DenysGonchar: added

src/mongooseim.erl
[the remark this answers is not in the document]
chrzaszcz: > The main reason is that an old Host concept uses `jid:lserver()` and I want to make the wrong code look wrong.

`jid:lserver()` is still the best option in many places. How would you differentiate between something not changed yet and not needing to be changed at all? If your answer is "change `jid:lserver` to `domain_name`" everywhere in the code, this doesn't make sense to me.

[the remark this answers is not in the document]
DenysGonchar: why everywhere, functions that process stanzas/JIDs/etc. doesn't have to change anything. But every `Host`/`jid:lserver()` input parameter must be replaced with `HostType`/`mongooseim:host_type()` or `Domain/mongooseim:domain_name()` depending on the logic behind the code. It must be clear if that parameter is about configuration or routing entity.

src/config/mongoose_config_validator.erl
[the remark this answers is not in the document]
arcusfelis: maybe. I still think dns check is important for many cases. at least as a warning.

src/mongooseim.erl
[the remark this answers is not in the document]
chrzaszcz: Let's say function `a` takes `#jid{lserver = LServer}` as an argument and passes `LServer` to function `b`.

You would have code like this:

```
a(#jid{lserver = LServer}) ->
   b(LServer).

b(LServer) ->
  do_something(...).
```

No if you change the argument of function `b` to `Domain` with the type `domain_name()`, how would you address that in function `a`? Neither`a(#jid{lserver = Domain})` nor calling `b(LServer)` look good to me.

The functions in this example were made-up, but there are many real cases like this in the code. If we agree on a rule which we both accept and understand, I would agree to proceed with the change.

The situation for host types is completely different and I agree with you on that one.

PR #3106, opened 2021-04-29
Support dynamic domains in auth backends

leszke:
Rework the auth backends to support host types and dynamic domains.
- New `mongoose_gen_auth` behaviour is introduced with updated and simplified callbacks and a new API.
- The backend-handling logic in `mongoose_auth` is unified and updated.
- Backend modules support multiple domains now.
- Unit tests use host types.

The following is  intentionally **not** done yet, but it can be done in upcoming PR's:
- Handling `remove_domain` in backends.
- Rework of the `ejabberd_auth` API functions e.g. `get_vh_*` to make them more consistent with the backend API.
- Similar rework of admin commands that are using 'vhosts'.
- Rework of error handling in `ejabberd_auth`, which differs between API functions and can be unified. After unification, add more tests for `ejabberd_auth`.
- Unit tests for backends including `ejabberd_auth_anonymous`, `ejabberd_auth_external`
- Migration guide for RDBMS users.
- Warnings in docs that users number estimates don't support multiple domains and LDAP registration works only for a single domain
- Experimented with, but not done (not convinced): putting the `LUser, LServer, HostType` triplet in a map, maybe extended with optional password data.
- Remove `supported_features/0` - why have it when all backends support dynamic domains?

For details, see individual commits.


src/auth/ejabberd_auth_dummy.erl
```diff
@@ -52,50 +48,40 @@ check_password(_HostType, User, Server, _Password, _Digest, _DigestGen) ->
...
 
 %% @spec (User, Server) -> ok
 %% @doc Remove user.
 %% Note: it returns ok even if there was some problem removing the user.
-remove_user(_User, _Server) ->
+-spec remove_user(mongooseim:host_type(), jid:luser(), jid:lserver()) -> ok.
+remove_user(_HostType, _User, _Server) ->
     ok.
 
+-spec remove_domain(mongooseim:host_type(), jid:lserver()) -> ok.
```
gustawlippa, line 78: I think this is not entirely correct typing - there is a `simple_acc` type in `mongoose_hooks`, which is just a map, and it's used only for the remove domain hooks. I would personally propose getting rid of this type from `mongoose_hooks`, but it may serve some purpose. It's hard to tell, because the hook doesn't seem to be run from any code apart from the tests.
The other comment is that this handler is present only in the dummy module out of the auth modules, right? Is it an oversight?

gustawlippa: Oh, sorry, for some reason I misread the spec... Obviously this is fine - I'll edit the comment. But still it's a callback that is present only in the dummy module - is it deliberate?

chrzaszcz: Handling domain removal has to be added to the TODO list. Doing it now.

DenysGonchar (approved):
Generally looks good, the only thing that make me worrying that we don't have tests for `ejabberd_auth` module itself. We should definitely create some unit tests for it with multiple mocked auth methods that return different values to ensure that checks are done until the first successful and in the same sequence as methods configured. Also, that would be nice to ensure that `mongoose_gen_auth` handles properly missing optional callbacks (can be done in the same ejabberd_auth unit tests). Being honest, I would trust such unit tests more than any code review, because it's a bit hard to follow all the changes.

Approving this PR, but please let's create a ticket for ejabberd_auth unit tests.

chrzaszcz:
> Generally looks good, the only thing that make me worrying that we don't have tests for `ejabberd_auth` module itself. We should definitely create some unit tests for it with multiple mocked auth methods that return different values to ensure that checks are done until the first successful and in the same sequence as methods configured. Also, that would be nice to ensure that `mongoose_gen_auth` handles properly missing optional callbacks (can be done in the same ejabberd_auth unit tests). Being honest, I would trust such unit tests more than any code review, because it's a bit hard to follow all the changes.
> 
> Approving this PR, but please let's create a ticket for ejabberd_auth unit tests.

I think we could add more tests when the and unification rework of error handling in `ejabbed_auth` is done. We would need to look at the case with multiple backends configured once again, because IMO it does not make much sense for most backends and the behaviour is very confusing, especially when one backend returns an error. The good thing is that most of the functionality is already covered by functional tests. Added to the TODO list above. I will create tasks from that list when this PR is merged.

big_tests/src/ct_mongoose_hook.erl
```diff
@@ -134,14 +134,15 @@ check_server_purity(Suite, Config) ->
...
-        fun check_roster/0],
+            fun check_registered_users/0,
+            fun check_registered_users_count/0,
+            fun check_offline_messages/0,
+            fun check_active_users/0,
+            fun check_privacy/0,
+            fun check_private/0,
+            fun check_vcard/0,
+            fun check_roster/0,
+            fun check_carboncopy/0],
```
NelsonVides, line 145: @chrzaszcz I've just realised, this is checking an ets table called `carboncopy` that doesn't really exists in mim 🤔

PR #3109, opened 2021-05-05
Prepare for dynamic domains in big tests

chrzaszcz:
This PR makes it possible to run big tests with the dynamic domains test spec.

It includes a few separate changes, but they are all connected:
- Introducing helpers for host types that are used in `acc_e2e_SUITE`.
- Replacing `get_host_type` with `get_domain_host_type`...
- ...which makes it safe to temporarily enable `mongoose_subhosts`...
- ...which is needed to run `filter_local_packet` for host types...
- ...which is needed by the tests.

The last commit is optional. The code can work without it, but after spending some time on testing different versions this one seems to be the best. I explained the motivation in the commit message.

This PR does **not** include tests for the bug related to the use in `get_host_type` in too many places. I think it can be done in a follow-up PR. Manual test shows that the bug is not present anymore.

### DIFF
--- a/big_tests/dynamic_domains.spec
+++ b/big_tests/dynamic_domains.spec
@@ -9,7 +9,7 @@
 %% http://www.erlang.org/doc/apps/common_test/run_test_chapter.html#test_specifications
 {include, "tests"}.
 
-{suites, "tests", domain_isolation_SUITE}.
+{suites, "tests", acc_e2e_SUITE}.
 
 {config, ["dynamic_domains.config", "test.config"]}.
 {logdir, "ct_report"}.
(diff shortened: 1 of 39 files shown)


big_tests/dynamic_domains.spec
```diff
@@ -9,7 +9,7 @@
 %% http://www.erlang.org/doc/apps/common_test/run_test_chapter.html#test_specifications
 {include, "tests"}.
 
-{suites, "tests", domain_isolation_SUITE}.
+{suites, "tests", acc_e2e_SUITE}.
```
DenysGonchar, line 12: why do we remove domain_isolation_SUITE?

DenysGonchar (approved):
looks fine

big_tests/dynamic_domains.spec
chrzaszcz: This suite was passing because `filter_local_packet` was executed per domain. I could update the test suite to use host types correctly, but the MUC test would fail. Can I do it in another PR?

PR #3110, opened 2021-05-05
Prevent infinite loop when domain isolation is on on both domains

arcusfelis:
This PR addresses MIM-1395.

Proposed changes include:
* Fixes "Domain isolation causes an infinite loop".
* A test.

**Description:**
Step to reproduce:

    Enable domain isolation for two domains, e.g.: `localhost`, `localhost.bis`

    Send any message between the two domains, e.g. from `alice@localhost.bis` (A) to `bob@localhost` (B)

`mod_domain_isolation` handles message from A to B by sending an error reply from B to A, which in turn results in an error reply etc... error elements are accumulating forever in the packet (actually: until the hook timeout passes), eating memory and time.

Add test to `domain_isolation_SUITE`, fix the issue by not bouncing the error message back.

### DIFF
--- a/big_tests/tests/domain_isolation_SUITE.erl
+++ b/big_tests/tests/domain_isolation_SUITE.erl
@@ -12,10 +12,13 @@ suite() ->
 all() ->
     [
      isolation_works_for_one2one,
+     isolation_works_for_one2one_2domains,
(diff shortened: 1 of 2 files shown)


chrzaszcz (approved):
Looks good

PR #3112, opened 2021-05-07
Don't set cluster_id if not needed

arcusfelis:
This PR addresses "MIM-1396 Prepared query in mongoose_cluster_id causes errors on startup"

Proposed changes include:
* Don't insert, if not needed.
* Insert only when we generate a new ID.

Basically, we implement "select and just use it. If not found, create new and insert" logic.

It is a bit less robust: 
- if a cluster got started without RDBMS, but after it got a new node with RDBMS, we would not get an update in DB.
- but this use case is _pretty weird_

There is still a race condition possible, when we try to insert, but someone has already inserted a ClusterID. we could handle it and retry, but it makes logic less clean. And I am pretty sure it's a rare case when two nodes go live for the first time. Could be avoided by adding retries, but nothing bad would happen anyway (we would see a warning in this case in logs, but that's it).

### DIFF
--- a/src/mongoose_cluster_id.erl
+++ b/src/mongoose_cluster_id.erl
@@ -20,22 +20,25 @@
 -spec start() -> maybe_cluster_id().
 start() ->
     init_mnesia_cache(),
-    maybe_prepare_queries(which_backend_available()),
-    run_steps(
-      [fun get_cached_cluster_id/0,
-       fun get_backend_cluster_id/0,
-       fun make_and_set_new_cluster_id/0],
-      {error, no_cluster_id_yet}).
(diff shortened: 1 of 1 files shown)


chrzaszcz (commented):
This looks cleaner than before, the only issue is with the statement from the description:

> we would see a warning in this case in logs, but that's it

This isn't completely true: we will get an error **and** a warning because of `execute_successfully`. I logged the original issue because the error was logged always, now it can happen _sometimes_, but I think it's acceptable, I just want to let you know.

chrzaszcz (approved):

PR #3118, opened 2021-05-14
Multi tenancy iq handlers

DenysGonchar:
this PR introduces iq handling  for dynamic domains. here is the list of all changes:
* introduction of `mongoose_iq_handlers`
* enforcing map as type of `packet_handler`'s extra field and extending it with `host_type` for dynamic domains
* introducing `mongoose_lazy_routing` module that controls dynamic domains/subdomains routing and iq handlers registration.
* splitting gen_iq_handler into 3 different modules:
   * iq handler registration API
   * iq component behaviour
   * iq handlers' workers
* introducing demo module
   
TODO: introduce unit testing for `mongoose_lazy_routing` module


big_tests/tests/dynamic_domains_pm_SUITE.erl
```diff
@@ -1,8 +1,12 @@
 -module(dynamic_domains_pm_SUITE).
 
+-include_lib("exml/include/exml.hrl").
+
 %% API
 -compile(export_all).
--import(distributed_helper, [mim/0, mim2/0, require_rpc_nodes/1, rpc/4]).
+-import(distributed_helper, [mim/0, mim2/0, rpc/4,
+require_rpc_nodes/1,
```
arcusfelis, line 8: indention

big_tests/tests/dynamic_domains_pm_SUITE.erl
```diff
@@ -85,3 +143,6 @@ uncluster_nodes([], Config) -> Config;
 uncluster_nodes([Node | T], Config) ->
     NewConfig = distributed_helper:remove_node_from_cluster(Node, Config),
     cluster_nodes(T, NewConfig).
+
+dump_ets_tables(Tables) ->
```
arcusfelis, line 147: remove if not used.

big_tests/tests/dynamic_domains_pm_SUITE.erl
```diff
@@ -69,6 +74,59 @@ auth_domain_removal_is_triggered_on_hook(_Config) ->
...
+            [begin
+                 IQ = escalus_stanza:iq(Domain, <<"get">>, [QueryEl]),
+                 escalus:send_iq_and_wait_for_result(Alice, IQ)
+             end || Domain <- ?DOMAINS],
+            2 = rpc(mim(), meck, num_calls, [mod_dynamic_domains_test, process_iq, 5]),
+            0 = rpc(mim(), meck, num_calls, [mod_dynamic_domains_test, process_packet, 5]),
+            %% check that process_iq/5 is called from one and the same worker process
+            History = rpc(mim(), meck, history, [mod_dynamic_domains_test]),
+            rpc(mim(), meck, reset, [mod_dynamic_domains_test]),
+            Pids = [Pid||{Pid,{_,process_iq,_},_}<-History],
```
arcusfelis, line 114:
```
            Pids = [Pid || {Pid,{_,process_iq,_},_} <- History],
```

src/ejabberd_local.erl
```diff
@@ -333,38 +320,27 @@ handle_cast(_Msg, State) ->
...
     catch mod_disco:register_feature(Host, XMLNS),
     {noreply, State};
 handle_info({unregister_iq_handler, Host, XMLNS}, State) ->
     case ets:lookup(?IQTABLE, {XMLNS, Host}) of
-        [{_, Module, Function, Opts}] ->
-            gen_iq_handler:stop_iq_handler(Module, Function, Opts);
+        [{_, IQHandler}] ->
+            gen_iq_component:stop_iq_handler(IQHandler),
+            ets:delete(?IQTABLE, {XMLNS, Host}),
+            catch mod_disco:unregister_feature(Host, XMLNS);
```
arcusfelis, line 334: these catches would make life harder...
`try ... catch` with logging would be better.

src/gen_iq_handler.erl
```diff
@@ -26,204 +26,116 @@
...
+                     Namespace :: binary(),
+                     Module :: atom(),
+                     Function :: atom(),
+                     ExecutionType :: mongoose_iq_handler:execution_type()) -> any().
+add_iq_handler(Component, Domain, Namespace, Module, Function, ExecutionType) ->
+    Extra = #{delete_on_unregister => true, module => Module, function => Function},
+    IQHandlerFn = make_iq_handler_fn(Module, Function),
+    IQHandler = mongoose_iq_handler:new(IQHandlerFn, Extra, ExecutionType),
+    gen_iq_component:register_iq_handler(Component, Domain, Namespace, IQHandler).
+
```
arcusfelis, line 57: one new line maximum, please.

arcusfelis (commented):
added some style comments, but good

src/domain/mongoose_lazy_routing.erl
```diff
@@ -0,0 +1,397 @@
+%% Generally, you should not call anything from this module directly.
```
chrzaszcz, line 1: I'd rather see a comment documenting the purpose of this module at the top. Could you add two sentences describing why the laziness is needed and what it means?

src/domain/mongoose_lazy_routing.erl
```diff
@@ -0,0 +1,397 @@
...
+                       namespace = '_' :: binary() | '_',
+                       component = '_' :: module() | '_'}).
+
+-type subdomain_pattern() :: mongoose_subdomain_utils:subdomain_pattern().
+-type subdomain_info() :: mongoose_subdomain_core:subdomain_info().
+-type host_type() :: mongooseim:host_type().
+-type domain() :: mongooseim:domain_name().
+
+-type iq_record() :: {#iq_table_key{}, mongoose_iq_handler:t()}.
+-type domain_record() :: {domain(), host_type() | {host_type(), subdomain_pattern()}}.
```
chrzaszcz, line 45: Could you rename these types? The `_record` suffix is misleading, they are not records. Could be `_entry`.

src/mongoose_router_dynamic_domains.erl
```diff
@@ -27,10 +26,9 @@ filter(From, To, Acc, Packet) ->
 
 route(From, To, Acc, Packet) ->
     LDstDomain = To#jid.lserver,
-    case mongoose_domain_api:get_host_type(LDstDomain) of
-        {ok, _} ->
-            ejabberd_local:register_host(LDstDomain),
+    case mongoose_lazy_routing:maybe_add_domain_or_subdomain(LDstDomain) of
```
chrzaszcz, line 29: We can't create bottlenecks like this. If someone sends a lot of messages to unknown domains, this will time out. We should verify the host type **before** any synchronous code is called.

src/domain/mongoose_lazy_routing.erl
```diff
@@ -0,0 +1,397 @@
...
+    end.
+
+-spec add_domain(host_type(), domain()) -> ok.
+add_domain(HostType, Domain) ->
+    DomainElement = {Domain, HostType},
+    case ets:insert_new(?ROUTING_TABLE, DomainElement) of
+        true ->
+            IQs = get_iqs(#iq_table_key{host_type = HostType}),
+            register_iqs(IQs, [DomainElement]),
+            ejabberd_local:register_host(Domain),
```
chrzaszcz, line 303: This is going to add hook handlers per domain, not host type, see the code in `ejabberd_local`

src/domain/mongoose_lazy_routing.erl
```diff
@@ -0,0 +1,397 @@
...
+            ok;
+        false ->
+            %% we should never get here, but it's ok to just ignore this.
```
chrzaszcz, line 306: So if this gets called multiple times (because of a bug), we would silently ignore this fact.

big_tests/tests/dynamic_domains_pm_SUITE.erl
```diff
@@ -69,6 +74,59 @@ auth_domain_removal_is_triggered_on_hook(_Config) ->
     1 = rpc(mim(), meck, num_calls, [ejabberd_auth_dummy, remove_domain, Params]),
     rpc(mim(), meck, unload, [ejabberd_auth_dummy]).
 
+test_routing(Config) ->
```
chrzaszcz, line 77: Please split into 3 test cases.

big_tests/tests/dynamic_domains_pm_SUITE.erl
chrzaszcz, line 114: I'd even say
```suggestion
            Pids = [Pid || {Pid, {_, process_iq, _}, _} <- History],
```

chrzaszcz (commented):
Looks good and quite complicated at the same time... There are some `gen_server`s introduced, which can always become bottlenecks. Please double-check that all the calls really need to be synchronized this way.

src/ejabberd_local.erl
DenysGonchar: This is out of scope of this PR. mod_disco rework is planned for the future.

src/mongoose_packet_handler.erl
```diff
@@ -59,3 +59,9 @@ module(#packet_handler{ module = Module }) ->
 
 extra(#packet_handler{ extra = Extra }) ->
     Extra.
+
+add_extra(#packet_handler{ extra = OldExtra } = Handler, Extra) ->
+    %% KV pairs from the OldExtra map will remain unchanged, only
+    %% the new keys from Extra map will be added to the NewExtra map
```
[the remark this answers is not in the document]
DenysGonchar: I find it much more readable when all the lines of the block comment have more or less the same length.

src/mod_muc_iq.erl
```diff
@@ -92,11 +98,17 @@ handle_call(_Request, _From, State) ->
...
-handle_cast({register_iq_handler, Host, XMLNS, Module, Function, Opts}, State) ->
-    ets:insert(tbl_name(), {{XMLNS, Host}, Module, Function, Opts}),
+handle_cast({register_iq_handler, Host, XMLNS, IQHandler}, State) ->
+    ets:insert(tbl_name(), {{XMLNS, Host}, IQHandler}),
     {noreply, State};
 handle_cast({unregister_iq_handler, Host, XMLNS}, State) ->
-    ets:delete(tbl_name(), {XMLNS, Host}),
+    case ets:lookup(tbl_name(), {XMLNS, Host}) of
+        [{_, IQHandler}] ->
+            gen_iq_componentr:stop_iq_handler(IQHandler),
```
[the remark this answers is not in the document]
DenysGonchar: I'm surprised that dialyzer haven't caught it.  We must get rid of dynamically compiled modules and run xref check as part of CI testing.

src/domain/mongoose_lazy_routing.erl
DenysGonchar: ejabberd_local is about routing. It works with domains, not host types. the way how ejabberd_local triggers hooks must be reworked (and we can supply host_types here if required), but that is a topic for another PR.

NelsonVides:
I've seen that ejabberd_local has now its ets table as protected, why can't we do the same for ejabberd_sm' ets table? 🤔

src/mongoose_iq_handler.erl
```diff
@@ -0,0 +1,133 @@
...
+
+%%----------------------------------------------------------------------
+%% Types
+%%----------------------------------------------------------------------
+
+-record(iq_handler, {iq_handler_fn :: iq_handler(),
+                     extra :: map(),
+                     execution_method :: execution_method()}).
+
+-type t() :: #iq_handler{}.
```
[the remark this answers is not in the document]
NelsonVides: This one is a common pattern, when the module name already says everything about the type, and the type is exported and referenced as `mongoose_iq_handler:t()`, so I'm kinda ok with this convention 🤔

DenysGonchar:
> I've seen that ejabberd_local has now its ets table as protected, why can't we do the same for ejabberd_sm' ets table? 🤔

it is protected. see http://erlang.org/doc/man/ets.html#new-2 :
```
protected

    The owner process can read and write to the table. Other processes can only read the table. This is the default setting for the access rights.
```

NelsonVides:
> > I've seen that ejabberd_local has now its ets table as protected, why can't we do the same for ejabberd_sm' ets table? 🤔
> 
> it is protected. see http://erlang.org/doc/man/ets.html#new-2 :
> 
> ```
> protected
> 
>     The owner process can read and write to the table. Other processes can only read the table. This is the default setting for the access rights.
> ```

Ah, protected is the _default_, I see, thanks!

src/domain/mongoose_subdomain_core.erl
```diff
@@ -245,15 +248,16 @@ handle_remove_domain(HostType, Domain) ->
...
          end,
     lists:foreach(Fn, SubdomainItems).
 
--spec remove_subdomain(host_type(), domain()) -> true.
-remove_subdomain(HostType, Subdomain) ->
-    mongoose_hooks:disable_subdomain(HostType, Subdomain),
-    ets:delete(?SUBDOMAINS_TABLE, Subdomain).
+-spec remove_subdomain(subdomain_item()) -> ok.
+remove_subdomain(#subdomain_item{subdomain = Subdomain} = SubdomainItem) ->
+    ets:delete(?SUBDOMAINS_TABLE, Subdomain),
```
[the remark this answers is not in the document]
DenysGonchar: Actually the current sequence is exactly about avoiding the race condition. When we add domain/subdomains in 'mongoose_lazy_routing', we only do async check of the ETS table for it. So we must ensure that there's no record in ETS table by the time we remove domain/subdomain from 'mongoose_lazy_routing', and there is no way it's re-added before we remove domain/subdomains from ETS.

src/mongoose_iq_handler.erl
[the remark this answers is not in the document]
DenysGonchar: Fully agree with @NelsonVides, it's a common pattern for the modules that suppose to be a wrapper for just one structure.

src/gen_iq_component.erl
```diff
@@ -0,0 +1,57 @@
...
+                            Domain :: jid:server(),
+                            Namespace :: binary()) -> ok.
+unregister_iq_handler(Component, Domain, Namespace) ->
+    Component:unregister_iq_handler(Domain, Namespace).
+
+-spec sync(Component :: module()) -> ok.
+sync(Component) ->
+    Component:sync().
+
+-spec stop_iq_handler(IQHandler :: mongoose_iq_handler:t()) -> any().
```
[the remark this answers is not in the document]
DenysGonchar: `stop_iq_handler/1` must be removed completely in the nearest future, `delete_on_unregister` field is an ugly hack which I don't want to move anywhere else. And `handle/5` indeed can be eliminated.

src/mod_dynamic_domains_test.erl
```diff
@@ -0,0 +1,98 @@
...
+start(HostType, Opts) ->
+    Namespace = gen_mod:get_opt(namespace, Opts),
+
+    %% if we need to intercept traffic to some subdomain, we can create custom packet
+    %% handler and register it. not that IQ will be delivered to this packet handler
+    %% as well
+    SubdomainPattern1 = gen_mod:get_opt(host1, Opts),
+    Handler1 = mongoose_packet_handler:new(?MODULE),
+    mongoose_domain_api:register_subdomain(HostType, SubdomainPattern1, Handler1),
+    %% the call below makes no sense, it's added for demo/testing purposes only
```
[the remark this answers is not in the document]
DenysGonchar: actually, it's added here to explicitly verify/demonstrate in the test case that it has no affect and IQs are landing in `process_packet/5`. 

The misunderstanding of this collision resulted in hack with `mongoose_subhosts` triggering in `mongoose_router_localdomain`, which I had to fix in this PR as well.

src/mongoose_packet_handler.erl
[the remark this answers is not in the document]
chrzaszcz: I am trying to follow the ideas in https://sembr.org/ which really help with reading the comments. It's mostly for markup languages, but it helps me a lot with comments. I don't see a benefit of that stray "only" at the end of the sentence, but keep it if you must.

src/mongoose_router_dynamic_domains.erl
DenysGonchar: Makes sense, we can check ETS tables before running gen_server:call/2

src/mongoose_iq_handler.erl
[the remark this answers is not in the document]
chrzaszcz: If I was to guess what the `t` means, I would fail, especially that there is another type named `iq_handler`. For me it is a bad convention anyway, type names should be descriptive, we don't call functions 'f', variables 'v' and records 'r' for the very same reason. Let's not break one window, because others will follow.

src/mod_dynamic_domains_test.erl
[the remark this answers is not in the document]
chrzaszcz: Please reword the comment then and state what you just said, it's way more understandable than the comment in the module.

src/mongoose_iq_handler.erl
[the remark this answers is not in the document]
DenysGonchar: I don't see how the type `mongoose_iq_handler:t()` is better than `mongoose_iq_handler:iq_handler()`.

But you are absolutely right, the type that is currently called `iq_handler()` must be renamed into `iq_handler_fn()` to avoid collisions with `#iq_handler{}` record.

arcusfelis (approved):
k

PR #3120, opened 2021-05-17
Metrics per host type

gustawlippa:
This PR updates the metrics to use host types instead of hosts. There are still some places left that use host types and should be updated when the modules are updated (for example `mongoose_wpool_rdbms`).

Because the host type may be any string, the spaces `" "` are changed to underscores `"_"` in the metrics modules in a simple manner on every call. Because of that, when calling functions like `mongoose_metrics:get_metric_value/1`, which expect the _metric name_, one has to provide a sanitized name - this is not needed when calling the metrics API expecting host types.

There are also some small fixes, like a bugfix for a test suite that occurred seemingly only on my computer, and getting rid of the "Failing the test due failed rerun groups" message that was printed even if the tests were successful.

### DIFF
--- a/big_tests/tests/metrics_helper.erl
+++ b/big_tests/tests/metrics_helper.erl
@@ -12,10 +12,11 @@
 -define(ONLY_GLOBAL_METRICS_GROUP_USERS, [clusterguy, clusterbuddy]).
 
 get_counter_value(CounterName) ->
-    get_counter_value(ct:get_config({hosts, mim, domain}), CounterName).
+    get_counter_value(domain_helper:host_type(mim), CounterName).
 
-get_counter_value(Host, Metric) ->
-    case rpc(mim(), mongoose_metrics, get_metric_value, [Host, Metric]) of
+get_counter_value(HostType, Metric) ->
+    HostTypeName = make_host_type_name(HostType),
+    case rpc(mim(), mongoose_metrics, get_metric_value, [HostTypeName, Metric]) of
         {ok, [{count, Total}, {one, _}]} ->
             {value, Total};
         {ok, [{value, Value} | _]} when is_integer(Value) ->
@@ -29,8 +30,9 @@ get_counter_value(Host, Metric) ->
 assert_counter(Value, CounterName) ->
     assert_counter(ct:get_config({hosts, mim, domain}), Value, CounterName).
 
-assert_counter(Host, Value, CounterName) ->
-    {value, Value} = get_counter_value(Host, CounterName).
+assert_counter(HostType, Value, CounterName) ->
+    HostTypeName = make_host_type_name(HostType),
+    {value, Value} = get_counter_value(HostTypeName, CounterName).
 
 -spec prepare_by_all_metrics_are_global(Config :: list(), UseAllMetricsAreGlobal :: boolean()) ->
     list().
@@ -90,9 +92,14 @@ user_ids(Config) ->
 wait_for_counter(ExpectedValue, Counter) ->
         wait_for_counter(ct:get_config({hosts, mim, domain}), ExpectedValue, Counter).
 
-wait_for_counter(Host, ExpectedValue, Counter) ->
-        mongoose_helper:wait_until(fun() ->
-                                       assert_counter(Host, ExpectedValue, Counter)
+wait_for_counter(HostType, ExpectedValue, Counter) ->
+    mongoose_helper:wait_until(fun() ->
+                                       assert_counter(HostType, ExpectedValue, Counter)
                                    end,
                                    {value, ExpectedValue},
                                    #{name => Counter, time_left => ?WAIT_TIME, sleep_time => 20}).
+
+make_host_type_name(HT) when is_atom(HT) ->
+    HT;
+make_host_type_name(HT) when is_binary(HT) ->
+    binary:replace(HT, <<" ">>, <<"_">>, [global]).
--- a/big_tests/tests/metrics_helper.erl
+++ b/big_tests/tests/metrics_helper.erl
@@ -28,7 +28,7 @@ get_counter_value(HostType, Metric) ->
     end.
 
 assert_counter(Value, CounterName) ->
-    assert_counter(ct:get_config({hosts, mim, domain}), Value, CounterName).
+    assert_counter(domain_helper:host_type(mim), Value, CounterName).
 
 assert_counter(HostType, Value, CounterName) ->
     HostTypeName = make_host_type_name(HostType),
@@ -90,7 +90,7 @@ user_ids(Config) ->
     end.
 
 wait_for_counter(ExpectedValue, Counter) ->
-        wait_for_counter(ct:get_config({hosts, mim, domain}), ExpectedValue, Counter).
+        wait_for_counter(domain_helper:host_type(mim), ExpectedValue, Counter).
 
 wait_for_counter(HostType, ExpectedValue, Counter) ->
     mongoose_helper:wait_until(fun() ->
--- a/doc/operation-and-maintenance/MongooseIM-metrics.md
+++ b/doc/operation-and-maintenance/MongooseIM-metrics.md
@@ -6,7 +6,7 @@ MongooseIM uses [ESL's fork of this project](https://github.com/esl/exometer/tre
 
 All metrics are divided into the following groups:
 
-* Per host metrics: Gathered separately for every XMPP host supported by the cluster.
+* Per host type metrics: Gathered separately for every host type supported by the cluster.
  **Warning:** If a cluster supports many (thousands or more) domains, performance issues might occur.
  To avoid this, use global equivalents of the metrics with `all_metrics_are_global` config option.
     * Hook metrics.
@@ -60,7 +60,7 @@ A histogram collects values over a sliding window of 60s and exposes the followi
 * `median`
 * `50`, `75`, `90`, `95`, `99`, `999` - 50th, 75th, 90th, 95th, 99th and 99.9th percentile
 
-## Per host metrics
+## Per host type metrics
 
 ### Hook metrics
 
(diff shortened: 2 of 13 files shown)


doc/operation-and-maintenance/MongooseIM-metrics.md
```diff
@@ -6,7 +6,7 @@ MongooseIM uses [ESL's fork of this project](https://github.com/esl/exometer/tre
 
 All metrics are divided into the following groups:
 
-* Per host metrics: Gathered separately for every XMPP host supported by the cluster.
+* Per host type metrics: Gathered separately for every host type supported by the cluster.
  **Warning:** If a cluster supports many (thousands or more) domains, performance issues might occur.
  To avoid this, use global equivalents of the metrics with `all_metrics_are_global` config option.
```
NelsonVides, line 11: That warning is not exactly true anymore, it might as well be a "if a cluster supports many many host types", which, who the heck would do that? 🤷🏽‍♂️

gustawlippa: Yeah, that is true, I was also wondering about this line. My reasoning was, well, if there is no possibility of performance issues occurring... why do we keep the `all_metrics_are_global` actually? What would you propose writing here? Something like just "There is also a way to group metrics for all host types using the `all_metrics_are_global` config option" seems weak.

NelsonVides: I was thinking the same, but it is still possible to configure thousands of host_types, even if nobody should ever do that. Dunno, maybe somebody do want to serve 5k domains and have metrics granular for each, so he configures 5k host_types with just one domain each... and if somebody actually wants that then the `all_metrics_are_global` does nothing for him, so yeah, I'm failing to see why do we still want to keep that flag 🤔 Let's ask the rest of the team 🤔

NelsonVides (approved):
All seems good to me, but I'm not confident about merging this one, yet, so maybe let's see about more opinions @chrzaszcz @DenysGonchar

src/global_distrib/mod_global_distrib_utils.erl
```diff
@@ -43,6 +43,7 @@
 binary_to_metric_atom(Binary) ->
     List = lists:filtermap(fun
                                ($.) -> {true, $_};
+                               ($ ) -> {true, $_};
```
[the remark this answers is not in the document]
gustawlippa: I was thinking about it, but I think it would not match - `$ ` is a number and `" "` is a string (a list with one number). An alternative would be just writing 32...
obraz

doc/rest-api/Metrics-backend.md
```diff
@@ -61,39 +61,39 @@ Example response:
 
 Returns ```200 OK``` and two elements:
```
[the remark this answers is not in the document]
gustawlippa: I can add a note, but my thinking was the same - it's obsolete since 2017.

src/global_distrib/mod_global_distrib_utils.erl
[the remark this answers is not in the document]
NelsonVides: Right, good one 😄

big_tests/tests/metrics_helper.erl
```diff
@@ -29,8 +30,9 @@ get_counter_value(Host, Metric) ->
 assert_counter(Value, CounterName) ->
     assert_counter(ct:get_config({hosts, mim, domain}), Value, CounterName).
```
chrzaszcz, line 31: Shouldn't it be the host type?

big_tests/tests/metrics_helper.erl
```diff
@@ -90,9 +92,14 @@ user_ids(Config) ->
 wait_for_counter(ExpectedValue, Counter) ->
         wait_for_counter(ct:get_config({hosts, mim, domain}), ExpectedValue, Counter).
```
chrzaszcz, line 93: Same here...

src/metrics/mongoose_metrics_hooks.erl
```diff
@@ -153,89 +154,99 @@ xmpp_send_element(Acc, _El) ->
...
     Acc.
 
--spec roster_set(Acc :: map(), JID :: jid:jid(), tuple(), tuple()) ->
-    metrics_notify_return().
-roster_set(Acc, #jid{server = Server}, _, _) ->
-    mongoose_metrics:update(Server, modRosterSets, 1),
+-spec roster_set(Acc :: any(), JID :: jid:jid(), tuple(), tuple()) ->
+    any().
+roster_set(Acc, #jid{lserver = LServer}, _, _) ->
+    {ok, HostType} = mongoose_domain_api:get_host_type(LServer),
```
chrzaszcz, line 166: These are so inconsistent... ensuring host types in non-global hook Acc's seems like a good solution as @DenysGonchar suggested. No need to address it in this PR of course.

test/carbon_cache_server.erl
```diff
@@ -1,25 +1,19 @@
...
 -export([server/2]).
 -export([wait_for_accepting/0]).
 
-start(Num,LPort) ->
-    case gen_tcp:listen(LPort,[{active, false},{packet,2}]) of
+start() ->
+    case gen_tcp:listen(0,[{active, false},{packet,2}]) of
         {ok, ListenSock} ->
-            start_servers(Num,ListenSock),
+            spawn(?MODULE,server,[ListenSock, self()]),
```
chrzaszcz, line 10:
```suggestion
            spawn(?MODULE, server, [ListenSock, self()]),
```

chrzaszcz (commented):
Looks good, mostly minor comments from me. Some hooks needs fixing soon as they are not converted to host types yet.

I would like to see a test suite for metrics added to `dynamic_domains.spec`. Would it be possible to do it now or not yet?

src/metrics/mongoose_metrics_hooks.erl
gustawlippa: True, but these is one of the hooks in which the Acc is `ok`. There are a few more similar cases in `mongoose_metrics_hooks` sadly.

big_tests/tests/metrics_helper.erl
gustawlippa: It seems so, I will double check, but probably an oversight, good catch!

arcusfelis:
```

=== Ended at 2021-05-31 14:05:25
=== Location: [{escalus_mongooseim,metric_type,100},
              {escalus_mongooseim,read_metric_initial_value,60},
              {lists,foldl,1263},
              {escalus_mongooseim,maybe_read_initial_metric_values,38},
              {escalus_story,story,68},
              {test_server,ts_tc,1754},
              {test_server,run_test_case_eval1,1263},
              {test_server,run_test_case_eval,1195}]
=== === Reason: no match of right hand side value []
  in function  escalus_mongooseim:metric_type/1 (/Users/mikhailuvarov/erlang/esl/MongooseIM/big_tests/_build/default/lib/escalus/src/escalus_mongooseim.erl, line 100)
  in call from escalus_mongooseim:read_metric_initial_value/2 (/Users/mikhailuvarov/erlang/esl/MongooseIM/big_tests/_build/default/lib/escalus/src/escalus_mongooseim.erl, line 60)
  in call from lists:foldl/3 (lists.erl, line 1263)
  in call from escalus_mongooseim:maybe_read_initial_metric_values/1 (/Users/mikhailuvarov/erlang/esl/MongooseIM/big_tests/_build/default/lib/escalus/src/escalus_mongooseim.erl, line 38)
  in call from escalus_story:story/4 (/Users/mikhailuvarov/erlang/esl/MongooseIM/big_tests/_build/default/lib/escalus/src/escalus_story.erl, line 68)
  in call from test_server:ts_tc/3 (test_server.erl, line 1754)
  in call from test_server:run_test_case_eval1/6 (test_server.erl, line 1263)
  in call from test_server:run_test_case_eval/9 (test_server.erl, line 1195)
  ```

DenysGonchar:
> ```
> 
> === Ended at 2021-05-31 14:05:25
> === Location: [{escalus_mongooseim,metric_type,100},
>               {escalus_mongooseim,read_metric_initial_value,60},
>               {lists,foldl,1263},
>               {escalus_mongooseim,maybe_read_initial_metric_values,38},
>               {escalus_story,story,68},
>               {test_server,ts_tc,1754},
>               {test_server,run_test_case_eval1,1263},
>               {test_server,run_test_case_eval,1195}]
> === === Reason: no match of right hand side value []
>   in function  escalus_mongooseim:metric_type/1 (/Users/mikhailuvarov/erlang/esl/MongooseIM/big_tests/_build/default/lib/escalus/src/escalus_mongooseim.erl, line 100)
>   in call from escalus_mongooseim:read_metric_initial_value/2 (/Users/mikhailuvarov/erlang/esl/MongooseIM/big_tests/_build/default/lib/escalus/src/escalus_mongooseim.erl, line 60)
>   in call from lists:foldl/3 (lists.erl, line 1263)
>   in call from escalus_mongooseim:maybe_read_initial_metric_values/1 (/Users/mikhailuvarov/erlang/esl/MongooseIM/big_tests/_build/default/lib/escalus/src/escalus_mongooseim.erl, line 38)
>   in call from escalus_story:story/4 (/Users/mikhailuvarov/erlang/esl/MongooseIM/big_tests/_build/default/lib/escalus/src/escalus_story.erl, line 68)
>   in call from test_server:ts_tc/3 (test_server.erl, line 1754)
>   in call from test_server:run_test_case_eval1/6 (test_server.erl, line 1263)
>   in call from test_server:run_test_case_eval/9 (test_server.erl, line 1195)
> ```

are you sure that the config provided to `escalus:story/3` contains host type specific metrics? I had the similar problem in mod_ping_SUITE, see changes for `mongoose_metrics` parameter in this [PR](https://github.com/esl/MongooseIM/pull/3136/files#diff-822f77bafc36001b89c77dc657258dd2eed6a9f300c901ff91f67d97ff9c13f4R164)

PR #3122, opened 2021-05-18
Remove '.travis_deploy_rsa.enc'

vkatsuba:
Proposed changes include:
* Looks like the `.travis_deploy_rsa.enc` is not used anymore. Was used in `tools/travis-upload-ct-reports.sh` before in `1.6.2` version.

**Tests Result**: https://github.com/vkatsuba/MongooseIM/actions/runs/852819153


NelsonVides (approved):
Good catch! 👌🏽

PR #3123, opened 2021-05-18
Multitenancy for MAM/MUC

arcusfelis:
This PR addresses "MAM/MUC test suites are passing for dynamic domains".

Proposed changes include:
* hook refactoring

* Common Tests Hooks for big tests fail, if init_per_suite fails. Added some more strictness there. Separate PR: https://github.com/esl/MongooseIM/pull/3133
* More granular rules in domain_isolation_SUITE, also converted it for HostTypes.

* We COULD pass hosttype as a separate arg to muclight DB backend (and we probably should), but it would increase complexity of the current PR, I would like to do it in a separate one.

* ejabberd_users refactoring would be a separate PR.

* REST API should be done in a separate PR.

* commands API should be done in a separate PR.

Moved into #3137:
* Add reason to some XMPP and HTTP errors. 
* New module mongoose_record_pp to pretty print records in logs (it should be called manually though).


big_tests/dynamic_domains.config
```diff
@@ -38,5 +56,121 @@
         {username, <<"bOb">>},
         {server, <<"domain.example.com">>},
         {host, <<"localhost">>},
-        {password, <<"makrolika">>}]}
+        {password, <<"makrolika">>}]},
+    {carol, [
```
chrzaszcz, line 60: Please only add users that are actually used by the enabled test suites.

arcusfelis: done

big_tests/dynamic_domains.config
```diff
@@ -1,11 +1,21 @@
...
+{disable_disco_tests, true}.
+{disable_s2s_tests, true}.  %% Requires mod_register
+{disable_carbons_tests, true}.
+{disable_roster_tests, true}.
+
 {hosts, [{mim,  [{node, mongooseim@localhost},
                  {domain, <<"domain.example.com">>},
+                 {host_type, <<"test type">>},
                  {secondary_domain, <<"domain.example.org">>},
+                 {secondary_host_type, <<"test type">>},
```
NelsonVides, line 13: Why's this the same than the first host type?

big_tests/src/ct_mongoose_log_hook.erl
```diff
@@ -163,28 +163,28 @@ make_content(CurrentLineNum, Line) ->
...
-            ct:pal("issue=\"ct_mongoose_log_hook created log file\", "
-                   "ct_node=~p, reader=~p, reader_node=~p, writer=~p, out_file=~p",
-                   [node(), Reader, node(Reader), Writer, OutFile]),
-            State#state{reader=Reader, writer=Writer, out_file=OutFile,
-                        current_line_num=CurrentLineNum, url_file=UrlFile};
-        Reason ->
-            ct:pal("issue=\"Failed to init ct_mongoose_log_hook\", node=~p, reason=~p",
-                   [Node, Reason]),
-            State#state{out_file=OutFile}
+    try
```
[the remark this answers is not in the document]
arcusfelis: it fails here when suite init_per_suite fails. Fails because RunDir directory is not created at this point.

big_tests/dynamic_domains.config
arcusfelis: idk, it works like that xdd
I think it's a value, you get if you try to convert "domain.example.org" to host type.

big_tests/tests/muc_SUITE.erl
```diff
@@ -85,15 +84,37 @@
...
+all() -> 
+    case disco_disabled() of
+        true ->
+            [];
+        false ->
+            [{group, disco},
+             {group, disco_non_parallel},
+             {group, disco_rsm},
+             {group, disco_rsm_with_offline}]
+    end ++ [
```
NelsonVides, line 117: What about doing a trick like Denys did at https://github.com/esl/MongooseIM/blob/5166ab32fbeef8f1e369c2ec3b3e74a8d2114663/big_tests/dynamic_domains.spec#L15 seems clearer, smaller diff, doesn't make a mess of the test-suite, and it's easier to add and remove in the future 🤔

big_tests/tests/muc_SUITE.erl
```diff
@@ -319,16 +341,20 @@ init_per_suite(Config) ->
     %% For mocking with unnamed functions
     mongoose_helper:inject_module(?MODULE),
     Config2 = escalus:init_per_suite(Config),
-    Config3 = dynamic_modules:save_modules(domain(), Config2),
-    load_muc(muc_host()),
+    Config3 = dynamic_modules:save_modules(host_type(), Config2),
+    HostType = ct:get_config({hosts, mim, host_type}),
```
NelsonVides, line 345: You've just defined a helper `host_type()` function up there, could be defined in `domains_helper` and be used everywhere around big_tests btw.

src/muc_light/mod_muc_light_db_rdbms.erl
```diff
@@ -774,28 +773,27 @@ modify_aff_users_transaction(MainHost, RoomUS, RoomID, AffUsersChanges,
...
 %% ------------------------ Common ------------------------
 
--spec main_host(JIDOrServer :: jid:simple_bare_jid() | binary()) -> jid:lserver().
+-spec main_host(JIDOrServer :: jid:simple_bare_jid() | binary()) -> mongooseim:host_type().
 main_host({_, RoomS}) ->
     main_host(RoomS);
 main_host(MUCServer) ->
-    {ok, MainHost} = mongoose_subhosts:get_host(MUCServer),
-    MainHost.
+    mod_muc_light_utils:muc_host_to_host_type(MUCServer).
```
NelsonVides, line 799: I think that this function can change its name, it doesn't really reflect anymore what it used to do.

big_tests/tests/muc_SUITE.erl
arcusfelis: done

src/mongoose_api_client.erl
```diff
@@ -63,6 +63,8 @@ cowboy_router_paths(Base, _Opts) ->
 
 init(Req, Opts) ->
     Bindings = maps:to_list(cowboy_req:bindings(Req)),
+    ?LOG_ERROR(#{what => rest_init,
+                 cowboy_req => Req}),
```
NelsonVides, line 67: `ERROR`? This seems like a debug left-over

src/mongoose_api_client.erl
```diff
@@ -112,6 +114,8 @@ delete_resource(Req, #http_api_state{command_category = Category,
 to_json(Req, #http_api_state{command_category = Category,
                              command_subcategory = SubCategory,
                              bindings = B} = State) ->
+    ?LOG_ERROR(#{what => rest_to_json,
+                 cowboy_req => Req}),
```
NelsonVides, line 118: Likewise here

src/mongoose_api_client.erl
```diff
@@ -122,6 +126,8 @@ to_json(Req, #http_api_state{command_category = Category,
 from_json(Req, #http_api_state{command_category = Category,
                                command_subcategory = SubCategory,
                                bindings = B} = State) ->
+    ?LOG_ERROR(#{what => rest_from_json,
+                 cowboy_req => Req}),
```
NelsonVides, line 130: And similarly

src/mongoose_api_common.erl
```diff
@@ -327,6 +327,8 @@ error_response(ErrorType, Reason, Req, State) ->
                     L when is_list(L) -> list_to_binary(L);
                     Other -> list_to_binary(io_lib:format("~p", [Other]))
                 end,
+    ?LOG_ERROR(#{what => rest_common_error,
+                 error_type => ErrorType, reason => Reason, cowboy_req => Req}),
```
NelsonVides, line 331: Now the only thing that concerns me is the fact that we log here `cowboy_req`. Doesn't that might contain private data? We might leak here.

src/mongoose_client_api/mongoose_client_api.erl
```diff
@@ -77,6 +78,14 @@ set_resp_body_if_missing(Body, Req) ->
             cowboy_req:set_resp_body(Body, Req)
     end.
 
+maybe_report_error(StatusCode, Req, Body) when StatusCode >= 400 ->
+    ?LOG_WARNING(#{what => reply_error,
+                   stacktrace => erlang:process_info(self(), current_stacktrace),
+                   status_code => StatusCode,
+                   cowboy_req => Req, body => Body});
```
NelsonVides, line 85: Same issue I had before, logging the `cowboy_req` and the `body` might leak private information.

src/mongoose_client_api/mongoose_client_api_rooms.erl
```diff
@@ -111,20 +111,29 @@ from_json(Req, State) ->
...
-            {false, Req, State}
+            mongoose_client_api:bad_request(Req, <<"Failed to parse parameters">>, State)
     end.
 
 handle_request(Method, JSONData, Req, State) ->
     case handle_request_by_method(Method, JSONData, Req, State) of
-        {error, _, _} ->
-            {false, Req, State};
+        {error, Short, Desc} = Other ->
+            mongoose_client_api:bad_request(Req, format_error(Short, Desc), State);
```
NelsonVides, line 120: Can we have all this "better reporting" in a separate PR? This one is sufficiently enormous, and rest reporting errors is a different topic. There's also the question of whether we want to log the entire requests as I mentioned before, which I doubt we want. So these things could be moved to a different PR where we discuss such thing without interrupting mam's multitenancy.

src/muc_light/mod_muc_light_room_config.erl
```diff
@@ -73,7 +73,7 @@
 %%====================================================================
 
 -spec schema_from_definition(UserDefinedSchema :: user_defined_schema()) -> schema().
-schema_from_definition(UserDefinedSchema) ->
+schema_from_definition(UserDefinedSchema) when is_list(UserDefinedSchema) ->
```
NelsonVides, line 76: Why this numb change? `lists:foldl` will already check that it is a list, and this function should then be on the stacktrace. Did that go wrong somewhere?

src/domain/mongoose_lazy_routing.erl
```diff
@@ -408,9 +422,12 @@ check_that_domain_and_iq_match({#iq_table_key{host_type = HostType,
                                {_, {HostType, Pattern}}) when is_binary(Namespace),
                                                               Component =/= '_' ->
     true;
-check_that_domain_and_iq_match(IQ, Domain) ->
+check_that_domain_and_iq_match({Key=#iq_table_key{}, _} = IQEntry, Domain) ->
```
NelsonVides, line 425:
```suggestion
check_that_domain_and_iq_match({Key = #iq_table_key{}, _} = IQEntry, Domain) ->
```

src/ejabberd_c2s.erl
```diff
@@ -2276,7 +2276,8 @@ process_privacy_iq(Acc1, To, StateData) ->
             From = mongoose_acc:from_jid(Acc2),
             {Acc3, NewStateData} = process_privacy_iq(Acc2, Type, To, StateData),
             Res = mongoose_acc:get(hook, result,
-                                   {error, mongoose_xmpp_errors:feature_not_implemented()}, Acc3),
+                                   {error, mongoose_xmpp_errors:feature_not_implemented(
+                                             <<"en">>, <<"Privacy IQ from c2s">>)}, Acc3),
```
NelsonVides, line 2280: Not sure this error message is very meaningful for a client. Maybe explain something about what was violated here.

src/ejabberd_local.erl
```diff
@@ -114,7 +114,8 @@ process_iq(#iq{ xmlns = XMLNS } = IQ, Acc, From, To, _El) ->
         [{_, IQHandler}] ->
             gen_iq_component:handle(IQHandler, Acc, From, To, IQ);
         [] ->
-            ejabberd_router:route_error_reply(To, From, Acc, mongoose_xmpp_errors:feature_not_implemented())
+            ejabberd_router:route_error_reply(To, From, Acc,
+                mongoose_xmpp_errors:feature_not_implemented(<<"en">>, <<"From local handler">>))
```
NelsonVides, line 118: This error message, could be something like... `Local server does not implement this feature`, where the keyboard is the word `Local`. Should be a string that means something to the end user, and that points us to the place in code that generated it, at the same time.

NelsonVides: But again, maybe all these extended error messages could also be moved to a separate PR, this is a different topic to mam.

src/ejabberd_sm.erl
```diff
@@ -540,15 +541,21 @@ handle_info({route, From, To, Packet}, State) ->
     route(From, To, Packet),
     {noreply, State};
 handle_info({register_iq_handler, Host, XMLNS, IQHandler}, State) ->
-    ets:insert(sm_iqtable, {{XMLNS, Host}, IQHandler}),
+    try
+        ets:insert_new(sm_iqtable, {{XMLNS, Host}, IQHandler})
```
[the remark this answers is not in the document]
arcusfelis: good

src/muc_light/mod_muc_light_room_config.erl
arcusfelis: It makes stacktraces better, failing in foldl is less pretty.
Function clause on schema_from_definition is better.

(it's here because test cases use that function).

src/mongoose_client_api/mongoose_client_api.erl
arcusfelis: Users would have to write log filters to filter out private info from the request.
We have cowboy_req field in docs.

src/mongoose_api_client.erl
arcusfelis: leftovers

src/mod_smart_markers.erl
```diff
@@ -144,8 +144,7 @@ get_chat_markers(ChatType, #jid{lserver = LServer} = To, Thread, TS) ->
     Host = case ChatType of
                one2one -> LServer;
                groupchat ->
-                   {ok, H} = mongoose_subhosts:get_host(LServer),
-                   H
+                   mod_muc_light_utils:muc_host_to_host_type(LServer)
```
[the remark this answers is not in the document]
arcusfelis: I don't know how, but we have test cases for chat markers in suites, which need this.

src/mam/mod_mam.erl
```diff
@@ -160,64 +161,57 @@
...
-    %% `parallel' is the only one recommended here.
-    IQDisc = gen_mod:get_opt(iqdisc, Opts, parallel), %% Type
-    [mod_disco:register_feature(Host, Feature) || Feature <- features(?MODULE, Host)],
-    gen_iq_handler:add_iq_handler(ejabberd_sm, Host, ?NS_MAM_04,
-                                  ?MODULE, process_mam_iq, IQDisc),
-    gen_iq_handler:add_iq_handler(ejabberd_sm, Host, ?NS_MAM_06,
-                                  ?MODULE, process_mam_iq, IQDisc),
-    ejabberd_hooks:add(hooks(Host)),
-    ensure_metrics(Host),
+    ?LOG_ERROR(#{what => mam_starting, host_type => HostType}),
```
NelsonVides, line 201: Leftovers here

src/mam/mod_mam.erl
```diff
@@ -160,64 +161,57 @@
...
--spec stop(Host :: jid:server()) -> any().
-stop(Host) ->
+-spec stop(HostType :: host_type()) -> any().
+stop(HostType) ->
     ?LOG_INFO(#{what => mam_stopping}),
-    ejabberd_hooks:delete(hooks(Host)),
-    gen_iq_handler:remove_iq_handler(ejabberd_sm, Host, ?NS_MAM_04),
-    gen_iq_handler:remove_iq_handler(ejabberd_sm, Host, ?NS_MAM_06),
-    [mod_disco:unregister_feature(Host, Feature) || Feature <- features(?MODULE, Host)],
+    ?LOG_ERROR(#{what => mam_stopping, host_type => HostType}),
```
NelsonVides, line 211: More leftovers. The host_type could actually be a field of the previous LOG_INFO. But surely this shouldn't be an error.

NelsonVides (commented):
A couple of left-overs.

src/muc_light/mod_muc_light_db_rdbms.erl
```diff
@@ -71,13 +71,13 @@
 
 %% ------------------------ Backend start/stop ------------------------
 
--spec start(Host :: jid:server(), MUCHost :: jid:server()) -> ok.
-start(_Host, _MUCHost) ->
+-spec start(Host :: jid:server()) -> ok.
+start(_Host) ->
```
NelsonVides, line 75: Aren't these `HostTypes` instead? If only, for good notation.

src/muc_light/mod_muc_light_db_rdbms.erl
```diff
@@ -71,13 +71,13 @@
...
     prepare_queries(),
     ok.
 
--spec stop(Host :: jid:server(), MUCHost :: jid:server()) -> ok.
-stop(_Host, _MUCHost) ->
+-spec stop(Host :: jid:server()) -> ok.
+stop(_Host) ->
```
NelsonVides, line 80: As above

src/muc_light/mod_muc_light.erl
```diff
@@ -138,66 +164,60 @@ delete_room(RoomUS) ->
...
 
-    ok.
+set_dynamic_opts(HostType, Opts) ->
+    %% Prepare config schema
+    Def = gen_mod:get_opt(config_schema, Opts, default_schema_definition()),
+    ConfigSchema = mod_muc_light_room_config:schema_from_definition(Def),
+    %% XXX That is a bad style
+    gen_mod:set_module_opt(HostType, ?MODULE, computed_config_schema, ConfigSchema),
+    %% Prepare default config
+    DefaultConfig = mod_muc_light_room_config:default_from_schema(ConfigSchema),
```
NelsonVides, line 217: What is this, and why is it a bad style?

arcusfelis: modules should not set their own options.

NelsonVides (approved):
It's impossible to be sure that an almost 3k lines big PR is correct, and that is an amazing way to introduce issues, this is impossible to truly review. Let's see what we come up fixing in the future.

chrzaszcz:
> It's impossible to be sure that an almost 3k lines big PR is correct, and that is an amazing way to introduce issues, this is impossible to truly review. Let's see what we come up fixing in the future.

We should have a second reviewer in such cases.

PR #3125, opened 2021-05-19
Carbons upgrade to 0.13.2

NelsonVides:
Including appending the MAM stanza-id as required by the lastest mam xep, and moving the mam hooks order earlier in the priority line.


big_tests/tests/carboncopy_SUITE.erl
```diff
@@ -235,26 +240,19 @@ all_bobs_other_resources_get_sent_carbons([Alice, Bob1 | Bobs], Msg) ->
...
 true_story(Config, UserSpecs, TestFun) ->
-    try   escalus_fresh:story_with_client_list(Config, UserSpecs, TestFun), true
+    try escalus_fresh:story_with_client_list(Config, UserSpecs, TestFun), true
     catch E ->
               {error, E}
     end.
 
 %% Number of resources per users
-no_of_resources() -> rand:uniform(4).
+no_of_resources() -> 1 + rand:uniform(4).
```
chrzaszcz, line 255: Just a note: we could set it to deterministic values to avoid randomly passing/failing tests in case of a bug, e.g. `list:seq(2,5)`

src/mod_carboncopy.erl
```diff
@@ -25,16 +25,16 @@
 %%%----------------------------------------------------------------------
 -module (mod_carboncopy).
 -author ('ecestari@process-one.net').
--xep([{xep, 280}, {version, "0.10"}]).
+-xep([{xep, 280}, {version, "0.6"}]).
+-xep([{xep, 280}, {version, "0.13.3"}]).
```
chrzaszcz, line 29: Two different versions for the same xep?

test/ejabberd_sm_SUITE.erl
```diff
@@ -412,9 +412,9 @@ given_session_opened(Sid, {U, S, R}, Priority, Info) ->
 when_session_opened(Sid, {U, S, R}, Priority, Info) ->
     given_session_opened(Sid, {U, S, R}, Priority, Info).
 
-when_session_info_stored(U, S, R, {_,_}=KV) ->
+when_session_info_stored(U, S, R, {K,V}) ->
```
chrzaszcz, line 415:
```suggestion
when_session_info_stored(U, S, R, {K, V}) ->
```

test/carboncopy_proper_tests_SUITE.erl
```diff
@@ -99,5 +117,26 @@ sent_message() ->
 simple_chat_message() ->
     xmlel("message", [{<<"type">>,<<"chat">>}],[]).
 
+chat_state_notification() ->
```
chrzaszcz, line 120: Would it be possible to add missing spaces to the functions below?

test/carboncopy_proper_tests_SUITE.erl
```diff
@@ -19,50 +19,73 @@ all() ->
...
+     no_copy_type_test,
+     empty_message_test,
+     received_type_test,
+     sent_message_test,
+     simple_badarg_test,
+     simple_chat_message_test,
+     has_chat_state_notifications,
+     has_delivery_receipts,
+     is_muc_invitation,
+     is_direct_invitation
```
chrzaszcz, line 33: I know that adding `_test` seems a bit awkward, but maybe either do it always or never in a single test suite? Just to stay consistent.

src/mod_carboncopy.erl
```diff
@@ -123,69 +126,102 @@ user_receive_packet(Acc, JID, _From, To, Packet) ->
...
-        SubTag -> is_forwarded(SubTag)
-    end.
+    undefined =/= exml_query:subelement_with_name_and_ns(Packet, <<"sent">>, ?NS_CC_2).
 
--spec is_forwarded(_) -> classification().
-is_forwarded(SubTag) ->
-    case xml:get_subtag(SubTag, <<"forwarded">>) of
-        false -> forward;
-        _ -> ignore
-    end.
```
chrzaszcz, line 185: Why is this one removed? Would it be possible to list the changed rules in a commit message, PR description or somewhere else? The refactoring makes it a bit hard to figure out which changes are intentional. According to the XEP:

> The sending server MUST NOT deliver forwarded <message/>s to the other Carbons-enabled resources of the sender.
The receiving server MUST NOT deliver forwarded <message/>s to the other Carbons-enabled resource of the recipient

chrzaszcz (commented):
Looks good! I have a few minor questions.

big_tests/tests/carboncopy_SUITE.erl
NelsonVides: The idea is that tests should pass regardless of the number of resources, that's what proper is doing 🤔

src/mod_carboncopy.erl
NelsonVides: Yes, because 0.6 had different rules on what should be forwarded as a carbon.

src/mod_carboncopy.erl
NelsonVides: Also, as far as I understand the xep, the only way the xep proposes to have incomplete conversations across carbons-enabled devices y by the `<private/>` tag 🤔

chrzaszcz: Ok, I trust you on this one. There are some unclear things in the XEP but in general I agree that the rules you chose look reasonable.

big_tests/tests/carboncopy_SUITE.erl
chrzaszcz: Ok let's keep it as it is.

src/mod_carboncopy.erl
chrzaszcz: Discussed the NS issue, after the fix it looks ok.

chrzaszcz (approved):

PR #3127, opened 2021-05-24
unit tests for mongoose_lazy_routing module

DenysGonchar:
This PR introduces unit tests for mongoose_lazy_routing module

### DIFF
--- a/test/mongoose_lazy_routing_SUITE.erl
+++ b/test/mongoose_lazy_routing_SUITE.erl
@@ -16,9 +16,8 @@
 -define(SUBDOMAIN_2, <<"sub.", (?DOMAIN_2)/binary>>).
 -define(SUBDOMAIN_3, <<"sub.", (?DOMAIN_3)/binary>>).
 
-%% these domain and subdomain are required for testing conflict resolution.
+%% this domain is required for testing domain/subdomain conflict resolution.
 -define(DOMAIN_X, <<"domain_x.test">>).
--define(SUBDOMAIN_X, ?DOMAIN_X).
 
 %% required for handles_missing_domain_or_subdomain test case
 -define(MISSING_DOMAIN, <<"missing.domain.test">>).
@@ -45,7 +44,7 @@ all() ->
      handles_double_iq_handler_registration_deregistration_for_domain,
      handles_double_iq_handler_registration_deregistration_for_subdomain].
 
-init_per_testcase(_,Config) ->
+init_per_testcase(_, Config) ->
     mongoose_lazy_routing:start(),
     setup_meck(),
     Config.
@@ -59,15 +58,15 @@ end_per_testcase(_, Config) ->
 %% test cases
 %%-------------------------------------------------------------------
 can_add_and_remove_domain_or_subdomain(_Config) ->
-    %add 2 domains, one of them add twice
+    %% add 2 domains, one of them add twice
     ?assertEqual(true, maybe_add_domain_or_subdomain(?DOMAIN_1)),
     ?assertEqual(true, maybe_add_domain_or_subdomain(?DOMAIN_2)),
     ?assertEqual(true, maybe_add_domain_or_subdomain(?DOMAIN_2)),
-    %add 2 subdomains, one of them add twice
+    %% add 2 subdomains, one of them add twice
     ?assertEqual(true, maybe_add_domain_or_subdomain(?SUBDOMAIN_1)),
     ?assertEqual(true, maybe_add_domain_or_subdomain(?SUBDOMAIN_2)),
     ?assertEqual(true, maybe_add_domain_or_subdomain(?SUBDOMAIN_2)),
-    %check that 2 domains and 2 subdomains added properly
+    %% check that 2 domains and 2 subdomains added properly
     ?assertEqualLists([?DOMAIN_1, ?DOMAIN_2], get_all_registered_domains()),
     ?assertEqualLists([{?SUBDOMAIN_1, packet_handler(?SUBDOMAIN_1)},
                        {?SUBDOMAIN_2, packet_handler(?SUBDOMAIN_2)}],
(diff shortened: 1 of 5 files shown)


test/mongoose_lazy_routing_SUITE.erl
```diff
@@ -0,0 +1,471 @@
...
+    IQHandlerWithHostType1 = create_iq_handler_and_register(<<"IQH1">>, ?HOST_TYPE_2,
+                                                            Pattern, ?NAMESPACE_1,
+                                                            ?COMPONENT),
+    ?assertEqualLists([{?COMPONENT, ?SUBDOMAIN_2, ?NAMESPACE_1, IQHandlerWithHostType1},
+                       {?COMPONENT, ?SUBDOMAIN_3, ?NAMESPACE_1, IQHandlerWithHostType1}],
+                      get_all_registered_iqs()),
+    ?assertEqual([], get_all_unregistered_iqs()),
+    meck:reset(gen_iq_component),
+    %% register 2 IQ handlers for ?HOST_TYPE_1 subdomains, then add one subdomain and
+    %% one domain (just to ensure that domain doesn't affect anything)
```
arcusfelis, line 228: I prefer tests to follow this convention:

```
can_register_iq_handler_for_subdomain() ->
       %% GIVEN idk, for me nothing is given in this case
       %% WHEN Register IQ handler for subdomain
       ?assertEqual(true, maybe_add_domain_or_subdomain(?SUBDOMAIN_2)), %% ONE subdomain
       %% THEN registered
       assert_subdomain_registered(?SUBDOMAIN_2).
```

If we want to test registering two subdomains, we do another case:

```
can_register_iq_handler_for_two_subdomains() ->
       %% GIVEN idk, for me nothing is given in this case
       %% WHEN Register IQ handler for subdomains
       ?assertEqual(true, maybe_add_domain_or_subdomain(?SUBDOMAIN_1)),
       ?assertEqual(true, maybe_add_domain_or_subdomain(?SUBDOMAIN_2)),
       %% THEN registered
       assert_subdomain_registered(?SUBDOMAIN_1),
       assert_subdomain_registered(?SUBDOMAIN_2).
```

If we want to test unregister:

```
can_unregister_iq_handler_for_two_subdomains() ->
       %% GIVEN two registered domains
       ?assertEqual(true, maybe_add_domain_or_subdomain(?SUBDOMAIN_1)),
       ?assertEqual(true, maybe_add_domain_or_subdomain(?SUBDOMAIN_2)),
       %% WHEN Register IQ handler for subdomains
       unregister_subdomain(?SUBDOMAIN_1),
       unregister_subdomain(?SUBDOMAIN_2),
       %% THEN registered
       assert_subdomain_not_registered(?SUBDOMAIN_1),
       assert_subdomain_not_registered(?SUBDOMAIN_2).
```

I am not a big fun of asserts in the given part, so:

```
can_unregister_iq_handler_for_two_subdomains() ->
       %% GIVEN two registered domains
       given_subdomains([?SUBDOMAIN_1, ?SUBDOMAIN_2]),
       %% WHEN Unegister IQ handler for subdomains
       unregister_subdomain(?SUBDOMAIN_1),
       unregister_subdomain(?SUBDOMAIN_2),
       %% THEN not registered
       assert_subdomain_not_registered(?SUBDOMAIN_1),
       assert_subdomain_not_registered(?SUBDOMAIN_2).
```

If you check more than two things in a case, it is still ok, unless test fails. After that it's hard to understand what is actually failing.
If you do something that is not described in test name - it's very hard to understand if the test is valid or not.

Overall, it's a good idea to follow `GIVEN... WHEN... THAN...` structure. 
And make helper functions for GIVEN and THAN conditions.

And generally, it's good idea to think about post conditions. They are our side effects. What does it mean subdomain not registered? It could be encapsulated inside that `assert_subdomain_not_registered` function.

Side effects dictate the name of the test case. Action (WHEN block) in the test is the secondary, but also should be present in the name.

```
%% Was can_unregister_iq_handler_for_two_subdomains
unregistered_subdomain_is_not_listed() ->
       %% GIVEN
       given_subdomains([?SUBDOMAIN_1]),
       %% WHEN
       unregister_subdomain(?SUBDOMAIN_1),
       %% THEN not registered
       assert_subdomain_not_registered(?SUBDOMAIN_1).
       
unregistered_subdomain_looses_iq_handlers() ->
       %% GIVEN
       given_subdomains([?SUBDOMAIN_1]),
       given_iq_handler(...),
       %% WHEN
       unregister_subdomain(?SUBDOMAIN_1),
       %% THEN not registered
       assert_iq_handler_not_registered(...).
```

Also, testing with more complex preconditions is cool, but harder to read. Still can be done, but in a separate test case.

```
%% Was can_unregister_iq_handler_for_two_subdomains
unregistered_subdomain_is_not_listed_when_there_is_another_domain() ->
       %% GIVEN
       given_subdomains([?SUBDOMAIN_2]),
       unregistered_subdomain_looses_iq_handlers().
```

arcusfelis (commented):
looks fine, I almost understand the tests.
Added comments about whitespaces though.

test/mongoose_lazy_routing_SUITE.erl
chrzaszcz: I agree that splitting this huge test case could help with readability.
Personally I think that naming functions `given_*` is not a good practice - I'd rather keep the names`register_subdomains`, `register_iq_handler` etc. to be explicit about what they are exactly doing.
IMO there is no need for a significant rework here, splitting could help (test sequence in a group could suffice), but it's not a must.

chrzaszcz (commented):
Looks good! I have a few minor comments.

test/mongoose_lazy_routing_SUITE.erl
DenysGonchar: `given, when, then` template doesn't fit well for the cases that test `correct setup/correct rollback` behaviour.
I will try to split `can_register_and_unregister_iq_handler_for_domain` and `can_register_and_unregister_iq_handler_for_subdomain` test case into 4 smaller

test/mongoose_lazy_routing_SUITE.erl
```diff
@@ -0,0 +1,485 @@
...
+         meck:reset(gen_iq_component)
+     end || _ <- lists:seq(1, 2)],
+    %% remove 2 domains and one subdomain for ?HOST_TYPE_2
+    maybe_remove_domain(domain_host_type(?DOMAIN_2), ?DOMAIN_2),
+    maybe_remove_domain(domain_host_type(?DOMAIN_3), ?DOMAIN_3),
+    maybe_remove_subdomain(subdomain_info(?SUBDOMAIN_2)),
+    ?assertEqual([], get_all_registered_iqs()),
+    ?assertEqual([], get_all_unregistered_iqs()).
+
+
```
chrzaszcz, line 158: Minor: double empty line.

chrzaszcz (approved):
Looks good!

PR #3131, opened 2021-05-26
Make dummy auth delays configurable

NelsonVides:
This makes the delays configurable, which by default are too long and that's something that can be noticed in CI.

### DIFF
--- a/src/config/mongoose_config_spec.erl
+++ b/src/config/mongoose_config_spec.erl
@@ -490,7 +490,8 @@ auth() ->
                  <<"jwt">> => auth_jwt(),
                  <<"ldap">> => auth_ldap(),
                  <<"riak">> => auth_riak(),
-                 <<"rdbms">> => auth_rdbms()},
+                 <<"rdbms">> => auth_rdbms(),
+                 <<"dummy">> => auth_dummy()},
        process = fun ?MODULE:process_auth/1,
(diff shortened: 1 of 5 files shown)


src/config/mongoose_config_spec.erl
```diff
@@ -646,6 +647,17 @@ auth_rdbms() ->
        format = none
       }.
 
+%% path: (host_config[].)auth.dummy
+auth_dummy() ->
```
DenysGonchar, line 651: how about to test it? e.g. we can set this configuration options for "test type" host_type.

NelsonVides: Pushed so :)

NelsonVides: Here, now carboncopy for dynamic domains is pretty fast: https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3131/55777/pgsql_mnesia.23.0.3-1/big/ct_run.test@default-44683ef9-9b3c-4852-b093-e90d801f50e2.2021-05-27_09.27.12/big_tests.tests.carboncopy_SUITE.logs/run.2021-05-27_09.27.14/suite.log.html 😄

DenysGonchar (approved):
Nice. I like this change.

PR #3132, opened 2021-05-26
Multitenancy/inbox

NelsonVides:
This PR attempts to prepare inbox to be multitenancy ready.

NOTE, important TODOs for the near future:
* Muc, muclight, and pools, aren't multitenancy ready yet.
* Pools need to be called by their hosttype, not their lserver.

### DIFF
--- a/big_tests/test.config
+++ b/big_tests/test.config
@@ -18,7 +18,6 @@
 %%       so that we rein the "bag of things" approach
 {hosts, [{mim,  [{node, mongooseim@localhost},
                  {domain, <<"localhost">>},
-                 {host_type, <<"localhost">>},
                  {vars, "mim1"},
                  {cluster, mim},
                  {secondary_domain, <<"localhost.bis">>},
--- a/big_tests/test.config
+++ b/big_tests/test.config
@@ -18,6 +18,7 @@
 %%       so that we rein the "bag of things" approach
 {hosts, [{mim,  [{node, mongooseim@localhost},
                  {domain, <<"localhost">>},
+                 {host_type, <<"localhost">>},
                  {vars, "mim1"},
                  {cluster, mim},
                  {secondary_domain, <<"localhost.bis">>},
--- a/big_tests/tests/domain_helper.erl
+++ b/big_tests/tests/domain_helper.erl
@@ -4,14 +4,10 @@
          delete_configured_domains/0,
          insert_domain/3,
          delete_domain/2,
-         host_type/0,
          host_type/1,
          host_type/2]).
 
 -import(distributed_helper, [get_or_fail/1, rpc/4]).
-
-host_type() ->
-    ct:get_config({hosts, mim, host_type}).
 
 host_type(NodeKey) ->
     host_type(NodeKey, domain).
--- a/big_tests/tests/domain_helper.erl
+++ b/big_tests/tests/domain_helper.erl
@@ -4,10 +4,14 @@
          delete_configured_domains/0,
          insert_domain/3,
          delete_domain/2,
+         host_type/0,
          host_type/1,
          host_type/2]).
 
 -import(distributed_helper, [get_or_fail/1, rpc/4]).
+
+host_type() ->
+    ct:get_config({hosts, mim, host_type}).
 
 host_type(NodeKey) ->
     host_type(NodeKey, domain).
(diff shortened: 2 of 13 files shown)


big_tests/tests/dynamic_modules.erl
```diff
@@ -30,65 +30,65 @@ to_replace([RequiredModule | Rest], CurrentModules, ReplaceAcc, ReplaceWithAcc)
...
         Str ->
-            stop(Domain, list_to_atom(lists:reverse(Str)++"_rdbms"))
+            stop(HostType, list_to_atom(lists:reverse(Str)++"_rdbms"))
     end,
-    start(Domain, Mod, Args).
+    start(HostType, Mod, Args).
 
 start_running(Config) ->
-    Domain = ct:get_config({hosts, mim, domain}),
+    HostType = domain_helper:host_type(),
```
gustawlippa, line 88: Shouldn't `stop_running/2` also be changed to use HostType?

big_tests/test.config
```diff
@@ -18,6 +18,7 @@
 %%       so that we rein the "bag of things" approach
 {hosts, [{mim,  [{node, mongooseim@localhost},
                  {domain, <<"localhost">>},
+                 {host_type, <<"localhost">>},
```
chrzaszcz, line 21: My idea was to use `host_type(mim)` instead - already implemented in `domain_helper`.

big_tests/tests/domain_helper.erl
```diff
@@ -4,11 +4,15 @@
          delete_configured_domains/0,
          insert_domain/3,
          delete_domain/2,
+         host_type/0,
          host_type/1,
          host_type/2]).
 
 -import(distributed_helper, [get_or_fail/1, rpc/4]).
 
+host_type() ->
```
chrzaszcz, line 13: My idea was to use `host_type(mim)` - already implemented below.

big_tests/tests/inbox_SUITE.erl
```diff
@@ -190,23 +191,22 @@ suite() ->
 %%--------------------------------------------------------------------
 
 init_per_suite(Config) ->
-    ok = dynamic_modules:ensure_modules(inbox_helper:domain(), required_modules()),
+    ok = dynamic_modules:ensure_modules(domain_helper:host_type(), inbox_modules()),
```
chrzaszcz, line 194: Could we use `domain_helper:host_type(mim)`?

big_tests/tests/inbox_extensions_SUITE.erl
```diff
@@ -144,28 +145,30 @@ inbox_extensions_tests() ->
     ].
 
 init_per_suite(Config) ->
-    ok = dynamic_modules:ensure_modules(inbox_helper:domain(), required_modules()),
+    ok = dynamic_modules:ensure_modules(domain_helper:host_type(), inbox_modules()),
```
chrzaszcz, line 148: Could we use `domain_helper:host_type(mim)`?

src/inbox/mod_inbox.erl
```diff
@@ -559,16 +570,22 @@ get_inbox_unread(undefined, Acc, To) ->
...
+     {remove_user, HostType, ?MODULE, remove_user, 50},
+     {remove_domain, HostType, ?MODULE, remove_domain, 50},
+     {user_send_packet, HostType, ?MODULE, user_send_packet, 70},
+     {filter_local_packet, HostType, ?MODULE, filter_packet, 90},
+     {inbox_unread_count, HostType, ?MODULE, inbox_unread_count, 80},
+     {get_personal_data, HostType, ?MODULE, get_personal_data, 50}
     ].
 
+add_default_backend(Opts) ->
+    case lists:keyfind(backend, 2, Opts) of
```
chrzaszcz, line 584: I think that key should be first, not second. It would be good to cover this in tests.

chrzaszcz (commented):
The changes looks good, my main concern is that all the `execute_*` functions are not converted to host types yet.

big_tests/test.config
NelsonVides: Could do that, but I actually thought of using this domain_helper shortcut, Mikail did the same in his MAM PR, and I thought it'd be just faster to skip the rpc. Can do that instead anyways 🤷🏽‍♂️

NelsonVides:
@chrzaszcz Changes applied. When it comes to the `execute` queries, in reality the host passed is ignored later on, so it doesn't make any difference yet, until we can really configure more than one rdbms pool. So for now that's a big todo for the task of adapting rdbms pools.

chrzaszcz (approved):
We decided with @NelsonVides that adapting the backends to work with host types will be done in a follow-up PR.

PR #3134, opened 2021-05-27
Remove check_carboncopy CT hook

gustawlippa:
This PR removes the `Suite ... finished dirty.` messages appearing when running the tests. They appeared because of a failing common test hook.
This hook is a leftover from when `mod_carboncopy` was using its own mnesia table. See commit e0b1d9d8746e2de3ef137a21772f4bacf20d85f3 in which it was removed for details. It is safe to remove this hook, as other hooks like `check_sessions` or `check_registered_users` together with using escalus fresh stories should already ensure that no user configuration leaks between the test groups.


NelsonVides (approved):
Excellent! And good reference to the commit that changed this! 😄

PR #3135, opened 2021-05-27
Multitenancy for ejabberd users

NelsonVides:
This erases `ejabberd_users` and creates a new module called `mongoose_users` that implements pretty much the same cache idea, but creates a table per host_type, and implements the remove_domain hook.

It also chooses to extract the host_type from the accumulator in some places where we can be sure we do have a host_type, like, when run inside `user_send_packet`, `filter_local_packet`, or iq handlers.


big_tests/tests/ejabberdctl_SUITE.erl
```diff
@@ -242,12 +242,8 @@ end_per_group(Rosters, Config) when (Rosters == roster) or (Rosters == roster_ad
...
                 SB = string_to_binary(S),
                 UB = string_to_binary(U),
-                Acc = rpc(mim(), mongoose_acc, new,
-                          [#{ location => {?MODULE, ?FUNCTION_NAME, ?LINE},
-                              lserver => SB,
-                              host_type => SB,
-                              element => undefined }]),
-                rpc(mim(), mongoose_hooks, remove_user, [SB, Acc, UB]);
+                Acc = mongoose_helper:new_mongoose_acc(SB),
+                rpc(mim(), mongoose_hooks, remove_user, [SB, Acc, SB, UB]);
```
DenysGonchar, line 246: seems like you need to update this place as well.

DenysGonchar (approved):
looks fine for now, mongoose_users must be reimplemented with the proper caching solution once we pick one.

PR #3136, opened 2021-05-28
adding dynamic domains support to mod_ping

DenysGonchar:
This PR introduces dynamic domains support for mod_ping.
Service discovery doesn't work for dynamic domains (as mod_disco doesn't support it yet), but still works correctly for statically configured domains

for a local testing you can use these commands:
```
test-runner.sh --skip-small-tests --skip-cover --preset pgsql_mnesia --skip-stop-nodes -- mod_ping
test-runner.sh --skip-small-tests --skip-cover --preset pgsql_mnesia --skip-stop-nodes --spec dynamic_domains.spec -- mod_ping
```


src/mod_ping.erl
```diff
@@ -161,42 +173,44 @@ user_keep_alive(Acc, _JID) ->
...
+handle_remote_call(init, _JID, _Server, HostType, HandlerState) ->
+    start_ping_timer(HandlerState, HostType);
+handle_remote_call(send_ping, JID, Server, HostType, HandlerState) ->
+    route_ping_iq(JID, Server, HostType),
     start_ping_timer(HandlerState, Server);
-handle_remote_call(timeout, JID, Server, HandlerState) ->
-    mongoose_hooks:user_ping_timeout(Server, JID),
-    case gen_mod:get_module_opt(Server, ?MODULE, timeout_action, none) of
+handle_remote_call(timeout, JID, _Server, HostType, HandlerState) ->
+    mongoose_hooks:user_ping_timeout(HostType, JID),
```
NelsonVides, line 198: Is it me or this hook doesn't really have any subscribers?

NelsonVides (commented):
I think it's pretty much ready to merge, I just have some minor comments in the meantime 👌🏽

src/mod_ping.erl
```diff
@@ -140,6 +151,7 @@ handle_remote_hook(HandlerState, mod_ping, Args, C2SState) ->
     handle_remote_call(Args,
                        ejabberd_c2s_state:jid(C2SState),
                        ejabberd_c2s_state:server(C2SState),
```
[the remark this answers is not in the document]
DenysGonchar: don't see any difference.

big_tests/tests/mod_ping_SUITE.erl
```diff
@@ -83,7 +83,7 @@ init_per_group(server_ping_kill, Config) ->
     [{timeout_action, kill} | Config].
 
 end_per_group(_GroupName, Config) ->
-    Domain = ct:get_config({hosts, mim, domain}),
+    Domain = ct:get_config({hosts, mim, host_type}),
```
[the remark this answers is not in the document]
DenysGonchar: but none of these PR's is merged so far

src/mod_ping.erl
DenysGonchar: Not in the open source MIM, but it might be used in MIM forks.

src/mod_ping.erl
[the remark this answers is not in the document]
NelsonVides: One argument less, and less ignoring below. Only `route_ping_iq/3` is using the server here, and it's also taking the JID, so it can extract the server from it. Just an opinionated one.

big_tests/tests/mod_ping_SUITE.erl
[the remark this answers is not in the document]
chrzaszcz: You can just call `domain_helper:host_type(mim)`, it's made for this purpose and does not need to have the host type added to the config spec when static host types are used.

big_tests/test.config
```diff
@@ -18,6 +18,7 @@
 %%       so that we rein the "bag of things" approach
 {hosts, [{mim,  [{node, mongooseim@localhost},
                  {domain, <<"localhost">>},
+                 {host_type, <<"localhost">>},
```
chrzaszcz, line 21: This isn't necessary, see below.

big_tests/tests/mod_ping_SUITE.erl
chrzaszcz, line 86:
```suggestion
    Domain = domain_helper:host_type(mim)
```

src/mod_ping.erl
```diff
@@ -68,69 +72,76 @@ route_ping_iq(JID, Server) ->
...
+    mod_disco:unregister_feature(HostType, ?NS_PING).
 
 -spec config_spec() -> mongoose_config_spec:config_section().
 config_spec() ->
     #section{
        items = #{<<"send_pings">> => #option{type = boolean},
                  <<"ping_interval">> => #option{type = integer,
-                                                validate = positive},
+                                                validate = positive,
+                                                process = fun timer:seconds/1},
```
chrzaszcz, line 125: Why? I really want to reduce the usage of `process` as much as possible. Either unit is fine for me, just keep it the same and consistent. We have multiple timeouts configured in many options and here out of the sudden for one random option there is an unexpected conversion.

chrzaszcz (commented):
Lokks good in general, but I have a few comments.

src/mod_ping.erl
chrzaszcz: Ok we talked it through and I think it's not an issue, let's keep the change :)

big_tests/test.config
arcusfelis: I would rather have it defined like that, not as rpc-s.

chrzaszcz: For me it's better to have one helper that consistently reads it from `mongoose_domain_core` instead of defining it multiple times for multiple hosts. Anyway, as the helper is already there, let's use it consistently. This way we can change the implementation later when needed.

chrzaszcz (approved):
Looks good. The only thing that could be done would be the removal of host type from `test.config`.

NelsonVides (approved):
Excellent work 👌🏽

PR #3137, opened 2021-05-31
Better error reporting

arcusfelis:
This PR addresses "We have too many places that can return 503 error".

Proposed changes include:
* Add text to errors, so we can know where it comes from.
* REST API logs the errors. We are not concerned with cleaning up creds from cowboy_req in this PR, it's a big task on its own and can be done in one log filter.
* debugging helpers for gen_mod.

### DIFF
--- a/src/logger/mongoose_record_pp.erl
+++ b/src/logger/mongoose_record_pp.erl
@@ -1,0 +1,19 @@
+-module(mongoose_record_pp).
+-export([format/3]).
+
+format(Record, Name, FieldDefs) when element(1, Record) =:= Name ->
+    FieldVals = tl(tuple_to_list(Record)),
+    FieldKV = print_pairs(FieldVals, FieldDefs),
+    iolist_to_binary(["#", atom_to_list(Name), "{", FieldKV, "}"]);
+format(Other, _, _) ->
+    iolist_to_binary(io_lib:format("~500p", [Other])).
+
+print_pairs([Val], [Def]) ->
+    %% trailing comma
+    [pair(Val, Def)];
+print_pairs([Val | VRest], [Def | DRest]) ->
+    [pair(Val, Def), ", " | print_pairs(VRest, DRest)];
+print_pairs([], []) -> [].
+
+pair(Val, Def) ->
+    io_lib:format("~p = ~500p", [Def, Val]).
--- a/src/logger/mongoose_record_pp.erl
+++ b/src/logger/mongoose_record_pp.erl
@@ -1,6 +1,13 @@
+%% Pretty prints a record including field names.
+%% It's useful to print records with many fields.
 -module(mongoose_record_pp).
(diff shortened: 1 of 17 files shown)


NelsonVides (commented):
Two comments here. Also, I'd like to raise concern for the privacy-related consequences of logging things like the `cowboy_req` element; otherwise this PR is a brilliant idea.

src/logger/mongoose_record_pp.erl
```diff
@@ -0,0 +1,19 @@
+-module(mongoose_record_pp).
```
NelsonVides, line 1: Maybe some documentation about what this new module is for?

src/mongoose_client_api/mongoose_client_api.erl
```diff
@@ -53,7 +53,7 @@ to_json(Req, User) ->
     {<<"{}">>, Req, User}.
 
 bad_request(Req, State) ->
-    bad_request(Req, <<>>, State).
+    bad_request(Req, <<"Bad request without a reason">>, State).
```
NelsonVides, line 56: This looks unfriendly for the end user, should have a better text error.

src/logger/mongoose_record_pp.erl
arcusfelis: done

src/mongoose_client_api/mongoose_client_api.erl
```diff
@@ -77,6 +78,13 @@ set_resp_body_if_missing(Body, Req) ->
             cowboy_req:set_resp_body(Body, Req)
     end.
 
+maybe_report_error(StatusCode, Req, Body) when StatusCode >= 400 ->
+    ?LOG_WARNING(#{what => reply_error,
+                   stacktrace => element(2, erlang:process_info(self(), current_stacktrace)),
+                   code => StatusCode, req => Req, reply_body => Body});
```
NelsonVides, line 84: This one is logging the `Body` with `reply_body` as a field, should be `body`, right?

NelsonVides (commented):
Now it's almost perfect, just one more comment.

src/mongoose_client_api/mongoose_client_api.erl
arcusfelis: nope, there is no body field documented yet.
So, reply_body. Because there could be also request_body.

NelsonVides:
```erlang
           {remove_fields_filter, {fun mongoose_log_filter:remove_fields_filter/2, [req, body]}},
```
From the config, I understand that `req` and `body` are the only ones removed.

arcusfelis: fixed

NelsonVides (approved):
All right, all clean now 👍🏽

PR #3139, opened 2021-05-31
Remove 'local_send_to_resource' hook

chrzaszcz:
It was registered per domain, not per host type, what caused exometer to create one metric per domain.

It can be removed as 'ejabberd_local' was the only module handling it.
This unexpected type of stanza shouldn't be handled by modules as well.


DenysGonchar (approved):
looks good

PR #3140, opened 2021-05-31
GDPR get data takes host type as a parameter

NelsonVides:
This will be needed in order to correctly query the correct pools where is needed. We can see that it was already done for `mod_mam` in #3123, which in-place queries the host_type. We can just use it as a parameter to the function.


chrzaszcz (approved):
Looks good!

PR #3141, opened 2021-05-31
Multitenancy/inbox

NelsonVides:
This PR most importantly enables (and fixes) muc/muclight tests for inbox and dynamic domains.

It also migrates inbox's backend to HostTypes in an orderly manner, following the pattern of the auth modules. As a side-note, some functions started taking too many parameters, so I made a type `{LUser, LServer, RemBareJid}`, which replicates the primary key that identifies inbox entries.

### DIFF
--- a/big_tests/dynamic_domains.spec
+++ b/big_tests/dynamic_domains.spec
@@ -10,32 +10,22 @@
 {include, "tests"}.
 
 {suites, "tests", acc_e2e_SUITE}.
-{suites, "tests", domain_isolation_SUITE}.
-{suites, "tests", muc_SUITE}.
-{suites, "tests", muc_light_SUITE}.
-{suites, "tests", muc_light_legacy_SUITE}.
-{suites, "tests", mam_SUITE}.
 
 {suites, "tests", carboncopy_SUITE}.
 {skip_cases, "tests", carboncopy_SUITE, [discovering_support],
  "at the moment mod_disco doesn't support dynamic domains"}.
 
-{suites, "tests", mod_ping_SUITE}.
-{skip_cases, "tests", mod_ping_SUITE, [disco],
- "at the moment mod_disco doesn't support dynamic domains"}.
+{suites, "tests", domain_isolation_SUITE}.
 
 {suites, "tests", inbox_SUITE}.
 {skip_cases, "tests", inbox_SUITE, [disco_service],
  "at the moment mod_disco doesn't support dynamic domains"}.
 {skip_cases, "tests", inbox_SUITE, [msg_sent_to_offline_user],
  "at the moment mod_offline doesn't support dynamic domains"}.
-{skip_groups, "tests", inbox_SUITE, [muclight, muc],
- "at the moment muc/muclight doesn't support dynamic domains"}.
 
 {suites, "tests", inbox_extensions_SUITE}.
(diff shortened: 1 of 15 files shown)


big_tests/dynamic_domains.spec
```diff
@@ -10,41 +10,10 @@
 {include, "tests"}.
 
 {suites, "tests", acc_e2e_SUITE}.
-{suites, "tests", domain_isolation_SUITE}.
```
chrzaszcz, line 13: Would it be possible to have alphabetical order of suites here, like for `default.spec`?

src/inbox/mod_inbox_muc.erl
```diff
@@ -121,4 +120,6 @@ write_to_receiver_inbox(Server, User, Remote, Packet, Acc) ->
 %% A local host can be used to fire hooks or write into database on this node.
 -spec is_local_xmpp_host(jid:lserver()) -> boolean().
 is_local_xmpp_host(LServer) ->
-    lists:member(LServer, ?MYHOSTS).
+    F = fun mongoose_domain_api:get_domains_by_host_type/1,
+    ServedDomains = lists:flatmap(F, ?ALL_HOST_TYPES),
+    lists:member(LServer, ServedDomains).
```
chrzaszcz, line 125: The correct way to do it would be:
```suggestion
    case mongoose_domain_api:get_domain_host_type(LServer) of
      {ok, _HostType} -> true;
      {error, not_found} -> false
    end
```
However, I think the returned host type shouldn't be ignored, see the comment above.

src/inbox/mod_inbox_muc.erl
```diff
@@ -60,18 +58,19 @@ update_inbox_for_muc(
...
+      HostType :: mongooseim:host_type(),
       Direction :: incoming | outgoing,
-      Host :: receiver_host(),
       Room :: room_bare_jid(),
       To :: receiver_bare_user_jid(),
       Packet :: packet().
-update_inbox_for_user(Direction, Host, Room, To, Packet) ->
+update_inbox_for_user(HostType, Direction, Room, To, Packet) ->
+    Host = To#jid.lserver,
     case {is_local_xmpp_host(Host), Direction} of
```
chrzaszcz, line 69: Here we are checking if `Host` matches **any** host type, but below we are passing **the** host type that we got as the first argument. Do we still need the check? Is it possible to run this hook for a host type that does not match `Host`? Please double check that. If so, we should check if both host types are identical.

src/inbox/mod_inbox.erl
```diff
@@ -38,72 +38,9 @@
...
-                      LServer :: jid:lserver(),
-                      EntryJID :: jid:literal_jid(),
-                      Params :: entry_properties(),
-                      Ret :: entry_properties() | {error, binary()}.
-
--callback remove_domain(HostType :: mongooseim:host_type(),
-                        LServer :: jid:lserver()) -> ok.
+-type entry_key() :: {LUser :: jid:luser(),
+                            LServer :: jid:lserver(),
+                            ToBareJid :: jid:literal_jid()}.
```
chrzaszcz, line 43: Minor: indentation

chrzaszcz (commented):
Good changes! I have a few comments.

chrzaszcz (approved):

PR #3143, opened 2021-06-01
Fix metrics in mam suite

gustawlippa:
This should fix the tests using metrics. Because we allow any string to be a host type, and in testing its `<<"test type">>` one has to ensure that the metrics are checked against a sanitised host type prefix, because metrics cannot contain spaces in their names.


chrzaszcz (approved):
Great!

PR #3146, opened 2021-06-01
Further mod_disco rework

chrzaszcz:
Further preparation and cleanup of disco-related code.

Some functionality was untested (and even broken). Tests added, bugs fixed. This will help assure that after introducing host types everything is still working.

Planned for the next PR: introduce host types to `mod_disco` and all the hooks to support dynamic domains.

Skipped for now: unit tests for `mongoose_disco`. It is a simple module covered by multiple functional tests.

Not sure: more pubsub tests. It's a deep rabbit hole, so I needed to stop at some point. A few lines I touched remain untested, just as before.

### DIFF
--- a/src/http_upload/mod_http_upload.erl
+++ b/src/http_upload/mod_http_upload.erl
@@ -114,7 +114,7 @@ disco_iq_handler(_From, To, Acc, #iq{type = get, lang = Lang, sub_el = SubEl} = 
     Node = xml:get_tag_attr_s(<<"node">>, SubEl),
     case Node of
         <<>> ->
-            Identity = disco_identity(Lang),
+            Identity = mongoose_disco:identities_to_xml(disco_identity(Lang)),
             Info = disco_info(LServer),
             Features = mongoose_disco:features_to_xml([?NS_HTTP_UPLOAD_030]),
             {Acc, IQ#iq{type = result,
@@ -125,7 +125,6 @@ disco_iq_handler(_From, To, Acc, #iq{type = get, lang = Lang, sub_el = SubEl} = 
             Error = mongoose_xmpp_errors:item_not_found(),
             {Acc, IQ#iq{type = error, sub_el = [SubEl, Error]}}
     end.
-
 
 -spec get_urls(Host :: jid:lserver(), Filename :: binary(), Size :: pos_integer(),
                ContentType :: binary() | undefined, Timeout :: pos_integer()) ->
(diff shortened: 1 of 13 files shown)


src/http_upload/mod_http_upload.erl
```diff
@@ -110,6 +104,27 @@ iq_handler(_From, _To = #jid{lserver = SubHost}, Acc, IQ = #iq{type = get, sub_e
     end,
     {Acc, Res}.
 
+-spec disco_iq_handler(From :: jid:jid(), To :: jid:jid(), Acc :: mongoose_acc:t(),
+                       IQ :: jlib:iq()) ->
+          {mongoose_acc:t(), jlib:iq()}.
+disco_iq_handler(_From, _To, Acc, #iq{type = set, sub_el = SubEl} = IQ) ->
+    {Acc, IQ#iq{type = error, sub_el = [SubEl, mongoose_xmpp_errors:not_allowed()]}};
```
arcusfelis, line 111: add some text to not_allowed. "Disco IQ with type set are not allowed", for example.

src/http_upload/mod_http_upload.erl
```diff
@@ -110,6 +104,27 @@ iq_handler(_From, _To = #jid{lserver = SubHost}, Acc, IQ = #iq{type = get, sub_e
...
+disco_iq_handler(_From, To, Acc, #iq{type = get, lang = Lang, sub_el = SubEl} = IQ) ->
+    LServer = To#jid.lserver,
```
arcusfelis, line 113: could be matched in the function clause.

src/pubsub/mod_pubsub.erl
```diff
@@ -796,10 +751,8 @@ disco_items(Host, Node, From) ->
                      Owners = node_owners_call(Host, Type, Nidx, O),
                      case get_allowed_items_call(Host, Nidx, From, Type, Options, Owners) of
                          {result, Items} ->
-                             {result, [#xmlel{name = <<"item">>,
-                                              attrs = [{<<"jid">>, jid:to_binary(Host)},
-                                                       {<<"name">>, ItemId}]}
-                                       || #pubsub_item{itemid = {ItemId, _}} <- Items]};
+                             {result, [disco_item(Host, ItemId) ||
+                                          #pubsub_item{itemid = {ItemId, _}} <- Items]};
```
arcusfelis, line 755: should be nice to have `[pubsub_to_disco_item(Host, Item) || Item Item <- Items]`. So, we would start crashing, if itemid format changes.

arcusfelis (commented):
added some coomens

DenysGonchar (approved):
LGTM

src/http_upload/mod_http_upload.erl
chrzaszcz: I'd rather limit the clause head to fit in one line less than 100 characters.

src/http_upload/mod_http_upload.erl
```diff
@@ -110,6 +104,27 @@ iq_handler(_From, _To = #jid{lserver = SubHost}, Acc, IQ = #iq{type = get, sub_e
...
+    case Node of
+        <<>> ->
+            Identity = mongoose_disco:identities_to_xml(disco_identity(Lang)),
+            Info = disco_info(LServer),
+            Features = mongoose_disco:features_to_xml([?NS_HTTP_UPLOAD_030]),
+            {Acc, IQ#iq{type = result,
+                        sub_el = [#xmlel{name = <<"query">>,
+                                         attrs = [{<<"xmlns">>, ?NS_DISCO_INFO}],
+                                         children = Identity ++ Info ++ Features}]}};
+        _ ->
```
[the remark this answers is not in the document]
chrzaszcz: It has to be empty. I will add a more descriptive error.

src/pubsub/mod_pubsub.erl
chrzaszcz: Good idea, but I think I will limit the changes for now. That code might return items that are not matching the pattern and the filtering might be done on purpose - I don't want to dive deeper into pubsub code now.

PR #3147, opened 2021-06-01
Multitenancy for the Client REST API for MAM and MUC-light

arcusfelis:
This PR addresses "Multitenancy for the Client REST API for MAM and MUC-light".

Proposed changes include:
* rooms and messages endpoints

* roster not included, because roster does not support MT yet.

* We also log more in rest_helper in tests now (similar to the way we log stanzas).

### DIFF
--- a/big_tests/tests/rest_client_SUITE.erl
+++ b/big_tests/tests/rest_client_SUITE.erl
@@ -115,14 +115,9 @@ security_test_cases() ->
 
 init_per_suite(Config) ->
     application:ensure_all_started(shotgun),
-    HostType = host_type(),
-    Config1 = dynamic_modules:save_modules(HostType, Config),
-    Config2 = rest_helper:maybe_enable_mam(mam_helper:backend(), HostType, Config1),
-    MucPattern = distributed_helper:subhost_pattern(muc_light_helper:muc_host_pattern()),
-    dynamic_modules:restart(HostType, mod_muc_light,
-                          [{host, MucPattern},
-                           {rooms_in_rosters, true}]),
-    [{muc_light_host, muc_light_helper:muc_host()} | escalus:init_per_suite(Config2)].
+    Config1 = init_modules(Config),
+    [{muc_light_host, muc_light_helper:muc_host()}
+     | escalus:init_per_suite(Config1)].
 
 end_per_suite(Config) ->
     escalus_fresh:clean(),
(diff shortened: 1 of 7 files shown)


big_tests/tests/rest_client_SUITE.erl
```diff
@@ -119,7 +119,7 @@ init_per_suite(Config) ->
     Config1 = dynamic_modules:save_modules(HostType, Config),
     Config2 = rest_helper:maybe_enable_mam(mam_helper:backend(), HostType, Config1),
     MucPattern = distributed_helper:subhost_pattern(muc_light_helper:muc_host_pattern()),
-    dynamic_modules:start(HostType, mod_muc_light,
+    dynamic_modules:restart(HostType, mod_muc_light,
```
chrzaszcz, line 122: Do we always need to restart? Could we call `ensure_modules` instead?

chrzaszcz, line 122: Could we use `ensure_modules` instead not to rely on the previous state?

src/muc_light/mod_muc_light_commands.erl
```diff
@@ -195,8 +196,8 @@ change_affiliation(Domain, RoomID, Sender, Recipient0, Affiliation) ->
 change_room_config(Domain, RoomID, RoomName, User, Subject) ->
     LServer = jid:nameprep(Domain),
     HostType = lserver_to_host_type(LServer),
-    MUCLightDomain = gen_mod:get_module_opt_subhost(
-                       Domain, mod_muc_light, mod_muc_light:default_host()),
+    HostType = mod_muc_light_utils:server_host_to_host_type(LServer),
```
chrzaszcz, line 199: This looks very strange, it basically does the same as the line above and pattern-matches the results. After cleaning this up, please check if we even need `lserver_to_host_type` as this functions does the same as `mod_muc_light_utils:server_host_to_host_type`.

chrzaszcz (commented):
Looks good in general, I have a few comments.

big_tests/tests/rest_client_SUITE.erl
arcusfelis: restart does not rely. It will stop, if it's running. But it will start if not running.

It's kinda like ensure_modules, but easier to migrate from `dynamic_modules:restart`.

Should be ensure_modules though eventually everywhere

src/muc_light/mod_muc_light_commands.erl
arcusfelis: yeah, was intended to be just one "HostType = "

chrzaszcz (approved):
Looks good!

PR #3152, opened 2021-06-09
Cleaning up subdomains management

DenysGonchar:
this PR introduces the following changes:
* all subdomains are registered using `mongoose_domain_api` interfaces - this change is required for mod_domain_isolation module.
* removal of mongoose_subhosts
* fix of mod_muc_iq generic iq component


src/mod_muc_room.erl
```diff
@@ -4581,12 +4581,12 @@ route_iq(Acc, #routed_iq{iq = #iq{type = Type, xmlns = ?NS_DISCO_ITEMS, lang = L
...
 route_iq(Acc, #routed_iq{iq = IQ = #iq{}, packet = Packet, from = From},
-         #state{host_type = HostType, jid = RoomJID} = StateData) ->
+         #state{host = Host, jid = RoomJID} = StateData) ->
     %% Custom IQ, addressed to this room's JID.
-    case mod_muc_iq:process_iq(HostType, From, RoomJID, Acc, IQ) of
+    case mod_muc_iq:process_iq(Host, From, RoomJID, Acc, IQ) of
         {Acc1, error} ->
             ?LOG_WARNING(#{what => muc_process_iq_failed, acc => Acc,
-                           host_type => HostType, room_jid => RoomJID}),
+                           host => Host, room_jid => RoomJID}),
```
arcusfelis, line 4589: log host_type field too.
host field is called `server`

src/mod_muc_iq.erl
```diff
@@ -36,24 +36,24 @@ start_link() ->
 
 %% @doc Handle custom IQ.
 %% Called from mod_muc_room.
--spec process_iq(mongooseim:host_type(), jid:jid(), jid:jid(), mongoose_acc:t(),
+-spec process_iq(jid:server(), jid:jid(), jid:jid(), mongoose_acc:t(),
         jlib:iq()) -> mongoose_acc:t() | {mongoose_acc:t(), error}.
-process_iq(HostType, From, RoomJID, Acc, IQ = #iq{xmlns = XMLNS}) ->
-    case ets:lookup(tbl_name(), {XMLNS, HostType}) of
+process_iq(Host, From, RoomJID, Acc, IQ = #iq{xmlns = XMLNS}) ->
```
arcusfelis, line 41: can we name it MucHost?
also type is probably jid:lserver()

arcusfelis (commented):
added some comments

src/ejabberd_router.erl
```diff
@@ -268,9 +277,9 @@ check_component_global(LDomain, Node) ->
     %% check that there is no component registered globally for this node
     case get_global_component(LDomain, Node) of
         undefined ->
-            ok;
+            false;
         _ ->
-            false
+            true
```
chrzaszcz, line 282: Ok, so we had a hidden bug, maybe we need to add a test?

chrzaszcz (approved):
Looks good, I have one minor comment.

src/ejabberd_router.erl
DenysGonchar: Not a bug, it was working correctly. But I think that having a validation function that returns `boolean() | ok`, when only `ok`value indicates a successful check, is super confusing. So I unified and reworked it a bit.

PR #3156, opened 2021-06-16
Support dynamic domains in mod_caps

chrzaszcz:
Add support for dynamic domains to mod_caps.

Hooks handled or triggered by `mod_caps` are refactored according to the following rule:
- If c2s state or `mongoose_acc:t()` is passed as an argument, do not add host type to the hook and handlers.
- Otherwise, add host type to the hook and its handlers.

Some lines remain untested as they are quite difficult to hit in the tests. Caps are mostly needed for PEP and it does not support dynamic domains, so the coverage for `dynamic_domains` is even lower. It could be increased in the future with the conversion of PubSub and PEP.

See individual commits for more details.

### DIFF
--- a/src/mod_caps.erl
+++ b/src/mod_caps.erl
@@ -362,26 +362,7 @@ init([HostType, Opts]) ->
                                timer:hours(24) div 1000),
     cache_tab:new(caps_features,
                   [{max_size, MaxSize}, {life_time, LifeTime}]),
-    ejabberd_hooks:add(c2s_presence_in, HostType, ?MODULE,
-                       c2s_presence_in, 75),
-    ejabberd_hooks:add(c2s_filter_packet, HostType, ?MODULE,
-                       c2s_filter_packet, 75),
-    ejabberd_hooks:add(c2s_broadcast_recipients, HostType,
-                       ?MODULE, c2s_broadcast_recipients, 75),
-    ejabberd_hooks:add(user_send_packet, HostType, ?MODULE,
-                       user_send_packet, 75),
-    ejabberd_hooks:add(user_receive_packet, HostType, ?MODULE,
-                       user_receive_packet, 75),
-    ejabberd_hooks:add(c2s_stream_features, HostType, ?MODULE,
-                       caps_stream_features, 75),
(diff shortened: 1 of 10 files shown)


src/mod_caps.erl
```diff
@@ -225,17 +239,18 @@ disco_info(Acc = #{node := Node}) ->
...
     Type = xml:get_attr_s(<<"type">>, Attrs),
     Subscription = ejabberd_c2s:get_subscription(From, C2SState),
-    Insert = ((Type == <<"">>) or (Type == <<"available">>))
-        and ((Subscription == both) or (Subscription == to)),
-    Delete = (Type == <<"unavailable">>) or (Type == <<"error">>),
-    case Insert or Delete of
+    Insert = (Type == <<>> orelse Type == <<"available">>)
+        and (Subscription == both orelse Subscription == to),
+    Delete = Type == <<"unavailable">> orelse Type == <<"error">>,
+    case Insert orelse Delete of
```
arcusfelis, line 253: I usually prefer `or`, when operations on booleans are made. because of better type checking at runtime
```
5> true or 4.
** exception error: bad argument
     in operator  or/2
        called as true or 4
6> true orelse 4.
true
```

chrzaszcz: On the other hand, short-circuit expressions are tail-recursive, quicker to execute and they make it obvious to the reader that the side-effects are not important. In this case I could go with both, `orelse` is just more natural to me, there is no complex type-checking involved here as the expressions are simple. I could revert, no problem for me.

src/mongoose_hooks.erl
```diff
@@ -1189,12 +1185,12 @@ s2s_send_packet(Acc, From, To, Packet) ->
 
 %%% @doc `s2s_stream_features' hook is used to extract
 %%% the stream management features supported by the server.
--spec s2s_stream_features(HostType, Server) -> Result when
+-spec s2s_stream_features(HostType, LServer) -> Result when
     HostType :: binary(),
```
arcusfelis, line 1189: host_type() type could be used here

arcusfelis (approved):
looks good

PR #3157, opened 2021-06-16
Use HostType variable in mod_mam_rdbms_prefs/user

arcusfelis:
This PR addresses MIM-1412 (kinda)

Proposed changes include:
* prefs/user and utils modules use HostType variable and host_type type where needed.
* no functional changes.
* ah, and I've refactored hooks there.

### DIFF
--- a/src/mam/mod_mam_cache_user.erl
+++ b/src/mam/mod_mam_cache_user.erl
@@ -78,18 +78,13 @@ start_server() ->
 hooks(HostType) ->
     PM = gen_mod:get_module_opt(HostType, ?MODULE, pm, false),
     MUC = gen_mod:get_module_opt(HostType, ?MODULE, muc, false),
-    case PM of
-        true ->
-            pm_hooks(HostType);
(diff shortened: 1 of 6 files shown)


src/mam/mod_mam_cache_user.erl
```diff
@@ -49,113 +48,70 @@ tbl_name_archive_id() ->
...
+hooks(HostType) ->
+    PM = gen_mod:get_module_opt(HostType, ?MODULE, pm, false),
+    MUC = gen_mod:get_module_opt(HostType, ?MODULE, muc, false),
+    case PM of
+        true ->
+            pm_hooks(HostType);
+        false ->
+            []
+    end ++
+    case MUC of
```
chrzaszcz, line 87: It's good, the only problem I see is that the part after `++` should be indented and with correct indentation it looks worse.

chrzaszcz (approved):
Looks good, I only have one minor style-related remark. Fixing is optional.

PR #3159, opened 2021-06-18
Support dynamic domains in mod_roster

chrzaszcz:
Main changes:
- Update roster-related hooks and their handlers to use host types.
- Update callbacks in `mod_roster` to use host types.
- Update all roster backends to implement the new callbacks and support multiple domains and host types.
- Enable roster-related tests with dynamic domains.

Side changes:
- Remove deprecated functions from `mod_roster`. They started becoming tedious to maintain.
- Change the templating of `mod_roster` to make the `test type` section consistent with the top-level ones by excluding the header from the template. This makes the templating itself a bit inconsistent, but IMO other modules can be converted as well. It makes sense to exclude constant parts from the template.
- Change handling of unexpected roster IQs directed to other JIDs - return errors.
- Make IQ handlers always respond with error 500 when they throw exceptions.
- Fix a bug in ejabberd_local (it caused errors for empty feature lists).
- Fix a bug in inbox tests (not enough cleanup, it was breaking subsequent tests).
- Fix and enable shared roster tests, they were always skipped.

Not changed:
- The `roster_groups` hook is only used by `mod_pubsub` and has no handlers. It can be removed, but we should figure out whether it needs a fix, as the resulting config field is always an empty list.
- Migration guide - needs to be addressed later.

Take a quick look at the commit messages for more information.


arcusfelis (approved):
looks good

rel/files/mongooseim.toml
```diff
@@ -256,7 +256,7 @@
   access = "register"
 {{#mod_roster}}
 
-{{{mod_roster}}}
+[modules.mod_roster]{{{mod_roster}}}
```
DenysGonchar, line 259: why do we still need templating here if mod_roster is explicitly enabled?

chrzaszcz: `{{#mod_roster}} ... {{/mod_roster}}` is a conditional block that will be inserted only if the `mod_roster` variable is defined.

DenysGonchar: does empty value count as defined or not?
```
{mod_roster, ""}.
```

DenysGonchar: Cause https://mustache.github.io/mustache.5.html specifies only behaviour for an empty list and `false` value

chrzaszcz: `{mod_roster, ""}` counts as defined - module enabled without any option.

chrzaszcz: This conditional expression is already used in many places to generate a nice TOML layout without empty lines. I think we could make it work for other modules work like for `mod_roster`.

DenysGonchar: Thanks for clarification.
I would probably prefer to have a new line in a template file:
```
[modules.mod_roster]
{{{mod_roster}}}
```
A loose a bit of readability in resulting file but have a bit more readable template 
but it can stay like that as well.

PR #3160, opened 2021-06-21
Add auth for mongoose_domain_handler

arcusfelis:
This PR addresses #

Proposed changes include:
* auth
* tests

### DIFF
--- a/big_tests/tests/service_domain_db_SUITE.erl
+++ b/big_tests/tests/service_domain_db_SUITE.erl
@@ -7,6 +7,11 @@
 -import(distributed_helper, [mim/0, mim2/0, require_rpc_nodes/1, rpc/4]).
 -import(ejabberdctl_helper, [ejabberdctl/3]).
 
+-define(TEST_PORT, 8866).
+-define(INV_PWD, <<"basic auth provided, invalid password">>).
+-define(NO_PWD, <<"basic auth is required">>).
+-define(UNWANTED_PWD, <<"basic auth provided, but not configured">>).
+
 suite() ->
     require_rpc_nodes([mim, mim2]).
 
@@ -23,7 +28,11 @@ all() ->
     ].
 
 groups() ->
-    [{db, [], db_cases()}].
+    [{db, [], db_cases()},
+     {rest_with_auth, [], [{group, rest_with_auth_parallel}|rest_cases()]},
+     {rest_without_auth, [], [{group, rest_without_auth_parallel}|rest_cases()]},
+     {rest_with_auth_parallel, [parallel], rest_auth_cases(true)},
+     {rest_without_auth_parallel, [parallel], rest_auth_cases(false)}].
 
 db_cases() -> [
      db_inserted_domain_is_in_db,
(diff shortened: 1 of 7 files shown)


src/domain/mongoose_domain_handler.erl
```diff
@@ -38,6 +40,28 @@ content_types_accepted(Req, State) ->
...
+        true ->
+            {true, Req, State};
+        false ->
+            {false, reply_error(403, <<"valid auth is required">>, Req), State}
+    end.
+
+state_to_details(#{username := User, password := Pass}) ->
+    {basic, User, Pass};
+state_to_details(_) ->
+    not_configured.
```
chrzaszcz, line 56: Please cover this case in the tests as this is the default.

rel/files/mongooseim.toml
```diff
@@ -62,6 +62,8 @@
   [[listen.http.handlers.mongoose_domain_handler]]
     host = "localhost"
     path = "/api"
+    username = "admin"
+    password = "secret"
```
chrzaszcz, line 66: I am not sure if it's good to set up such defaults in the config file... they aren't more secure than having no auth. Maybe we could leave out these two options? This way the user would have to consciously set something here. This would be also more in line with the docs, which docuemnt these two fields as empty by default. I know that this is not equal to them being empty in the default config file... but in this case I think they can be.

chrzaszcz (commented):
The code looks good in general. I have some minor comments.

chrzaszcz:
One more thing, it would be good to add a test case to `config_parser_SUITE` similar to `listen_http_handlers_api`, called e.g. `listen_http_handlers_domain`.

rel/files/mongooseim.toml
arcusfelis: do we have a way to dynamically start cowboy handlers?

big_tests/tests/service_domain_db_SUITE.erl
```diff
@@ -578,190 +618,233 @@ cli_disable_domain_fails_if_service_disabled(Config) ->
...
+rest_cannot_delete_domain_with_wrong_pass(Config) ->
+    {{<<"403">>,<<"Forbidden">>}, {[{<<"what">>, ?INV_PWD}]}} =
+        rest_delete_domain(set_invalid_creds(Config), <<"example.db">>, <<"type1">>).
 
-rest_cannot_insert_domain_without_valid_auth(Config) ->
-    {{<<"403">>,<<"Forbidden">>}, {[{<<"what">>, <<"valid auth is required">>}]}} =
-        rest_put_domain_with_invalid_auth(<<"example.db">>, <<"type1">>).
+rest_cannot_enable_domain_with_wrong_pass(Config) ->
+    {{<<"403">>,<<"Forbidden">>}, {[{<<"what">>, ?INV_PWD}]}} =
+        rest_patch_enabled(set_invalid_creds(Config), <<"example.db">>, false).
```
chrzaszcz, line 698: Should be enabled:
```suggestion
    {{<<"403">>,<<"Forbidden">>}, {[{<<"what">>, ?INV_PWD}]}} =
        rest_patch_enabled(set_invalid_creds(Config), <<"example.db">>, true).
```

big_tests/tests/service_domain_db_SUITE.erl
```diff
@@ -578,190 +618,233 @@ cli_disable_domain_fails_if_service_disabled(Config) ->
...
+    {{<<"403">>,<<"Forbidden">>}, {[{<<"what">>, ?NO_PWD}]}} =
+        rest_put_domain(set_no_creds(Config), <<"example.db">>, <<"type1">>).
+
+rest_cannot_delete_domain_without_auth(Config) ->
+    {{<<"403">>,<<"Forbidden">>}, {[{<<"what">>, ?NO_PWD}]}} =
+        rest_delete_domain(set_no_creds(Config), <<"example.db">>, <<"type1">>).
+
+rest_cannot_enable_domain_without_auth(Config) ->
+    {{<<"403">>,<<"Forbidden">>}, {[{<<"what">>, ?NO_PWD}]}} =
+        rest_patch_enabled(set_no_creds(Config), <<"example.db">>, false).
```
chrzaszcz, line 719: Should be enabled:
```suggestion
        rest_patch_enabled(set_no_creds(Config), <<"example.db">>, true).
```

chrzaszcz (commented):
Looks good! I have two minor remarks.

chrzaszcz (approved):
Great!

PR #3161, opened 2021-06-23
Fix user cache

NelsonVides:
As the tables are created per-host-type, a table can have more than one
domain within, and so there's no unique key properly defined,
considering that the same namepart can be used in two different domains,
and the same domain contains many nameparts.

The cache was broken because, as a set where the domain is the key, inserted users would always override previous. If namepart is the key, two users with the same namepart in different domains would also override each other. So the whole pair needed to be made the key as a 1-tuple.


chrzaszcz (approved):
Great. I think it's best to squash these two commits on merge.

PR #3162, opened 2021-06-25
Fix rendering of domain management docs, note table of contents

NelsonVides:
mkdocs-material loses the table of contents if we start from the top index (single #).


arcusfelis (approved):
a

PR #3163, opened 2021-06-28
Create a configurable FIFO cache

NelsonVides:
This uses the strategy of keeping a set of ets tables that are rotated
periodically. This means that TTLs have a very coarse-grained precision,
but book-keeping is considerably simpler.

Many things can be done from here, like whether we want to create a
`does_user_exist` hook where this and ejabberd_auth hook and inbox and
mam query, or if we want to add to the cache the calls from
ejabberd_auth and mam and inbox query here directly, or through a hook,
etcetera.

TODOs:
- [x] Make `does_user_exist` a hook a described above.
- [x] Introduce usage of this module instead of `mongoose_users`.
- [ ] Introduce API to just clear cache entries.


src/mod_cache_users.erl
```diff
@@ -0,0 +1,237 @@
...
+start_cache(HostType, Opts) ->
+    Tab = tbl_name(HostType),
+    N   = gen_mod:get_opt(n, Opts, 1),
+    TTL = gen_mod:get_opt(ttl, Opts, infinity),
+    RC  = gen_mod:get_opt(read_concurrency, Opts, true),
+    WC  = gen_mod:get_opt(write_concurrency, Opts, true),
+    ParentTab = ets:new(Tab, [ordered_set, named_table, public, {read_concurrency, true}]),
+    lists:foreach(
+      fun(I) ->
+              Ref = ets:new(Tab, [set, public, {read_concurrency, RC}, {write_concurrency, WC}]),
```
arcusfelis, line 184: maybe just use named tables, like `Tab ++ SegmentNumStr`?

Also, if you check logs carefully, it would complain that you are creating ets tables from inside of `gen_mod:start_module`, which is big no. We can't guarantee that the caller of the callback would stay alive. Basically, it stops working with RPC-s. So, create the tables from the inside of the gen_server.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,237 @@
...
+%%--------------------------------------------------------------------
+%% gen_server callbacks
+%%--------------------------------------------------------------------
+
+-record(cache_state, {tab :: atom()}).
+
+purge_tables(#cache_state{tab = Tab}) ->
+    First = ets:first(Tab),
+    Last = ets:last(Tab),
+    Ets = case ets:info(Tab, size) of
```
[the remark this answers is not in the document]
NelsonVides: This is rotating the tables and then taking who's the last table and dropping all info on that one.

src/mod_cache_users.erl
NelsonVides: Don't want to use named tables, I want the parent table to be the interface to the segments.

Creating the table from gen_mod, I think I know what to do there.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,236 @@
...
+    end.
+
+%%====================================================================
+%% Hooks
+%%====================================================================
+
+-spec remove_user(Acc :: mongoose_acc:t(),
+                  LUser :: jid:luser(),
+                  LServer :: jid:lserver() | string()) -> mongoose_acc:t().
+remove_user(Acc, LUser, LServer) ->
```
arcusfelis, line 90: that would not work across the cluster. You have to remove from all nodes (it was broken in the original ejabberd_users version too, btw)

src/mod_cache_users.erl
```diff
@@ -0,0 +1,236 @@
...
+    EtsOpts = [ordered_set, named_table, protected, {read_concurrency, true}],
+    ParentTab = ets:new(ParentTab, EtsOpts),
+    lists:foreach(
+      fun(I) ->
+              SegmentOpts = [set, public, {read_concurrency, true}, {write_concurrency, true}],
+              Ref = ets:new(ParentTab, SegmentOpts),
+              ets:insert(ParentTab, {I, Ref})
+      end, lists:seq(1, N)),
+    erlang:send_after(TTL, self(), {purge, TTL}),
+    {ok, #cache_state{tab = ParentTab}}.
```
arcusfelis, line 220: It's a good idea to keep ParentTab name instead of tab/Tab everywhere. Even if it's a longer version.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,236 @@
...
+
+handle_call(Msg, From, State) ->
+    ?UNEXPECTED_CALL(Msg, From),
+    {reply, ok, State}.
+
+handle_cast(Msg, State) ->
+    ?UNEXPECTED_CAST(Msg),
+    {noreply, State}.
+
+handle_info({purge, TTL}, State) ->
```
arcusfelis, line 230: we usually keep TTL inside of the process state.
Also, it could be a good idea to reset the old timer, if someone wants to trigger purging manually.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,236 @@
...
+                true -> true
+            end;
+        [] ->
+           is_member(Tab, Key, ets:prev(Tab, Table))
+    end.
+
+%% There's a chance that by the time we insert in the ets table, this table is not
+%% the first anymore because the cleaner has taken action and pushed it behind.
+%% That's fine, worst case this record will live a segment less than expected.
+put_member(Tab, Key, Table) ->
```
arcusfelis, line 145: Tab, Table... Poor people trying to read that later XDD
ParentTab, SegmentTab (ends with Tab - we know its a tab).

src/mod_cache_users.erl
NelsonVides: Keeping it inside the state, ofc. Reset the old timer, only if someone wants to trigger purging manually, but I haven't exposed any API for it yet.

src/mod_cache_users.erl
NelsonVides: Why wouldn't it work? The hook is called on a single node? How would I fix that in all others?

arcusfelis: The hooks is called on the single node. It's for cleaning database, not for cleaning the local node state.
 Do an rpc call to other nodes. Or send a message.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,236 @@
...
+
+init([ParentTab, Opts]) ->
+    TTL = gen_mod:get_opt(ttl, Opts, infinity),
+    N   = gen_mod:get_opt(number_of_segments, Opts, 1),
+    EtsOpts = [ordered_set, named_table, protected, {read_concurrency, true}],
+    ParentTab = ets:new(ParentTab, EtsOpts),
+    lists:foreach(
+      fun(I) ->
+              SegmentOpts = [set, public, {read_concurrency, true}, {write_concurrency, true}],
+              Ref = ets:new(ParentTab, SegmentOpts),
```
arcusfelis, line 216: that line is confusing. Could you use a DIFFERENT names for the parent (named) table and segment tables? 
Even if there is no logic in your code using that, the debugging of unnamed becomes kinda a pain. i.e. simple `ets:info(users_cache_my_cool_host_1).` would not work. Of course, there could be a work around, but it's not simple.

I would rather name the resources, that could potentially eat some memory ;)

src/mod_cache_users.erl
NelsonVides: Made a pg2 group.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,258 @@
...
+      end, lists:seq(1, N)),
+    erlang:send_after(TTL, self(), purge),
+    {ok, #cache_state{tab = ParentName, ttl = TTL}}.
+
+handle_call(Msg, From, State) ->
+    ?UNEXPECTED_CALL(Msg, From),
+    {reply, ok, State}.
+
+handle_cast({remove_user, Key}, #cache_state{tab = ParentTab} = State) ->
+    del_member(ParentTab, Key, ets:first(ParentTab)),
```
arcusfelis, line 243: we could wrap handle_remove_user and handle_remove_domain into try...catch with error reporting I think. No need to crash the process, if something is wrong in the removal code.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,258 @@
...
+            ok
+    end,
+    Acc.
+
+-spec remove_domain(mongoose_hooks:simple_acc(),
+                    mongooseim:host_type(), jid:lserver()) ->
+    mongoose_hooks:simple_acc().
+remove_domain(Acc, HostType, Domain) ->
+    ParentName = tbl_name(HostType),
+    case pg2:get_members(ParentName) of
```
arcusfelis, line 107: send_to_group could be a helper ;)

arcusfelis (approved):
looks ok to me, just few tiny naming comments

big_tests/tests/mam_SUITE.erl
```diff
@@ -3068,9 +3068,11 @@ check_user_exist(Config) ->
   ok = rpc(mim(), ejabberd_auth, try_register, [JID, AdminP]),
   %% admin user already registered
   {ok, HostType} = rpc(mim(), mongoose_domain_core, get_host_type, [AdminS]),
-  true = rpc(mim(), mongoose_users, does_user_exist, [HostType, JID]),
-  false = rpc(mim(), mongoose_users, does_user_exist, [HostType, mongoose_helper:make_jid(<<"fake-user">>, AdminS)]),
-  false = rpc(mim(), mongoose_users, does_user_exist, [HostType, mongoose_helper:make_jid(AdminU, <<"fake-domain">>)]),
+  #{result := true} = rpc(mim(), mongoose_hooks, does_user_exist, [HostType, JID, stored]),
```
DenysGonchar, line 3071: Not sure if this is the best place to add `does_user_exist` hook testing, but I don't have any better suggestions either.

src/auth/ejabberd_auth.erl
```diff
@@ -328,6 +333,21 @@ does_stored_user_exist(HostType, #jid{luser = LUser, lserver = LServer}) ->
 does_stored_user_exist(_HostType, error) ->
     false.
 
+-spec does_user_exist(Status :: mongoose_hooks:simple_acc(),
+                      HostType :: mongooseim:host_type(),
+                      Jid :: jid:jid()) -> does_user_exist().
+does_user_exist(#{result := true} = Status, _, _) ->
+    Status;
+does_user_exist(#{type := stored} = Status, HostType, Jid) ->
```
DenysGonchar, line 341: it seems that `type` should be a hook argument, not an acc field. we are not supposed to change it.

src/auth/ejabberd_auth.erl
```diff
@@ -328,6 +333,21 @@ does_stored_user_exist(HostType, #jid{luser = LUser, lserver = LServer}) ->
 does_stored_user_exist(_HostType, error) ->
     false.
 
+-spec does_user_exist(Status :: mongoose_hooks:simple_acc(),
+                      HostType :: mongooseim:host_type(),
+                      Jid :: jid:jid()) -> does_user_exist().
+does_user_exist(#{result := true} = Status, _, _) ->
```
DenysGonchar, line 339: Instead of having this manual `result` check, we should just stop hook processing (return `{stop, NewAcc}` from the hook) if there's a cache record found

src/auth/ejabberd_auth.erl
```diff
@@ -61,15 +61,18 @@
...
 -include("jlib.hrl").
 
 -export_type([authmodule/0,
-              passterm/0]).
+              passterm/0,
+              does_user_exist/0]).
 
 -type authmodule() :: module().
 -type passterm() :: binary() | mongoose_scram:scram_tuple() | mongoose_scram:scram_map().
+-type does_user_exist() :: #{result := boolean(), cached := boolean(), '_' := '_'}.
```
DenysGonchar, line 75: do we really use this `cached` field anywhere?

src/mongoose_hooks.erl
```diff
@@ -207,6 +208,16 @@ anonymous_purge_hook(LServer, Acc, LUser) ->
 auth_failed(HostType, Server, Username) ->
     ejabberd_hooks:run_for_host_type(auth_failed, HostType, ok, [Username, Server]).
 
+-spec does_user_exist(HostType, Jid, Type) -> Result when
+    HostType :: binary(),
+    Jid :: jid:jid(),
+    Type :: stored | with_anonymous,
+    Result :: ejabberd_auth:does_user_exist().
+does_user_exist(HostType, Jid, Type) ->
```
DenysGonchar, line 216: Type should be an argument of the hook, not a field of the Acc

src/event_pusher/mod_event_pusher_push_plugin_defaults.erl
```diff
@@ -80,10 +80,10 @@ publish_notification(Acc, _, Payload, Services) ->
 -spec should_publish(Acc :: mongoose_acc:t(), To :: jid:jid()) -> boolean().
 should_publish(Acc, #jid{} = To) ->
     HostType = mongoose_acc:host_type(Acc),
-    try mongoose_users:does_user_exist(HostType, To) of
-        false ->
+    try mongoose_hooks:does_user_exist(HostType, To, stored) of
```
DenysGonchar, line 83: Should we have a generic wrapper interface in ejabberd_auth module (e.g. `ejabberd_hook:does_user_exist/3`) that triggers the hook and returns just a boolean value? From my perspective, ejabberd_auth is the right place for such an interface.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,268 @@
...
+-spec stop(mongooseim:host_type()) -> ok.
+stop(HostType) ->
+    ejabberd_hooks:delete(hooks(HostType)),
+    stop_server(HostType),
+    ok.
+
+-spec config_spec() -> mongoose_config_spec:config_section().
+config_spec() ->
+    #section{items = #{
+                       <<"time_to_live">> => #option{type = int_or_infinity,
```
DenysGonchar, line 56: `infinity` doesn't make any sense if we don't have any size limit for a section

src/auth/ejabberd_auth.erl
NelsonVides: It's not changing it, it's matching on it only.

src/auth/ejabberd_auth.erl
NelsonVides: On the third subscriber, to know that we don't need to cache it again.

src/event_pusher/mod_event_pusher_push_plugin_defaults.erl
NelsonVides: I like this idea very much actually.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,268 @@
...
+-spec remove_user(Acc :: mongoose_acc:t(),
+                  LUser :: jid:luser(),
+                  LServer :: jid:lserver() | string()) -> mongoose_acc:t().
+remove_user(Acc, LUser, LServer) ->
+    Key = key(LUser, LServer),
+    HostType = mongoose_acc:host_type(Acc),
+    send_to_group(HostType, {remove_user, Key}),
+    Acc.
+
+-spec remove_domain(mongoose_hooks:simple_acc(),
```
DenysGonchar, line 87: just a side not, not related to this PR.
we have introduced `remove_domain` but we haven't added `mongoose_hooks:remove_domain/2` execution on `mongoose_domain_api:delete_domain/2` call

src/mod_cache_users.erl
NelsonVides: infinity is just to keep the same behaviour than before, in case somebody wants it — might be that some deployment only has a few thousand users and then the entire cache kept forever is just a few megas of memory, somebody might want that.

src/auth/ejabberd_auth.erl
DenysGonchar: so that's an argument, not a part of the Acc.

src/auth/ejabberd_auth.erl
DenysGonchar: don't get it, do you want to have a multi layer cache? I don't think that we need it.

src/mod_cache_users.erl
DenysGonchar: well, I hope we will add size limitation for sections in the future.

src/auth/ejabberd_auth.erl
NelsonVides: If the first subscriber returned true, the third (which is the same module, mod_cache_users) should not cache it again, so the third subscriber (put_user_into_cache or sth like that), needs to know who said it's true, the cache, or the backend, to know if it should cache it or not.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,268 @@
...
+
+-spec supported_features() -> [atom()].
+supported_features() ->
+    [dynamic_domains].
+
+hooks(HostType) ->
+    [
+     {does_user_exist, HostType, ?MODULE, does_cached_user_exist, 30},
+     {does_user_exist, HostType, ?MODULE, maybe_put_user_into_cache, 70},
+     {remove_user, HostType, ?MODULE, remove_user, 30},
```
DenysGonchar, line 70: We must ensure that `remove_user` and `remove_domain` hooks are executed first for  `ejabberd_auth`, after that for `mod_cache_users` and only then for all other modules.
So there is no chance that we asynchronously re-insert the removed user in the cache, or insert any permanent data (e.g. MAM, Inbox) for the removed but still cached user.
I suggest to change priorities for  `remove_user` & `remove_domain` hooks in the following way:
* ejabberd_auth should have prio=1
* mod_cache_users should have prio=2
* all other modules - prio>30

src/mod_cache_users.erl
```diff
@@ -0,0 +1,268 @@
...
+     {remove_domain, HostType, ?MODULE, remove_domain, 30}
```
DenysGonchar, line 71: Please note that `remove_domain` hook is executed on one node in the MIM cluster. This is desired behaviour for data removal from external DBs, but it requires extra work to ensure that local cache is cleaned on all the nodes in the MIM cluster.

src/auth/ejabberd_auth.erl
DenysGonchar: again, why don't we just stop the execution of the hook?

src/mod_cache_users.erl
NelsonVides: Very good point.

src/auth/ejabberd_auth.erl
NelsonVides: Ok, we're going to think nobody's inserting a handler _after_ the cache has already said `true`, so I'm doing a stop here. This allows me to simplify the accumulator and the second and third handlers, as they won't even be called if the cache was already called.

src/mod_cache_users.erl
NelsonVides: Ouch, that seems like it should be done, really. Wanna make a ticket for it and put it first in our backlog?

src/mod_cache_users.erl
NelsonVides: In the case of `remove_user`, `mongoose_hooks:remove_user` is only called by `ejabberd_auth:remove_user`, after `ejabberd_auth` has called `remove_user` on all of its backends. So we're safe there.
For `remove_domain`, gonna do that.

NelsonVides: For `remove_domain`, all modules have priority 50. I've set ejabberd_auth to have 10 and the cache to have 20, just to leave space for other handlers, were any to be desired.

doc/modules/mod_cache_users.md
```diff
@@ -0,0 +1,33 @@
+## Module Description
+
+`mod_cache_users` is a module that caches whether a user exists. This is useful for example to decide if a message should be stored in [MAM] or [Inbox] — for example, the receiver might not exist, so no message should be stored on his archive nor his inbox.
```
chrzaszcz, line 3:
```suggestion
`mod_cache_users` is a module that caches whether a user exists. This is useful for example to decide if a message should be stored in [MAM] or [Inbox] — for example, the receiver might not exist, so no message should be stored in his archive nor his inbox.
```

src/auth/ejabberd_auth.erl
```diff
@@ -316,6 +328,30 @@ does_user_exist(#jid{luser = LUser, lserver = LServer}) ->
 does_user_exist(error) ->
     false.
 
+%% Hook interface
+-spec does_user_exist(HostType, Jid, RequestType) -> Res when
+      HostType :: mongooseim:host_type(),
+      Jid :: jid:jid(),
+      RequestType :: stored | with_anonymous,
```
chrzaszcz, line 335: Why not use `exist_type`? I also think that the variable names don't give any extra information in this case, only more lines.

src/auth/ejabberd_auth.erl
```diff
@@ -316,6 +328,30 @@ does_user_exist(#jid{luser = LUser, lserver = LServer}) ->
...
+      Res :: boolean().
+does_user_exist(HostType, Jid, RequestType) ->
+    mongoose_hooks:does_user_exist(HostType, Jid, RequestType).
+
+%% Hook handler
+-spec does_user_exist(Status :: boolean(),
```
chrzaszcz, line 341: Same as above.

src/auth/ejabberd_auth.erl
```diff
@@ -316,6 +328,30 @@ does_user_exist(#jid{luser = LUser, lserver = LServer}) ->
...
+                      HostType :: mongooseim:host_type(),
+                      Jid :: jid:jid(),
+                      RequestType :: stored | with_anonymous) -> boolean().
+does_user_exist(false, HostType, Jid, stored) ->
+    does_stored_user_exist(HostType, Jid);
```
chrzaszcz, line 346: `does_user_exist` is supposed to return a boolean, but it calls a function that can return `{error, any()}`. In this case we should treat errors as `false` and add a comment that says so to the function header.

```suggestion
    true =:= does_stored_user_exist(HostType, Jid);
```

src/mod_cache_users.erl
```diff
@@ -0,0 +1,262 @@
...
+    stop_server(HostType),
+    ok.
+
+-spec config_spec() -> mongoose_config_spec:config_section().
+config_spec() ->
+    #section{items = #{
+                       <<"time_to_live">> => #option{type = int_or_infinity,
+                                                     validate = non_negative,
+                                                     format = {kv, ttl}},
+                       <<"number_of_segments">> => #option{type = integer}
```
chrzaszcz, line 59:
```suggestion
                       <<"number_of_segments">> => #option{type = integer,
                                                           validate = positive}
```

chrzaszcz:
Review in progress, please **do not merge** yet.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,262 @@
...
+                         true = ets:insert_new(ParentTab, {Last + 1, E}),
+                         E
+                 end,
+    ets:delete_all_objects(EtsSegment),
+    ok.
+
+iterate_through_tables(ParentTab, Action) when is_atom(ParentTab), is_function(Action, 1) ->
+    iterate_(ParentTab, Action, ets:first(ParentTab)).
+
+iterate_(_, _, '$end_of_table') ->
```
chrzaszcz, line 192: Minor, style, please do not end function names with underscores.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,262 @@
...
+-spec stop_server(mongooseim:host_type()) -> any().
+stop_server(HostType) ->
+    ok = ejabberd_sup:stop_child(tbl_name(HostType)).
+
+%%--------------------------------------------------------------------
+%% gen_server callbacks
+%%--------------------------------------------------------------------
+
+-record(cache_state, {tab :: atom(),
+                      ttl = timer:hours(8) :: timeout()}).
```
chrzaszcz, line 173: It looks like a duplicated default, do we need to have the "8" in two places? They can get out of sync.

src/mod_cache_users.erl
```diff
@@ -0,0 +1,262 @@
...
+            iterate_(ParentTab, Action, ets:next(ParentTab, SegmentKey))
+    end.
+
+-spec start_link(atom(), gen_mod:module_opts()) -> {ok, pid()}.
+start_link(ParentTab, Opts) ->
+    gen_server:start_link({local, ParentTab}, ?MODULE, [ParentTab, Opts], []).
+
+init([ParentName, Opts]) ->
+    TTL0 = gen_mod:get_opt(ttl, Opts, 8 * 60), %% 8h
+    N    = gen_mod:get_opt(number_of_segments, Opts, 3),
```
chrzaszcz, line 209: Minor, style
```suggestion
    N = gen_mod:get_opt(number_of_segments, Opts, 3),
```

src/mod_cache_users.erl
```diff
@@ -0,0 +1,262 @@
...
+    EtsOpts = [ordered_set, named_table, protected, {read_concurrency, true}],
+    ets:new(ParentName, EtsOpts),
+    lists:foreach(
+      fun(I) ->
+              %% Note, these names are only for debugging purposes,
```
chrzaszcz, line 214: What are "these names"? The tables are not named.

chrzaszcz (commented):
Nice improvements in the cache. I have a few comments.

There is one more thing that is really important: please add a test to `config_parser_SUITE` in small tests. We worked really hard to test all config options for all modules, let's keep it that way.

Another question: is there a test that checks how everything works **without** the cache? It is enabled for all dev nodes.

src/mod_cache_users.erl
NelsonVides: Ah, they were named before and then when un-naming I forgot to remove the comment

NelsonVides:
> Nice improvements in the cache. I have a few comments.
> 
> There is one more thing that is really important: please add a test to `config_parser_SUITE` in small tests. We worked really hard to test all config options for all modules, let's keep it that way.
> 
> Another question: is there a test that checks how everything works **without** the cache? It is enabled for all dev nodes.

All other comments solved and pushed. These ones, gonna take action, good ones 👍🏽

chrzaszcz (approved):

PR #3164, opened 2021-06-30
Support dynamic domains in mod_offline

chrzaszcz:
**Main changes:**
- Support host types in `mod_offline` and its backends.
- Enable offline-related tests for dynamic domains.
- Change module templating to make the template and the resulting config readable at the same time. There are some whitespace manipulations, but they don't change the functionality. This can be still improved, I have ideas (e.g. mustache lists), but it would be a separate story. Right now the resulting config file always has a single empty line separating module sections. All module templates should be unified in the near future.

**Side changes:**
- Unify the implementation of `remove_*_messages` to only remove messages for the domain that was provided. The backends were behaving inconsistently before (it worked as expected only for Riak).

**Not included in this PR:**
- DB Query optimization with indexes. It does not seem to be required now, but message removal per domain could be sped up by using indexes.
- Support for dynamic domains in `mod_offline_chatmarkers`. This is an undocumented module which is disabled by default.
- Better error handling in backends. Sometimes DB errors can be missed and `ok` is returned. I didn't want to change too much and these issues don't seem to be critical. It's a bit out of scope as well.

### DIFF
--- a/rel/files/mongooseim.toml
+++ b/rel/files/mongooseim.toml
@@ -211,46 +211,46 @@
   periodic_report = 10_800_000
 
 [modules.mod_adhoc]
+
 {{#mod_amp}}
-
 {{{mod_amp}}}
-{{/mod_amp}}
 
+{{/mod_amp}}
 [modules.mod_disco]
   users_can_see_hidden_services = false
 
 [modules.mod_commands]
+
 {{#mod_cache_users}}
-
-[modules.mod_cache_users]{{{mod_cache_users}}}
+[modules.mod_cache_users]
+{{{mod_cache_users}}}
 {{/mod_cache_users}}
-
 [modules.mod_muc_commands]
 
 [modules.mod_muc_light_commands]
+
 {{#mod_last}}
-
 {{{mod_last}}}
+
 {{/mod_last}}
-
 [modules.mod_stream_management]
-{{#mod_offline}}
 
+{{#mod_offline}}
+[modules.mod_offline]
 {{{mod_offline}}}
 {{/mod_offline}}
 {{#mod_privacy}}
-
 {{{mod_privacy}}}
+
 {{/mod_privacy}}
 {{#mod_blocking}}
-
 {{{mod_blocking}}}
+
 {{/mod_blocking}}
 {{#mod_private}}
-
 {{{mod_private}}}
-{{/mod_private}}
 
+{{/mod_private}}
 [modules.mod_register]
   welcome_message = {body = "", subject = ""}
   ip_access = [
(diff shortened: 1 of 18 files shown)


rel/files/mongooseim.toml
```diff
@@ -211,64 +211,64 @@
...
 [modules.mod_disco]
   users_can_see_hidden_services = false
 
 [modules.mod_commands]
-{{#mod_cache_users}}
 
-[modules.mod_cache_users]{{{mod_cache_users}}}
+{{#mod_cache_users}}
+[modules.mod_cache_users]
+{{{mod_cache_users}}}
```
NelsonVides, line 226: This disables this module in the prod config.

src/ejabberd_admin.erl
```diff
@@ -442,22 +442,33 @@ get_loglevel() ->
 %%% Purge DB
 %%%
 
--spec delete_expired_messages(jid:lserver()) -> {ok, iolist()} | {error, iolist()}.
+-spec delete_expired_messages(mongooseim:host_type()) -> {ok, iolist()} | {error, iolist()}.
```
NelsonVides, line 445: This is still taking a `lserver`, so the type shouldn't have changed.

src/ejabberd_admin.erl
```diff
@@ -442,22 +442,33 @@ get_loglevel() ->
...
+                    {ok, io_lib:format("Removed ~p messages", [C])};
+                {error, Reason} ->
+                    {error, io_lib:format("Can't delete expired messages: ~n~p", [Reason])}
+            end;
+        {error, not_found} ->
+            {error, "Unknown domain"}
     end.
 
--spec delete_old_messages(jid:lserver(), Days :: integer()) -> {ok, iolist()} | {error, iolist()}.
+-spec delete_old_messages(mongooseim:host_type(), Days :: integer()) ->
```
NelsonVides, line 459: As above.

rel/files/mongooseim.toml
chrzaszcz: It is only a whitespace change, it cannot disable anything.

src/ejabberd_admin.erl
chrzaszcz: Good catch! I changed it back and forth and forgot to revert this one.

src/offline/mod_offline.erl
```diff
@@ -212,23 +210,24 @@ write_messages(LUser, LServer, Msgs) ->
             discard_warn_sender(Msgs)
     end.
 
--spec is_message_count_threshold_reached(integer(), jid:luser(),
-                                         jid:lserver(), integer()) ->
+-spec is_message_count_threshold_reached(host_type(), integer(), jid:luser(), jid:lserver(),
```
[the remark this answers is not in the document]
chrzaszcz: Indeed :-) I did not check that part.

rel/files/mongooseim.toml
chrzaszcz: It was disabled before and remains unchanged. Let's change this in a separate PR.

NelsonVides (approved):
LGTM

arcusfelis (commented):
ok

PR #3165, opened 2021-06-30
Mam inbox unification

NelsonVides:
If inbox is nothing more than a low-cost summary of mam, we can expect inbox to store pretty much what is stored in mam, and nothing else.

This is currently not true, as inbox is not taking into account `no-store` hints for example, and it is storing body-less messages, like for example it would store carbon copies.

Additionally, make both take the `<store>` hint as enforcing storage.

### DIFF
--- a/big_tests/tests/inbox_SUITE.erl
+++ b/big_tests/tests/inbox_SUITE.erl
@@ -30,6 +30,7 @@
         ]).
 -export([msg_sent_stored_in_inbox/1,
          msg_with_no_store_is_not_stored_in_inbox/1,
+         msg_with_store_hint_is_always_stored/1,
          carbons_are_not_stored/1,
          user_has_empty_inbox/1,
          user_has_two_unread_messages/1,
@@ -133,6 +134,7 @@ groups() ->
            user_has_empty_inbox,
            msg_sent_stored_in_inbox,
            msg_with_no_store_is_not_stored_in_inbox,
+           msg_with_store_hint_is_always_stored,
            carbons_are_not_stored,
            user_has_two_conversations,
            msg_sent_to_offline_user,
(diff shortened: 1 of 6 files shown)


big_tests/tests/inbox_SUITE.erl
```diff
@@ -427,6 +433,45 @@ msg_sent_stored_in_inbox(Config) ->
...
+        escalus:wait_for_stanza(Bob),
+        %% Alice and Bob has a body-less message in their inbox
+        check_inbox(Bob, [#conv{unread = 1, from = Alice, to = Bob, content = <<>>}]),
+        check_inbox(Alice, [#conv{unread = 0, from = Alice, to = Bob, content = <<>>}])
+      end).
+
+carbons_are_not_stored(Config) ->
+    escalus:fresh_story(Config, [{alice, 2}, {bob, 2}], fun(Alice1, Alice2, Bob1, Bob2) ->
+        mongoose_helper:enable_carbons([Alice1, Alice2, Bob1, Bob2]),
+        %% Alice1 sends a message to Bob with a no-store hint
```
chrzaszcz, line 468: Where is the no-store hint? Probably a copy-pasted artifact?

src/mam/mod_mam_utils.erl
```diff
@@ -323,13 +324,19 @@ is_archivable_message(Mod, Dir, Packet=#xmlel{name = <<"message">>}, ArchiveChat
...
 is_valid_message_type(_, _, <<"normal">>) -> true;
 is_valid_message_type(_, _, <<"chat">>) -> true;
 is_valid_message_type(_, incoming, <<"groupchat">>) -> true;
 is_valid_message_type(_, _, <<"error">>) -> false;
 is_valid_message_type(_, _, _) -> false.
 
 is_valid_message(_Mod, _Dir, Packet, ArchiveChatMarkers) ->
+    %% Message Processing Hints (XEP-0334)
+    Store = exml_query:path(Packet, [{element_with_ns, <<"store">>, ?NS_HINTS}], false) =/= false,
+    Store orelse is_valid_message(Packet, ArchiveChatMarkers).
```
chrzaszcz, line 337: Why not add `store` to the `has_any` list in line 351 below to stay consistent with the check for `no-store`?

chrzaszcz (commented):
Looks good in general, I have two minor remarks.

src/mam/mod_mam_utils.erl
NelsonVides: Because store means always store, regardless of everything else, so I thought of saving the computation to check all the other stuff and simply say yes to that one.

big_tests/tests/inbox_SUITE.erl
NelsonVides: This is not related to the hints, but to the fact that carbons aren't stored in inbox, the same way they aren't stored in MAM, else we would have a terrible pollution.

NelsonVides: fixed

src/mam/mod_mam_utils.erl
NelsonVides: fixed

chrzaszcz (approved):
Nice, it's good to see the things unified.

PR #3166, opened 2021-07-01
Use the new official MSSQL docker image

chrzaszcz:
See https://hub.docker.com/_/microsoft-mssql-server


NelsonVides (approved):
Great fix, thanks! 🎉

PR #3224, opened 2021-08-18
Restart db service once domain core is down

arcusfelis:
This PR addresses MIM-1470

Proposed changes include:
- Restart db service once domain core is down
- Add logging for initial sync
-  Add test demonstrating that Core process is restarted, but domain service is not

PR #3226, opened 2021-08-20
Reset last_event_id once mongoose_domain_core starts

arcusfelis:
This tells to the service, that it needs to read the complete database.
But it assumes that the service process is down, so it would not
overwrite the last_event_id we set here.

To ensure that the service is down if the core is down,
the core and service processes should be supervised with rest_for_one strategy.

Which should be done in a separate PR. Though, it does not required to fix the "join cluster" issue.

This PR addresses MIM-1470

### DIFF
--- a/big_tests/tests/service_domain_db_SUITE.erl
+++ b/big_tests/tests/service_domain_db_SUITE.erl
@@ -63,6 +63,7 @@ db_cases() -> [
      db_out_of_sync_restarts_service,
      db_crash_on_initial_load_restarts_service,
      db_restarts_properly,
+     db_keeps_syncing_after_cluster_join,
      cli_can_insert_domain,
      cli_can_disable_domain,
      cli_can_enable_domain,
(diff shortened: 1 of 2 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/68058) / small_tests / d0880a6cb6d3cca18f30cd2610eac9e342bbdf1d
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3226/68058/small_tests.24.0.4) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3226/68058/small_tests.24.0.4/small/ct_run.mongooseim@localhost.2021-08-20_09.48.54/lib.mongooseim.logs/run.2021-08-20_09.48.54/suite.log.html)

NelsonVides:
Can you explain how this proposal solves my question at https://github.com/esl/MongooseIM/pull/3224#discussion_r692822530? I'm not sure I understand how the new testcase deals with the issue I had.

NelsonVides (approved):
Ok, after internal discussion this seems to be the right answer, this will get QA-ed soon enough anyway.

PR #3245, opened 2021-09-06
Update Swagger dynamic domains doc

gustawlippa:
This PR adds tags to the dynamic domains Swagger file, so that the operations are grouped under a "Dynamic domains" header, rather than a "default" one.


vkatsuba (approved):
LGTM.

PR #3191, opened 2021-07-21
Mam/iq fin transformations

NelsonVides:
Sometimes it can be useful to be able to add information to a mam query, that doesn't belong to any specific message but to all of them. So being able to add subelements to the iq-fin element will solve the extensibility issue.

PR #3329, opened 2021-10-08
Document low-level MAM options

gustawlippa:
This PR addresses https://github.com/esl/MongooseIM/pull/3191#pullrequestreview-712659773.
It adds documentation for low-level MAM options. Although some are for example RDBMS only, I think it's better to keep them all in their own group, as these are probably really rarely used.

`async_writer_rdbms_pool` is not documented - if I am not mistaken this option is not read by anything, but I left it for now, because I'm not sure.
`simple` seems weird and from what I understand it doubles what `db_message_format` already does. It may be useful maybe for setups with multiple backends with different formats? If not, the code for `mod_mam_cassandra_arch` and `mod_mam_muc_cassandra_arch` could be simplified.

The options and their structure in the TOML config could use a refactor in my opinion, but that may be done in the future.

Some other minor changes like typos fixed (there is no `"mod_mam_rdbms_async_writer"`).

### DIFF
--- a/doc/configuration/database-backends-configuration.md
+++ b/doc/configuration/database-backends-configuration.md
@@ -91,7 +91,7 @@ Please refer to the [RDBMS options](outgoing-connections.md#rdbms-options) for m
 **Version notice**
 
 The required minimum version of MySQL is `5.7.9`.
-This is mainly to benefit from the JSON data type and the default settings which works out of the box with MongooseIM.
+This is mainly to benefit from the JSON data type, and the default settings which work out of the box with MongooseIM.
 
 #### MySQL 8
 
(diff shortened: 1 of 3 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/78741) / small_tests / f46b97355fbe9aabeff3f6f6fa23e45ffd70522b
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3329/78741/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3329/78741/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2021-10-12_08.52.48/lib.mongooseim.logs/run.2021-10-12_08.52.48/suite.log.html)

PR #3331, opened 2021-10-11
Hook funs

NelsonVides:
It's a bit pointless to be passing to gen_hook a function reference, even as to make gen_hook throw an error and prevent altogether the registration of a hook if the given fun is not strictly external, only to then extract the module and function atoms and use them to call the function through an `erlang:apply/3` proxied through another module. Instead we can store the function reference, have the hook record that will be stored in the ets table be a word smaller (one field less in the tuple), and use the ref to call the function, which has a theoretical performance advantage.
In line with this, get rid of the `safely` module, that is being barely used anywhere.

### DIFF
--- a/src/gen_hook.erl
+++ b/src/gen_hook.erl
@@ -52,11 +52,8 @@
 
 -export_type([hook_fn/0, hook_list/0]).
 
--record(hook_handler, {%% 'prio' field must be the first
-                       %% this is required for the proper sorting.
-                       prio :: pos_integer(),
-                       module :: module(),
-                       function :: atom(), %% function name
+-record(hook_handler, {prio :: pos_integer(),
+                       hook_fn :: hook_fn(),
                        extra :: map()}).
 
 -define(TABLE, ?MODULE).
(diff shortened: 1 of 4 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/78502) / small_tests / e9bc6bc4deb9bcba9c8a781c8b805a6a9891fa56
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3331/78502/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3331/78502/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2021-10-11_07.18.33/lib.mongooseim.logs/run.2021-10-11_07.18.33/suite.log.html)

DenysGonchar:
> So after discussing this with @DenysGonchar, action points are:
> 
>     * `safely:apply` returns, but with better distinguishing of error types, currently it returns the stacktrace only for the `error` case, and makes all exception classes an `'EXIT'` tagged tuple, which is misleading, that tag is used for real exits elsewhere in Erlang/OTP.
> 
>     * `mongoose_domain_core` should at least log errors when doing `safely:apply` and not just ignore all return values.
> 
>     * hook removal needs to do manual comparison of function objects, by running `fun_info` on the comparison strategy, ensuring handlers comparison will be systematic.

Thanks for putting here the summary of the discussion.

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/78793) / small_tests / 60502a357ac553e4d7c48b06be06beb3749e336f
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3331/78793/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3331/78793/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2021-10-12_10.31.07/lib.mongooseim.logs/run.2021-10-12_10.31.08/suite.log.html)

PR #3334, opened 2021-10-13
Fix missing "/" for make_request

imcyee:
Currently, `FullPath` is missing '/' in MongooseIM/src/mongoose_http_client.erl, because fix_path has removed '\'.
A prefix like `localhost:5005` and a path of `/notification` will be `localhost:5005notification` instead of `localhost:5005/notification`
Here are the flow:
https://github.com/esl/MongooseIM/blob/0497dd8d632baa357af2b4655c7afcd32ca768d7/src/event_pusher/mod_event_pusher_http.erl#L97-L98

request path is stripped. 
https://github.com/esl/MongooseIM/blob/0497dd8d632baa357af2b4655c7afcd32ca768d7/src/event_pusher/mod_event_pusher_http.erl#L138-L139

but it is missing the '/' at
https://github.com/esl/MongooseIM/blob/0497dd8d632baa357af2b4655c7afcd32ca768d7/src/mongoose_http_client.erl#L53-L54

Do we need to remove the strip_path or add '/' to make_request like:
`FullPath = <<PathPrefix/binary, "/", Path/binary>>,`

### DIFF
--- a/src/mongoose_http_client.erl
+++ b/src/mongoose_http_client.erl
@@ -51,7 +51,8 @@ make_request(Host, PoolTag, Path, Method, Headers, Query) ->
     end.
 
 make_request(Host, PoolTag, PathPrefix, RequestTimeout, Path, Method, Headers, Query) ->
-    FullPath = <<PathPrefix/binary, "/", Path/binary>>,
+    FixedPath = fix_path(Path),
+    FullPath = <<PathPrefix/binary, FixedPath/binary>>,
     Req = {request, FullPath, Method, Headers, Query, 2, RequestTimeout},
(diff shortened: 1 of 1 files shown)


imcyee:
@vkatsuba  will get back to this later. Currently hands are tied.

gustawlippa:
Maybe the code changed in the meantime, but the issue seems to be resolved. You shouldn't put the host inside the prefix parameter - please see https://esl.github.io/MongooseDocs/latest/configuration/outgoing-connections/#http-options for examples. The `"/"` prefix _after_ the domain part (possibly with a port) will be added by default.

PR #3345, opened 2021-10-19
Add mongoose_backend:get_backend_name

gustawlippa:
This is analogous to the function exposed in the dynamically compiled backends. It is used in some places in MIM, and while the "backend_name" (the last part of the backend module name, like `pgsql`, `mnesia` etc.) can be extracted via string manipulation, it seems cleaner to just reintroduce this function.

This change goes together with an update in `mongoose_helper` for the big tests. It is used to check if the suites finished dirty or not. After all backends will have been refactored, we can get rid of the `*_old` functions of course.

### DIFF
--- a/src/mongoose_backend.erl
+++ b/src/mongoose_backend.erl
@@ -6,10 +6,13 @@
          call_tracked/4,
          get_backend_name/2]).
 
+%% remove after mongoose_rdbms is refactored not to use dynamically compiled backend
+-ignore_xref([get_backend_name/2]).
+
 %% Legacy call from backend_module
 -export([ensure_backend_metrics/2]).
 
-%% For debugging
+%% For debugging and tests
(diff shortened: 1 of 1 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/80702) / small_tests / 01fb11a7ca8a5d78d3905b82e5e92aed35ee8723
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3345/80702/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3345/80702/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2021-10-20_08.57.05/lib.mongooseim.logs/run.2021-10-20_08.57.06/suite.log.html)

PR #3340, opened 2021-10-14
Add `mod_event_pusher_push_backend`

gustawlippa:
This PR removes dynamically compiled `mod_event_pusher_push` backend.

PR #3347, opened 2021-10-20
Migrate mod_privacy to mongoose_backend

NelsonVides:


gustawlippa:
There were two PRs opened, #3340 ran CI, passed, another PR (#3339) made changes to the API, the tests for it passed and it was merged first. Then the first one was merged with incorrect call but with green CI. A fix is merged already to the feature branch, sorry for the confusion.

PR #3348, opened 2021-10-20
Fix function call

NelsonVides:
This function with arity 3 was not actually defined :)

### DIFF
--- a/src/event_pusher/mod_event_pusher_push_backend.erl
+++ b/src/event_pusher/mod_event_pusher_push_backend.erl
@@ -40,7 +40,7 @@
 -spec init(Host :: jid:server(), Opts :: list()) -> ok.
 init(Host, Opts) ->
     TrackedFuns = [enable, disable, get_publish_services],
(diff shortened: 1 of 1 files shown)


vkatsuba:
@gustawlippa before merge something, need wait for finishing CI... even if these are obvious changes ...

NelsonVides:
> @gustawlippa before merge something, need wait for finishing CI... even if these are obvious changes ...

By fault as well, I suggested to merge promptly. We just waited for xref and dialyzer to be green.

vkatsuba:
> By fault as well, I suggested to merge promptly. We just waited for xref and dialyzer to be green.

Clear.

vkatsuba:
> feature branch with incorrect code and search for bugs in their changes

If developer didn't change code from where provided error in `xref` - he don't need find any issue in his code.

> Sure we would not do that normally, but keep in mind this is all WIP.

Following this logic - it no make sense also to create PR's at all then for current branch till time when this branch will finished and ready for PR, but this is wrong approach.

PR #3379, opened 2021-10-29
Use persistent terms to store backend_module in ejabberd_sm

arcusfelis:
Proposed changes include:
* Also fix acl:match_rule call to use hosttypes
* Because we support only one global backend, we don't need to pass HostTypes (that could be done later, if needed, but would have to have tests and rewriting API a bit)

### DIFF
--- a/big_tests/tests/distributed_helper.erl
+++ b/big_tests/tests/distributed_helper.erl
@@ -14,7 +14,7 @@
 -export_type([rpc_spec/0]).
 
 is_sm_distributed() ->
-    Backend = rpc(mim(), ejabberd_sm_backend, backend, []),
+    Backend = rpc(mim(), ejabberd_sm, sm_backend, []),
     is_sm_backend_distributed(Backend).
 
 is_sm_backend_distributed(ejabberd_sm_mnesia) -> true;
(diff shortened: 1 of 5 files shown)


chrzaszcz:
For consistency, could we make `mongoose_backend` support such non-host-type (global) backends? Such pattern will occur a few more times and I think that it would be good to use one backend management module for this instead of creating persistent terms directly, which is a bit ad-hoc.

arcusfelis:
@chrzaszcz Yea, possible.  What's about call_tracked/call API? ejabberd_gen_sm works kinda differently, it wants us to provide a backend module name. While mongoose_backend wants us to provide a HostType, and it gets a backend name internally.

Also, some operations, that would require several calls to the backend modules would be a little bit slower, because we currently get a backend module name once and use it in several functions (basically see below):

```erlang
open_session(HostType, SID, JID, Priority, Info) ->
    BackendModule = sm_backend(),
    set_session(SID, JID, Priority, Info),
    ReplacedPIDs = check_for_sessions_to_replace(HostType, BackendModule, JID),
    mongoose_hooks:sm_register_connection_hook(HostType, SID, JID, Info),
    ReplacedPIDs.
```

We could also add HostType support into SM API. It asks for it :D

PR #3386, opened 2021-11-04
MIM-1539 Allow global to be passed into mongoose_backend as an arg

arcusfelis:
This PR addresses MIM-1539

Proposed changes include:
* Removed behaviour_info, idk why it is commented out.
* Renamed init_per_host_type (so many files to change).

### DIFF
--- a/src/ejabberd_gen_sm.erl
+++ b/src/ejabberd_gen_sm.erl
@@ -1,3 +1,6 @@
+%% Generic module to access SM backend modules
+%% Pass HostType to the functions as an argument, if you can
+%% Pass global, when you cannot
 -module(ejabberd_gen_sm).
 
 -callback start(list()) ->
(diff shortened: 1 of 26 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/84908) / small_tests / 6c47fe4931ec515793d91d9442ce3a4feb335c3f
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3386/84908/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3386/84908/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2021-11-04_14.53.56/lib.mongooseim.logs/run.2021-11-04_14.53.57/suite.log.html)

src/ejabberd_gen_sm.erl
```diff
@@ -32,78 +35,76 @@
...
          total_count/1, unique_count/1]).
 
--ignore_xref([behaviour_info/1, cleanup/2]).
+-ignore_xref([cleanup/2, behaviour_info/1]).
 
--spec start(module(), list()) -> any().
-start(Mod, Opts) ->
-    Mod:start(Opts).
+-define(MAIN_MODULE, ejabberd_sm).
+-type host_type() :: mongooseim:host_type() | global.
```
NelsonVides, line 41: This is definition is repeated with many names in many many places, like gen_hook, acl, metrics... Shouldn't we by now just make it a global type defined in `mongooseim.erl` together with the original host_type(), like
```erl
-type host_type_or_global() :: mongooseim:host_type() | global.
```
And just use it everywhere?
No need for this PR to retroactively modify all places where this is already used, that would be a too big diff for this PR, but the definition can already be created and we'll update everything else later.

src/mongoose_backend.erl
```diff
@@ -20,12 +20,13 @@
 -type function_name() :: atom().
 -type main_module() :: module().
 -type backend_module() :: module().
+-type host_type_or_global() :: host_type_or_global() | global.
```
NelsonVides, line 23: Like here, it is again repeated here.

src/mongoose_backend.erl
```diff
@@ -20,12 +20,13 @@
...
 
--spec init_per_host_type(HostType :: mongooseim:host_type(),
+-spec init(HostType :: host_type_or_global(),
                          MainModule :: main_module(),
                          TrackedFuns :: [function_name()],
                          Opts :: gen_mod:module_opts()) -> ok.
```
NelsonVides, line 28: Indentation.

src/mongoose_backend.erl
```diff
@@ -20,12 +20,13 @@
 -type function_name() :: atom().
 -type main_module() :: module().
 -type backend_module() :: module().
+-type host_type_or_global() :: host_type_or_global() | global.
 
--spec init_per_host_type(HostType :: mongooseim:host_type(),
+-spec init(HostType :: host_type_or_global(),
```
NelsonVides, line 25: Also, the name of the variable is by now misleading, this is not a host_type any more, so just like it is done for `gen_hook`, all these `HostType` vars can be renamed to `Tag` or something like that.

src/ejabberd_sm.erl
```diff
@@ -1074,9 +1068,4 @@ get_cached_unique_count() ->
 
 -spec sm_backend() -> backend().
 sm_backend() ->
```
gustawlippa, line 1070: With the new changes, is this function even used? If it's not, I don't see a point keeping it, as HostTypes (or, until it's changed, `global`) should be enough to determine which backend is used.

NelsonVides (changes requested):
Honestly I find this a mess, what are we trying to do? Where is the place where we acknowledge this is not host_type ready and it only accepts the `global` key? I think that that should be well defined somewhere in a unique place. Right now some places in ejabberd_sm are ignoring the `HostType` vars and passing `global`and then again that is done here. Then this is also duplicating knowledge of which one is the backend both in `ejabberd_sm` and in `ejabberd_gen_sm`, that should also happen in only one place.

src/ejabberd_gen_sm.erl
gustawlippa, line 41: I'd call it `host_type_or_global` like in `mongoose_backend` so that it's less surprising, but this may be a matter of taste.

gustawlippa: Maybe not in this PR, I'd personally go with just using `mongooseim:host_type() | global` in type specs, but I think just renaming this `host_type()` to `host_type_or_global()` would be fine for now.

gustawlippa:
To answer @NelsonVides , what we are trying to do, is to allow `mongoose_backend` to use `global` and not just `HostType`s, as there are some modules which need that. The first module, that could use this option is `ejabberd_sm`, and it is changed here to use it. It was noted in a previous PR, that it should not keep the backend module in some ad-hoc created persistent term and follow suit with other MIM modules and use `mongoose_backend`.
I also don't like, that `ejabberd_gen_sm` discards `HostType`s, but I'm not sure if other comments should come into the scope of this PR?

NelsonVides:
> To answer @NelsonVides , what we are trying to do, is to allow `mongoose_backend` to use `global` and not just `HostType`s, as there are some modules which need that. The first module, that could use this option is `ejabberd_sm`, and it is changed here to use it. It was noted in a previous PR, that it should not keep the backend module in some ad-hoc created persistent term and follow suit with other MIM modules and use `mongoose_backend`. I also don't like, that `ejabberd_gen_sm` discards `HostType`s, but I'm not sure if other comments should come into the scope of this PR?

Yeah, I know, I'm aware of all of that excatly, the thing is that introducing `host_type_or_global()` to `mongoose_backend` is well done here, except that that type definition could be placed in `mongooseim.erl` instead of here (as it is used in many other places, we can reuse later), but how's that's all used by ejabberd_sm seems very messy to me, where both `ejabberd_sm` and `ejabberd_gen_sm` know about the fact that this is all a global, and mix actual `HostType`s with `global`s everywhere.

Also, `ejabberd_gen_sm` is the same proxy-module we establish in many other places as being named with the `_backend` suffix. So I think it wouldn't hurt to actually rename the module to `ejabberd_sm_backend` and follow exactly the same code patterns we follow everywhere else. It's boring, but very predictable, if we just follow so everywhere.

src/ejabberd_sm.erl
arcusfelis: it is used in tests

arcusfelis:
Not renaming HostType argument in mongoose_backend.
It is done similar in ACL and in Wpool.
Just look at spec.
Could be a Tag, but you would have to do it consistently in this case.

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/85061) / small_tests / 34fc43bc4c974dedafc1b28d707d9880b80e28e3
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3386/85061/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3386/85061/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2021-11-05_10.24.53/lib.mongooseim.logs/run.2021-11-05_10.24.53/suite.log.html)

src/ejabberd_sm.erl
```diff
@@ -499,18 +496,13 @@ init([]) ->
                           undefined -> {mnesia, []};
                           Value -> Value
                       end,
-    store_backend(Backend),
-
+    mongoose_backend:init(global, ?MODULE, [], [{backend, Backend}]),
```
NelsonVides, line 499: Just to replicate what all other modules do, here we could be doing
```suggestion
    ejabberd_sm_backend:start(Opts),
```
And `ejabberd_sm_backend` can do all the `mongoose_backend:init` together with the `ejabberd_sm_mnesia:start` callback and so on.

NelsonVides (commented):
Now this seems more organized, and it's clear that this is simply a global module, and if we want to make it per-host-type in the future, all the necessary changes are clear and straightforward (albeit long and mechanical, yeah, big diff, but very predictable and organised diff).

Just have one more comment 👌🏽

NelsonVides (approved):
Excellent for me, thanks for all the applied requests :)

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/85154) / small_tests / 9299add6cb0636d99273b30a7ecadf8d05332231
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3386/85154/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3386/85154/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2021-11-05_14.59.30/lib.mongooseim.logs/run.2021-11-05_14.59.31/suite.log.html)

chrzaszcz (approved):
Looks good!

PR #3528, opened 2022-02-01
Add config flag to be able to set sni hostname matching

NelsonVides:
As it turns out, the ssl:connect/4 default fails to match against websites that offer wildcard certificates, because apparently Erlang's SSL team finds them less secure: https://www.erlang.org/doc/man/public_key.html#pkix_verify_hostname_match_fun-1
> Currently supported https fun will allow wildcard certificate matching as specified by the HTTP standard. Note that for instance LDAP have a different set of wildcard matching rules. **If you do not want to allow wildcard certificates (recommended from a security perspective)** or otherwise customize the hostname match the default match function used by ssl application will be sufficient.

But in reality many web deployments do use wildcard certificates, so it is indeed recommended to allow for this in the configuration: https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/ssl.html

See below a screenshot from my local machine. If providing the SNI hostname as an FQDN against a server with wildcard certificates, the SSL verification will fail, so it will require setting the SNI hostname as a wildcard as well, which is not what the usual deployment would expect from the [specification point of view](https://datatracker.ietf.org/doc/html/rfc6125#section-6), and I've also been told a myriad of other languages expect as well. So instead, we'll need to pass that configuration parameter to the underlying SSL protocol.

Screenshot 2022-02-01 at 14 42 35

See also how setting this configuration does indeed prepares fusco's state with the right configuration:
image
Screenshot 2022-02-01 at 14 51 00


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/104163) / small_tests / 4b08f9569a0d44fa14c2bce5375421b1360a283e
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3528/104163/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3528/104163/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2022-02-01_13.55.33/lib.mongooseim.logs/run.2022-02-01_13.55.34/suite.log.html)

doc/configuration/outgoing-connections.md
```diff
@@ -446,3 +446,10 @@ Enables SNI extension to TLS protocol. If set to `true`, the `server_name_indica
 * **Example:** `tls.server_name_indication_host = "domain.com"`
 
 Domain against which the certificates will be checked, using SNI. It can be specified only when `server_name_indication` is set to `true`.
+
+### `outgoing_pools.*.*.connection.tls.server_name_indication_protocol`
+* **Syntax:** string, one of "default" or "https"
+* **Default:** not set
```
gustawlippa, line 452: I hate how a literal "default" option is not set by default. And from what I see setting it to "default" is the same as not setting it, so maybe skip the "default" possibility altogether?

NelsonVides: More or less fixed, and also fixed testing coverage.

arcusfelis (approved):
fine

gustawlippa (approved):
Looks good! I'm surprised this is not the default `ssl` behaviour.

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/104237) / small_tests / 322eb82a7135adb10d3e0f3d951a7b3474ea8a68
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3528/104237/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3528/104237/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2022-02-01_15.00.48/lib.mongooseim.logs/run.2022-02-01_15.00.49/suite.log.html)

PR #3536, opened 2022-02-09
Use circleci convenience images

NelsonVides:
Apparently there's new circleci images for postgres and mysql, that as usual they promise to always have cached in all their nodes so it "should all be faster and more stable".


mongoose-im:
[mysql_redis_24](https://circleci.com/gh/esl/MongooseIM/105227) / mysql_redis / bea050f62cdb6ae79e53487ca1dc25f1207c470d
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3536/105227/mysql_redis.24.0.5)

PR #3560, opened 2022-02-23
Don't use mnesia for IQ response tracking

arcusfelis:
The code I am touching here is a bit old and a bit hairy.
I honestly didn't read that one and it took sometime to understand why we use mnesia there to begin with.

So, TLDR:
- the code I am touching allows to route an IQ from the SERVER to the CLIENT (the opposite way they usually go).
- it is used in mod_ping.
- though, the last thing I want is to **replicate** data, that is used by mod_ping in a big cluster (because of how often it is written).
- in the real life, that iq_response would never do RPC calls (at least for mod_ping, because the process sending IQ and receiving IQ is a c2s process). We still need to test it though :(

Proposed changes include:
- This change removes replication for IQ response tracking (useful in the big clusters, that use mod_ping)
- We use IQ IDs to locate the metadata for the IQ instead

### DIFF
--- a/src/ejabberd_local.erl
+++ b/src/ejabberd_local.erl
@@ -138,27 +138,16 @@ process_iq_reply(From, To, Acc, #iq{id = ID} = IQ) ->
 -spec get_iq_callback_in_cluster(id(), mongoose_acc:t()) ->
         {ok, callback()} | {error, term()}.
 get_iq_callback_in_cluster(ID, Acc) ->
-    {ok, NodeId} = ejabberd_node_id:node_id(),
-    BinNodeId = integer_to_binary(NodeId),
-    case binary:split(ID, <<"_">>, []) of
-        [BinNodeId, _Rest] ->
+    %% We store information from which node the request is originating in the ID
+    case parse_iq_id(ID) of
+        local_node ->
             get_iq_callback(ID);
-        [OtherBinNodeId, _Rest] ->
-            OtherNodeId = binary_to_integer(OtherBinNodeId),
-            case ejabberd_node_id:node_id_to_name(OtherNodeId) of
-                {ok, Name} ->
-                    rpc:call(Name, ?MODULE, get_iq_callback, [ID]);
-                {error, Reason} ->
-                    ?LOG_ERROR(#{what => process_iq_reply_failed,
-                                 text => <<"Unknown node_id">>,
-                                 reason => Reason, acc => Acc}),
-                    {error, unknown_node_id}
-            end;
-        _ ->
-            ?LOG_ERROR(#{what => process_iq_reply_failed,
-                         text => <<"Bad IQ ID format">>,
-                         acc => Acc}),
-            {error, bad_iq_format}
+        {remote_node, NodeName} ->
+            rpc:call(NodeName, ?MODULE, get_iq_callback, [ID]);
+        {error, Reason} ->
+            ?LOG_ERROR(#{what => parse_iq_id_failed,
+                         reason => Reason, acc => Acc}),
+            {error, Reason}
     end.
 
 -spec process_packet(Acc :: mongoose_acc:t(),
@@ -204,12 +193,6 @@ route_iq(From, To, Acc, #iq{type = Type} = IQ, Callback, Timeout)
                      jlib:iq_to_xml(IQ)
              end,
     ejabberd_router:route(From, To, Acc, Packet).
-
-make_iq_id() ->
-    %% Attach NodeId, so we know to which node to forward the response
-    {ok, NodeId} = ejabberd_node_id:node_id(),
-    Rand = mongoose_bin:gen_from_crypto(),
-    <<(integer_to_binary(NodeId))/binary, "_", Rand/binary>>.
 
 -spec register_iq_response_handler(
            ID :: id(),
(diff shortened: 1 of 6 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/109168) / small_tests / 51a1be39e919b3ce2d57d849292df5b93d55077a
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3560/109168/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3560/109168/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2022-02-23_20.13.01/lib.mongooseim.logs/run.2022-02-23_20.13.02/suite.log.html)

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/109189) / small_tests / 82809ecf9fb4ef7567fed1d39db49f05d6dc5cba
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3560/109189/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3560/109189/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2022-02-23_20.34.29/lib.mongooseim.logs/run.2022-02-23_20.34.29/suite.log.html)

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/112750) / small_tests / 962079a1e1b4a6c6578bdd4b2becc0ededee0f7f
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3560/112750/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3560/112750/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2022-03-07_19.23.46/lib.mongooseim.logs/run.2022-03-07_19.23.47/suite.log.html)

src/ejabberd_local.erl
```diff
@@ -127,18 +125,41 @@ process_iq(_, Acc, From, To, El) ->
...
         _ ->
             Acc
     end.
 
+-spec get_iq_callback_in_cluster(id(), mongoose_acc:t()) ->
+        {ok, callback()} | {error, term()}.
+get_iq_callback_in_cluster(ID, Acc) ->
+    {ok, NodeId} = ejabberd_node_id:node_id(),
+    BinNodeId = integer_to_binary(NodeId),
+    case binary:split(ID, <<"_">>, []) of
```
NelsonVides, line 143: I'd make this some kind of `parse_iq_id` next to `make_iq_id`, so that knowledge about the structured of this ID is encapsulated in two related and close-to-each-other functions, makes it a bit less magic.

But this still feels like a hack, so I'd say hacks are welcome but only if we really need one, still thinking about this one 🤔

NelsonVides:
Yeah, I like the idea of one less mnesia here, I didn't even know ping was replicating all its handlers across all nodes, that seems like a waaaaaaste.

But I'd also say, if we don't have any module doing rpcs nor using this cluster logic, we can throw away that code to begin with. It feels a bit like a hack to have those prefixes in the ID and to use rpcs, and also, I imagine `ejabberd_local` to do local stuff, non-clustered logic, I'd make ejabberd_sm a bit fatter for that instead. When we need distributed iqs, we'll implement it, until then, less hacky code to maintain.

src/ejabberd_local.erl
arcusfelis: Could do `parse_iq_id` for sure.

yea, it is also used in pusher push to route a pubsub requests to a remote servers. D:
Which means that response could arrive from any s2s connection (meh).

So, the issue here is too generic API, which is now just writes data to other nodes which is not used :)

mod_caps/mod_ping - just local routes.
pusher_push - remote calls (s2s is possible).


BTW, ejabberd_local means jids like `servername` (without `user@` part).

That pusher_push code, will try to read it one more time:
```erlang
    ResponseHandler =
    fun(_From, _To, FAcc, Response) ->
            mod_event_pusher_push:cast(Host, fun handle_publish_response/5,
                                       [Host, To, PubsubJID, Node, Response]),
            FAcc
    end,
    %% The IQ is routed from the recipient's server JID to pubsub JID
    %% This is recommended in the XEP and also helps process replies to this IQ
    NotificationFrom = jid:make(<<>>, Host, <<>>),
    mod_event_pusher_push:cast(Host, fun ejabberd_local:route_iq/5,
                               [NotificationFrom, PubsubJID, Acc, Stanza, ResponseHandler]).
```

mongoose-im:
[riak_mnesia_24](https://circleci.com/gh/esl/MongooseIM/116014) / riak_mnesia / f476beb1f58020ba77b16886605fdd03d8c65a17
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3560/116014/riak_mnesia.24.0.5)

src/ejabberd_local.erl
```diff
@@ -411,57 +388,26 @@ directed_to(<<>>, _) ->
...
 get_iq_callback(ID) ->
-    case mnesia:dirty_read(iq_response, ID) of
-        [#iq_response{module = Module, timer = TRef,
-                      function = Function}] ->
-            cancel_timer(TRef),
-            mnesia:dirty_delete(iq_response, ID),
-            {ok, Module, Function};
+    case ets:lookup(?IQRESPONSE, ID) of
+        [{ID, Callback, TRef}] ->
+            erlang:cancel_timer(TRef),
```
NelsonVides, line 395: This makes me think, the ping IQs timeouts are handled twice here... c2s also registers its own. Well This PR solves _not using mnesia_, which is a huge improvement, and everything else actually doesn't change, so it seems fine. But hell, there's an important tech debt waiting here. Let's keep the PRs small and work on that on the next one.

NelsonVides (approved):
One less mnesia! 💪🏽

PR #3601, opened 2022-03-21
Fix error in cleanup of registration_timeout big test

amccormick93:
Trivial fix. The cleanup stage of the registration_timeout test was hitting an error because it tried to delete Alice and Bob, but user Bob was never created by the test. This is because the test was creating Alice twice, instead of creating Bob. 

The test now does not hit an error in cleanup.


chrzaszcz (approved):
Looks good

PR #3611, opened 2022-03-26
Async rdbms order

NelsonVides:
If we have asynchronous inserts for inbox or smart_markers, there's a risk of different nodes aggregating different data, and then inserting it in undefined orders, so an earlier event could be flushed by one node _after_ a latter event from another node. For example in the case of a conversation between Alice and Bob, when Alice and Bob are connected to different nodes on a cluster.

So what we want here, is for the DB to ignore updates, if the timestamp of such update is not strictly higher than the timestamp from the previous update.

For this, we add one more parameter to the preparing of queries, that constructs the appropriate query for Postgres and MySQL. This is currently not implemented for MSSQL.

With this, now the asynchronous backends for inbox and markers are much closer to being production ready, the only thing missing is the remove events, which don't have a timestamp so they won't be reordered correctly with updates. I plan to make removals tagged updates instead, like in the case of inbox, moving them to a hidden box instead.

Note the commit message for f8876363762640c7a0b78cce8ff42c735df49fc5 for how the trick works for MySQL.

### DIFF
--- a/big_tests/tests/rdbms_SUITE.erl
+++ b/big_tests/tests/rdbms_SUITE.erl
@@ -67,6 +67,7 @@ rdbms_queries_cases() ->
      insert_batch_with_null_case,
      test_cast_insert,
      test_request_insert,
+     test_incremental_upsert,
      arguments_from_two_tables].
 
 suite() ->
@@ -85,9 +86,15 @@ init_per_suite(Config) ->
 end_per_suite(Config) ->
     escalus:end_per_suite(Config).
 
+init_per_testcase(test_incremental_upsert, Config) ->
+    sql_query(Config, <<"TRUNCATE TABLE inbox">>),
+    escalus:init_per_testcase(test_incremental_upsert, Config);
 init_per_testcase(CaseName, Config) ->
     escalus:init_per_testcase(CaseName, Config).
 
+end_per_testcase(test_incremental_upsert, Config) ->
+    sql_query(Config, <<"TRUNCATE TABLE inbox">>),
+    escalus:end_per_testcase(test_incremental_upsert, Config);
 end_per_testcase(CaseName, Config) ->
     escalus:end_per_testcase(CaseName, Config).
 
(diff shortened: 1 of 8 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/117019) / small_tests / b47f53b1471098013594f73b44770bfebc611397
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3611/117019/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3611/117019/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2022-03-26_11.35.25/lib.mongooseim.logs/run.2022-03-26_11.35.25/suite.log.html)

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/117066) / small_tests / 32a7f94ac0556293c7cf955752e193fc4dcc9a1a
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3611/117066/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3611/117066/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2022-03-26_17.47.24/lib.mongooseim.logs/run.2022-03-26_17.47.24/suite.log.html)

chrzaszcz (commented):
The code looks good, I only have some general notes:
- How do we check that it indeed works as intended? Maybe add a test in `rdbms_SUITE`?
- Of course it would be good to add it for MS SQL. Do you know if this is even possible?

NelsonVides:
>     * How do we check that it indeed works as intended? Maybe add a test in `rdbms_SUITE`?
Test added 👌🏽
>     * Of course it would be good to add it for MS SQL. Do you know if this is even possible?
Probably, but honestly I have no idea how, and I'd like to unblock inbox now. Perhaps we can leave the mssql case as a ticket for the future? Perhaps to when a client actually requests it?
>     Just a note: I think we should make this a map of arguments, especially that some of them are optional.
May be, but there's plenty of modules using this function and I didn't want to introduce a big diff (albein mechanical) in this PR, dunno, but could do 🤔

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/117749) / small_tests / 7db782e1281f18434e7272b30ee3212aa27b43ef
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3611/117749/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3611/117749/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2022-03-30_11.08.36/lib.mongooseim.logs/run.2022-03-30_11.08.36/suite.log.html)

chrzaszcz:
> > ```
> > * How do we check that it indeed works as intended? Maybe add a test in `rdbms_SUITE`?
> > ```
> 
> Test added 👌🏽

Nice!

> 
> > ```
> > * Of course it would be good to add it for MS SQL. Do you know if this is even possible?
> > ```
> 
> Probably, but honestly I have no idea how, and I'd like to unblock inbox now. Perhaps we can leave the mssql case as a ticket for the future? Perhaps to when a client actually requests it?
> 
Ok, let's do it like this

> > ```
> > Just a note: I think we should make this a map of arguments, especially that some of them are optional.
> > ```
> 
> May be, but there's plenty of modules using this function and I didn't want to introduce a big diff (albein mechanical) in this PR, dunno, but could do 🤔
Yes, let's do it separately.

PR #3630, opened 2022-04-13
Inbox flaky tests fixes

NelsonVides:
The command as it was currently implemented would not move a
conversation to the bin when aggregating it onto the set_inbox commands,
as the sql query was so that it would only unarchive entries on request,
but the entry was not on the archive, so the box was not moved.

This happens because we enforce automatic moves from the archive to the
inbox, and no other box is to be moved automatically.


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/119458) / small_tests / e8e43854faa4f9206152776094fa0ba676072d94
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3630/119458/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3630/119458/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2022-04-13_13.31.56/lib.mongooseim.logs/run.2022-04-13_13.31.57/suite.log.html)

big_tests/tests/inbox_extensions_SUITE.erl
```diff
@@ -312,6 +313,23 @@ box_active_entry_gets_archived(Config) ->
...
+    escalus:fresh_story(Config, [{alice, 1}, {bob, 1}], fun(Alice, Bob) ->
+        % Alice sends a message to Bob and then she archives the conversation
+        Body1 = <<"Hi Bob">>,
+        inbox_helper:send_msg(Alice, Bob, Body1),
+        inbox_helper:check_inbox(Alice, [#conv{unread = 0, from = Alice, to = Bob, content = Body1}]),
+        set_inbox_properties(Alice, Bob, [{box, other}]),
+        % But then Alice keeps writing
+        Body2 = <<"Hi Bob again">>,
+        inbox_helper:send_msg(Alice, Bob, Body2),
+        inbox_helper:get_inbox(Alice, #{box => all, hidden_read => false }, #{count => 1}),
```
chrzaszcz, line 326: Minor: the spacing looks a bit awkward to me with the extra space after `false`

big_tests/tests/inbox_extensions_SUITE.erl
```diff
@@ -312,6 +313,23 @@ box_active_entry_gets_archived(Config) ->
...
+        % Then the conversation is still in the same box
+        inbox_helper:check_inbox(Alice, [], #{box => archive}),
```
chrzaszcz, line 328: Just to confirm, sholdn't this check for `inbox` instead of `archive`?

big_tests/tests/inbox_extensions_SUITE.erl
```diff
@@ -312,6 +313,23 @@ box_active_entry_gets_archived(Config) ->
                                  #{box => archive})
     end).
 
+box_other_entry_does_not_get_unarchived(Config) ->
+    escalus:fresh_story(Config, [{alice, 1}, {bob, 1}], fun(Alice, Bob) ->
+        % Alice sends a message to Bob and then she archives the conversation
```
chrzaszcz, line 318: By `archives` we mean `moves to the 'other' box`, right?

src/inbox/mod_inbox_rdbms.erl
```diff
@@ -72,15 +72,19 @@ init(HostType, _Options) ->
     rdbms_queries:prepare_upsert(HostType, inbox_upsert, inbox,
                                  InsertFields,
                                  [<<"msg_id">>,
-                                  {expression, <<"box">>, <<"CASE WHEN inbox.box='archive' THEN ? ELSE inbox.box END">>},
+                                  {expression, <<"box">>, <<"CASE WHEN ?='bin' THEN 'bin'",
+                                                                " WHEN inbox.box='archive' THEN 'inbox'",
+                                                                " ELSE inbox.box END">>},
```
chrzaszcz, line 77: Could you help me understand this logic? It seems like we can only set `inbox`, `bin` or leave the previous value unchanged, while previously we could set other boxes. What if moving to some other box is aggregated asynchronously? Am I missing something?

chrzaszcz (commented):
Looks good, could you answer some questions?

src/inbox/mod_inbox_rdbms.erl
NelsonVides: The trick is that setting the box is still a synchronous request, I haven't added it to the list of aggregates, partially because is simpler that way, partially because I can assume setting the box to be an operation that doesn't happen in a tight loop. So aggregates are:
* only updating the inbox record, which will try to do inbox if it was archived and otherwise no box change
* resetting the count which does no box change
* removing the record, which does box change to the bin.
So at the end only moves to the bin, or moves from-archive-to-inbox, are attempted.

NelsonVides: To be a bit more pedantic, now that I think about it, one can also argue that the remove-record is also an infrequent operation and can stay synchronous. It also does not interfere with the others, once in the bin an automatic archive-to-inbox won't change anything. That would make the aggregation much simpler. Didn't think about it before...
Making set-inbox-properties async would also be an option, perhaps even better performance with one less operation being sync, but it would make the aggregation quite more complex. Still, I have a bazillion extensions in client code that make the aggregation several orders of magnitude more complex than properties and I managed well, so I can try adding these operations to the aggregation as well.

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/119806) / small_tests / e81a92f44622ac51b3b5eced36838f686970e3fd
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3630/119806/small_tests.24.0.5) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3630/119806/small_tests.24.0.5/small/ct_run.mongooseim@localhost.2022-04-20_07.44.36/lib.mongooseim.logs/run.2022-04-20_07.44.36/suite.log.html)

chrzaszcz (approved):
Looks good 👍

PR #3631, opened 2022-04-16
Ensure maybe_restart_rabbit_connection/1 returns two pids in the tuple.

ColaCheng:
When the rabbit connection is still alive the return will be `{pid(), {ok, pid()}}`. The worker will encounter the error when the next `amqp_channel:call/2`.

This PR ensures the return is the correct data format and adds a test for that.


chrzaszcz (approved):
Looks good

PR #3632, opened 2022-04-19
Force cert cache renewal

chrzaszcz:
CircleCI jobs started failing and the cache has one-week retention time. Changing the key forces it to be rebuilt now.


Premwoik (approved):
Looks good

Issue #3690, opened 2022-06-22
Incorrect type of jwt_secret from persistent_term:get


imcyee:
MongooseIM version: 5.1.0
Installed from: source 
Erlang/OTP version: Erlang/OTP 24 [erts-12.3.1]

https://github.com/esl/MongooseIM/blob/c213d44a0ccdf38fcec5e0d5b2ccab70cbb9407f/src/auth/ejabberd_auth_jwt.erl#L101-L105

It is weird that the get secret in my case clause doesn't match with `is_binary(Key1)`. After some debugging, it seems it is matching `io_lib:char_list`, `is_list`, which could be a string instead of binary. Hence, `check-password` crashed with invalid clause. I could have removed the `is_binary` check and return the `Key1`, everything works fine as below.

```
Key = case persistent_term:get({?MODULE, HostType, jwt_secret}) of
              Key1 -> Key1;
              {env, Var} -> list_to_binary(os:getenv(Var))
          end,
```

I have checked the issue seems to be jwt_secret is not saved correctly to `persistent_terms` in `get_jwt_secret(HostType)`

My config
```
[auth]
  methods = ["jwt"]

  [auth.jwt]
    username_key = "sub"
    secret.value = "VERY_DANGEROUS_SECRET"
    algorithm = "HS256"
```

chrzaszcz:
Fixed

PR #3691, opened 2022-06-22
Fix invalid clause for case for issue 3690

imcyee:
This PR addresses [Issue 3690](https://github.com/esl/MongooseIM/issues/3690) 

Proposed changes include:
Remove `is_binary` for checking string, it will always fail, instead convert to binary.


imcyee:
I have checked the previous version. The issue is in the 'get_jwt_secret' it is missing the check for list/binary. Hence a missing, list_to_binary', which caused it to store incorrectly in persistent_term. I will be away for two days. Will submit again once I have my comp to run again.

imcyee:
Closed. Make way for https://github.com/esl/MongooseIM/pull/3692

PR #3692, opened 2022-06-23
Fix jwt secret

imcyee:
This PR addresses https://github.com/esl/MongooseIM/issues/3690

Proposed changes include:
Add check for `list` type before secret is saved in `persistent_terms`


imcyee:
This is the github action test result, [Test action](https://github.com/imcyee/MongooseIM/actions/runs/2552610535). It failed with pgql_mnesia which i have no idea why it failed while others are ok.

PR #3693, opened 2022-06-24
Add affiliation change REST API for muc lights

apontini:
Muc-light APIs were missing a way to update participants
affiliations. Now this exposes an API to allow just for that.

Swagger documentation was updated.

Tests are missing, unfortunately I'm not really used to Erlang
and Erlang tests

### DIFF
--- a/doc/rest-api/Administration-backend_swagger.yml
+++ b/doc/rest-api/Administration-backend_swagger.yml
@@ -536,6 +536,44 @@ paths:
       responses:
         204:
           description: Message was sent to the MUC Light room
+  /muc-lights/{XMPPMUCHost}/{roomName}/{user}/affiliation:
+    parameters:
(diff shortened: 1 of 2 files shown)


apontini:
The api is actually missing and error response if something goes wrong, I'll try adding it

chrzaszcz:
Thanks for the suggestion. Affiliation management will be added to the new [GraphQL API](https://esl.github.io/MongooseDocs/latest/graphql-api/Admin-GraphQL/) we are developing. It is already possible to change the room owner, to invite a user, and to remove a user from a room. You can use the `master` branch to do so.

We are planning to deprecate the REST API, so if you need this functionality there, the only option would be to sponsor such a feature or to [contact us](https://www.erlang-solutions.com/contact/) to find out an individual solution to suit your needs.

PR #3768, opened 2022-09-22
Unified REST API: domains

chrzaszcz:
Integrate the domain REST handler with `mongoose_admin_api`

The domain handler had some custom code:
- The auth failed if creds were provided, but not configured. I decided to apply this behaviour to the whole Admin API.
- Authorization and request parsing. I just used the functionality already provided by `mongoose_admin_api`.
- `{"what": "message"}` error messages. I decided to use the plain-text ones as in `mongoose_admin_api`.
- Custom test helpers. I also decided to use the ones for `mongoose_admin_api`.

PR #3770, opened 2022-09-22
Async rest endpoint

NelsonVides:
Most often, tenant deletion can incur in many many DB operations, which can introduce very long delays and timeout any HTTP connection. And when the HTTP connection times out, the cowboy process running the tenant deletion hook will timeout, and could potentially leave the DB in an inconsistent state.

This PR introduces a parameter in the delete body, that transforms the request in asynchronous. So far, no special changes are made to the deletion itself, which if it fails it is not retried and the half-committed deletion leaves everything inconsistent. Future changes will make DB operations more incremental, and failures possible to retry, but for now only unblocking the HTTP endpoint is done.

### DIFF
--- a/big_tests/tests/domain_rest_helper.erl
+++ b/big_tests/tests/domain_rest_helper.erl
@@ -11,6 +11,7 @@
          putt_domain_with_custom_body/2,
          rest_select_domain/2,
          rest_delete_domain/3,
+         request_delete_domain/3,
          delete_custom/4,
          patch_custom/4]).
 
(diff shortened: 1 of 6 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/138533) / small_tests / 18308766f485538f72fe212628aae2eb2fc32ead
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3770/138533/small_tests.24.3.4) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3770/138533/small_tests.24.3.4/small/ct_run.mongooseim@localhost.2022-09-22_16.39.02/lib.mongooseim.logs/run.2022-09-22_16.39.02/suite.log.html)

chrzaszcz (changes requested):
This is in confict with https://github.com/esl/MongooseIM/pull/3768, and because that one has priority (it is one of the sprint goals), I don't think we can merge this one.

We need a unified way to handle such long-running operations across GraphQL and REST.

PR #3813, opened 2022-10-17
Remove domain/async

NelsonVides:
Most often, tenant deletion can incur in many many DB operations, which can introduce very long delays and timeout any HTTP connection. And when the HTTP connection times out, the cowboy process running the tenant deletion hook will timeout, and could potentially leave the DB in an inconsistent state.

### DIFF
--- a/big_tests/tests/graphql_domain_SUITE.erl
+++ b/big_tests/tests/graphql_domain_SUITE.erl
@@ -41,6 +41,7 @@ domain_tests() ->
      get_domains_by_host_type,
      get_domain_details,
      delete_domain,
+     request_delete_domain,
      get_domains_after_deletion,
      set_domain_password,
      set_nonexistent_domain_password,
@@ -177,6 +178,20 @@ delete_domain(Config) ->
                    <<"domain">> := #{<<"domain">> := ?SECOND_EXAMPLE_DOMAIN}},
                  ParsedResult2).
 
+request_delete_domain(Config) ->
+    Domain = <<"exampleDomain">>,
+    Result1 = request_remove_domain(Domain, ?HOST_TYPE, Config),
+    ParsedResult1 = get_ok_value([data, domain, requestRemoveDomain], Result1),
+    ?assertMatch(#{<<"msg">> := <<"Domain disabled and enqueued for deletion">>,
+                   <<"domain">> := #{<<"domain">> := Domain,
+                                     <<"status">> := <<"DELETING">>}},
+                 ParsedResult1),
+    F = fun() ->
+                Result = get_domain_details(Domain, Config),
+                domain_not_found_error_formatting(Result)
+        end,
+    mongoose_helper:wait_until(F, ok, #{time_left => timer:seconds(5)}).
+
 get_domains_after_deletion(Config) ->
     Result = get_domains_by_host_type(?HOST_TYPE, Config),
     ParsedResult = get_ok_value([data, domain, domainsByHostType], Result),
@@ -262,6 +277,10 @@ get_domain_details(Domain, Config) ->
 remove_domain(Domain, HostType, Config) ->
     Vars = #{domain => Domain, hostType => HostType},
     execute_command(<<"domain">>, <<"removeDomain">>, Vars, Config).
+
+request_remove_domain(Domain, HostType, Config) ->
+    Vars = #{domain => Domain, hostType => HostType},
+    execute_command(<<"domain">>, <<"requestRemoveDomain">>, Vars, Config).
 
 get_domains_by_host_type(HostType, Config) ->
     Vars = #{hostType => HostType},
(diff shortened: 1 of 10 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/145081) / small_tests / ec4787865f058735aa234caa0a075d58e3d6d3a1
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3813/145081/small_tests.24.3.4) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3813/145081/small_tests.24.3.4/small/ct_run.mongooseim@localhost.2022-10-17_11.14.54/lib.mongooseim.logs/run.2022-10-17_11.14.55/suite.log.html)

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/145105) / small_tests / 6e3fc6db4f250f029b83721044e956aa72c51e74
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3813/145105/small_tests.24.3.4) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3813/145105/small_tests.24.3.4/small/ct_run.mongooseim@localhost.2022-10-17_11.47.51/lib.mongooseim.logs/run.2022-10-17_11.47.52/suite.log.html)

priv/graphql/schemas/admin/domain.gql
```diff
@@ -14,6 +14,9 @@ type DomainAdminMutation @protected{
   "Remove domain. Only for global admin"
   removeDomain(domain: String!, hostType: String!): RemoveDomainPayload
     @protected(type: GLOBAL)
+  "Remove domain. Only for global admin"
```
kamilwaz, line 17:
```suggestion
  "Remove domain asynchronously. Only for global admin"
```

It should help to understand the difference between this and previous one.

src/domain/mongoose_domain_api.erl
```diff
@@ -98,8 +107,10 @@ delete_domain(Domain, HostType) ->
 
 %% This is ran only in the context of `do_delete_domain',
 %% so it can already skip some checks
--spec do_delete_domain_in_progress(domain(), host_type()) -> delete_domain_return().
-do_delete_domain_in_progress(Domain, HostType) ->
+-spec do_delete_domain_in_progress(domain(), host_type(), sync | async) -> delete_domain_return().
+do_delete_domain_in_progress(Domain, HostType, async) ->
```
kamilwaz, line 111: Shouldn't async operation block sync operation? Currently, we allow to do both operations at the same time (or am I missing something?) and it's hard to say what output would we receive.

Generally, I wonder if it wouldn't be better to replace the old endpoint with a new one. After merging this PR, we will have two endpoints for the same operation where one (sync) isn't working correctly in all cases.

kamilwaz (changes requested):

big_tests/tests/graphql_domain_SUITE.erl
```diff
@@ -177,6 +178,21 @@ delete_domain(Config) ->
                    <<"domain">> := #{<<"domain">> := ?SECOND_EXAMPLE_DOMAIN}},
                  ParsedResult2).
 
+request_delete_domain(Config) ->
+    Domain = <<"exampleDomain">>,
+    Result1 = request_remove_domain(Domain, ?HOST_TYPE, Config),
+    ParsedResult1 = get_ok_value([data, domain, requestRemoveDomain], Result1),
+    ct:pal("Value ~p~n", [ParsedResult1]),
```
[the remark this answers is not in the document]
NelsonVides: Huh, debugging artefact, sorry 😅

src/domain/mongoose_domain_api.erl
NelsonVides: We allow both, but they happen in different processes. A sync operation would run on the cowboy process that was spawned for this request, while in an async operation the cowboy process would only synchronously mark the domain as `deleting` and then submit a request to the cleaner to do the cleaning offline, so that the cowboy process can finish its work and return. If you were to do both operations concurrently, the same caveats apply than doing two deletes in the old sync way concurrently, after all they also spawn two parallel cowboy processes to handle the request.

The sync operation is theoretically correct, the problem is that most often HTTP clients will have a timeout and kill the request before the sync operation can finish. But in that case the operation can be retried, that was added in a previous PR. Or the client can submit a much longer timeout than usual.

About whether to remove the sync operation, perhaps, dunno, I'm kinda in favour of leaving it because a client might want to know promptly that the operation is done, but I don't really have an opinion here, don't know what's the best API practice for this matter.

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/145244) / small_tests / aa6e24d273c7ff3adc67ba71866742cc518ee3c8
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3813/145244/small_tests.24.3.4) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3813/145244/small_tests.24.3.4/small/ct_run.mongooseim@localhost.2022-10-18_12.31.05/lib.mongooseim.logs/run.2022-10-18_12.31.06/suite.log.html)

kamilwaz (approved):

PR #3927, opened 2023-01-10
C2S/fix muc_light_legacy tests

chrzaszcz:
The goal is to make `muc_light_legacy_SUITE` pass with the new `mongoose_c2s`.

The new c2s implementation requires proper addressing of the stanzas, so the tests are amended according to https://esl.github.io/MongooseDocs/latest/open-extensions/muc_light/#451-requesting-a-blocking-list

There is also a small fix in the code.


NelsonVides (approved):
Excellent 😄

PR #3928, opened 2023-01-11
Traffic monitor

bartekgorny:
Browser-based tool for watching XMPP traffic.  Rebased, fixed and improved.


src/mongoose_traffic.erl
```diff
@@ -0,0 +1,127 @@
...
+         end,
+    traffic(in, To, El),
+    {ok, Acc}.
+
+traffic(Dir, Jid, El) ->
+    St = fix_and_format(El),
+    UserPid = self(),
+    case whereis(?MODULE) of
+        undefined -> ok;
+        Pid -> gen_server:cast(Pid, {message, Dir, {UserPid, Jid}, St})
```
fenek, line 67: `cast` will always return `ok`, so this `case` is not necessary.

src/mongoose_traffic.erl
```diff
@@ -0,0 +1,127 @@
...
+
+hooks(_HostType) ->
+    [{c2s_debug, global, fun ?MODULE:trace_traffic/3, #{}, 50}].
+
+
+trace_traffic(Acc, #{arg := {out, From, El}}, _) ->
+    traffic(out, From, El),
+    {ok, Acc};
+trace_traffic(Acc, #{arg := {in, El}}, _) ->
+    To = case get_jid(stanza, El) of
```
fenek, line 55: This might be simplified to `get_jid(El, Acc)` with clauses that match on mongoose_acc first, then try to retrieve JID from `#xmlel` and return undefined if it's not there.

web/traffic/src/Traffic.elm
```diff
@@ -0,0 +1,314 @@
+port module Traffic exposing (main)
```
fenek, line 1: The only major problem I see with Elm is that the maintainability of this tool may drop to near-zero if you ever decide to leave ESL.

src/mongoose_traffic_channel.erl
```diff
@@ -0,0 +1,241 @@
...
+    end;
+websocket_info(stop, State) ->
+    {stop, State};
+websocket_info(Info, State) ->
+    ?DEBUG("unknown info: ~p", [Info]),
+    {ok, State}.
+
+force_stop_tracing(State) ->
+    State1 = State#state{tracing = false},
+    M = reply(<<"error">>, #{<<"reason">> => <<"too_many_accounts">>}),
```
fenek, line 103: About the naming: JID is not necessarily equal to an account in XMPP.

src/mongoose_traffic_channel.erl
```diff
@@ -0,0 +1,241 @@
...
+handle(<<"clear_all">>, _, State) ->
+    {<<"cleared_all">>,
+     State#state{traces = #{}, current = <<>>}};
+handle(<<"heartbeat">>, _, State) ->
+    {<<"heartbeat_ok">>,
+     <<>>,
+     State};
+handle(Event, Payload, State) ->
+    ?ERROR_MSG("unknown event: ~p carrying ~p", [Event, Payload]),
+    {<<"zwrotka">>, <<"bzzzzzzzz">>, State}.
```
fenek, line 159: :D

fenek: Why not "dupa"?

src/mongoose_traffic_channel.erl
fenek: Oh, wait, or are these "accounts" the traced events?

src/mongoose_traffic_channel.erl
```diff
@@ -0,0 +1,241 @@
...
+-include("mongoose.hrl").
+-include("jlib.hrl").
+
+-define(LISTENER, ?MODULE).
+-define(MAX_ITEMS, 500).
+-define(MAX_ACCOUNTS, 100).
+
+-record(state, {traces = #{},
+                tracing = false,
+                current = <<>>,
```
fenek, line 26: What's the use of `current`? Difficult to tell after a quick read.

rel/files/mongooseim.toml
```diff
@@ -37,6 +37,18 @@
       shaper_rule = "fast"
       password = "secret"
 
+# [[listen.http]]
+#   port = {{traffic_channel_port}}
```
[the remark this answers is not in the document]
bartekgorny: They were meant to make life easier, so that the generated configs contain the proper code, and all you need to do is uncomment it.

web/traffic/js/tracer_connect.js
```diff
@@ -0,0 +1,6 @@
+
+function open_websocket() {
+    var adr = "ws://" + window.location.host + "/ws-traffic";
```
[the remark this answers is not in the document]
bartekgorny: I don't understand. Websocket is available on the same host that you opened the page from, so this code always works.

web/traffic/js/traffic.js
```diff
@@ -0,0 +1,6174 @@
+(function(scope){
```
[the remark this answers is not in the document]
bartekgorny: Perhaps. If this file is in the repo, it is immediately available, but it clutters the repo. If it is not, you have to install Elm to compile it. Some people are even opposed to placing Elm code in this repo. There was a discussion today about making the monitor a subrepo or a dependency.

src/mongoose_traffic_channel.erl
bartekgorny: Technically, an account here is a c2s process, which obviously maps 1to1 to full jid.

src/mongoose_traffic_channel.erl
bartekgorny: This is pid you are currently following. The monitor is accumulating stanzas from all traces pids, but one of them is the current one (you have it on your screen) and its stanzas are being streamed to you.

big_tests/tests/mod_blocking_SUITE.erl
```diff
@@ -132,6 +134,42 @@ end_per_testcase(CaseName, Config) ->
 %% Tests
 %%--------------------------------------------------------------------
 
+test_sending_short(Config) ->
```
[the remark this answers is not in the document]
bartekgorny: Come to think of that, it can probably be removed altogether.

src/http_upload/mod_http_upload_api.erl
```diff
@@ -19,7 +19,7 @@ get_urls_mongooseimctl(Domain, Filename, Size, ContentType, Timeout) ->
             {error, Message}
     end.
 
--spec get_urls(Domain :: jid:lserver(), Filename :: nonempty_binary(), Size :: pos_integer(),
+-spec get_urls(Domain :: jid:lserver(), Filename :: binary(), Size :: pos_integer(),
```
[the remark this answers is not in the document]
bartekgorny: It caused a weird dialyzer fail.

rel/files/tracer_connect.js
```diff
@@ -0,0 +1,5 @@
+
+function open_websocket() {
```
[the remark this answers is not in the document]
bartekgorny: Yes, but it is supposed to end up in the other place because it is there. We need to decide how the build and deployment process should look like and what goes where.

web/traffic/js/tracer_connect.js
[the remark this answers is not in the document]
fenek: I think it will be more future proof - it would decouple the monitor frontend and allow to even use a local copy of it to connect to various MIM instances in the network.

src/mongoose_traffic_channel.erl
fenek: Yes, c2s maps to a full JID but I've never seen it called "account", which is usually a set of user data in the DB. Unnecessary breach of naming convention.

bartekgorny:
We still need to decide how to make it available without storing compiled javascript in the main repo.

PR #3857, opened 2022-11-15
C2s/refactor user send packet

NelsonVides:
user_send_packet subscribers can really be split into different handlers for messages or presences as needed, considering most often they only need messages like in the case of MAM, inbox, or carbon-copies.

This PR also fixes usages of types that don't exist anymore: `mongoose_c2s:hook_result()` was moved to `mongoose_c2s_hooks:hook_result()`, which in turn I here abbreviate to `mongoose_c2s_hooks:result()`, as the module name already contains the fact that we're referring to hooks. Similarly for other types defined in the `c2s_hooks` module.

PR #3955, opened 2023-02-01
Use more granular c2s hooks

jacekwegr:
This PR changes the hooks used in the `mod_event_pusher` and `mod_mam` to more granular ones.

### DIFF
--- a/src/event_pusher/mod_event_pusher_hook_translator.erl
+++ b/src/event_pusher/mod_event_pusher_hook_translator.erl
@@ -22,7 +22,7 @@
 
 -export([add_hooks/1, delete_hooks/1]).
 
--export([user_send_packet/3,
+-export([user_send_message/3,
          filter_local_packet/3,
          user_present/3,
          user_not_present/3,
@@ -54,9 +54,9 @@ filter_local_packet({From, To, Acc0, Packet}, _, _) ->
           end,
     {ok, {From, To, Acc, Packet}}.
 
--spec user_send_packet(mongoose_acc:t(), mongoose_c2s_hooks:params(), gen_hook:extra()) ->
+-spec user_send_message(mongoose_acc:t(), mongoose_c2s_hooks:params(), gen_hook:extra()) ->
     mongoose_c2s_hooks:result().
-user_send_packet(Acc, _, _) ->
+user_send_message(Acc, _, _) ->
     Packet = mongoose_acc:packet(Acc),
     ChatType = chat_type(Acc),
     ResultAcc = if
@@ -132,6 +132,6 @@ hooks(HostType) ->
         {filter_local_packet, HostType, fun ?MODULE:filter_local_packet/3, #{}, 80},
         {unset_presence_hook, HostType, fun ?MODULE:user_not_present/3, #{}, 90},
         {user_available_hook, HostType, fun ?MODULE:user_present/3, #{}, 90},
-        {user_send_packet, HostType, fun ?MODULE:user_send_packet/3, #{}, 90},
+        {user_send_message, HostType, fun ?MODULE:user_send_message/3, #{}, 90},
         {unacknowledged_message, HostType, fun ?MODULE:unacknowledged_message/3, #{}, 90}
     ].
--- a/src/mam/mod_mam_pm.erl
+++ b/src/mam/mod_mam_pm.erl
@@ -480,7 +480,7 @@ should_archive_if_groupchat(_, _) ->
 %% Only store messages sent to yourself in user_send_packet.
 should_archive_if_sent_to_yourself(LocJID, RemJID, incoming) ->
     not jid:are_bare_equal(LocJID, RemJID);
-should_archive_if_sent_to_yourself(LocJID, RemJID, _Dir) ->
+should_archive_if_sent_to_yourself(_LocJID, _RemJID, _Dir) ->
     true.
 
 -spec return_external_message_id_if_ok(ReturnMessID :: boolean(),
--- a/src/mam/mod_mam_pm.erl
+++ b/src/mam/mod_mam_pm.erl
@@ -47,7 +47,7 @@
 
 %% hook handlers
 -export([disco_local_features/3,
-         user_send_packet/3,
+         user_send_message/3,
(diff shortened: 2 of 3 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/164662) / small_tests / b2907f60627d8884f89b6d08cab5558cce92ba9f
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3955/164662/small_tests.24.3.4) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3955/164662/small_tests.24.3.4/small/ct_run.mongooseim@localhost.2023-02-01_10.01.43/lib.mongooseim.logs/run.2023-02-01_10.01.44/suite.log.html)

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/164821) / small_tests / c851084217adf1c07cf377aaa5a6e55f44a2ba30
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3955/164821/small_tests.24.3.4) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3955/164821/small_tests.24.3.4/small/ct_run.mongooseim@localhost.2023-02-01_12.38.37/lib.mongooseim.logs/run.2023-02-01_12.38.38/suite.log.html)

NelsonVides (commented):
The two modules still subscribed to `user_send_packet`:
* MAM: don't we want to move it to `user_send_message` as well?
* event pusher: it is probably right, but, have we verified that down the line the event pusher only cares about messages? I'd be happy with a pointer to the line of code were we can know that :)

src/event_pusher/mod_event_pusher_hook_translator.erl
```diff
@@ -22,7 +22,7 @@
 
 -export([add_hooks/1, delete_hooks/1]).
 
--export([user_send_packet/3,
+-export([user_send_message/3,
```
DenysGonchar, line 25: I'm not 100% sure if really want to change the hook type here.

DenysGonchar:
What about other places, e.g. mam, inbox, mod_offline (actually all the places that use user_send_packet hook)? Even if we don't need to change anything there, that would be nice to see some analysis and conclusions.

jacekwegr:
> The two modules still subscribed to `user_send_packet`:
> 
>     * MAM: don't we want to move it to `user_send_message` as well?
> 
>     * event pusher: it is probably right, but, have we verified that down the line the event pusher only cares about messages? I'd be happy with a pointer to the line of code were we can know that :)

* I think that MAM is also used for querying for messages, and when you change the hook to `user_send_message`, it stops working, and tests like `graphql_stanza_SUITE:admin_stanza_http:admin_mam:admin_get_last_messages` are going to fail, so I decided to leave it as is.

* Correct me if I'm wrong, but the records found in `include/mod_event_pusher_events.hrl` and which are used by `src/event_pusher/mod_event_pusher_hook_translator.erl` can suggest that it is used only for messages.

jacekwegr:
> What about other places, e.g. mam, inbox, mod_offline (actually all the places that use user_send_packet hook)? Even if we don't need to change anything there, that would be nice to see some analysis and conclusions.

- Changing the mod_mam hook to `user_send_message` causes some trouble while querying for messages, so I think it is better to leave it as is
- mod_inbox, mod_presence, mod_smart_markers and mod_carboncopy were already changed to use `user_send_message` and `user_receive_message` in this PR https://github.com/esl/MongooseIM/pull/3857
- mod_ping uses `user_send_packet`, but it makes sense because it is not used for messages
- mod_stream_management uses `user_send_packet` and `user_receive_packet`, which also makes sense to me because it is not used for handling messages
- metrics use `user_send_packet` and `user_receive_packet` to then divide by the type of message, which seems fine

NelsonVides:
ping, stream management, and metrics, are the ones that actually make sense they should capture all packets.

About the event_pusher, I also understood it cares only for messages: in the hook handler, it checks for `chat_type` and will act only when the type is `chat`, `groupchat`, `headline`, `normal`, or not defined. This means that it surely can't be:
* an IQ because IQ types are required and can only be `get`, `set`, `error`, `result`: https://www.rfc-editor.org/rfc/rfc6120.html#section-8.2.3
* a presence because presence types can only be `error`, `probe`, `subscribe`, `subscribed`, `unavailable`, `unsubscribe`, `unsubscribed` and when not provided it means `available`: https://www.rfc-editor.org/rfc/rfc6121#section-4.7.1. This means that a presence without type will be passed, but then all callbacks seem to filter that out, like in the case of https://github.com/esl/MongooseIM/blob/997ce8cc01dacf8bf1f1f4e3a984ee10f0ce5dd6/src/event_pusher/mod_event_pusher_push.erl#L115. And in any case, it doesn't seem sensible to me to push only available presences without pushing at least the unavailable counterpart so it seems like a sensible change here.

About MAM, this one's weird, because as far as I understand https://xmpp.org/extensions/xep-0313.html it seems to me to require only messages, so I don't know why MAM would be concerned about the other types of messages. If a test fails I find that suspicious 🙄 MAM is also used for fetching messages, but the fetch happens in the IQ handler, here the hook is only capturing the messages that will be stored 🤔 

So I'd only ask, do we want to investigate and fix that MAM failure in this PR or in some other case? If leaving that for later I'd approve this one for now 👌🏽

mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/166358) / small_tests / 3f042fc42444ae4aede2d4f085ae9b46f52b80fe
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3955/166358/small_tests.24.3.4) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3955/166358/small_tests.24.3.4/small/ct_run.mongooseim@localhost.2023-02-17_14.19.50/lib.mongooseim.logs/run.2023-02-17_14.19.51/suite.log.html)

DenysGonchar (approved):
looks ok

PR #3966, opened 2023-02-18
 Fix typo in doc [iOS tutorial]

askarsyzdykov:
There is incorrect markdown parsing issue in the iOS tutorial page https://esl.github.io/MongooseDocs/latest/tutorials/iOS_tutorial/. I deleted the space that can lead to the issue.

изображение


NelsonVides (approved):
Good catch! Thanks 🎉

PR #3983, opened 2023-03-09
Avoid duplicated close_session

NelsonVides:
Ensuring that the session is closed should be the responsibility of the c2s code, which will handle the case on an external state machine state. Running close_session when the session is resumed means running close_session twice.

This PR is a quickfix for that case.

### DIFF
--- a/src/c2s/mongoose_c2s.erl
+++ b/src/c2s/mongoose_c2s.erl
@@ -17,7 +17,7 @@
          get_mod_state/2, get_listener_opts/1, merge_mod_state/2, remove_mod_state/2,
          get_ip/1, get_socket/1, get_lang/1, get_stream_id/1, hook_arg/5]).
 -export([filter_mechanism/2, c2s_stream_error/2, maybe_retry_state/1, merge_states/2]).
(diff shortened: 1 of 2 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/168882) / small_tests / e8c15cb7a615aea46ecce94980cf2479bedc9433
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/3983/168882/small_tests.24.3.4) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/3983/168882/small_tests.24.3.4/small/ct_run.mongooseim@localhost.2023-03-09_21.22.15/lib.mongooseim.logs/run.2023-03-09_21.22.15/suite.log.html)

pawlooss1 (commented):
Looks good 👍 I'm just wondering if we can come up with a test case wich would reproduce it, fail on previous version and pass on the new version. What do you think?

pawlooss1:
After discussion with @NelsonVides it turned out that such case is not straightforward to reproduce and we decided to merge it without such test.

pawlooss1 (approved):

arcusfelis (commented):
ok

PR #4077, opened 2023-08-04
C2S features small optimisation

NelsonVides:
Two things here:
- First is that the hook `c2s_stream_features` can already take the initial list of pre-created features as an argument, instead of giving an empty list and then appending two lists. This way, the initial list can also be picked up from subsequent handlers and modified – which SASL2 will do with SASL1 ascribed earlier.
- `cyrsasl:listmech/1` output is given to `mongoose_c2s:filter_mechanism/1` twice, once in c2s itself, and another one when creating the features in c2s_stanza. Unify them.


chrzaszcz (approved):
👍

arcusfelis (approved):
looks fine, probably would use Mech variable instead of M.

PR #4088, opened 2023-08-09
Add prototype for backend = auto support [DO NOT MERGE]

arcusfelis:
This PR addresses "experimenting how to set auto backend". Probably will do inside the config parser instead (?)

Proposed changes include:
* Do not merge

Auto-selecting backend

Allow to set backend = auto (i.e. disables validation that auto module exists).

mongoose_backend:init would have to take an extra option ModuleType = external_db | internal_db.

We actually do not call config module directly from mongoose_backend module.
So, if we wanna know which backend is running - use mongoose_backend:get_backend_module/2.
And config API would return backend = auto, which we are fine with (we don't have logic which would use backend name anyway).

Also, mongoose_backend could call is_backend_present/1 which would do check:
- for rdbms - that rdbms is configured in outgoing pools
- for mnesia - that mnesia is configured in internal_databases
- for cets - that cets is configured in internal_databases
- for unknown backends it will do nothing

(it is just quick config check, so should be fast)

Autoselecting would just use mnesia for both internal and persistent data, if
preferred_persistent_db and preferred_internal_db options are not set
(so, we would not have issues with old configs, that assume that the default backend is mnesia).

But, we could do default backend = auto now, so config should be really clean
(i.e. no need to specify backend = "rdbms" for every module, could just say preferred_persistent_db = "rdbms" in the general options).

Naming

Could be auto_persistent_database or preferred_persistent_db.
While we could try allowing a list there `auto_persistent_database = [cets, mnesia]`, a single database is fine and simple.

Enabling:

```
[general]
  default_volotile_db = "cets"
  default_persistent_db = "rdbms"

[modules.mod_private]
```

Check:

```
mongoose_backend:get_backend_module(<<"localhost">>, mod_private).
mod_private_rdbms
```

Motivation:

Removes these three options, but adds two new:
```
[general]
  sm_backend = "cets"
  component_backend = "cets"
  s2s_backend = "cets"
```

Also, we don't need to specify `backend = "rdbms"` anymore in modules, which is good (it is a common bug to have it unspecified).

The issue it does not reduces the size of the default config that much, sadly.
But it could be a first step :)

We could try to autocreate shapers as the next step.

### DIFF
--- a/src/config/mongoose_config_spec.erl
+++ b/src/config/mongoose_config_spec.erl
@@ -201,7 +201,13 @@ general() ->
                                                     wrap = global_config},
                  <<"domain_certfile">> => #list{items = domain_cert(),
                                                 format_items = map,
-                                                wrap = global_config}
+                                                wrap = global_config},
+                 <<"default_persistent_db">> => #option{type = atom,
+                                                        validate = {enum, [rdbms, mnesia]},
+                                                        wrap = global_config},
+                 <<"default_volotile_db">> => #option{type = atom,
+                                                      validate = {enum, [cets, mnesia]},
+                                                      wrap = global_config}
                 },
        wrap = none,
        format_items = list
@@ -221,7 +227,9 @@ general_defaults() ->
       <<"mongooseimctl_access_commands">> => #{},
       <<"routing_modules">> => mongoose_router:default_routing_modules(),
       <<"replaced_wait_timeout">> => 2000,
-      <<"hide_service_name">> => false}.
+      <<"hide_service_name">> => false,
+      <<"default_volotile_db">> => mnesia,
+      <<"default_persistent_db">> => mnesia}.
 
 ctl_access_rule() ->
     #section{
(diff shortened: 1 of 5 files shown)


mongoose-im:
[small_tests_24](https://circleci.com/gh/esl/MongooseIM/185807) / small_tests / b81d8d1c6de893ffd8d7db857ee7a8f2ac5be78f
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/4088/185807/small_tests.24.3.4.9-amd64) / [small](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/4088/185807/small_tests.24.3.4.9-amd64/small/ct_run.mongooseim@localhost.2023-08-09_08.31.39/lib.mongooseim.logs/run.2023-08-09_08.31.40/suite.log.html)

arcusfelis:
Close for now, we still have the code.
Use Helm charts for now to do configuration for all backends.

PR #4122, opened 2023-09-11
Return a proper type from mod_muc:node_cleanup_for_host_type/3

arcusfelis:
This PR addresses MIM-2039 - Process crashes when another node stops during the upgrade.

Error:

```json
when=2023-09-11T12:12:54.123796+00:00 level=error reason="{{case_clause,#{}},[{gen_hook,run_hook,4,[{file,\"/home/circleci/project/src/gen_hook.erl\"},{line,237}]},{mongoose_hooks,run_fold,4,[{file,\"/home/circleci/project/src/hooks/mongoose_hooks.erl\"},{line,1388}]},{mongoose_cleaner,'-run_node_cleanup/1-lc$^0/1-0-',2,[{file,\"/home/circleci/project/src/mongoose_cleaner.erl\"},{line,92}]},{mongoose_cleaner,'-run_node_cleanup/1-fun-1-',1,[{file,\"/home/circleci/project/src/mongoose_cleaner.erl\"},{line,92}]},{timer,tc,1,[{file,\"timer.erl\"},{line,235}]},{mongoose_cleaner,run_node_cleanup,1,[{file,\"/home/circleci/project/src/mongoose_cleaner.erl\"},{line,90}]},{global,trans,4,[{file,\"global.erl\"},{line,477}]},{mongoose_cleaner,cleanup_modules,1,[{file,\"/home/circleci/project/src/mongoose_cleaner.erl\"},{line,79}]}]}" pid=<0.652.0> at=gen_server:error_info/8:1402 state={state} name=mongoose_cleaner log= last_message={nodedown,'mongooseim@mongooseim-0.mongooseim.default.svc.cluster.local'} label={gen_server,terminate} client_info=undefined
[...]

PR #4126, opened 2023-09-13
Dialyzer extra checks

arcusfelis:
This PR addresses "PoC for more dialyzer checks".

Proposed changes include:
* The bug like that https://github.com/esl/MongooseIM/pull/4122 could be easily avoided if we had `missing_return` option enabled for dialyzer.
* Except, enabling it would produce 200 warnings.
* Enabling `extra_return` would make 200 more warnings.
* OTP 25 had no support to disable warnings for functions (it had nowarn attribute, but not per function per check).
* OTP 26 adds extra attributes we use in this PR.

Proposed idea allows to enable missing checks per module basis.
This PR enables the checks for `mod_stream_management`.
The idea is to enable the checks gradually for more and more modules, till we enable the checks for all modules.
After that we could remove the attributes, enable the check for all modules in rebar.config and benifit from proper static type analysis.

These attributes are only supported by OTP 26, so no CI yet (we could add a job for Circle CI though without waiting for all tests to pass on OTP 26).

missing_return would catch the errors like:
- we said we return binary() in spec but we return a string() instead.
- we said we return `{ok, Type}` in spec, but we return `Type` instead.
- sadly, if we use some generic API like mongoose_config, we have to disable the check for that specific function (but it is kinda ok, we usually have getter functions to get options which do not contain any logic. So, disabling the check per function would say dialyzer "yes, trust me, it is that type").

unmatched_returns would force us to write `_ = function_call()`, if the function makes some complex result like `ok | {error, _}` and we silently ignore the result (so, reader would know that the place could require some branching logic).

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -3,7 +3,8 @@
             {d, xml_nif},
             {d, namespaced_types},
             {platform_define, "^(20)", fun_stacktrace},
-            {platform_define, "^(25)", gen_server_request_id}]}.
+            {platform_define, "^(25)", gen_server_request_id},
+            {platform_define, "^(26)", dialyzer_has_missing_return_attribute}]}.
 
 %% For behaviour info
 {erl_first_files, [
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -1,3 +1,4 @@
+% vim: syntax=erlang
 os:putenv("EXOMETER_PACKAGES", "(minimal)"),
 
 MaybeReadFromConfig =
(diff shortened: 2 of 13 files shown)


mongoose-im:
[elasticsearch_and_cassandra_25](https://circleci.com/gh/esl/MongooseIM/191452) / elasticsearch_and_cassandra_mnesia / a285b658ecee122550a841e475acb1242716a6ab
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/4126/191452/elasticsearch_and_cassandra_mnesia.25.3-amd64)/ [big](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/4126/191452/elasticsearch_and_cassandra_mnesia.25.3-amd64/big/ct_run.test%4001a5c43c5404.2023-09-13_13.40.43/index.html)
**OK: 369** / Failed: 0 / **User-skipped: 38** / Auto-skipped: 0

PR #4154, opened 2023-10-27
Traffic monitor

bartekgorny:
Browser-based tool for watching XMPP traffic. Now downloaded as a dependency.


rel/files/tracer_connect.js
```diff
@@ -0,0 +1,5 @@
+
+function open_websocket() {
```
bartekgorny, line 2: this file will be removed, it is not needed anymore

bartekgorny:
I don't know why tests are failing, I don't seem to have access to CI

Issue #4173, opened 2023-11-20
Change config in muclight replacing old config


jaspreet-android:
MongooseIM version: 6.1
Installed from: source
Erlang/OTP version: 25

All config parameters that are not provided in the request are reverted to their default values, which should not happen.

`<iq from='jaspreet@localhost'
    id='subject1'
    to='new_room@muclight.localhost'
    type='set'>
    <query xmlns='urn:xmpp:muclight:0#configuration'>
        <roomname>roomname changed</roomname>
    </query>
</iq>`

This request is to change the room name of muc light room but it also sets the room subject to default value.

PR #4175, opened 2023-11-21
Change config in muclight was replacing old config.

jaspreet-android:
This PR addresses [#4173 ](https://github.com/esl/MongooseIM/issues/4173)

Proposed changes include: 
This change will fix the config in muc-light which was replacing the old config to the default value.

### DIFF
--- a/src/muc_light/mod_muc_light_room_config.erl
+++ b/src/muc_light/mod_muc_light_room_config.erl
@@ -61,10 +61,10 @@ from_binary_kv_diff(RawConfig, ConfigSchema, Config) ->
         {error, Reason} ->
             {error, Reason};
         {value, RRawConfig, RConfigSchema, KV} ->
-            from_binary_kv(RRawConfig, RConfigSchema, [KV | Config]);
-        {default, _, _, _} ->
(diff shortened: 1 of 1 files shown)


jaspreet-android:
> Hi, thanks for the pull request. We would have to have tests for the fix inside muc_light suite in the big tests directory: https://github.com/esl/MongooseIM/blob/master/big_tests/tests/graphql_muc_light_SUITE.erl

Screenshot 2023-11-21 at 9 01 23 PM
I think test is passing on my side.

NelsonVides (changes requested):
Hi @jaspreet-android 
Thanks a lot for this PR, seems like sensible code, however, it lacks *new* tests, so we can't accept it without a proof that the issue was there and the source code changes indeed fixed it. Please add tests for any bugfix submitted. I'll work on this sooner than later within the scheduled sprint planning of MIM in case you didn't manage to do so earlier.

NelsonVides:
@jaspreet-android I just tested locally, and `muc_light_SUITE` (not `graphql_muc_light_SUITE` as you pasted in your screenshot), fails in my machine. This needs fixing, old tests can't fail and new functionality must come with new tests.

jaspreet-android:
This is fixed as in master, so I am closing it.

PR #4250, opened 2024-03-26
Make CETS node cleanup consistent

chrzaszcz:
Previously, during node cleanup for CETS:
- Some modules removed only local content
- Others performed cluster-wide CETS operations

It is not possible to always remove local content because of `session_cleanup` and the need for insertions - see e.g. stream management. Meanwhile, Mnesia cleanup consistently removes all content across the cluster.

After this change, CETS does the same. An immediate benefit is that after the first `global:trans/2` finishes in `mongoose_cleaner:run_node_cleanup/1`, cluster state is consistent. Previously, cleanup would need to succeed on all nodes (sequentially), and before that the cluster would be in an inconsistent state.

Also:
- Refactor cleanup tests to make them more maintainable.
- Improve test coverage for CETS cleanup.

PR #4251, opened 2024-03-27
Revert dist blocker

chrzaszcz:
The `dist_blocker` feature was introduced to protect a disconnected node from reconnecting before other nodes finished the cleanup to avoid issues with inconsistent CETS tables. The solution was to change the cookie value on `nodeup` (sic!) for the particular node until the cleanup for that node is finished.

There are following issues with that solution:
- Sometimes there is a delay before `nodedown` is received, and cleanup is started. However, when the node comes back online, `nodedown` would be received, followed by `nodeup`. The problem is that `dist_blocker` is preventing the node from connecting, which might delay reconnection. In [MongooseHelm tests](https://github.com/esl/MongooseHelm/blob/master/test/mim_kind_SUITE.erl#L97) this was causing a 15-second delay before reconnection, but it could be more in some configurations. There is even potential for a deadlock, but I couldn't trigger it in tests.
- When `dist_blocker` kicks in, multiple error messages would appear on several nodes, informing about the blocked connections. These unwanted and surprising logs could confuse the user.
- Changing cookies looks like a very radical solution to me, and the recent improvement of cleanup makes it avoidable.

Because of these concerns, `dist_blocker` is disabled in this PR. It is enough to disable it in MIM, because CETS does not enable it by default.

The MongooseHelm tests have shown that after https://github.com/esl/MongooseIM/pull/4250 it is difficult to reproduce the issue - it occurred only once in about 50 tests, which try to quickly restart MIM to trigger the issue. Keep in mind, that with `dist_blocker` it has also [failed](https://app.circleci.com/pipelines/github/esl/MongooseHelm/221/workflows/148fd51f-52db-49f7-a5df-a4c2274e6cdc/jobs/298) at least twice on CI.

We can prevent the issues in different ways, and we can do so in separate PR's.

### DIFF
--- a/src/ejabberd_sup.erl
+++ b/src/ejabberd_sup.erl
@@ -38,7 +38,6 @@ start_link() ->
     supervisor:start_link({local, ?MODULE}, ?MODULE, []).
 
 init([]) ->
-    DistBlocker = worker_spec(cets_dist_blocker),
     Hooks = worker_spec(gen_hook),
     Cleaner = worker_spec(mongoose_cleaner),
     Router = worker_spec(ejabberd_router),
@@ -65,8 +64,7 @@ init([]) ->
     IQSupervisor =
         ejabberd_tmp_sup_spec(ejabberd_iq_sup, [ejabberd_iq_sup, mongoose_iq_worker]),
     {ok, {{one_for_one, 10, 1},
-          [DistBlocker,
-           StartIdServer,
+          [StartIdServer,
            PG,
            Hooks,
            Cleaner,
--- a/src/mongoose_cleaner.erl
+++ b/src/mongoose_cleaner.erl
@@ -34,7 +34,6 @@ start_link() ->
 %%%===================================================================
 
 init([]) ->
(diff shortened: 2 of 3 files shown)


mongoose-im:
[elasticsearch_and_cassandra_26](https://circleci.com/gh/esl/MongooseIM/213560) / elasticsearch_and_cassandra_mnesia / 509759e917f4c152f12e50a6b42488c7a51bf0fc
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/4251/213560/elasticsearch_and_cassandra_mnesia.26.1.2-amd64)/ [big](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/4251/213560/elasticsearch_and_cassandra_mnesia.26.1.2-amd64/big/ct_run.test%400865fff3807b.2024-03-27_08.45.18/index.html)
**OK: 437** / Failed: 0 / **User-skipped: 43** / Auto-skipped: 0

jacekwegr (approved):
Looks good to me! 👍

PR #4262, opened 2024-04-11
Make commit hash in system status fail-safe

chrzaszcz:
This hotfix is needed to make the build for the 6.2.1 release pass. The tests were failing only for GH tags, because the commit hash was not available.

For now - default to an empty string. We could make the field optional in the future to clean this up.

Big tests checked manually with and without the commit hash.


JanuszJakubiec (approved):
Thanks for the changes

PR #4273, opened 2024-04-23
Fix and update package building scripts

chrzaszcz:
This PR fixes and updates the packages built with CircleCI.

The original issue was that the `packages.erlang-solutions.com/...` repos stopped working. I reported this, and they are almost fully functional again (except for Centos 7).

In the meantime, I researched possible alternatives, because the repos are now considered old and discontinued.
- Using the new repos `binaries2.erlang-solutions.com/...` would be a good option, but they don't work. I reported this, but it will take time to fix these issues.
- Using directly downloaded packages from https://www.erlang-solutions.com/downloads/. This option worked correctly, so I used it.

With the new packages it was possible to update the Erlang versions, and add support for the newer Ubuntu 22.04 Jammy Jellyfish.

Notes:
- I had to use a few different Erlang versions, because of the availability of downloadable packages. This should be fixed, because we need to support all distros for each new Erlang version, and currently it is random. I reported all the issues with ESL Erlang packages.
- In the future we might want to support Arm64, but Erlang packages are not available atm, so it's a task for the future.
- I had to remove RPaths from the `rpm` packages to avoid errors, just like here https://github.com/esl/packages-pipeline/blob/master/esl_erlang/ctx/scripts/rpm/script.sh#L100-L101.

PR #4276, opened 2024-05-07
Fix Erlang version for GitHub Actions

gustawlippa:
Currently, Erlang does not download on GitHub Actions, which we use to test the master branch. See 
https://github.com/esl/MongooseIM/actions/runs/8971880579 for example.
The issue was with the new scripts that download the Erlang package from our website.
The fix is taken from 05f03eb from https://github.com/esl/MongooseIM/pull/4273.

I've run GH Actions for this branch manually, and the packages jobs have passed: https://github.com/esl/MongooseIM/actions/runs/8984746064.
There are some red results, but they should be unrelated (flaky tests?).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,7 +191,7 @@ jobs:
         pkg: [centos_7, ubuntu_xenial]
     runs-on: ubuntu-22.04
     env:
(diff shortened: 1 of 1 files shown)


arcusfelis (approved):
good

PR #4286, opened 2024-05-22
Fix slow MAM muc_prefs_cases tests

gustawlippa:
This PR fixes three testcase added recently. I'm having issues with this case in my other PR, and noticed that it's surprisingly slow.
Time before fix:
obraz
Time after fix:
obraz

Explanation:
Since the room is reused across `muc_run_prefs_case` calls, there is no need to resend presence every time - the users are room participants all the time. The issue is that when (re)entering, Alice would wait for 4 stanzas - 3 presences and 1 room subject message. However, only in the beginning, as a new user would she receive the room subject - subsequent waits for 4 stanzas failed silently after only 3 presences, but only after waiting the 5s timeout. This has happend 6 times per testcase, and there were 3 tescases using this mechanism, so there is 3*6*5=90s less pointless waiting now.

I think maybe Escalus should fail loudly when less then N stanzas arrive when waiting for N stanzas.

### DIFF
--- a/big_tests/tests/mam_SUITE.erl
+++ b/big_tests/tests/mam_SUITE.erl
@@ -3573,6 +3573,13 @@ muc_run_prefs_cases(DefaultPolicy, ConfigIn) ->
     F = fun(Config, Alice, Bob, Kate) ->
         %% Just send messages for each prefs configuration
         Namespace = mam_ns_binary_v04(),
+
+        Room = ?config(room, Config),
+        escalus:send(Alice, stanza_muc_enter_room(Room, nick(Alice))),
+        escalus:send(Bob, stanza_muc_enter_room(Room, nick(Bob))),
+        escalus:send(Kate, stanza_muc_enter_room(Room, nick(Kate))),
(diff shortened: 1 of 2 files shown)


mongoose-im:
[elasticsearch_and_cassandra_26](https://circleci.com/gh/esl/MongooseIM/219822) / elasticsearch_and_cassandra_mnesia / 41863b7a88b392803113f15ebc6a4d5c0ad9264b
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/4286/219822/elasticsearch_and_cassandra_mnesia.26.2.2-amd64)/ [big](https://circleci-mim-results.s3.eu-central-1.amazonaws.com/PR/4286/219822/elasticsearch_and_cassandra_mnesia.26.2.2-amd64/big/ct_run.test%403ed805ef64b6.2024-05-22_10.48.08/index.html)
**OK: 457** / Failed: 0 / **User-skipped: 41** / Auto-skipped: 0

jacekwegr (approved):
Looks good! Thanks for the fix!

PR #4397, opened 2024-11-13
Remove goldrush (unused now) from rebar.lock

chrzaszcz:
```
$ rebar3 get-deps
===> Verifying dependencies...
===> App goldrush is no longer needed and can be deleted.
```


JanuszJakubiec (approved):
Looks good 👍🏻

PR #4414, opened 2024-12-03
Limit building packages to tags and master

chrzaszcz:
Previously it was for every branch.

Main motivation is to save CI credits and to limit unwanted failures in case of issues only with packages.


NelsonVides (approved):
Great 👌🏽

PR #4467, opened 2025-01-20
Improve the consistency of just_tls helpers

NelsonVides:
Note that now we can use `erlang:alias/1` to ensure that this message doesn't come from anywhere else, so the `Ref` parameter is not needed to ensure uniqueness. This allows us to unify the c2s preparing options from all other submodules preparing so, and we can fix naming convention by simply `make_client_opts` vs `make_server_opts`.

Note that the given `verify_fun` is just the default for the cases when `verify_mode` is `peer` or `none`, so only for `selfsigned_peer` it is needed to be provided explicitly. Also, the `verify_fun_wrapper` gets quite more readable this way.

We also here extend tests to ensure we are testing the case of unsupported critical extensions.

### DIFF
--- a/big_tests/tests/ca_certificate_helper.erl
+++ b/big_tests/tests/ca_certificate_helper.erl
@@ -14,9 +14,12 @@ generate_cert(Config, #{cn := User} = CertSpec, BasicTemplateValues) ->
     ct:log("OpenSSL config: ~ts~n~ts", [UserConfig, OpenSSLConfig]),
     file:write_file(UserConfig, OpenSSLConfig),
     UserKey = filename:join(?config(priv_dir, Config), User ++ "_key.pem"),
+    MaybeCritical = maps:get(with_critical_extension, CertSpec, false),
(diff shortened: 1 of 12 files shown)


mongoose-im:
[elasticsearch_and_cassandra_27](https://circleci.com/gh/esl/MongooseIM/256088) / elasticsearch_and_cassandra_mnesia / 3bb6599cfcd8c7f78735b17a7d114be852e254fe
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/4467/256088/elasticsearch_and_cassandra_mnesia.27.1.2-amd64)/ [big](https://esl.github.io/html-zip-reader/PR/4467/256088/elasticsearch_and_cassandra_mnesia.27.1.2-amd64/big.tar.gz//ct_run.test%404ad9d20161d9.2025-01-20_14.53.22/index.html)
**OK: 472** / Failed: 0 / **User-skipped: 49** / Auto-skipped: 0

arcusfelis (approved):
ok

PR #4474, opened 2025-01-28
Migrated graphiql to v3

xvnds:
Current implemented GraphiQL version in this repo still uses React UMD build. This caused the IDE to be stuck at "Loading" since React 19 doesn't have UMD builds anymore: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#umd-builds-removed

Updated GraphiQL to v3 so it uses the latest tech stack.

Reference: https://github.com/graphql/graphiql/tree/main/examples/graphiql-cdn


DenysGonchar (approved):
Looks good, thanks for the contribution.

PR #4529, opened 2025-05-09
Fix flaky test in bosh_SUITE

chrzaszcz:
The instrumentation event `xmpp_element_in` checked in `create_and_terminate_session` was actually NOT emitted in this test, but in the previous ones. This is why the test has become extremely flaky. It wasn't flaky before reworking the XMPP element events, because they used to be emitted by stream opening/ending tags as well.

Solution:
- Simplify event assertions - check them once in `interleave_requests`, where they are all emitted indeed.
- Make sure the asserted events are emitted in the same test by using timestamps.

### DIFF
--- a/big_tests/tests/bosh_SUITE.erl
+++ b/big_tests/tests/bosh_SUITE.erl
@@ -191,10 +191,6 @@ create_and_terminate_session(Config) ->
     Terminate = escalus_bosh:session_termination_body(get_bosh_rid(Conn), Sid),
     ok = bosh_send_raw(Conn, Terminate),
 
-    % Assert that correct events have been executed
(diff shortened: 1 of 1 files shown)


mongoose-im:
[mssql_mnesia_27](https://circleci.com/gh/esl/MongooseIM/266113) / odbc_mssql_mnesia / c1759634280e89bf09fd06ebfd09ddbef80776e4
Reports [root](http://esl.github.io/circleci-mim-results/s3_reports.html?prefix=PR/4529/266113/odbc_mssql_mnesia.27.1.2-amd64)

pawlooss1 (approved):
Looks good 👍

PR #4672, opened 2026-03-11
Send pings with correct interval in `mod_ping`

niecore:
The current implementation does not match the documentation. Following configuration options are available: `ping_interval` and `ping_req_timout`. The expectation would be that a server to client ping is send after `ping_interval` seconds after the last user activity and then the user has `ping_req_timeout` seconds to answer the ping before ping action is executed. But the ping is only sent after `ping_interval` + `ping_req_timeout` This change removes the additional timer for `ping_req_timeout`.

### DIFF
--- a/big_tests/tests/mod_ping_SUITE.erl
+++ b/big_tests/tests/mod_ping_SUITE.erl
@@ -258,7 +258,7 @@ check_connection(_, Client) ->
     true = escalus_connection:is_connected(Client).
 
 wait_for_ping_req(Alice) ->
-    PingReq = escalus_client:wait_for_stanza(Alice, timer:seconds(10)),
+    PingReq = escalus_client:wait_for_stanza(Alice, ping_interval() + timer:seconds(1)),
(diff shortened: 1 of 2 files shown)


big_tests/tests/mod_ping_SUITE.erl
```diff
@@ -258,7 +258,7 @@ check_connection(_, Client) ->
     true = escalus_connection:is_connected(Client).
 
 wait_for_ping_req(Alice) ->
-    PingReq = escalus_client:wait_for_stanza(Alice, timer:seconds(10)),
+    PingReq = escalus_client:wait_for_stanza(Alice, ping_interval() + timer:seconds(1)),
```
niecore, line 261: This test change makes the issue visible in the test results. I am not sure how stable that timing would be in CI but locally the tests were stable. If there are timing flakiness concerns, let me know and I can look for other solutions.

chrzaszcz:
Thanks for finding and fixing this. I am merging it into a branch to check it with CI.

